PR #29, opened 2026-04-11
fix(telegram): refuse to start with invalid/empty TELEGRAM_ALLOWED_USERS

WingedGuardian:
## Summary

- **Refuse to start** Telegram adapter when `TELEGRAM_ALLOWED_USERS` is empty or contains no valid numeric IDs — previously the bot would start silently and allow messages from _all_ users (security issue)
- **Dashboard validation** rejects `TELEGRAM_ALLOWED_USERS` values that look like a bot token (contain `:`) or aren't numeric, with a clear error pointing to `@userinfobot`
- **secrets.env.example** now documents the expected format for each Telegram field

## Background

A user configured `TELEGRAM_BOT_TOKEN` correctly but accidentally put the bot token value into `TELEGRAM_ALLOWED_USERS`. The system logged a quiet WARNING and started the bot with zero allowed users. Empty `allowed_users` in `_handler_context.py` means allow-all — so the bot was effectively open to anyone.

## Test plan

- [ ] `pytest tests/test_channels/test_bridge_wiring.py -v` — 7 tests pass (2 new: empty users → None, invalid UIDs → None)
- [ ] Dashboard PUT `/api/genesis/secrets` with `TELEGRAM_ALLOWED_USERS=123:abc` returns 422
- [ ] Server started with token but no allowed users logs ERROR and skips Telegram (dashboard-only mode)

### DIFF
--- a/secrets.env.example
+++ b/secrets.env.example
@@ -8,6 +8,10 @@
 
 # ─── Telegram ────────────────────────────────────────────────────────────
 # Required for the Telegram bridge. Create a bot via @BotFather.
+# TELEGRAM_BOT_TOKEN: paste the token from BotFather (format: 123456:ABC-DEF...)
+# TELEGRAM_ALLOWED_USERS: comma-separated numeric user IDs (NOT the bot token).
+#   Get your ID: message @userinfobot on Telegram.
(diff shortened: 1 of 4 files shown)


WingedGuardian:
Landed via v3.0a3-hf1 (cherry-picked into private main, released through the public release pipeline). Thanks for the fix.

PR #78, opened 2026-04-18
docs: add financial transaction approval rule

WingedGuardian:
## Summary
- Adds a hard rule to CLAUDE.md: Genesis must never initiate financial transactions without explicit per-transaction user approval
- Prior approval does not carry forward to subsequent transactions
- Only exception: dedicated accounts with explicit autonomous spending authorization

## Context
Learned from Conway Cloud setup where credits were transferred to a wrong wallet address without user confirmation, resulting in lost funds.

## Test plan
- [ ] Verify rule appears in CLAUDE.md Rules section
- [ ] Confirm Genesis sessions load the rule via system prompt

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -226,6 +226,12 @@ To send the user a future Telegram reminder, use `mcp__genesis-outreach__outreac
 
 ## Rules
 
+- **No unsanctioned financial transactions.** Genesis must NEVER send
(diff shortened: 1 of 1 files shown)


WingedGuardian:
Superseded by #80 which includes this rule plus the execute-don't-delegate rule.

PR #119, opened 2026-04-22
fix(mcp): wire DirectSessionRunner into standalone health MCP

WingedGuardian:
## Summary

- Wires DirectSessionRunner into standalone health MCP bootstrap (genesis_mcp_server.py)
- MCP servers are CC child processes that don't share Genesis runtime state — init_direct_session_tools() was never called, causing "not initialized" error
- Creates dedicated CCInvoker + SessionManager in the MCP lifespan hook

**Known limitation:** FastMCP uses anyio, but tracked_task uses asyncio.Task. Spawned sessions may not execute in the anyio loop. Core runner verified working via asyncio.run() directly. Needs anyio-compatible task spawning for full MCP integration — follow-up item.

## Test plan

- [x] `direct_session_run` MCP tool returns spawned session_id (no longer "not initialized")
- [x] `DirectSessionRunner.spawn()` runs end-to-end: session completes, metadata stored, status=completed
- [x] Profile enforcement validated (observe profile blocks write tools)

### DIFF
--- a/scripts/genesis_mcp_server.py
+++ b/scripts/genesis_mcp_server.py
@@ -106,6 +106,40 @@ def _bootstrap_health() -> None:
             tracker = ProviderActivityTracker()
             tracker.set_db(db)
             init_health_mcp(svc, activity_tracker=tracker)
+
+            # Wire direct session runner for background CC sessions.
+            # Standalone MCP has no outreach pipeline — notifications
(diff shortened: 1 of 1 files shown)


WingedGuardian:
Superseded by #121 (DB queue approach). Standalone runner wiring replaced by DB-backed dispatch.

PR #180, opened 2026-04-28
fix(executor): expand tilde in plan path before reading

WingedGuardian:
## Summary

- `Path('~/.genesis/plans/...').read_text()` raises `FileNotFoundError` because Python's `Path` doesn't automatically expand `~`
- Added `.expanduser()` before `.read_text()` in `CCSessionExecutor.execute()`
- This bug caused the mac-refactor task (`t-3575f95040bc`) to fail immediately on startup with "Cannot read plan" even though the file existed

## Test plan
- [ ] Submit a task with a `~/.genesis/plans/...` path — executor should read it correctly
- [ ] Existing plan paths (absolute) unaffected by `.expanduser()` (it's a no-op on absolute paths)

### DIFF
--- a/src/genesis/autonomy/executor/engine.py
+++ b/src/genesis/autonomy/executor/engine.py
@@ -123,7 +123,7 @@ class CCSessionExecutor:
             return False
 
         try:
-            plan_content = Path(plan_path).read_text(encoding="utf-8")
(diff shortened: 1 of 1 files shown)


WingedGuardian:
Superseded by #178 which included the same expanduser fix.

PR #212, opened 2026-05-03
fix(migrations): duplicate prefix guard + 0008→0009 rename

WingedGuardian:
## Summary

- **Renames** `0008_intake_token_enforcement.py` → `0009_intake_token_enforcement.py` to resolve the duplicate prefix that caused the v3.0b1 update failure
- **Adds pre-flight duplicate detection** in `MigrationRunner.run_pending()` — scans all available migrations for duplicate NNNN prefixes before running any, failing with a clear diagnostic naming both conflicting files
- **Adds tests** for the duplicate detection (both positive detection and no false-positive cases)

## Context

The v3.0b1 update failed because two PRs independently created `0008_*.py` migration files. The runner's `INSERT INTO schema_migrations` hit a UNIQUE constraint on the second file, triggering an automatic rollback. This guard catches the problem before any migration runs.

A CI job to catch this at PR time is ready but requires `workflow` scope to push — will be added separately.

## Test plan

- [ ] `pytest tests/test_db/test_migrations_runner.py::TestDuplicatePrefixDetection -v` — both new tests pass
- [ ] Existing migration runner tests unaffected (pre-existing failures unchanged)
- [ ] Re-run `scripts/update.sh` after merge to verify clean update path

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,6 +128,29 @@ jobs:
           fi
           echo "Email scan: CLEAN"
 
+  migration-check:
+    # Catch duplicate migration prefixes before merge — two PRs creating
+    # the same NNNN_*.py prefix will collide at runtime.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Check migration prefix uniqueness
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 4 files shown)


WingedGuardian:
All changes are already on main (via #219 and #220). Closing as superseded.

PR #524, opened 2026-06-02
feat(voice): Phase 2 S2S voice pipeline — Wyoming + GPT-Realtime

WingedGuardian:
## Summary

Phase 2 speech-to-speech voice pipeline. The S2S model (GPT-Realtime-1.5) acts as a
conversational front-end between the user and Genesis. Simple queries answered directly.
Memory/knowledge/tasks dispatched to Genesis via 2-tool function calling.

**Architecture:**
- Wyoming STT server (:10300) receives audio from HA's Assist pipeline
- Wyoming TTS server (:10301) sends S2S response audio back to HA
- GPT-Realtime-1.5 session with 2 tools: `ask_genesis` + `web_search`
- GenesisBridge dispatches tool calls to memory retriever, knowledge base, web search
- Phase 1 text pipeline preserved as fallback when S2S unavailable

**New files:**
- `config.py` — voice pipeline configuration
- `genesis_bridge.py` — 2-tool bridge for S2S function calls
- `s2s_session.py` — GPT-Realtime session manager
- `wyoming_stt.py` — Wyoming STT server
- `wyoming_tts.py` — Wyoming TTS server

**Verified via spikes:**
- GPT-Realtime-1.5 function calling: 4/4 tests passed (memory→ask_genesis, knowledge→no tool, weather→web_search, follow-ups→ask_genesis)
- Wyoming STT server: Describe/Info handshake works, audio events flow correctly
- Wyoming TTS server: synthesize→audio-chunk flow works

## Test plan
- [x] 19 new tests (config, bridge, session manager, TTS queue)
- [x] 25 existing voice tests pass (no regressions)
- [ ] CI green
- [ ] E2E: wake word → S2S model → spoken response (blocked on port forwarding)

## Prerequisites
Incus proxy devices on host VM for ports 10300/10301 (same as port 5000):
```
incus config device add genesis wyoming-stt proxy listen=tcp:0.0.0.0:10300 connect=tcp:127.0.0.1:10300
incus config device add genesis wyoming-tts proxy listen=tcp:0.0.0.0:10301 connect=tcp:127.0.0.1:10301
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #525, opened 2026-06-02
fix(voice): enable S2S audio output + fix response handling

WingedGuardian:
## Summary

First E2E test of the S2S voice pipeline (PR #524) revealed GPT-Realtime was returning text-only responses — no audio reaching the speaker. Three root causes identified and fixed:

- **Missing `modalities: ["text", "audio"]`** in session config — GPT-Realtime defaulted to text-only output. Added modalities, voice ("alloy"), and explicit audio format (pcm16)
- **`receive_response()` broke on first `response.done` after function calls** — the function-call response completes before the audio response starts. Added `awaiting_tool_response` flag to continue listening through the follow-up audio response
- **"Buffer too small" crash on short audio clips** — HA sometimes sends <100ms audio on rapid wake word retrigger. Added 150ms minimum guard with graceful fallback to Groq Whisper
- **Input audio sample rate mismatch** — Wyoming sends 16kHz, Realtime API expects 24kHz. Added resampling via existing `_resample_pcm()` utility

## Test plan

- [x] `pytest tests/test_channels/test_voice_s2s.py -v` — 19/19 pass
- [x] `ruff check` — clean
- [ ] Server restart + live E2E test: "okay nabu" → GPT-Realtime → audio plays on speaker
- [ ] Function call path: "what did we work on today?" → ask_genesis tool call → audio response
- [ ] Short audio rejection: rapid double wake word → falls back to Whisper instead of crashing

### DIFF
--- a/src/genesis/channels/voice/s2s_session.py
+++ b/src/genesis/channels/voice/s2s_session.py
@@ -134,11 +134,16 @@ class S2SSessionManager:
         session.connection = conn
         session._conn_mgr = conn_mgr
 
-        # Configure session with tools and system prompt
+        # Configure session with tools, system prompt, and audio output
         system_prompt = self._bridge.get_system_prompt()
+        voice = voice_config.s2s_voice()
         await conn.session.update(session={
-            "type": "realtime",
+            "modalities": ["text", "audio"],
             "instructions": system_prompt,
+            "voice": voice,
+            "input_audio_format": "pcm16",
+            "output_audio_format": "pcm16",
+            "input_audio_transcription": {"model": "gpt-4o-mini-transcribe"},
             "tools": TOOL_DECLARATIONS,
         })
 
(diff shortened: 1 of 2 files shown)


WingedGuardian:

Scanned added lines in `s2s_session.py` and `wyoming_stt.py`, plus PR title and description. No API keys, tokens, private IPs, personal emails, private hostnames, absolute paths with usernames, or incident-disclosure language found.

---

WingedGuardian (commented):
## Structural review: PR #525 — fix(voice): S2S audio delivery

**Files:** `s2s_session.py`, `wyoming_stt.py` | **No ERRORs found.**

---

### WARNING — Import of private function across sibling modules

**`s2s_session.py` inside `send_audio()`:**
```python
from genesis.channels.voice.wyoming_tts import _resample_pcm
```

`_resample_pcm` is a private function (underscore prefix, defined at `wyoming_tts.py:50`). Importing private symbols from a sibling module breaks encapsulation — if `wyoming_tts` is ever refactored, this breaks silently with no contract violation.

Additionally, `wyoming_tts.py` does not import from `s2s_session.py` at runtime, so there is **no circular import being avoided** — the deferred (in-function) placement has no structural justification. The import can be moved to the top of the file.

Recommended fix: extract `_resample_pcm` to a shared `voice_utils.py` as a public function and import it from both `s2s_session.py` and `wyoming_tts.py`.

---

### WARNING — `_MIN_AUDIO_BYTES` hardcodes a value derived from mutable instance state

**`wyoming_stt.py:124`:**
```python
_MIN_AUDIO_BYTES = 4800  # 16kHz * 2 bytes * 0.15s
```

`self._rate` (default 16000) and `self._width` (default 2) are set dynamically from the Wyoming `AudioStart` event (`wyoming_stt.py:85-86`). The class-level constant assumes 16kHz/16-bit/mono, but if the upstream satellite sends a different format, the guard silently uses the wrong threshold while the `duration_ms` log line (which uses `self._rate * self._width`) still reports the correct duration.

Recommended fix — compute from instance state (in `__init__` or inline):
```python
# In __init__, after self._rate / self._width are set:
self._min_audio_bytes = int(self._rate * self._width * 0.15)

# Or inline:
if len(audio) < int(self._rate * self._width * 0.15):
```

---

### WARNING — Missing test coverage for `wyoming_stt.py` behavioral changes

The PR test plan cites `tests/test_channels/test_voice_s2s.py`. Neither `test_voice_s2s.py` nor `test_voice.py` covers `_handle_s2s` in `wyoming_stt.py`. The new short-audio rejection path (guard + fallback) is an untested code path. A test asserting that audio of `< _MIN_AUDIO_BYTES` triggers `_handle_fallback` and that the `duration_ms` warning is logged should be added to `tests/test_channels/test_voice.py` or a new `test_voice_wyoming_stt.py`.

---

PR #526, opened 2026-06-02
fix(routing,surplus): DeepSeek V4 cost tracking + dream cycle hardening

WingedGuardian:
## Summary

- **DeepSeek V4 cost tracking**: Registers V4 Pro, V4 Flash, and NIM V4 Pro costs in `litellm.model_cost` at import time, stopping `cost_unknown` events on every surplus call-site invocation. Adds matching model profiles to `model_profiles.yaml`.
- **Dream cycle hardening**: Adds `record_job_start()` with a separate DB persist path (no `total_runs` double-count), APScheduler error/missed listener for silent failure observability, and explicit `timezone="UTC"` on all 9 CronTrigger instances.
- **Stale comment fix**: Corrects wrong line-number reference in dream cycle `finally` block.

## Context

Driven by 10 pending ego proposals reviewed in foreground session. Two were genuine code tasks; the rest were stale notifications or meta-work. Fix 1A (chunked dedup for large buckets) deferred to separate PR — requires design work since the O(n) search loop is structural.

## Files changed

| File | Change |
|------|--------|
| `src/genesis/routing/litellm_delegate.py` | Custom model cost injection at import time |
| `config/model_profiles.yaml` | 3 new profiles (V4 Pro, V4 Flash, NIM V4 Pro) |
| `src/genesis/surplus/scheduler.py` | CronTrigger tz, error listener, record_job_start call |
| `src/genesis/runtime/_job_health.py` | `record_job_start()` + `_persist_job_start()` |
| `src/genesis/runtime/_core.py` | Import + wrapper for `record_job_start` |

## Verification

- LiteLLM cost injection confirmed working E2E (`completion_cost()` returns correct values)
- All 3 model profiles parse correctly from YAML
- `_persist_job_start` verified to NOT increment `total_runs`
- Scheduler handler methods exist, imports resolve
- 128 targeted tests passed, 0 failed
- Lint clean (`ruff check`)

## Test plan

- [ ] CI passes
- [ ] After deploy, grep logs for `cost_unknown` on deepseek-v4 providers — should stop
- [ ] Next dream cycle run (June 7) should show `last_run` in `job_health` even if it crashes
- [ ] Verify CronTrigger jobs fire at correct UTC times

PR #528, opened 2026-06-02
fix(memory): chunked dedup for large dream cycle buckets

WingedGuardian:
## Summary

- Replaces the 500-point bucket skip-guard in entity resolution with a chunking loop (`DEDUP_CHUNK_SIZE=200`). Large buckets are now deduplicated instead of skipped.
- Cross-chunk pair dedup via shared `processed_pairs` set (Qdrant searches globally)
- Deprecated point guard on both search sources and search results to prevent double-counting
- Companion to PR #526 (dream cycle hardening)

## Context

PR #517 added a temporary `MAX_BUCKET_FOR_DEDUP=500` guard that skipped large buckets entirely to prevent OOM. This meant 6 buckets (including one with 14,887 points) were never deduplicated. This PR replaces the skip with bounded-memory chunking.

## Verified on real data

- 690-point bucket (`infrastructure/database`): processed in 4 chunks, **578 candidates found** (previously 0 — bucket was skipped), completed in 15.3 seconds
- Extrapolated: 14,887-point bucket ~5.5 minutes
- Zero errors, no OOM

## Test plan

- [x] 37 existing tests pass (including `TestBucketChunking`)
- [x] Lint clean (`ruff check`)
- [x] Code review: deprecated `point_b` guard added, `DEDUP_CHUNK_SIZE` hoisted to module level
- [x] E2E dry-run on real Qdrant data (690 points, 4 chunks, 578 candidates)
- [ ] CI passes
- [ ] June 7 dream cycle run completes without OOM

### DIFF
--- a/src/genesis/memory/dream_entity_scan.py
+++ b/src/genesis/memory/dream_entity_scan.py
@@ -27,6 +27,7 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 COLLECTION: str = "episodic_memory"
+DEDUP_CHUNK_SIZE: int = 200
 
 
 def _parse_ts(ts: str) -> datetime:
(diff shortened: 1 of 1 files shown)


WingedGuardian:
## Structural Review — PR #528 `fix(memory): chunked dedup for large dream cycle buckets`

**File reviewed:** `src/genesis/memory/dream_entity_scan.py` (1 file, +186/−148)

---

### ERRORs (must fix)

None.

---

### WARNINGs

**[WARNING] Missing test coverage for changed behavior — `run_entity_resolution`**

The entire dedup loop in `run_entity_resolution` is restructured (skip-guard replaced by chunking loop, `processed_pairs`/`deprecated_this_run` cross-chunk guards added). No test file appears in this PR's diff. The PR description references `TestBucketChunking` as passing, but the diff shows only the implementation file. If `TestBucketChunking` already exercises multi-chunk paths (buckets > 200 points, cross-chunk pair deduplication, deprecated-point filtering across chunks), this is fine — but if those tests were added separately, they should be referenced or included here.

What would close this: confirm `TestBucketChunking` covers (a) a bucket split across ≥2 chunks, (b) a cross-chunk duplicate pair, and (c) a point deprecated in chunk N being filtered in chunk N+1.

---

### NOTEs

**[NOTE] `run_entity_resolution` docstring (if any) may not reflect new behavior**

The function's contract changed significantly: it no longer skips large buckets, and the `report` dict no longer includes `skipped_large_buckets`. If a docstring or leading comment describes the old skip behavior, it should be updated.

**[NOTE] `DEDUP_CHUNK_SIZE = 200` — basis for the value undocumented**

The comment above the constant explains *what* it does but not *why 200*. Given the OOM context (PR #517), a one-line note on the memory/time tradeoff that motivated this number (e.g. "200 × ~1536-dim vectors ≈ ~1.2 MB per chunk") would help future readers calibrate whether to tune it.

---

### Clean checks

| Category | Result |
|---|---|
| `nohup` / bare process start | ✅ clean |
| Direct `task_states` insert | ✅ clean |
| `/tmp/` usage | ✅ clean |
| `git add .` / `git add -A` | ✅ clean |
| `git push --force` / `reset --hard` | ✅ clean |
| Bare `except:` clauses | ✅ clean — all handlers typed (`Exception`) |
| Mutable default arguments | ✅ clean |
| Raw SQL against genesis.db | ✅ clean |
| Star imports | ✅ clean |
| Circular imports | ✅ clean (inline `from genesis.db.crud import memory_links` unchanged from pre-existing code) |
| IntervalTrigger for >1h | ✅ n/a |
| Conventional commit prefix | ✅ `fix(memory):` |

---

**Overall: no pattern violations found.** Two notes for optional follow-up, one warning about test coverage verification.</body>
</invoke>

---

PR #536, opened 2026-06-03
fix(security): replace pickle with JSON in embedding disk cache

### LINKED ISSUE #519: Security [MEDIUM]: diskcache CVE-2025-69872 — pickle deserialization RCE in embedding cache
## Summary

**CVE-2025-69872** / GHSA-w8v5-vhqr-4h9v  
**Package:** `diskcache` (installed: 5.6.3)  
**Fix version:** None available upstream as of 2026-06-01  
**Severity:** MEDIUM (requires local write access to cache directory)

## Vulnerability

DiskCache uses Python `pickle` for serialization by default. An attacker with write access to the cache directory can plant malicious pickle data that achieves **arbitrary code execution** when the victim process reads from the cache.

## Genesis Usage — CONFIRMED

`src/genesis/memory/embeddings.py` uses `diskcache.Cache` to cache embedding vectors at `~/.genesis/embedding_cache/`. This is the L2 layer of the two-level embedding cache used for all memory recall operations.

```python
# embeddings.py ~line 244
self._disk_cache = diskcache.Cache(
    str(cache_dir), size_limit=100_000_000,  # 100 MB
)
```

## Attack Path

1. Attacker gains write access to `~/.genesis/embedding_cache/` (requires local filesystem access or a prior compromise)
2. Plants malicious pickle-serialized data in a cache file
3. Next memory recall triggers `diskcache` to deserialize the payload → **RCE as the Genesis process user**

## Why MEDIUM (not HIGH)

- Requires local write access to the user's home directory — attacker already needs meaningful system access
- No remote attack vector; exploit is a privilege-escalation / persistence mechanism post-initial-access
- Cache directory (`~/.genesis/embedding_cache/`) is not yet created in this environment (non-existent), so only active when embeddings are being used

## Recommended Actions

1. **Workaround (no upstream fix):** Override diskcache serialization to avoid pickle:
   ```python
   import diskcache, json, numpy as np
   
   class JsonDisk(diskcache.Disk):
       def put(self, key): return super().put(key)
       def store(self, value, read, **kwargs):
           return super().store(json.dumps(value), read, **kwargs)
       def fetch(self, mode, filename, value, read):
           raw = super().fetch(mode, filename, value, read)
           return json.loads(raw)
   
   cache = diskcache.Cache(str(cache_dir), disk=JsonDisk)
   ```
   Or use `diskcache.JSONDisk` if available in your version.

2. **Short-term:** Verify `~/.genesis/embedding_cache/` has mode `700` (owner-only), preventing lateral-movement exploitation even if a co-tenant process is compromised.

3. **Medium-term:** Evaluate replacing `diskcache` with a serialization-safe alternative (e.g., `shelve` with JSON, or a Redis-backed cache with explicit JSON encoding) for the embedding L2 layer.

4. **Track upstream:** Watch https://github.com/grantjenks/python-diskcache/issues for a fix or a `disk=SafeDisk` option in a future release.

## Falsifiability

This finding would be **disproven** if diskcache is confirmed to use a non-pickle serializer for the stored values (e.g., if embedding vectors stored as raw bytes bypass the pickle path). Verify by inspecting `diskcache.UNKNOWN` handling for `list[float]` values.

WingedGuardian:
## Summary

- Mitigates **CVE-2025-69872** (diskcache pickle deserialization RCE) by subclassing `diskcache.Disk` to use JSON serialization for embedding vectors
- Old pickle-format cache entries are silently rejected (treated as misses, re-embedded as JSON on next access)
- No dependency changes — same diskcache version, just a serializer override

## Changes

- `src/genesis/memory/embeddings.py`: Added `_SafeDisk` subclass that JSON-encodes `list[float]` values on write and rejects `MODE_PICKLE` entries on read
- `tests/test_memory/test_embedding_cache.py`: Two new tests verifying JSON storage mode and pickle rejection

## Test plan

- [x] `test_l2_stores_as_json_not_pickle` — verifies stored mode is not MODE_PICKLE (4)
- [x] `test_l2_rejects_pickle_entries` — pre-populates pickle entry, verifies it returns None
- [x] All 22 existing embedding cache tests pass (no regressions)

Closes #519

### DIFF
--- a/src/genesis/memory/embeddings.py
+++ b/src/genesis/memory/embeddings.py
@@ -239,10 +239,34 @@ class EmbeddingProvider:
         self._disk_cache = None
         if cache_dir is not None:
             try:
+                import json as _json
+
                 import diskcache
+                import diskcache.core
+
+                class _SafeDisk(diskcache.Disk):
+                    """Disk using JSON instead of pickle (CVE-2025-69872)."""
+
+                    def store(self, value, read, key=diskcache.core.UNKNOWN):
+                        if isinstance(value, (list, dict)):
+                            value = _json.dumps(value)
+                        return super().store(value, read, key=key)
+
+                    def fetch(self, mode, filename, value, read):
+                        if mode == diskcache.core.MODE_PICKLE:
+                            return None
+                        data = super().fetch(mode, filename, value, read)
+                        if isinstance(data, str):
+                            try:
+                                return _json.loads(data)
+                            except (ValueError, _json.JSONDecodeError):
+                                return data
+                        return data
+
                 cache_dir.mkdir(parents=True, exist_ok=True)
                 self._disk_cache = diskcache.Cache(
                     str(cache_dir), size_limit=100_000_000,  # 100 MB
+                    disk=_SafeDisk,
                 )
             except Exception:
                 logger.warning(
--- a/tests/test_memory/test_embedding_cache.py
+++ b/tests/test_memory/test_embedding_cache.py
@@ -119,6 +119,38 @@ class TestL2DiskCache:
             assert key in provider_with_l2._cache
             assert provider_with_l2._disk_cache.get(key) == fake_vec
 
+    def test_l2_stores_as_json_not_pickle(self, provider_with_l2: EmbeddingProvider):
+        """CVE-2025-69872: verify embedding cache never uses pickle serialization."""
+        import sqlite3
+
+        vec = [1.0, 2.0, 3.0]
+        provider_with_l2._cache_put("pickle_test", vec)
+        key = provider_with_l2._cache_key("pickle_test")
+        # Inspect the raw mode in diskcache's SQLite DB
+        cache_dir = str(provider_with_l2._disk_cache.directory)
+        conn = sqlite3.connect(cache_dir + "/cache.db")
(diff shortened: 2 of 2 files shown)


WingedGuardian:
## First-pass review — PR #536

### Structural review

**WARNING — Import placement** (`src/genesis/memory/embeddings.py`)

`import json as _json` is stdlib and is imported inside the `try` block in `__init__`. The lazy-import pattern is justified for `diskcache` (optional dependency), but `json` is always available and should be at the module top alongside the other stdlib imports. Moving it there is safe and keeps import hygiene clean.

**NOTE — Missing docstrings on overridden methods** (`src/genesis/memory/embeddings.py`)

`_SafeDisk.store` and `_SafeDisk.fetch` are undocumented. The class docstring covers the why, but brief per-method docstrings clarifying the intentional behaviour (e.g. "JSON-encodes list/dict values to avoid pickle serialization" and "Returns None for MODE_PICKLE entries — treats them as cache misses") would help future readers understand the CVE mitigation at a glance.

---

**CLEAN** — no secrets, private IPs, personal emails, absolute paths with usernames, or hardware identifiers in the added lines. PR title and body use standard CVE disclosure language; no wording that could be read as advertising an internal credential leak or infrastructure exposure.

---

*No blocking findings. Both items above are optional improvements.*

---

PR #537, opened 2026-06-04
fix(bootstrap): cap systemd journal at 200MB to prevent disk bloat

WingedGuardian:
## Summary

- Adds an idempotent drop-in (`/etc/systemd/journald.conf.d/genesis-size-cap.conf`) that caps journal storage at 200MB (~7 days rolling retention)
- Runs during bootstrap and on every `update.sh` — skips silently if already applied
- Fixes the root cause of the May 30 sentinel disk alarm (journal had grown to 1.1GB unchecked)

## Test plan

- [ ] Fresh install: confirm drop-in file is created and `journalctl --disk-usage` reports ≤200MB after bootstrap
- [ ] Existing install: confirm re-running bootstrap prints "Journal cap already set (200MB)" (idempotency)
- [ ] Confirm `systemd-journald` restarts cleanly after the drop-in is written

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -345,6 +345,9 @@
             }
         ]
     },
+    "worktree": {
+        "baseRef": "head"
+    },
     "enabledPlugins": {
         "claude-code-setup@claude-plugins-official": true
     }
--- a/docs/reference/cc-compatibility.md
+++ b/docs/reference/cc-compatibility.md
@@ -9,13 +9,14 @@
 > through 8 evaluation lenses (see analyzer prompt for details). This document
 > is updated manually after each evaluation.
 >
-> Created: 2026-03-09 | Last updated: 2026-04-01
+> Created: 2026-03-09 | Last updated: 2026-06-02
 
 ---
 
 ## Current CC Version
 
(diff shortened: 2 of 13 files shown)


WingedGuardian:
## Structural review

### ERROR — Raw SQL against genesis.db in production code

`src/genesis/learning/signals/cc_version.py`, `_check_registry_version`:

```python
    "SELECT 1 FROM observations WHERE source = 'cc_version' "
    "AND type = 'cc_version_available' AND resolved = 0 LIMIT 1",
)
```

The same method already imports and calls `genesis.db.crud.observations` for `create` — the CRUD path is live. This `SELECT` must go through the CRUD module too. If `observations` doesn't expose a `get_unresolved_by_source_and_type()` (or equivalent) helper yet, add one there rather than embedding raw SQL in the signal collector.

---

### WARNING — In-method import in test

`tests/test_cc/test_invoker.py`, `test_build_args_clamping_emits_warning`:

```python
def test_build_args_clamping_emits_warning(self, invoker, caplog):
    import logging   # ← inside method body
```

Move `import logging` to the module-level import block at the top of the file.

---

**CLEAN** — no secrets, private IPs, credentials, or sensitive hostnames in added lines. PR title and body are clean; the "May 30 sentinel disk alarm" reference is routine operational context, not a security-incident disclosure.

---

PR #551, opened 2026-06-07
chore(eval): V4 scope + active-skill cross-ref before IGNORE

WingedGuardian:
## Summary
- Remove V3 from all evaluation scope options — V4 is current, V5 is next
- Add active-skill cross-reference rule: evaluators must check if the *topic* relates to an active skill before rating IGNORE
- Update architecture.md to reflect V4 work is active

Addresses inbox eval quality gap where a reCAPTCHA behavioral biometrics video was rated IGNORE despite being directly relevant to the stealth-browser skill.

## Files changed
- `INBOX_EVALUATE.md` — scope tags + cross-ref rule
- `evaluate/SKILL.md` — scope tags + IGNORE guardrail
- `RESEARCH_EVALUATION.md` — scope tags + action routing
- `evaluate.md` (command) — scope tags + action routing
- `architecture.md` — V4 is active

## Test plan
- [ ] Verify no remaining "V3" in scope tag definitions across eval files
- [ ] Verify INBOX_EVALUATE.md contains active-skill cross-reference instruction


WingedGuardian:
## First-pass code review

### Structural: ERROR — must fix

**Raw SQL against genesis.db in `_fetch_thread_messages`** (`src/genesis/cc/conversation.py`)

The new `_fetch_thread_messages` method issues two raw SQL queries directly against the `telegram_messages` table instead of going through a `genesis.db.crud` module:

```python
query = """SELECT sender, content, timestamp FROM telegram_messages
           WHERE chat_id = ? AND thread_id = ?
           ORDER BY timestamp DESC LIMIT ?"""
```

and the fallback path without `chat_id`. Note that `_build_topic_context` already has a pre-existing raw SQL call against `telegram_topics` — that's a pre-existing violation, but this PR adds a second raw SQL access point in the same file. Both should route through CRUD; at minimum the new method must.

---

### Structural: WARNING

**Missing test coverage for `_fetch_thread_messages`** (`src/genesis/cc/conversation.py`)

The new private method has two distinct DB query paths (with `chat_id` and without) and a defensive `except Exception` fallback. Neither path is tested. `tests/test_cc/` has no coverage for this method.

**`session_config` validation paths not covered** (`src/genesis/mcp/health/session_control.py`, `tests/test_mcp/test_health_mcp.py`)

`test_health_mcp.py` was updated only to check tool registration (`name in tools`). The new combined tool introduces several new logic paths with no coverage:
- both-`None` guard → error path
- invalid `model` value → error path
- invalid `effort` value → error path
- DB unavailable → error path
- partial update (model-only or effort-only) → correct result shape

---

No credentials, private IPs, or sensitive identifiers in added lines. PR title and body contain no security incident language.

---

PR #553, opened 2026-06-07
fix(ego): pending-proposal dedup in realist + content-hash guard

WingedGuardian:
## Summary

- Realist Rule #2 now rejects proposals that semantically duplicate a currently **pending** or **approved** proposal (previously only caught recycled/withdrawn/expired "zombies")
- `create_batch()` pre-insert content-hash check catches exact-text duplicates as a belt-and-suspenders guard
- Includes carve-out so related-but-distinct proposals (e.g., different paper sections) aren't falsely rejected

## Root Cause

The ego re-words proposals each cycle, so content_hash (exact match) didn't catch them. The realist had the pending proposal data in its history table but no instruction to flag semantic overlaps with pending items.

## Test plan

- [x] 3 new tests: dedup blocks exact duplicate, allows different content, allows re-proposal after rejection
- [x] All 125 ego tests pass
- [x] Lint clean
- [ ] CI green
- [ ] Monitor next ego cycle for duplicate proposals

### DIFF
--- a/docs/reference/models.md
+++ b/docs/reference/models.md
@@ -141,7 +141,7 @@ The daily drivers for professional productivity and general intelligence.
 - **Intelligence Tier:** A/S (Thoughtful reasoning)
 - **SWE-Bench:** ~62%
 - **Cost:** $1.00/$4.00 per MTok (input/output)
-- **Free Tier:** Available on Nvidia NIM (5,000 credits, 40 RPM)
+- **Free Tier:** Available on Nvidia NIM (5,000 credits, 40 RPM); also free on OpenRouter (`:free` suffix, 262k context)
 
 **Best At:**
 1. Needle in a Haystack: Specialized architecture for searching its own memory — king of long-context retrieval.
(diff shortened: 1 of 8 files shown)


WingedGuardian:
## First-pass review — 1 ERROR, 1 NOTE

### Structural

**ERROR — Raw SQL outside a migration (`src/genesis/ego/proposals.py`, dedup block)**

The new dedup check executes a hand-written query directly on the database connection:

```python
    "SELECT 1 FROM ego_proposals "
    "WHERE content_hash = ? AND status IN ('pending', 'approved') "
    "LIMIT 1",
    (hash_val,),
)
```

Per Genesis conventions, raw SQL against `genesis.db` must go through a `genesis.db.crud` module (exception is migrations only). Add a function such as `ego_crud.find_pending_by_hash(db, hash_val) -> bool` and call that instead.

---

**NOTE — Docstring for the new crud helper (once added)**

When you add the crud function, add a one-line docstring so the public surface stays documented.

---

### PII / Secrets / Wording

CLEAN — no secrets, private IPs, credentials, or sensitive wording found in the diff or PR metadata.

---

PR #561, opened 2026-06-07
fix(mail): increase misfire grace to 6h for weekly job

WingedGuardian:
## Summary

- Increase mail_monitor_batch misfire_grace_time from 1h to 6h
- Fix config comment: cron runs in user timezone, not UTC

## Why

Today's weekly mail run was missed because the server restarted 5 times
during the cron window (PR merges). The 1-hour grace wasn't enough to
survive the restart churn. 6 hours covers typical morning activity without
risk of double-firing (weekly cadence, single instance guard).

### DIFF
--- a/config/mail_monitor.yaml
+++ b/config/mail_monitor.yaml
@@ -1,6 +1,6 @@
 mail_monitor:
   enabled: true
-  cron_expression: "0 5 * * 0"        # Sunday 05:00 UTC
+  cron_expression: "0 5 * * 0"        # Sunday 05:00 user timezone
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

**NOTE:** `config/mail_monitor.yaml` — the updated comment `# Sunday 05:00 user timezone` is more accurate than the previous `UTC` claim, but "user timezone" is vague. If the scheduler's timezone is explicitly configured somewhere (e.g. `scheduler_timezone` in config), referencing it by name (e.g. `# Sunday 05:00 America/New_York`) would be clearer. Not a blocker.

---

PR #567, opened 2026-06-08
fix(mail): wire thread tracker in outreach init, not mail init

WingedGuardian:
## Summary
- Thread tracker was wired in mail init, but outreach pipeline doesn't exist yet at that point in the boot sequence (mail init runs before outreach init)
- Moved the `set_thread_tracker()` call to outreach init where `_outreach_pipeline` is already created and `_thread_tracker` is available on the runtime

## Test plan
- [x] `tests/test_mail/test_threads.py` — 19/19 pass
- [x] Lint clean
- [ ] Verify "Thread tracker wired into outreach pipeline" appears in server logs after restart

### DIFF
--- a/src/genesis/runtime/init/mail.py
+++ b/src/genesis/runtime/init/mail.py
@@ -101,11 +101,8 @@ async def init(rt: GenesisRuntime) -> None:
                 _register_reply_poll_job(rt._mail_monitor, reply_poller)
                 logger.info("Email reply poller registered (every 4h)")
 
-                # Wire thread tracker into outreach pipeline for auto-registration
-                pipeline = getattr(rt, "_outreach_pipeline", None)
-                if pipeline is not None and hasattr(pipeline, "set_thread_tracker"):
-                    pipeline.set_thread_tracker(thread_tracker)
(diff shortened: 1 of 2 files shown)


WingedGuardian:
**Structural review:** 1 NOTE — no ERRORs or WARNINGs.

---

**NOTE — `hasattr` guard dropped on `set_thread_tracker` (`src/genesis/runtime/init/outreach.py`)**

The removed mail-init code guarded defensively:
```python
if pipeline is not None and hasattr(pipeline, "set_thread_tracker"):
    pipeline.set_thread_tracker(thread_tracker)
```

The new outreach-init code calls the method directly:
```python
if thread_tracker is not None:
    rt._outreach_pipeline.set_thread_tracker(thread_tracker)
```

This is almost certainly correct — `_outreach_pipeline` is constructed earlier in the same `init()` and the tests confirm the method exists. The old `hasattr` was defensive scaffolding for the wrong boot order, not a real guard. Just flagging it so the drop is explicit: if `OutreachPipeline.set_thread_tracker` is ever renamed or removed, this will raise `AttributeError` rather than silently no-op.

---

---

PR #570, opened 2026-06-08
feat(voice): pre-announce chime + shutdown timing fix

WingedGuardian:
## Summary
- Switch voice adapter from `tts.speak` to `assist_satellite.announce` — plays a built-in pre-announce chime before TTS, giving the user a gentle audio cue before Genesis speaks unprompted
- Falls back to `tts.speak` if announce fails or satellite entity not configured
- `preannounce=False` parameter for rapid follow-up messages where chime would be redundant
- Fix shutdown notification timing: 4s delay after TTS dispatch so HA can deliver audio before Wyoming servers shut down

## Test plan
- [x] 28 voice adapter tests pass (4 new: announce, no-preannounce, fallback, no-satellite)
- [x] 41 S2S voice tests pass (unchanged)
- [ ] E2E: trigger proactive alert → hear chime + spoken message
- [ ] E2E: restart server → hear shutdown notification before device goes to error state

### DIFF
--- a/src/genesis/channels/voice/adapter.py
+++ b/src/genesis/channels/voice/adapter.py
@@ -1,14 +1,18 @@
-"""Voice channel adapter — HA TTS for outbound speech.
+"""Voice channel adapter — HA satellite announce for outbound speech.
 
 Implements ``ChannelAdapter`` for voice output through Home Assistant.
 Inbound voice (HA → Genesis) arrives via the Flask endpoint in
 ``dashboard/routes/voice_api.py``, not through this adapter.  This
 adapter handles outbound only: Genesis speaking to the user proactively.
 
-NOT registered with the outreach pipeline via ``register_channel`` —
-outreach delivery at arbitrary times (3am morning reports) would be
-disruptive.  Instead, held directly by the standalone server for
-explicit voice output when needed.
+Uses ``assist_satellite.announce`` for delivery — this plays a short
+pre-announce chime before the TTS message, giving the user a gentle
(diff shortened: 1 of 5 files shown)


WingedGuardian (commented):
## First-pass review: 1 WARNING · 1 SOFT WARNING

### Structural

**WARNING — Import inside function body** (`src/genesis/hosting/standalone.py`, shutdown method)

```python
import asyncio
await asyncio.sleep(4)
```

`asyncio` is a stdlib module and should be imported at the top of the file with the other stdlib imports, not deferred into the function body. Move it to the existing import block.

---

### PII / Secrets

**SOFT WARNING — Hardware identifier in default parameter** (`src/genesis/channels/voice/adapter.py:37`)

The new `satellite_entity` default value contains `home_assistant_voice_0a2841` — the `0a2841` fragment looks like a partial hardware identifier (MAC/serial fragment common in HA auto-generated entity IDs). Not a secret, but worth noting: if the default is device-specific it will silently fail on any other install. Consider making the default `""` (empty, like `satellite_entity=""`) and requiring explicit configuration, or documenting that callers must pass their own entity ID. Flagging as soft warning rather than hard block since there's no credential exposure.

---

### Clean areas

- No bare `except:`, no mutable defaults, no `/tmp/`, no raw SQL, no `task_states` direct inserts
- Commit prefix `feat(voice):` ✓
- All 4 new test cases cover the new paths (announce, no-preannounce, fallback, no-satellite) ✓
- PR title and body: no PII/secrets wording issues ✓

---

PR #571, opened 2026-06-08
fix(voice): media_player chime + approval notifications + time awareness

WingedGuardian:
## Summary
- Switch pre-chime from `assist_satellite.announce` to `media_player.play_media` — the announce service engages the voice assistant C++ state machine, which unpredictably reopens the microphone. The media player path stays entirely in the media layer.
- Custom chime audio file (gentle two-note ascending tone, 0.45s) uploaded to HA `/config/www/`
- Add `OutreachCategory.APPROVAL` to voice categories so pending approval requests are spoken aloud
- Include current local time in S2S system prompt so the model can answer "what time is it?"
- Named constants for shutdown drain delay and chime delay

## Test plan
- [x] 69 voice tests pass (28 adapter + 41 S2S)
- [x] E2E: chime + TTS plays, satellite stays idle (no mic reopen)
- [x] E2E: continue-conversation still works for wake-word conversations
- [ ] E2E: restart server → hear chime + "Server restarting" notification
- [ ] E2E: approval request triggers voice notification

### DIFF
--- a/src/genesis/channels/voice/adapter.py
+++ b/src/genesis/channels/voice/adapter.py
@@ -6,12 +6,15 @@ Inbound voice (HA → Genesis) arrives via the Flask endpoint in
 adapter handles outbound only: Genesis speaking to the user proactively.
 
 Two-step delivery:
-1. ``assist_satellite.announce`` with empty message — plays the built-in
-   pre-announce chime (gentle audio cue before Genesis speaks).
+1. ``media_player.play_media`` — plays a gentle chime audio file as an
+   audio cue before Genesis speaks.
 2. ``tts.speak`` — renders and plays the actual message via Piper.
 
-The announce service's own TTS rendering is unreliable (depends on
-pipeline TTS config), so we split chime and speech into separate calls.
+The chime uses ``media_player.play_media`` (not ``assist_satellite.announce``)
+because the announce service engages the voice assistant C++ state machine,
+which can unpredictably reopen the microphone via its internal
+``continue_conversation_`` flag.  The media player path stays entirely
+in the media layer — no voice assistant involvement.
 
 Registered with the outreach pipeline (time-gated to voice hours) for
 proactive alerts, and held by the standalone server for shutdown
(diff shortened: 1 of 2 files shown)


WingedGuardian:
## Code Review — PR #571

**Structural review: 2 WARNINGs**

---

### Structural

**WARNING — Import hygiene: deferred stdlib imports inside method body**

`src/genesis/channels/voice/genesis_bridge.py` — `get_system_prompt()` imports three modules inline:

```python
from datetime import datetime
from genesis.env import user_timezone
import zoneinfo  # placed even lower, after other statements
```

`datetime` and `zoneinfo` are stdlib with zero circular-import risk — they belong at the top of the file in the stdlib group. `user_timezone` is a local import; if it's safe at module level (check for cycles), move it there too. The same pattern appears in `src/genesis/hosting/standalone.py` where `VoiceChannelAdapter` is imported inline inside `shutdown()` — check whether that's guarding a circular import or just incidental.

---

**WARNING — Missing test coverage for 3 changed modules**

`adapter.py` coverage is excellent — 4 new tests covering the happy path, no-preannounce, chime failure, and service-path assertion. The other changed modules have no new tests in this diff:

- `genesis_bridge.py` `get_system_prompt()`: new time-injection path is untested. A test asserting `"Current time:"` appears in the returned prompt, and that a bad timezone string falls back cleanly, would cover the new branches.
- `outreach/pipeline.py` `_should_voice()`: the `APPROVAL` category addition has no accompanying test. Existing tests may cover it — flagging in case they don't.
- `hosting/standalone.py` shutdown drain: `SHUTDOWN_DRAIN_S` sleep has no test. Existing integration coverage may exist.

---

### PII / Secrets / Wording

**SOFT WARNING — Private device identifier hardcoded as default parameter** (`src/genesis/channels/voice/adapter.py`)

```python
media_player_entity: str = "media_player.home_assistant_voice_0a2841",
```

The `_0a2841` suffix reads like a device-specific hardware identifier (partial MAC or serial number). Hardcoding it as a default couples the code to a specific physical device and leaks infrastructure detail if the repo is ever shared. Consider matching the pattern used for `ha_url`/`ha_token`: default to `""` (or `None`) and require the caller to supply the entity ID from `genesis.yaml` config.

---

No API keys, tokens, private IPs, SSH keys, or credential strings found. PR title and body are clean — no leak-cleanup language.

---

PR #583, opened 2026-06-09
refactor(mail): consolidate reply handler profiles and prompt templates

WingedGuardian:
## Summary

- Extract dedicated mail reply session profile with streamlined MCP config
- Standardize reply and follow-up prompt construction with consistent content boundaries
- Update MAIL_REPLY.md with clearer decision framework and sending instructions
- Add thread-aware email delivery for per-recipient routing
- Simplify recon finding storage path (observations instead of intake pipeline)
- Add outbound content validation to email delivery pipeline
- Consolidate content processing modes in sanitizer with perimeter-aware blocking
- Clean up judge session configuration (strip unused tool access)

## Test plan

- [ ] Profile tests: 48 tests verify tool restrictions, addenda, and model behavior
- [ ] Output scanner: 14 tests covering detection accuracy and false-positive avoidance  
- [ ] Sanitizer: existing 49 tests pass with new blocking mode
- [ ] All 111 tests pass locally
- [ ] CI green

### DIFF
--- a/src/genesis/cc/direct_session.py
+++ b/src/genesis/cc/direct_session.py
@@ -123,6 +123,20 @@ _NO_RECON_WRITES = [
     "mcp__genesis-recon__recon_run_model_intelligence",
 ]
 
+# Perimeter sessions: block web tools to prevent second-stage content
+# retrieval from attacker-controlled URLs.
+_NO_WEB_TOOLS = [
+    "WebFetch",
+    "WebSearch",
+]
+
+# Perimeter sessions: block outreach tools beyond basic send.
+_NO_OUTREACH_EXTRAS = [
+    "mcp__genesis-outreach__outreach_send_and_wait",
+    "mcp__genesis-outreach__outreach_poll",
+    "mcp__genesis-outreach__outreach_digest",
+]
+
 PROFILES: dict[str, list[str]] = {
     "observe": (
         _UNIVERSAL_DISALLOW + _NO_FILE_WRITE + _NO_OUTREACH_SEND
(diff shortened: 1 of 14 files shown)


WingedGuardian (commented):
## First-pass review: 2 warnings, 1 note

**PII/Secrets/Wording scan: CLEAN** — no secrets, private IPs (test fixtures excluded per rules), or incident language in PR metadata.

---

### WARNING — Direction string case mismatch (`reply_handler.py`)

`_build_reply_prompt` checks `direction == "RECEIVED"` (uppercase) while `_build_follow_up_prompt` checks `msg["direction"] == "received"` (lowercase).

If the `direction` field is stored lowercase in the DB (which `"received"` in the follow-up function suggests it is), the branch in `_build_reply_prompt` never fires for received messages in thread history — those messages bypass the sanitizer wrapping silently. The security boundary is applied only to the current reply body, not to any received history messages shown in the prompt.

Needs a consistent comparison. Recommend verifying what the DB actually stores and unifying to match.

---

### WARNING — No test for `handle_reply` blocking path (`reply_handler.py`)

The new `sanitizer.should_block(check)` early return (added around line 195) is the perimeter entry gate, but it has no unit test. A false positive here silently drops a legitimate reply and returns `None` with no user notification. Given the security role of this code path, tests covering both the block case and the pass-through case are needed — at minimum one each.

---

### NOTE — `timeout_s=600` is below the documented 2h floor

The 10-minute timeout is well-justified (no memory/web tools on the mail profile), but CLAUDE.md documents a 2h default floor. Confirming that the policy allows explicit exceptions for stripped-down perimeter profiles is worthwhile, or the floor policy doc should note the exception class.

---

PR #589, opened 2026-06-09
fix(infra): prevent pipefail cascade in tmp_watchgod helpers

WingedGuardian:
## Summary

- Fix `dir_usage_mb`, `fs_free_mb`, and `tmp_usage_pct` helpers in `tmp_watchgod.sh` to capture pipeline output before applying defaults
- Under `set -o pipefail`, transient `du`/`df` errors (race with CC file ops) caused `|| echo 0` to **append** `0` to already-emitted output, producing multi-line values like `63\n0`
- This cascaded: arithmetic syntax errors → corrupted JSON state → Python `JSONDecodeError` → dashboard infrastructure RED

## Root cause

```
du -sm dir | awk '{print $1}' || echo 0
```
When `du` exits non-zero but `awk` already output `63`, `pipefail` makes the pipeline fail, so `|| echo 0` fires — stdout becomes `63\n0`.

## Fix

Capture in subshell first, then apply default on empty:
```bash
result=$(du -sm "$1" 2>/dev/null | awk '{print $1}') || true
echo "${result:-0}"
```

## Test plan

- [x] `bash -n` syntax check passes
- [x] Verified normal case returns clean integer
- [x] Verified missing-directory case returns default `0`
- [ ] Restart `genesis-tmp-watchgod` service and confirm journal is clean

### DIFF
--- a/scripts/tmp_watchgod.sh
+++ b/scripts/tmp_watchgod.sh
@@ -40,20 +40,27 @@ log() {
 # ── Helpers ──────────────────────────────────────────────────
 dir_usage_mb() {
     # Total disk usage of a directory in MB
-    du -sm "$1" 2>/dev/null | awk '{print $1}' || echo 0
+    # NOTE: capture first, then default — pipefail + '|| echo 0' appends
+    # a spurious '0' when du exits non-zero but awk already emitted output.
+    local result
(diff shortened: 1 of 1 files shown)


WingedGuardian:

---

PR #592, opened 2026-06-10
fix(eval): dashboard white flash + J9 eval session attribution bugs

WingedGuardian:
## Summary

- **Dashboard white flash**: All 4 page templates now have inline dark background (#131313) on `<html>` + `<meta name="color-scheme" content="dark">` to prevent white flash during full-page navigation
- **surplus.py raw JSON**: Deserialize `metrics_json` before returning to frontend (was returning raw string, breaking Neural Monitor)
- **eval.py null deref**: Defensive `.get("metrics", {})` on line 60, matching safe pattern on line 48
- **J9 session attribution**: `emit_recall_fired` and `emit_procedure_invoked` now resolve `session_id` via fallback chain (explicit → ContextVar → `GENESIS_SESSION_ID` env var), fixing cognitive dimension being permanently null
- **Session leak in `_compute_recall_used`**: Now queries `pending_embeddings.source_session_id` instead of all recent memories via `memory_metadata.created_at` — fixes inflated `usage_rate` metric
- **Foreground session context**: `ConversationLoop` sets ContextVar for both `handle_message` and `handle_message_streaming` paths
- **CCInvoker env propagation**: Propagates session ContextVar → `GENESIS_SESSION_ID` env var for child MCP processes, clears stale values

## Audit context

Full J9 eval system audit identified 7 bugs — 2 dead dimensions (cognitive, procedure), session data leak inflating usage_rate, raw JSON API response, null deref, and missing session attribution. This PR fixes the code bugs. Methodology changes (ego grade weights, awareness scoring, missing-data inflation) are deferred to a separate PR.

## Test plan

- [x] `ruff check` — all changed files pass
- [x] `pytest tests/test_eval/test_j9_eval.py tests/test_eval/test_recall_instrumentation.py` — 38 passed
- [x] `pytest tests/test_dashboard/` — 102 passed
- [ ] E2E: Navigate Dashboard → Error Log → Event Log → Neural Monitor — no white flash
- [ ] E2E: After server restart, trigger recall, verify `session_id` populated in `eval_events`
- [ ] E2E: Verify `/api/genesis/surplus/config` returns `metrics` (object) not `metrics_json` (string) in j9 data

### DIFF
--- a/src/genesis/cc/conversation.py
+++ b/src/genesis/cc/conversation.py
@@ -153,6 +153,13 @@ class ConversationLoop:
             thread_id=thread_id,
         )
 
+        # Set session context so downstream code (CCInvoker, eval hooks)
+        # can attribute work to this session without explicit threading.
+        # Uses set/clear rather than session_scope() to avoid re-indenting
+        # the entire lock block; follows the pattern in direct_session.py.
+        from genesis.observability.session_context import set_session_id
+        set_session_id(session["id"])
+
         async with self._get_lock(session["id"]):
             await self._persist_overrides(session, model, effort)
 
@@ -351,6 +358,10 @@ class ConversationLoop:
             user_id=user_id, channel=channel, model=model, effort=effort,
             thread_id=thread_id,
         )
+
+        # Set session context for eval attribution (same as handle_message).
+        from genesis.observability.session_context import set_session_id as _set_sid
+        _set_sid(session["id"])
 
         async with self._get_lock(session["id"]):
             # Capture old values before persisting overrides (for change feedback)
--- a/src/genesis/cc/invoker.py
+++ b/src/genesis/cc/invoker.py
@@ -179,6 +179,14 @@ class CCInvoker:
         # The genesis_session_context.py hook skips identity injection when set,
         # preventing double injection (identity is in the system prompt arg).
         env["GENESIS_CC_SESSION"] = "1"
+        # Propagate Genesis session_id to child CC + MCP server processes
+        # so eval hooks can attribute recall events to specific sessions.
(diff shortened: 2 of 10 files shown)


WingedGuardian:
## First-pass review — PR #592

### Structural review

**No ERRORs found.**

#### NOTE — `conversation.py`: redundant guard in `j9_batch.py`

`src/genesis/eval/j9_batch.py` — The dict `by_session` is built with:
```python
if sid:
    by_session.setdefault(sid, []).append(ev)
```
…so iterating `for session_id, events in by_session.items()` means `session_id` is already guaranteed truthy. The `if not session_id: continue` guard added immediately inside the loop is dead code. Safe to remove.

#### NOTE — `conversation.py`: `set_session_id` without restore token

`src/genesis/cc/conversation.py:161` and `:363` — Both `handle_message` and `handle_message_streaming` call `set_session_id(session["id"])` (raw set, no token saved) rather than using the existing `session_scope()` context manager. The comment correctly notes this avoids re-indenting the lock block, and in asyncio each Task owns a context copy so cross-request leakage is not a concern in practice.

The one edge case to be aware of: if `handle_message` is ever called from a coroutine that already holds a session_id in its ContextVar (e.g., a nested or reused Task), the previous value is silently clobbered and never restored — unlike `session_scope()` which uses a reset token. Not a bug today given the call sites, but the asymmetry with `session_scope()` is worth keeping in mind if call sites change.

---

**CLEAN** — No secrets, tokens, private IPs, private hostnames, or credentials in any added lines. PR title and body contain no language implying credential exposure, security incidents, or infrastructure cleanup.

---

PR #594, opened 2026-06-10
feat(infra): worktree lifecycle management — cross-session guard + trash bin

WingedGuardian:
## Summary

- **Cross-session guard hook**: Extends `worktree_cwd_guard.py` to scan `/proc/*/cwd` for OTHER sessions using a worktree before allowing removal. ALL direct worktree removal is now blocked — both `git worktree remove` (Bash) and ExitWorktree "remove" (CC built-in tool).
- **Lifecycle manager** (`scripts/worktree_lifecycle.py`): Automated stale worktree cleanup via daily cron. Worktrees are trashed (not deleted) when: no active processes, no activity for 14 days, and PR is merged. 7-day recovery window. CLI: `--dry-run`, `--list-trash`, `--recover`.
- **ExitWorktree interception**: New PreToolUse hook matcher for the ExitWorktree built-in tool. Blocks `action: "remove"` and guides to `action: "keep"`.

### Incident context
2026-06-09: Session B finished work in a shared worktree, user said "remove worktree" on `/exit`. Session A — still using that worktree — became a zombie. Every tool call failed with "Working directory no longer exists."

### Three safety layers
1. **Guard hook** — blocks removal if any process is using the worktree (hard guarantee)
2. **Block-all policy** — even without conflicts, redirects to lifecycle manager
3. **Trash bin** — 7-day recovery for worktrees reaped by the lifecycle manager

## Test plan
- [x] 16 new tests: passthrough, block-all, self-CWD, cross-session PID detection, ExitWorktree, fail-open (all passing)
- [x] 100 existing inline hook tests unaffected (all passing)
- [x] 360/360 total hook tests passing
- [x] E2E: cross-session detection correctly identifies PID 941743 in live worktree
- [x] E2E: ExitWorktree "keep" allowed, "remove" blocked
- [x] E2E: lifecycle `--dry-run` correctly skips active/recent worktrees
- [x] Code review: metadata write ordering fixed (staging file before move)
- [x] Daily cron installed: `0 4 * * *`

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -233,6 +233,16 @@
                         "timeout": 2000
                     }
                 ]
+            },
+            {
+                "matcher": "ExitWorktree",
+                "hooks": [
+                    {
+                        "type": "command",
+                        "command": "${CLAUDE_PROJECT_DIR}/.claude/hooks/genesis-hook hooks/worktree_cwd_guard.py --exit-worktree",
+                        "timeout": 2000
+                    }
+                ]
             }
         ],
         "Stop": [
--- a/scripts/hooks/worktree_cwd_guard.py
+++ b/scripts/hooks/worktree_cwd_guard.py
@@ -117,7 +117,7 @@ def _block_with_pids(target: str, pids: list[int]) -> int:
 def _block_no_direct_removal(target: str) -> int:
     """Print lifecycle-manager redirect and return exit code 2."""
     print(
-        f"BLOCKED: Direct worktree removal is disabled.",
+        "BLOCKED: Direct worktree removal is disabled.",
         file=sys.stderr,
     )
     print(
--- a/scripts/hooks/worktree_cwd_guard.py
+++ b/scripts/hooks/worktree_cwd_guard.py
@@ -1,12 +1,23 @@
 #!/usr/bin/env python3
-"""PreToolUse hook (Bash): block git worktree remove on the current working directory.
+"""PreToolUse hook: block worktree removal to protect active sessions.
 
-Removing a worktree that is the shell's CWD bricks the entire session —
-every subsequent Bash command fails with "Path does not exist" because the
-shell's persisted working directory no longer exists.
+Two modes:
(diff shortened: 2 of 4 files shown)


WingedGuardian (commented):
## First-pass review: structural + PII/secrets

### Structural Review

**WARNING — Unhandled `json.JSONDecodeError` in `_recover()` (`scripts/worktree_lifecycle.py`)**

`json.loads(meta_path.read_text())` is called bare after the `meta_path.exists()` check. The same pattern in `_purge_old_trash()` is wrapped in `try/except (json.JSONDecodeError, ValueError, OSError)`. A corrupted `.trash_meta.json` (partial write during a crash, disk error) will cause `--recover` to exit with an unhandled traceback rather than a graceful error. Suggest wrapping consistently:

```python
try:
    meta = json.loads(meta_path.read_text())
except (json.JSONDecodeError, OSError):
    print(f"Corrupted metadata in {meta_path}", file=sys.stderr)
    return False
```

---

**NOTE — `_find_processes_in_dir` is duplicated verbatim between `worktree_cwd_guard.py` and `worktree_lifecycle.py`**

Both scripts are stdlib-only by design (explicitly documented in each module docstring), so a shared import isn't possible. The duplication is intentional. Worth adding a one-liner cross-reference comment in each copy (e.g. `# Duplicated in worktree_lifecycle.py — both are stdlib-only`) so a future reader doesn't "clean it up" by introducing a non-stdlib import.

---

**NOTE — No test file for `scripts/worktree_lifecycle.py`**

The missing-coverage rule triggers for `src/` only, so this doesn't violate policy. The PR test checklist covers `--dry-run` E2E, which is the right approach for cron behavior. Flagging only because the lifecycle logic (`_branch_is_merged` fallback chain, `_last_activity_time` depth-limited walk, name-collision loop in `_trash_worktree`) has enough edge cases to benefit from unit tests in a follow-up.

---

**CLEAN** — No secrets, private IPs, private hostnames, personal identifiers, or sensitive wording found in added lines or PR metadata.

---

PR #601, opened 2026-06-10
chore: bump Claude Code 2.1.160 → 2.1.170 and fix background-session transcript path

WingedGuardian:
## What

- **Bump pinned Claude Code 2.1.160 → 2.1.170** (install scripts + setup docs). New installs get 2.1.170, which adds Claude Fable 5 access and fixes transcript saving for sessions launched from a shell that inherited Claude Code environment variables.
- **Fix background-session transcript path derivation.** The project-key encoding only replaced `/` and left the leading dot of `~/.genesis` intact, so the stored transcript path pointed at a nonexistent directory — downstream transcript readers (audit, bookmark enrichment) silently found nothing. Now replicates Claude Code's full encoding via a new `cc_project_key()` helper.
- **Correct the host-VM update docs** to the native-installer method (`claude install <version>`); npm conflicts with the native installer on hosts that use it.

## Why

- Keep installs on a current, verified Claude Code release.
- Background-session transcripts are now reliably located by Genesis.

## Verification

- 2.1.170 installed and verified on both reference environments (live `claude -p` round-trip + targeted CC integration tests; a real background session spawned and saved its transcript).
- Transcript-path fix written test-first (TDD): unit tests assert the encoding (incl. the dot case that the old code missed); CC test suite green; `ruff` clean.

### DIFF
--- a/docs/reference/cc-compatibility.md
+++ b/docs/reference/cc-compatibility.md
@@ -15,7 +15,7 @@
 
 ## Current CC Version
 
-**Installed:** Claude Code 2.1.170 (upgraded 2026-06-10 from 2.1.160 — adds Fable 5 access + inherited-env transcript fix)
+**Installed:** Claude Code 2.1.170 — **container + host VM** (both upgraded 2026-06-10 from 2.1.160 — adds Fable 5 access + inherited-env transcript fix). Container via npm (`--prefix /usr/local`); host via native installer (`claude install 2.1.170`).
 **Pin in scripts:** `CC_VERSION=2.1.170` in `scripts/install.sh` and `scripts/host-setup.sh`
 **Minimum required by Genesis:** Not yet formalized (all current code works with 2.0+).
 `requiredMinimumVersion`/`requiredMaximumVersion` (2.1.163) were evaluated as a way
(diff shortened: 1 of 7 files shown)


WingedGuardian:
**Structural review:** 1 WARNING — no ERRORs

---

### WARNING — Missing conventional commit prefix

The PR title `"Claude Code 2.1.170 + background-session transcript path fix"` does not begin with a Genesis conventional prefix (`feat:`, `fix:`, `refactor:`, `docs:`, `test:`, `chore:`). The branch name `chore/cc-2.1.170-upgrade` signals the intent, but the title (used as the squash-commit message on merge) is unprefixed.

Suggestion: `chore: bump Claude Code 2.1.160 → 2.1.170 and fix background-session transcript path` — or split into `chore:` (version bump) + `fix:` (encoding fix) if keeping two commits.

---

Everything else looks solid: `cc_project_key()` is correctly scoped, well-documented, and tested TDD-first (including the leading-dot edge case that the old code missed). Import order clean, no PII in added lines, test-fixture paths correctly use `ubuntu` as a placeholder rather than a real leaked username.

---

PR #596, opened 2026-06-10
feat(voice): upgrade pipecat-ai 0.0.97 → 1.3.0 — SemanticTurnDetection + noise reduction

WingedGuardian:
## Summary

- Upgrade pipecat-ai from 0.0.97 to 1.3.0 (major framework rewrite)
- Replace threshold-based server_vad with AI-driven SemanticTurnDetection
- Add server-side InputAudioNoiseReduction (near_field mode)
- Enable real-time InputAudioTranscription
- Simplify Dockerfile: gcompat for manylinux wheels (eliminates LLVM source build)
- Remove deprecated on_client_message interrupt handler (semantic VAD handles interrupts server-side)

Key improvements:
- Full-duplex conversations: semantic VAD distinguishes user speech from echo/noise
- Near-field noise reduction: OpenAI handles echo cancellation server-side
- Configurable eagerness: low/medium/high via addon config
- Cleaner dependency chain: no LLVM, no source compilation

## Test plan

- [ ] Docker image builds successfully on HA Supervisor
- [ ] Addon starts without import errors
- [ ] Wake word triggers conversation
- [ ] Tool calls (ask_genesis) work
- [ ] Bot speaks, user interrupts mid-speech — verify semantic VAD handles it
- [ ] Session disconnect/reconnect works (60-min clock deferred)
- [ ] No deprecation warnings in logs (or only non-blocking ones)
- [ ] Audio quality: near_field noise reduction noticeable vs previous

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #602, opened 2026-06-10
fix(voice): pipecat 1.3.0 audio path — transport audio flags, stale-socket guard, deploy fixes

WingedGuardian:
## Summary

Completes the pipecat 1.3.0 migration (#596) by fixing the audio-path regressions found during HA deployment. Root causes verified against pipecat source and live addon logs.

- **Enable transport audio**: `audio_in_enabled`/`audio_out_enabled` default to `False` on `TransportParams` (inherited, not deprecated). With them unset, the input audio task is never created and all audio is silently dropped in both directions — the root cause of the post-upgrade silence. This also caused the satellite firmware to give up and reconnect a few seconds into each session (no audio coming back).
- **Stale-socket disconnect guard**: defensive fix — pipecat allows one client and evicts the old socket when a replacement connects; the evicted socket's disconnect event would tear down the live OpenAI session and null the output client connection. Track the active websocket, ignore stale-socket disconnects, and restore the output client connection the transport clears.
- **External turn strategies**: user aggregator now accepts turn signals broadcast by the OpenAI Realtime service (server-side semantic VAD) instead of loading local SmartTurn (onnxruntime unavailable on the addon base image).
- **Conversation setup flag**: set `_llm_needs_conversation_setup` before raw `_connect()` so instructions/tools are re-sent on reconnect.
- **Config wiring**: `run.sh` now exports `SEMANTIC_VAD_EAGERNESS` (was still reading the removed `vad_*` options).
- **Build fixes** (earlier commits): onnxruntime stub for the addon base image, pip-based dependency install.

## Test plan

- [x] ruff + py_compile + bash -n on changed files
- [x] Deployed to HA, new code markers verified in logs
- [x] E2E: Genesis tools loaded (system prompt + 3 tool declarations fetched)
- [x] E2E: wake word → spoken audio response
- [x] E2E: conversation stable for ~60s with mid-conversation tool call (200 OK)
- [x] E2E: context cached on disconnect (7 messages)
- [ ] Interrupt/barge-in: known limitation — satellite firmware blocks mic audio during bot playback (half-duplex). Tracked as Phase 3C, out of scope here.

### DIFF
--- a/src/genesis/channels/voice/s2s_bridge/Dockerfile
+++ b/src/genesis/channels/voice/s2s_bridge/Dockerfile
@@ -3,13 +3,13 @@
 # Override with: --build-arg BUILD_FROM=ghcr.io/home-assistant/aarch64-base:latest
 ARG BUILD_FROM=ghcr.io/home-assistant/aarch64-base:latest
 
-# Build stage - install dependencies with pip using pre-built wheels
+# Build stage - compile native deps then install Python packages
 FROM ${BUILD_FROM} AS builder
 
 # Set shell (use sh for Alpine, bash will be installed)
 SHELL ["/bin/sh", "-o", "pipefail", "-c"]
 
-# Install build dependencies + gcompat (enables manylinux wheel support on Alpine)
+# Install build dependencies + LLVM 20 from Alpine edge (required by llvmlite 0.47.0)
 RUN \
     apk add --no-cache \
         bash \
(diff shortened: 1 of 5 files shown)


WingedGuardian:
## First-pass review — PR #602

### Structural review

**NOTE** — PR title says "LLVM 20 + onnxruntime stub" and the description says "Use LLVM 20 from Alpine edge for llvmlite 0.47.0 build compatibility", but the diff does the opposite: LLVM/gcompat are removed entirely. The Dockerfile comment even reads `# no LLVM — we skip numba/llvmlite, incompatible with Py3.14`. The description appears to be from an earlier iteration of this fix and doesn't reflect what landed. Consider updating the PR description to match the actual approach (pip + onnxruntime stub, no LLVM).

**WARNING** — `create_context_aggregator` in `src/genesis/channels/voice/s2s_bridge/app/session_manager.py` has a behavioral change: it now passes `user_params=LLMUserAggregatorParams(user_turn_strategies=None)` to `LLMContextAggregatorPair`. If tests exist for this function, they may need updating to account for the new parameter.

No structural ERRORs found (no nohup, no task_states direct inserts, no `/tmp/` misuse in Genesis services, no bare excepts, no star imports, conventional commit prefix `fix:` present).

---

PR #606, opened 2026-06-11
fix(install): auto-detect npm prefix instead of hardcoding /usr/local

WingedGuardian:
Cherry-picked from the in-flight `fix/cc-npm-prefix-detection` branch to land
this standalone install fix on its own, separate from the incomplete voice-HA
work on that branch.

## Problem
The `--prefix /usr/local` override assumed `which claude` resolves to
`/usr/local/bin/claude`, but machines with a user-level npm prefix
(`~/.npm-global`) have `claude` there instead — making the `/usr/local`
install invisible to PATH.

## Fix
Check `npm config get prefix` and install to the user prefix when it differs
from the system default, falling back to the explicit `/usr/local` prefix for
system-level installs. Updates `docs/reference/cc-compatibility.md` to match.

Touches only `scripts/install.sh` + the compat doc — no overlap with the
voice work it was previously stacked on. `bash -n` clean.


WingedGuardian:

---

PR #609, opened 2026-06-12
fix(dashboard): support folder uploads, fix drop-hang on directories

WingedGuardian:
## Problem

Dragging a **folder** onto the dashboard Files-tab upload zone hung forever, stuck showing the folder's name. `dataTransfer.files` delivers a dropped directory as a single unreadable `File`, which `FormData`/`fetch` then stalls on — the request never resolves, so the UI freezes.

## Fix

**Frontend** (`genesis_dashboard.html`): dropped folders are now walked via the File and Directory Entries API (`webkitGetAsEntry`/`readEntries`) into their individual files, each posted with a relative path. Drag-drop only; single-file and multi-file uploads are unchanged. Entries are captured synchronously before the first `await` (the `DataTransferItemList` is invalidated once the handler yields), and `readEntries` is looped to handle its ~100-entry pagination.

**Backend** (`routes/files.py`): `file_upload()` honors an optional `relpath` form field and recreates the folder structure under the uploads root. Each path segment is sanitized to be traversal-safe (reusing the existing filename charset filter); validation (containment + blocked-name/allowlist guards) runs **before** any filesystem write, and `mkdir`/`save` are wrapped so a subdir colliding with an existing file — or a write error — returns a clean error instead of a 500.

## Tests

`tests/test_dashboard/test_files_api.py` gains upload coverage: folder-structure preservation, traversal containment (`../../etc` stays inside uploads), absolute-path reinterpretation, dedup within a folder, blocked leaf name, subdir/file collision → clean 400, and the size cap.

## Verification

- 18 backend tests pass (real Flask WSGI dispatch).
- Real-HTTP E2E: a 3-level nested folder recreates exactly under uploads; `../../etc/evil.txt` lands contained with nothing escaping.
- Frontend traversal logic verified via a Node simulation of the exact functions; the page's inline script syntax-checks clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Dropping a folder onto the dashboard uploader hung forever** --- dragging a
+  folder onto the Files-tab upload zone got stuck showing the folder's name and
+  never finished, because folders weren't expanded into their contents. Folder
+  drops now upload every file inside, preserving the folder structure under the
+  uploads directory. Single-file and multi-file uploads are unchanged.
+
 - **Eval quality dashboard could stall** --- the nightly memory-scoring job
   timed out and retried from scratch, so the compounding-intelligence
   dashboard stopped getting fresh data and double-counted judgments. The job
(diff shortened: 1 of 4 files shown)


WingedGuardian (commented):

---

PR #614, opened 2026-06-13
feat(voice): Phase 3B — session rotation + response brevity + settings fix

WingedGuardian:
## Summary

Three independent improvements to the voice channel:

**Phase 3B — 55-min session rotation** (`s2s_bridge/app/main.py`):
OpenAI Realtime sessions expire after 60 minutes with no warning. Adds a background rotation loop that proactively calls `reset_conversation()` at 55 minutes — disconnects, reconnects (new 60-min clock), replays cached conversation context. User hears ~1-2s silence. Timer starts on client connect, cancels on disconnect, guarded by pipeline_lock. Configurable via `SESSION_ROTATION_SECONDS` env var for testing.

**Response length tuning** (`genesis_bridge.py`):
Tighten the tool-result escape hatch from "cover all key points" to "cover 2-3 most important points, then offer to go deeper." Prevents verbose tool-backed responses while preserving opt-in expansion.

**Settings fix** (`mcp/health/settings.py`):
The outreach domain had `dedicated_tool="outreach_preferences"` — a tool that doesn't exist — causing `settings_get("outreach")` to return a redirect instead of the config. Removed the dead reference so `voice_alert_ids` and all outreach settings are now accessible via the generic settings MCP tools.

## Test plan

- [x] ruff + py_compile on all changed files
- [ ] Deploy with short rotation timer (30s), verify rotation log + context preserved
- [ ] Verify `settings_get("outreach")` returns full config including voice.alert_ids
- [ ] Voice conversation: tool-backed answers ≤3 points + "go deeper?" offer

### DIFF
--- a/src/genesis/channels/voice/genesis_bridge.py
+++ b/src/genesis/channels/voice/genesis_bridge.py
@@ -111,8 +111,8 @@ Better to be thorough than to guess wrong.
 
 VOICE RULES:
 - Keep responses to 2-4 sentences for direct answers. When summarizing tool \
-results (memory, search), cover all key points — don't truncate to hit a \
-sentence target.
+results (memory, search), cover the most important 2-3 points. If there's \
+more to share, end with "Want me to go deeper?" and wait for the user.
 - Avoid unnecessary filler — don't restate the question, don't end with \
 "let me know if you need anything." Start with the answer.
 - Never use markdown, bullet points, or formatting. Speak naturally.
--- a/src/genesis/channels/voice/s2s_bridge/app/main.py
+++ b/src/genesis/channels/voice/s2s_bridge/app/main.py
@@ -53,6 +53,12 @@ class Application:
         self.session_manager: SessionManager | None = None
         self.current_task: PipelineTask | None = None
         self._pipeline_lock: asyncio.Lock | None = None
+        self._rotation_task: asyncio.Task | None = None
+        # Session rotation interval in seconds. OpenAI Realtime sessions
+        # expire at 60 minutes — rotate at 55 to leave margin for the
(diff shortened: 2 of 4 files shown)


WingedGuardian:
## First-pass review — PR #614

### Structural review

**No ERRORs.** Two items to note:

---

**WARNING — Missing test coverage for rotation logic** (`s2s_bridge/app/main.py`)

Three new async functions are added — `_session_rotation_loop`, `_start_rotation_timer`, `_cancel_rotation_timer` — but no test changes are included. The rotation logic has meaningful edge cases: what happens if `_pipeline_lock` is None, if `reset_conversation()` raises, or if `_cancel_rotation_timer` races with a mid-sleep task. The PR test plan covers deployment-level verification but unit tests for the cancellation and lock-guard paths would be valuable here.

---

**NOTE — Failed rotation retries after a full 55-min sleep** (`s2s_bridge/app/main.py`, `_session_rotation_loop`)

On `reset_conversation()` failure the exception is caught and the loop continues — sleeping for another full `_rotation_interval` (55 min) before the next attempt. If the failure occurs at the 55-min mark, the next retry lands at 110 min, which is 50 min past OpenAI's 60-min hard expiry. A shorter fallback interval on failure (e.g., 5 min) would recover before the session dies:

```python
except Exception as e:
    logger.warning("⚠️ Session rotation failed: %s — retrying in 5 min", e)
    await asyncio.sleep(300)  # short retry, then loop restarts full sleep
    continue
```

Not a hard block — the current behaviour avoids a tight retry storm — but worth a conscious decision either way.

---

**CLEAN** — no secrets, private IPs, personal data, or sensitive wording in diff or PR metadata.

---

PR #615, opened 2026-06-13
fix(voice): consistent context cache keying (Bug 5)

WingedGuardian:
## Summary

Pipeline builds once with `client_id="server"`, registering context aggregators under that key. But connect/disconnect handlers used the client's IP address from `extract_client_id()`, causing a key mismatch — the aggregator was never found, context caching fell through to `service._context` fallback, and the `No context aggregator found` warning logged on every disconnect.

Use `"server"` as the logical client_id for all SessionManager operations. Pipecat only allows one active client, so per-client keying was unused. The client IP is still logged for observability.

Closes follow-up 77941ede (identified during Phase 3A due diligence, Jun 11). Prerequisite for Phase 3D (voice conversation memory).

## Test plan

- [x] ruff + py_compile
- [ ] Deploy to HA, verify `Found context in aggregator` replaces `No context aggregator found` warning
- [ ] Disconnect + reconnect within 5 min → context restored with message count

### DIFF
--- a/src/genesis/channels/voice/s2s_bridge/app/websocket_handler.py
+++ b/src/genesis/channels/voice/s2s_bridge/app/websocket_handler.py
@@ -273,15 +273,18 @@ class WebSocketHandler:
                 and self._active_websocket is not websocket
             )
             self._active_websocket = websocket
-            client_id = self.extract_client_id(websocket)
-            logger.info(f"🔗 New WebSocket connection from IP: {client_id}")
+            client_ip = self.extract_client_id(websocket)
+            logger.info(f"🔗 New WebSocket connection from IP: {client_ip}")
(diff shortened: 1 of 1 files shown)


WingedGuardian (commented):

**NOTE:** The `if client_id:` guard was removed from the disconnect handler, so `on_client_disconnected_callback("server")` now fires unconditionally regardless of what `extract_client_id()` returns. The intent is correct (session cleanup should always run), but if `extract_client_id()` returns `None` the log line will read `🔌 Client None disconnected`. Consider `client_ip or "unknown"` in the format string to keep logs clean.

PII/Secrets/Wording scan: CLEAN — no credentials, private IPs, personal data, or infrastructure-revealing wording found.

---

PR #619, opened 2026-06-13
fix(cc): make the session status header fire reliably

WingedGuardian:
## Problem
The first-reply status header `[model version / effort]` (e.g. `[Opus 4.8 / high]`) is specified in `CONVERSATION.md` → *Session Start*, but it fired unreliably. That spec sits hundreds of lines deep in the injected identity block, so it got buried under the user's first message and was frequently not emitted. The high-salience top-of-injection `## Session Configuration` block carried only `effort:` and never mentioned the header.

## Fix
Echo the header directive into the top `## Session Configuration` block (the first thing the model reads in the SessionStart injection), so the trigger is co-located with the effort value at maximum salience.

Also:
- Reconcile `CONVERSATION.md`: the model is derived from the environment's "You are powered by the model named …" line (not the config block, which never carried it); emit on every session start (fresh, resume, or post-compact).
- Add a regression test asserting the emitted block contains the header directive, so this can't silently regress again.

## Testing
`pytest tests/test_cc/test_session_context.py tests/test_scripts/test_session_context_tier.py -v` → 24 passed (includes the new `test_session_config_block_carries_header_directive`). Ruff clean. No programmatic consumer parses the block (it is LLM-facing only), so the format change is safe.

### DIFF
--- a/scripts/genesis_session_context.py
+++ b/scripts/genesis_session_context.py
@@ -109,14 +109,17 @@ def main() -> None:
         _SESSION_START_FILE.parent.mkdir(parents=True, exist_ok=True)
         _SESSION_START_FILE.write_text(datetime.now(UTC).isoformat())
 
-        # 0.5. Session Configuration — inject effort level for the LLM's
-        # status header. Model is NOT injected here because the LLM can
-        # derive it from CC's "You are powered by..." system text, which
-        # is always accurate. The sidecar file goes stale when CC's native
-        # /model command is used (it doesn't call Genesis MCP tools).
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

WingedGuardian:

Injects the first-reply `[<model> / <effort>]` status-header directive into the high-salience top of the session-context hook (the deep CONVERSATION.md spec fired unreliably); model derived from CC's always-accurate system text, effort from the session_config sidecar. Regression test added.

---
_Generated by Claude Code_

PR #620, opened 2026-06-14
fix(voice): surface silent HA delivery failures + drop device-specific entity defaults

WingedGuardian:
## What

Harden the proactive voice path so silent delivery failures surface, and remove install-specific entity defaults that break silently.

## Why

A proactive voice chime was silently dead for an unknown period: Home Assistant accepted the service call (HTTP 200) but changed **zero** states because the configured `media_player` entity didn't exist (a device firmware reflash had renamed it). `VoiceChannelAdapter.send_message` swallowed this and returned a delivery id as if it succeeded — so nothing surfaced, and an automated E2E falsely passed. The runtime was fixed via config (`HA_MEDIA_PLAYER_ENTITY`); this PR fixes the **code** so the failure class can't hide again.

## Changes

- **Surface silent failures** (`channels/voice/adapter.py`): `_tts_speak` now warns when HA returns an empty changed-states list (no entity matched → no audio) instead of logging a false "delivered". The pre-announce chime checks HTTP status (`play_media` returns `[]` even on success, so status is its only reliable signal there).
- **No device-specific defaults**: drop the hardcoded `media_player` entity default; require `HA_MEDIA_PLAYER_ENTITY`, warn + skip delivery when unset (tts keeps the generic `tts.piper` default).
- **Shutdown notice** (`hosting/standalone.py`): switch from `assist_satellite.announce` (that satellite entity is often `unavailable`, so the call silently no-op'd) to the held `VoiceChannelAdapter` (media_player path), inheriting the working entity config + failure-surfacing.

## Testing

- `tests/test_channels/test_voice.py`: reconciled the two cases that assumed the old hardcoded default; added coverage for the no-media_player skip and the empty-changed-states warning. **39 voice + 31 hosting + 6 shutdown tests pass; ruff clean.**
- Live E2E against real HA: positive path delivers (no "changed no states" warning = states actually changed, audibly confirmed); negative path (no entity) skips with zero HA calls + warning.

### DIFF
--- a/src/genesis/channels/voice/adapter.py
+++ b/src/genesis/channels/voice/adapter.py
@@ -64,10 +64,18 @@ class VoiceChannelAdapter(ChannelAdapter):
         self._tts_entity = tts_entity or os.environ.get(
             "HA_TTS_ENTITY", "tts.piper",
         )
+        # No device-specific default — entity IDs are install-specific and a
+        # hardcoded default silently breaks when the device is renamed (a
+        # firmware reflash added a "_media_player" suffix on 2026-06-13,
+        # silently killing the chime). Require explicit config.
         self._media_player = media_player_entity or os.environ.get(
-            "HA_MEDIA_PLAYER_ENTITY",
-            "media_player.home_assistant_voice_0a2841",
+            "HA_MEDIA_PLAYER_ENTITY", "",
         )
+        if not self._media_player:
+            logger.warning(
+                "Voice adapter: HA_MEDIA_PLAYER_ENTITY is not set — proactive "
+                "voice (chime/TTS) is disabled until it is configured."
+            )
 
     async def start(self) -> None:
         """No-op — adapter is stateless, HA handles transport."""
@@ -99,6 +107,12 @@ class VoiceChannelAdapter(ChannelAdapter):
         if not self._ha_url or not self._ha_token:
             logger.warning("Voice adapter: HA not configured, skipping TTS")
             return ""
+        if not self._media_player:
+            logger.warning(
+                "Voice adapter: no media_player entity (set HA_MEDIA_PLAYER_ENTITY) "
+                "— skipping proactive voice delivery"
+            )
+            return ""
 
         delivery_id = str(uuid.uuid4())
         headers = {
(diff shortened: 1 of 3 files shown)


WingedGuardian:
## First-pass review — PR #620

### Structural review

**WARNING — Test coverage gap in `standalone.py`**

The shutdown path in `src/genesis/hosting/standalone.py` changed significantly: direct `httpx` calls to `assist_satellite.announce` were replaced by `voice_adapter.send_message()`. No test file changes for `standalone.py` appear in this diff.

The PR states "31 hosting + 6 shutdown tests pass", but if those tests were mocking `httpx.AsyncClient` at the standalone level (the old pattern), they would still pass while silently not exercising the new delegation path at all. Worth confirming the existing shutdown tests mock `voice_adapter.send_message` (or the adapter's underlying httpx calls), not the old direct call site that no longer exists.

Everything else is clean:
- No bare `except:` clauses — all uses are `except Exception:` ✓
- No mutable default arguments ✓
- No raw SQL, `/tmp/`, `nohup`, `task_states` inserts, or force-push patterns ✓
- Import hygiene: the inline `import httpx` inside the old shutdown block was actually removed — net improvement ✓
- Conventional commit prefix `fix(voice):` ✓
- New test cases in `tests/test_channels/test_voice.py` cover the adapter changes well ✓

---

**CLEAN**

- No API keys, tokens, private IPs, SSH keys, or connection strings in added lines ✓
- `home_assistant_voice_0a2841` and `assist_satellite.home_assistant_voice_0a2841_assist_satellite` appear **only in removed lines** — this PR is actively cleaning up install-specific identifiers, which is correct ✓
- Test fixtures use generic placeholders (`media_player.test`, `media_player.bogus`, `http://ha.local:8123`) ✓
- PR title and body do not use alarming language about leaks, credential exposure, or security incidents — "drop device-specific entity defaults" is accurate and appropriately neutral ✓

---

WingedGuardian:
Addressed in `466d6393` + `6c740380`:
- **WARNING (untested shutdown delegation):** confirmed your read — the existing 6 'shutdown tests' cover `GenesisRuntime.ashutdown()`, a different method, so the standalone delegation had zero coverage. Extracted the shutdown notice into `StandaloneAdapter._voice_last_breath()` and added isolated unit tests (adapter present → `send_message` awaited; no adapter / no app → no-op).
- Also from first-pass review: exercised the tts success branch in `test_send_message_with_chime` (was bypassed by a bare MagicMock) and removed a byte-equivalent duplicate test.

Will also live-verify the spoken shutdown notice on the next restart at deploy.

PR #624, opened 2026-06-14
fix(runtime): clear stale job_health failure state on recovery (WS-3b)

WingedGuardian:
## WS-3b · Clear stale `job_health` failure state on recovery

Bug fix from the 2026-06-13 audit (`sched-002`).

### The bug
`job_health`'s UPSERT (`runtime/_job_health.py`) used `COALESCE(excluded.last_failure, last_failure)` (and the same for `last_error`). But `record_job_success` and `clear_stale_job_failures` deliberately carry **NULL** in those columns to signal recovery, so COALESCE kept the *old* failure timestamp forever. A job that failed once showed a **permanent failure** in the health view even after it recovered.

### The fix
Drop `COALESCE` on `last_failure`/`last_error` (keep it on `last_success`, which the failure path legitimately omits). `total_failures` already keys off `excluded.last_failure IS NOT NULL`, so the counter stays correct.

**Caller audit (all 3 paths):** `record_job_success` and `clear_stale_job_failures` clear (carry NULL); `record_job_failure` always supplies both. `_persist_job_start` uses a *separate* INSERT that never touches the failure columns, so it is unaffected.

### Tests
First real-DB (aiosqlite) coverage of this UPSERT — previously only exercised with a mocked `_db`:
- `test_success_clears_persisted_failure` — fail→success clears both columns.
- `test_failure_path_unbroken_and_success_preserved` — failure still persists; prior `last_success` survives.
- `test_clear_stale_failures_clears_persisted_failure` — the bootstrap stale-failure sweep also clears.

Verified via `git stash`: both clearing tests fail on the pre-fix code (RED) and all three pass with the fix (GREEN). Regression suites green: `test_runtime`, job_health collector, `pause_guard`, operational-vitals e2e (50 + 13 tests).


WingedGuardian (commented):

---

PR #625, opened 2026-06-14
fix(modules): shell-quote remote CC dispatch params to stop injection (WS-1)

WingedGuardian:
## WS-1 · Stop SSH command injection in remote CC dispatch

Security fix from the 2026-06-13 audit (`R2-001`).

### The vulnerability
`SshIPCAdapter._send_cc` built the remote command by string-interpolating caller-supplied `model`/`effort` (and config-supplied `remote_working_dir`/`remote_claude_path`) into a `cd … && claude …` string. That string is handed to `ssh` and **re-parsed by the remote shell**, so a crafted `model`/`effort` could execute arbitrary commands on the remote host (career-ops). The `prompt` was already safe (piped via stdin).

### The fix
- Extracted **`_build_remote_command()`** that `shlex.quote`s every interpolated value (`model`, `effort`, `remote_working_dir`, `remote_claude_path`). Valid enum values (`sonnet`, `high`, normal paths) quote to themselves, so legitimate dispatch is **byte-identical** to before.
- **Lenient enum validation:** warn on an unrecognized `model`/`effort` but still pass the quoted value through, so full model ids (`claude-sonnet-4-6`) or future tiers don't break callers. `shlex.quote` neutralizes injection regardless, so nothing is lost on the security side.
- **Closed the same injection class in `health_check`** (`remote_claude_path` was interpolated unquoted into a `--version` call) — found during review.
- Documented `_send_shell`'s contract (it forwards raw shell commands by design; callers must pre-quote).

### Tests (`TestSshIPCAdapterCommandInjection`)
Behavioral: after `shlex.split` (what the remote shell does), a hostile value must remain a **single token**. Covers `;`, `$()`, backticks in `model`/`effort`, a malicious working dir, the `health_check` path, plus an end-to-end check that the quoted command reaches the actual `ssh` argv. Benign values verified unchanged. Full `test_ssh_ipc.py` suite (24) green; external-adapter + module-registry suites (39) green; no circular import from the new `cc.types` import.

### Notes
- Threat-model context: this path is reachable via internal dispatch, so live exploitability is bounded, but it's a genuine injection in publicly-shipped code — fixed on its own merit (audit decision D1/D13 keep this separate from the moot-on-private items).


WingedGuardian (commented):
## First-pass review — PR #625

### Structural review: CLEAN

No pattern violations found across all three changed files.

| Check | Result |
|---|---|
| Import order (`ipc.py`) | ✓ `shlex` in stdlib block; `genesis.cc.types` in local block |
| Docstrings | ✓ `_build_remote_command()` documented; `_send_shell()` contract clarified |
| Test coverage | ✓ `test_ssh_ipc.py` updated with injection-specific test class (6 cases) |
| Bare `except:` / mutable defaults / raw SQL / `/tmp/` / `nohup` | ✓ None found |
| Conventional commit prefix | ✓ `fix(modules):` |

---

### PII / Secrets / Wording scan: CLEAN (1 soft warning)

**Added-line scan:** No API keys, tokens, private IPs, SSH private keys, or credentials found. Test-fixture paths (`/home/user/project`, `user@host`) are known false positives per policy.

**CHANGELOG wording:** Generic and appropriate — describes the fix class without naming specific hosts, IPs, or credentials. No wording issue.

**SOFT WARNING — internal codename in PR body:**
> "so a crafted `model`/`effort` could execute arbitrary commands on the remote host **(career-ops)**"

`career-ops` reads as an internal module or service name. If it's a private service identifier, consider replacing with a generic term ("the remote host") in the PR description. This is a PR body note only — not in shipped code — and does not block merge.

---

PII/Secrets/Wording scan: **CLEAN** — 1 soft warning (internal codename in PR body, no action required).

---

PR #626, opened 2026-06-14
fix(routing): make circuit-breaker state durable across restarts (WS-3c)

WingedGuardian:
## WS-3c · Make circuit-breaker state durable across restarts

From the 2026-06-13 audit (`ROUTE-07` / `CONC-02`), plus a related bug found while writing the tests.

### Three defects in breaker-state persistence
1. **Non-atomic write.** `save_state` used `Path.write_text` (truncate-in-place), so a concurrent reader could see a partial file — and `load_state` silently discards a malformed file, losing **all** breaker state.
2. **MCP children clobbered the shared file.** Every MCP child process built the registry against the same default state file and wrote to it on each breaker transition, racing the server.
3. **OPEN never survived a restart (found in testing).** `load_state` compared `saved_state == "OPEN"`, but `save_state` writes the `StrEnum` value `"open"` (lowercase). So a provider that tripped came back **CLOSED** on every restart and got retried immediately instead of serving its backoff — defeating the whole point of persistence. A pre-existing test hid this with an unrealistic uppercase `"OPEN"` fixture.

### The fix
- New **`genesis/util/atomic.py` `atomic_write_text()`** (temp file + `os.replace` + fsync, cleanup on failure); `save_state` uses it.
- **`CircuitBreakerRegistry(persist=...)`**: `persist=False` loads state at construction but never writes (both `on_state_change=None` and a `save_state` early-return). `standalone.py` (MCP children) passes `persist=False`; the server keeps the default, so its behavior is byte-identical.
- **Casing fix**: `load_state` now compares `ProviderState.OPEN.value`; the pre-existing test fixture corrected to the real lowercase `"open"`.

### Tests
`tests/test_util/test_atomic.py` (content / overwrite / no temp leftovers / parent-dir creation) and breaker tests for: persist-on-trip, read-only registry never writes, **read-only still loads** the server's state, and a direct **OPEN-restore regression**. Full `tests/test_routing` (255) + util green. Independent code review: all three fixes confirmed correct, no high-confidence issues.

### Notes
- The audit found this via the non-atomic write; the restart-survival bug (#3) was a latent defect in the load path uncovered during TDD and folded in (it's the read half of the same durability mechanism).


WingedGuardian (commented):

---

PR #582, opened 2026-06-09
fix(routing): add 120s default timeout to litellm calls + fix watchdog messaging

WingedGuardian:
## Summary

- **LLM timeout**: Adds explicit `timeout=120` to all `litellm.acompletion()` calls via `LiteLLMDelegate.call()`. Prevents indefinite hangs when a provider accepts TCP but stalls on response (production incident: deepseek-v4-pro via OpenRouter hung for 20 minutes on 2026-06-08, blocking the surplus dispatch loop until the watchdog killed the server). Callers can override via kwargs.
- **Watchdog messaging**: Changes "consecutive failures" to "restarts" across dashboard, neural monitor, and health alerts — the watchdog detecting and restarting a broken server is a success, not a failure. Dashboard degraded threshold raised from >0 to >3 (matching existing alert threshold).

## Changes

| File | Change |
|------|--------|
| `src/genesis/routing/litellm_delegate.py` | `_DEFAULT_TIMEOUT_S = 120`, injected into `call_kwargs` |
| `src/genesis/mcp/health/errors.py` | Alert text: "failures" → "restarts" |
| `src/genesis/dashboard/templates/genesis_dashboard.html` | Threshold >0→>3, wording + reason display |
| `src/genesis/dashboard/templates/neural_monitor.html` | "WD Failures" → "WD Restarts" with reason |
| `az_plugins/genesis/templates/neural_monitor.html` | Same fix for AZ plugins copy |
| `tests/test_routing/test_litellm_delegate.py` | 2 new tests: default timeout applied, caller override works |

## Test plan

- [x] `pytest tests/test_routing/test_litellm_delegate.py` — 27/27 pass (25 existing + 2 new)
- [x] `ruff check` — clean
- [ ] CI green
- [ ] Dashboard shows "WD Restarts" with reason, not "fails"
- [ ] Automated review

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #627, opened 2026-06-14
fix(routing): fail fast on provider timeouts + fallback chains for dream cycle & reflections

WingedGuardian:
## Problem

LLM operations were dead-lettering with "All providers exhausted" and the dream
cycle / reflections were stalling. Root cause is a **two-layer retry stack on a
hung provider**:

1. A provider that hangs (accepts the connection but never responds) returns a
   timeout that `classify_error` treated as `TRANSIENT`.
2. `Router._try_with_retry` retried that **same hung provider** up to
   `max_retries + 1` times — `5` for the `background` profile.
3. litellm's own internal retries stacked on top of the per-attempt timeout
   (the 120s default added in #582 was not being honored — ~361s observed on a
   120s timeout).

Net: up to ~30 minutes burned on one hung provider before the router even
reached the next provider in the chain. Several chains had no working fallback
left, so they exhausted and dead-lettered.

## Fix

**Fail fast on timeouts (keystone):**
- New `ErrorCategory.TIMEOUT`. `classify_error` maps HTTP 408 and "timeout"
  messages to it (`connection` errors stay `TRANSIENT`).
- `Router._try_with_retry` returns immediately on `TIMEOUT` — no same-provider
  retry. `route_call` advances to the next provider and the **circuit breaker
  still records the failure**, so a repeatedly-hanging provider trips OPEN.
- `litellm_delegate`: wrap `acompletion` in a hard `asyncio.wait_for` cap and
  set `num_retries=0`, so a single attempt cannot exceed the timeout. Completes
  the 120s ceiling from #582.

**Fallback chains (`config/model_routing.yaml`):**
- `dream_cycle_synthesis_challenge`: paid Kimi via OpenRouter (`kimi-k2.5`) as a
  last resort — same model family as the free NIM primary, preserving the
  Kimi-vs-DeepSeek independence. Billed only during a free-tier outage.
- `3_micro_reflection` and `dream_cycle_entity_check`: more free providers
  (non-DeepSeek for entity_check to preserve independence from the entity
  challenge). No paid creep.
- `4_light_reflection` intentionally unchanged (routes via the Claude Code
  background process; keeps `never_pays`).

## Testing

- New unit tests: TIMEOUT classification (408 + message), fail-fast no
  same-provider retry, hard-timeout cancellation of a hung call.
- New **end-to-end integration test**: real `LiteLLMDelegate` + `Router` with a
  hung provider — verifies the call is capped at the timeout and the router
  fails over to the next provider (not the full hang, not 5x).
- Full routing suite (250 tests) green; config loads via the real loader with
  all appended providers resolving and keyed.

## Notes

- Cost impact is bounded: only `dream_cycle_synthesis_challenge` adds a paid
  provider, and only when both free tiers are down.
- Follow-ups (non-blocking): suppress cosmetic dead letters logged on the
  CC-backstopped API attempt in dual dispatch; watch for idle-connection
  accumulation on providers that hang frequently (cancellation cleanup).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,16 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   failure (was 100% only), so you hear about pressure and degradation
   sooner, on both Telegram and voice.
 
+### Fixed
+
+- **Provider hangs no longer stall reflections and the dream cycle** — when a
+  model provider hangs (accepts the connection but never responds), Genesis
+  now fails over to the next provider within its timeout instead of blocking
+  for minutes. Reflections and the nightly dream cycle stop piling up
+  dead-lettered work during provider outages, and adversarial review and
+  reflections keep running when free-tier providers are down (extra free
+  fallbacks added, plus a paid last-resort for the dream-cycle challenge).
+
 ---
 
 ## [v3.0b15] - 2026-06-12
--- a/config/model_routing.yaml
+++ b/config/model_routing.yaml
@@ -241,6 +241,14 @@ call_sites:
     - mistral-large-free
     - groq-free
     - gemini-free
+    # Additional free providers so an essential reflection (3_micro_reflection
+    # is in degradation._L3_KEEP) survives a broad free-tier outage without
+    # paid fallback. Healthy OpenRouter-free models first; NIM last (shared
+    # 40 RPM pool, currently flaky).
+    - openrouter-gemma4
+    - openrouter-trinity-free
+    - nvidia-nim-kimi
+    - nvidia-nim-deepseek
     retry_profile: background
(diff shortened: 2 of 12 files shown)


WingedGuardian:

---

WingedGuardian:
Review note: the cloud Codex reviewer was quota-limited (no findings posted), and local Codex/OpenCode were unavailable (account model restriction / no credits). A thorough local adversarial code review was run via the Claude code-reviewer agent: **0 critical (P1) findings**, 4 non-blocking advisories (exception-ordering confirmed correct; httpx idle-connection watch-item; budget-gated paid fallback consistent with existing dream_cycle_synthesis; test wall-clock now covered by the new e2e integration test). Structural review: CLEAN. CI: all green.

PR #631, opened 2026-06-14
fix(observability): make dropped events visible + durable (WS-17)

WingedGuardian:
## What & why

Event-bus persistence drops were **silent**. When the write queue (`maxsize=500`) filled during an incident storm, events were dropped with only a log line — so the dashboard unified-errors and `health_errors` MCP (which read the `events` table) **under-counted exactly when things were worst** (EH-03). Separately, one un-serializable `details` value aborted a whole batch insert (OBS-002).

## Changes
- **Visible, non-blocking drop counter** — `emit()` increments a cumulative counter on `QueueFull` (never blocks); surfaced on the health snapshot as `queues.events_dropped`.
- **Overflow meta-event** — the background writer task emits a rate-limited (60s) `event_queue_overflow` WARNING row via `events_crud.insert`, so the loss shows up in the same error surfaces that previously hid it. Emitted only from the writer (which owns the DB handle), never from the hot `emit()` path.
- **Bigger buffer** — queue `maxsize 500 → 5000` to absorb bursts before dropping.
- **OBS-002** — `insert`/`insert_batch` serialize via `json.dumps(default=str)` + a marker fallback, so a single un-serializable event degrades gracefully instead of dropping the whole batch.
- Threads the existing `rt._event_bus` into `HealthDataService` → `queues()` (init order verified: observability initializes before health-data).

**No blocking backpressure** (deliberate, per D14 intent + the project's anti-stall philosophy): `emit()` never awaits on a full queue — a slow DB must not stall hot cognitive paths. D14's "make the loss visible" is satisfied by the counter + meta-event.

## Tests
`tests/test_observability/` — 143 passed (incl. new: drop-counter non-blocking, queue maxsize, rate-limited overflow notice via in-memory DB, OBS-002 batch survival, snapshot surfacing). ruff clean. Independent code-reviewer pass: clean.

WS-17 / D14 + OBS-002 from the 2026-06-13 audit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   casing mismatch), so a failing provider got retried immediately instead of
   serving out its backoff. Breaker state is now also written atomically, and
   MCP helper processes no longer overwrite the shared state file.
+- **The error log no longer silently under-counts during incident storms.**
+  When the event-persistence queue filled up, events were dropped without a
+  trace — so the dashboard and health views under-reported errors exactly when
+  things were worst. Dropped events are now counted and made visible (an
+  "event queue overflow" warning in the same error views, plus a live counter
+  on the health snapshot), the buffer is 10× larger (500 → 5000) to absorb
+  bursts, and a single un-serializable event can no longer drop a whole batch.
 
 ### Security
 
--- a/src/genesis/db/crud/events.py
+++ b/src/genesis/db/crud/events.py
@@ -10,6 +10,26 @@ from datetime import UTC, datetime
 import aiosqlite
 
 
+def _safe_details_json(details: dict | None) -> str | None:
+    """Serialize event details to JSON, degrading gracefully (OBS-002).
+
+    ``default=str`` lets common non-JSON values (datetimes, Paths, enums)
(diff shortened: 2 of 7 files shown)


WingedGuardian (commented):

---

**Structural (7 files reviewed)**

- No `nohup`, no raw `task_states` inserts, no `/tmp/`, no `git add -A`, no force-push/reset, no bare `except:`, no mutable defaults, no `os.killpg`. All DB writes go through `events_crud` CRUD module.
- Import placement: new `GenesisEventBus` imports correctly behind `TYPE_CHECKING`. Deferred `from genesis.db.crud import events as events_crud` inside `_flush_dropped_notice` follows the pre-existing circular-import avoidance pattern already documented in this file. No star imports.
- Coverage: all three new behaviours (drop counter, `_flush_dropped_notice`, OBS-002 serialization fallback) have direct tests in `test_events.py`.
- Commit prefix `fix(observability):` — correct.

**One note (not a blocker):** `queues()` in `snapshots/queues.py` gains a new `event_bus` parameter; the diff doesn't show the existing docstring being updated to mention it. If there is one, worth a line.

**PII/Secrets/Wording**

- Added lines: no credentials, private IPs, SSH keys, or secret env values.
- PR title, body, and CHANGELOG entry: clean — no "fixed leak / removed credentials / infrastructure detail" language.

---

PR #637, opened 2026-06-15
fix(outreach): gate forum routing to telegram; run SMTP off event loop

WingedGuardian:
## Problem

Outreach on non-Telegram channels (email, Discord, voice) had its recipient overwritten with the Telegram forum chat id whenever a category routed to the supergroup (the default routing). Email sends were therefore addressed to a Telegram chat id, which the SMTP server rejects as an invalid address, so every such send failed and piled up in the deferred-work queue.

Compounding this, `EmailAdapter.send_message` ran synchronous `smtplib` directly inside an `async` function, so each (failing) send blocked the event loop. With the recovery worker retrying the queued failures, this starved heartbeats, health checks, and the awareness loop.

The routing override has been latent since the 2026-03-29 release; it only surfaced once email outreach for a supergroup-routed category became active.

## Fix

1. **`outreach/pipeline.py`** — gate the Telegram supergroup/forum routing (thread-id resolution + `forum_chat_id` override) on `channel == "telegram"`. Other channels now deliver to their own recipient with no thread id. Telegram supergroup/dm/both routing is unchanged.
2. **`channels/email_adapter.py`** — run the blocking `smtplib` send via `asyncio.to_thread` so a slow or rejected send can't freeze the event loop.

## Tests

- `test_pipeline.py`: email recipient is not clobbered by supergroup routing (RED→GREEN) + a Telegram supergroup-routing regression guard.
- `test_email_adapter.py`: a slow SMTP send no longer blocks the event loop (RED→GREEN).
- Full outreach + channels suites pass (382 tests); a live end-to-end send verified the correct recipient under the exact bug condition.

## Review

Independent code review passed with no findings. The external second-opinion reviewer was unavailable (usage limit) and is noted as quota-limited.

## Out of scope (follow-ups)

- Two pre-existing, unrelated test issues surfaced: a rare order-dependent flake in `test_pipeline.py` under random ordering, and `test_load_autonomous_cli_policy_from_yaml` reading local config. Tracked separately.
- Awareness-tick / session-observer lock scoping and the judge-call-site alerting are separate follow-up changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Outreach emails actually send now** — email (and Discord/voice) outreach was
+  being misaddressed to the Telegram forum chat for any category that routes to
+  the supergroup, so every such send failed and silently piled up as retries.
+  Forum/topic routing is now correctly Telegram-only; other channels deliver to
+  their own recipient.
+- **A slow or failed email can no longer stall Genesis** — SMTP sending now runs
+  off the event loop, so a hung or rejected send no longer freezes heartbeats,
+  health checks, or the awareness loop.
 - **Provider hangs no longer stall reflections and the dream cycle** — when a
   model provider hangs (accepts the connection but never responds), Genesis
   now fails over to the next provider within its timeout instead of blocking
--- a/src/genesis/channels/email_adapter.py
+++ b/src/genesis/channels/email_adapter.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import email.message
 import logging
 import smtplib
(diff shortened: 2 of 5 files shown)


WingedGuardian (commented):
**First-pass review — Structural + PII/Secrets/Wording**

No ERRORs, no hard blocks, no wording issues. Two findings below.

---

### NOTE — `tests/test_channels/test_email_adapter.py`

The new test `test_send_message_does_not_block_event_loop` is decorated with `@pytest.mark.asyncio`, while every adjacent test in the same class uses `@pytest.mark.anyio`. Mixed markers in the same class can cause the asyncio-marked test to be silently skipped or fail with an anyio-configured runner (depending on `asyncio_mode` and whether both plugins are active).

Recommend switching the new test to `@pytest.mark.anyio` for consistency with the rest of the class.

---

### SOFT WARNING — Real Telegram supergroup ID in test fixtures

The value `-1003741378738` appears hardcoded in:
- `tests/test_outreach/test_pipeline.py` — `pipeline.set_forum_chat_id(-1003741378738)` (two fixtures) and in the docstring of `test_email_delivery_uses_email_recipient_not_forum_chat_id`
- `tests/test_channels/test_email_adapter.py` — referenced in the docstring of the new event-loop test

If this is the real production Telegram supergroup ID from the 2026-06-14 incident, it embeds a private infrastructure identifier in the public test suite. Consider replacing with a clearly fictional placeholder (e.g., `-1001234567890`) so the tests document the bug class without leaking the live group identity. The functional coverage is identical either way.

---

PII/Secrets/Wording scan: 1 soft warning (see above); no hard blocks, no wording issues.

---

PR #639, opened 2026-06-15
feat(cc): unified host+container Claude Code version management (WS-16)

WingedGuardian:
## What & why (WS-16, audit remediation Phase 2 — final item)

Until now `update.sh` only updated the **container's** Claude Code; the **host VM's** CC (used by the Guardian) drifted behind silently — it sat at **2.1.87** while the container ran **2.1.173**. This makes CC version management **unified across both machines** from a single pin.

### Changes
- **`scripts/lib/cc_version.sh`** (new) — single source of truth for `CC_VERSION`, sourced by `install.sh`, `host-setup.sh`, and `update.sh` (replaces the duplicated `2.1.173` literals). Bump in one place.
- **`scripts/guardian-gateway.sh`** — new `update-cc <semver>` op: validates a bare, anchored semver (`^[0-9]+\.[0-9]+\.[0-9]+$`, mirroring the existing `redeploy` hash check), installs `@anthropic-ai/claude-code@<ver>` via the npm that owns the in-use `claude` under `sudo -n`, then verifies `claude --version`. Package name hardcoded; arg never reaches a shell.
- **`scripts/update.sh`** — after the Guardian redeploy, queries the host CC version and dispatches `update-cc` **only on drift** — idempotent, non-fatal, graceful with an old gateway that lacks the op.
- **`docs/reference/cc-compatibility.md`** — corrected + the deliberate design decision recorded (see below).
- **Tests** — `tests/test_scripts/test_cc_version_gateway.py` runs the **real** gateway with stubbed `claude`/`npm`/`sudo`: proves the accept path installs the exact pinned package, and that 11 malformed/injection args are rejected before npm runs.

### Design decision: "updater only, NO managed-settings floor"
A `requiredMinimumVersion` floor was evaluated and **deliberately rejected** — a hard floor removes the incident-recovery downgrade path (the project has used 2.1.90→2.1.87 before) and can brick CC. Governance is the npm pin + this unified updater + the existing `DISABLE_UPDATES`.

### Note for reviewers — a stale-doc correction
An automated review flagged that the doc said the host uses CC's *native installer* (do-not-npm). **Live inspection disproved that:** `npm ls -g` shows `@anthropic-ai/claude-code@2.1.87` at the `/usr` prefix, `claude` is `/usr/bin/claude`, and there is no `~/.local/share/claude/versions/` tree. The host is **npm-global**; the doc was stale (it had migrated since the "2026-06-10" note). This PR corrects that section, so the npm-based op is the right mechanism.

### Verification
- `bash -n` + `shellcheck` clean on all modified scripts (no new findings; the one pre-existing SC2015 is untouched).
- 13 gateway tests + full `test_scripts` suite green; `ruff` clean.
- Merging this is **inert on the live host** (`update.sh` is on no timer). The live host bump 2.1.87→2.1.173 is a deliberate follow-up step.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Changed
 
+- **`update.sh` now keeps Claude Code in sync on your host VM too** — if you run
+  Genesis with a Guardian on a separate host VM, updates previously only touched
+  the container's Claude Code, letting the host drift behind. `update.sh` now
+  checks the host's version against a single pin (`scripts/lib/cc_version.sh`)
+  and updates the host to match when it has drifted, so container and host never
+  fall out of step. It's skipped when already in sync and never fails your update
+  if the host is unreachable.
 - **Voice: you now choose exactly which alerts are spoken aloud** — the
   Voice PE only speaks alerts on an allowlist you control (`voice.alert_ids`
   in `outreach.yaml`) instead of chiming for every blocker, alert, and
--- a/docs/reference/cc-compatibility.md
+++ b/docs/reference/cc-compatibility.md
@@ -15,14 +15,37 @@
 
 ## Current CC Version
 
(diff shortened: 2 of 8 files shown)


WingedGuardian (commented):

---

PR #640, opened 2026-06-15
perf(health): fix /api/genesis/health hang (call_sites N+1, parallelize, bridge timeout)

WingedGuardian:
## Problem

`GET /api/genesis/health` was taking 9-20s and hanging the dashboard (spinning forever) and the host Guardian's health probe. Root cause was found by **instrumenting `snapshot()` and measuring** (not guessing — earlier theories about the guardian SSH probe and event-loop contention were disproved):

```
SNAPSHOT_TIMINGS total≈9s: call_sites 4.1-4.9s, memory_health ~2s, probe_all ~2s, awareness ~1s, rest <0.3s
```

`call_sites` ran a **per-call-site `message LIKE '%site_id%'` query** — ~30 non-indexable full scans of the `events` table per snapshot — and the ~17 sub-calls ran serially.

## Fix

1. **`call_sites` N+1 → one scan** (`observability/snapshots/call_sites.py`): pre-fetch the recent routing-failure events once, bucket by `site_id` in memory (same substring match as the old `LIKE`). ~30 scans → 1.
2. **Parallelize the snapshot** (`observability/health_data.py`): run the independent async sub-calls via `asyncio.gather` — total ≈ max(sub-call) instead of the sum; network-bound calls (Qdrant, MCP) overlap the serialized aiosqlite queries.
3. **Opt-in async-route timeout** (`dashboard/_blueprint.py`, `routes/health.py`): the health route returns HTTP 503 after 15s instead of blocking the Flask worker indefinitely on `future.result()`. Default (None) preserves unbounded behavior for long-running routes (reply waiters, approval waits). Timed-out coroutines are cancelled so they don't stack under polling.

Net: the health snapshot drops from ~9-20s to ~2s, and the endpoint can never hang regardless of any single sub-call's latency — so a degraded subsystem can't blind the dashboard or the Guardian.

## Tests

- `call_sites` bucketing test asserts correct per-site failure counts AND **exactly one** events query (guards against N+1 regression).
- `_async_route` timeout tests: slow handler → 503 promptly; bare decorator and fast handlers unaffected.
- Full sweep: 295 passed across `test_health_data` + `test_dashboard` + `test_observability`; ruff clean.

## Review

Independent code review: no P1; one P2 (timed-out coroutine should be cancelled) fixed in this branch. The external second-opinion reviewer is quota-limited and noted as such.

## Verification (post-merge)

Deploy + confirm against the live state: `/api/genesis/health` returns fast and the dashboard loads **while the Guardian is still down** — direct proof the snapshot no longer hangs.

### DIFF
--- a/src/genesis/dashboard/_blueprint.py
+++ b/src/genesis/dashboard/_blueprint.py
@@ -58,6 +58,11 @@ def _async_route(f=None, *, timeout: float | None = None):
                 try:
                     return future.result(timeout=timeout)
                 except FuturesTimeoutError:
+                    # Cancel the coroutine so it doesn't keep running on the loop
+                    # after we've returned — otherwise repeated polling stacks
+                    # concurrent snapshots (run_coroutine_threadsafe futures
+                    # schedule task cancellation on the loop thread-safely).
+                    future.cancel()
                     logger.warning(
                         "async route %s exceeded %.1fs timeout — returning 503",
                         getattr(fn, "__name__", "?"), timeout or 0.0,
--- a/src/genesis/dashboard/_blueprint.py
+++ b/src/genesis/dashboard/_blueprint.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+from concurrent.futures import TimeoutError as FuturesTimeoutError
 from functools import wraps
 from pathlib import Path
 
@@ -19,7 +20,7 @@ blueprint = Blueprint(
 )
 
 
-def _async_route(f):
+def _async_route(f=None, *, timeout: float | None = None):
     """Decorator to run async Flask route handlers.
 
     When the Genesis runtime loop is available (stored in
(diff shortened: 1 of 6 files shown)


WingedGuardian:
## First-Pass Code Review

> Note: `gh pr review --request-changes` is blocked on self-authored PRs — posting findings as a comment instead.

---

### Structural Review

**ERROR — Raw SQL against genesis.db in `src/genesis/observability/snapshots/call_sites.py`**

The new prefetch block introduces a raw `db.execute()` call with inline SQL:

```python
async with db.execute(
    "SELECT message, created_at FROM events "
    "WHERE event_type IN "
    "('all_exhausted', 'provider.fallback', 'breaker.tripped') "
    "AND created_at >= ?",
    (cutoff,),
) as cursor:
```

Per project conventions, queries against genesis.db must go through `genesis.db.crud` modules, not inline SQL (migrations only are excepted). Note the PR *reduces* the violation — N per-site raw queries → 1 pre-fetch — but the new `+` lines still introduce a raw SQL statement. Options:

- Extract to a crud helper, e.g. `crud.events.get_recent_routing_failures(db, cutoff)` (preferred)
- Or document explicitly why direct access is warranted in this observability path — the pre-existing raw SQL here suggests this module may have been grandfathered; a comment or tracked exception would keep it from becoming a silent pattern

---

**NOTE — Commit prefix `perf:`**

`perf(health):` is valid Conventional Commits but not in the project's explicit list (`feat:`, `fix:`, `refactor:`, `docs:`, `test:`, `chore:`). If `perf:` isn't established project-wide, consider `fix:` (corrects a hang) or `refactor:` (restructures for performance). Minor — adapt to whichever the project uses.

---

### PII / Secrets / Wording Scan: CLEAN

No secrets, private IPs, personal emails, hardcoded absolute paths, or sensitive wording found in code diff or PR metadata.

---

PR #642, opened 2026-06-15
fix(awareness): run session observer out-of-band of the tick lock

WingedGuardian:
## Problem (verified by measurement)

The awareness tick awaited `session_observer` INSIDE `_tick_lock`. Under provider exhaustion its LLM call (45s budget only checked *between* batches) ground through the chain for 16+ min, holding the lock and starving the heartbeat — APScheduler dropped ticks (`tick.max_instances`) and the host Guardian's health probe failed.

## Fix (architect-reviewed)

- **Out-of-band dispatch**: `session_observer` now runs via `tracked_task` (like reflections) so the tick fires on cadence regardless of provider state.
- **Single-flight guard** (`_session_observer_lock`): prevents overlapping runs from racing the observer's `.jsonl→.processing` renames. The `locked()`-then-`async with` pattern is asyncio-safe (no await in the gap).
- **Cancellation-safe cleanup**: the observer's processing-file restore is now in `try/finally`, so a shutdown mid-LLM-call restores `.processing` files instead of orphaning the observations.
- guardian-watchdog SSH (also in-lock, ~15s bounded) left as a follow-up.

## Tests
- Out-of-band dispatch (handed to tracked_task, not awaited inline), single-flight skip, error-swallowing. 136 passed across test_awareness + session_observer.

### DIFF
--- a/src/genesis/awareness/loop.py
+++ b/src/genesis/awareness/loop.py
@@ -371,6 +371,10 @@ class AwarenessLoop:
         self._interval = interval_minutes
         self._scheduler = AsyncIOScheduler()
         self._tick_lock = asyncio.Lock()
+        # Single-flight guard for the session observer, which now runs
+        # out-of-band (no longer serialized by _tick_lock). Prevents
+        # overlapping runs from racing the .jsonl→.processing file renames.
+        self._session_observer_lock = asyncio.Lock()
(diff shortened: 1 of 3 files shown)


WingedGuardian (commented):

---

PR #643, opened 2026-06-15
fix(memory): reference mirror renders reference.* only + refresh on boot

WingedGuardian:
## What

The `known-to-genesis.md` reference mirror rendered every `project_type='reference'` row — including bulk-curated knowledge mis-tagged with non-`reference.*` domains — and only regenerated on `reference_store`/`delete`/`export`. References added by other paths (the extraction job, bulk imports) never refreshed it, so the file drifted stale and polluted.

## Changes

- **`reference_mirror.py`** — the catch-all for non-ordered domains now skips anything whose domain isn't `reference.*`, so a row mis-tagged `project_type='reference'` with a non-reference domain can't leak into the human reference view.
- **`runtime/init/memory.py`** — regenerate the mirror once on boot (guarded; re-raises `CancelledError` so a mid-boot SIGTERM isn't swallowed), so references added by non-triggering paths surface on the next restart.
- **`test_reference_mirror.py`** — covers the mis-tagged-exclusion case.

## Verification

- `pytest tests/test_memory/test_reference_mirror.py` → 9 passed
- `ruff check` clean

## Scope note

This is the **code half** of a larger reference-store repair. The data cleanup (re-embedding a straggler reference so it's findable via `reference_lookup`, removing an orphaned vector, reclassifying mis-tagged curated rows) is intentionally **deferred to a separate change** while the production DB is stabilized — it requires a server stop + bulk SQLite/Qdrant writes and is not safe to run yet.

### DIFF
--- a/src/genesis/memory/reference_mirror.py
+++ b/src/genesis/memory/reference_mirror.py
@@ -1,7 +1,10 @@
 """Markdown mirror of the reference store at ~/.genesis/known-to-genesis.md.
 
 Read-only human view. Regenerated on reference_store / reference_delete /
-reference_export operations. NOT a source of truth — the database is.
+reference_export operations and once at memory init (boot) so references added
+by non-triggering paths (extraction job, bulk imports) still surface. Only
+genuine reference kinds (``domain`` like ``reference.*``) are rendered. NOT a
+source of truth — the database is.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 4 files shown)


WingedGuardian (commented):
## First-pass review — PR #643

### Structural review

**WARNING — Missing test coverage for boot-refresh path**

`src/genesis/runtime/init/memory.py` adds a new block to `init()` that calls `regenerate_mirror` at boot, re-raises `CancelledError`, and swallows all other exceptions with a `logger.warning`. None of that behaviour is exercised by the new test. The new `test_mistagged_non_reference_domain_excluded` covers the domain-filter logic in `reference_mirror.py` — good — but the `init()` path (happy path, swallowed-exception path, `CancelledError` propagation) is untested. Consider adding a test to `tests/test_memory/` (or the relevant `runtime/init/` suite if one exists) that mocks `regenerate_mirror` and asserts each branch.

**NOTE — Late import inside `try` block**

```python
# src/genesis/runtime/init/memory.py ~line 71
try:
    from genesis.memory.reference_mirror import regenerate_mirror
    await regenerate_mirror(rt._db)
```

Importing inside the `try` block rather than at the module top is unconventional. If this is intentional (e.g. to break a circular import that surfaces only at runtime), a short inline comment explaining why would help future readers. If it isn't load-order–sensitive, move the import to the module-level import block with the rest of the `genesis.*` imports.

---

No API keys, tokens, private IPs, credentials, hardcoded `/home/<user>/` paths, or private hostnames found in added lines. PR title and description contain no security-incident language, no leaked-credential references, and no private infrastructure details.

---

PR #644, opened 2026-06-15
fix(db): release MCP read snapshots + serialize connections to stop WAL pin (WS-15)

WingedGuardian:
## Problem
`memory_store` / `reference_store` were failing with **"database is locked"** while `follow_up_create` (a different connection) succeeded, and the SQLite WAL had grown to **~1.9 GB**. Root cause (verified in code): the four long-lived MCP connections (`scripts/genesis_mcp_server.py`) were opened as raw `aiosqlite.connect` in **deferred-isolation** mode, and read-only tool calls never committed — leaving a read transaction open that pinned the WAL checkpoint and blocked later writes. Being raw (not `SerializedConnection`) also let concurrent tool calls wedge the transaction state. This is the long-lived-connection case WS-15 (#634) deliberately deferred.

## Approach — staged (low-risk now, refactor later)
Phase 1 here keeps **write atomicity exactly as today** (writes already commit at the CRUD layer before a tool returns):

1. **Per-call read-snapshot release** — `InstrumentationMiddleware` takes the server's `db` and, at each tool call's `finally`, commits on success / rolls back on error. That closes the *trailing read* transaction so it can't pin the WAL. DB errors in the boundary never break the tool.
2. **Serialize the connections** — `get_db()` gains a `foreign_keys` param (default `True`); the 4 bootstrap connections now route through `get_db(foreign_keys=False)` → `SerializedConnection` (prior no-FK behavior preserved), preventing the interleave-wedge.
3. **Reclaim the 1.9 GB** — a one-time `wal_checkpoint(TRUNCATE)` after deploy, once the stale MCP readers cycle out (operational; no code).

**Phase 2 (separate, later):** `isolation_level=None` autocommit + a reentrant `transaction()` context manager for true concurrent multi-statement write atomicity.

## Review-driven fix
An inline code review caught that memory/recon/outreach bootstraps never passed an `activity_tracker`, so the middleware — and thus the new boundary — would never attach for them (including `memory_store`!). Fixed: those three now create a tracker so the boundary attaches for **all four** servers. (The tracker's own writes are batched + self-committing on a separate task, so they don't entangle with the boundary.)

## Tests / verification
- TDD: `test_mcp_middleware.py` (commit-on-success, rollback-on-error, no-db backward-compat, commit/rollback errors swallowed, original-error not masked); `test_connection.py` (FK on/off).
- `test_mcp/` 336 passed; observability+db+mcp 1061 passed (the only 3 failures are a pre-existing cross-dir logging-pollution in `test_manifest_tool.py`, visible only in a sequential combined run — passes alone; CI runs parallel workers).
- ruff clean; import smoke OK.
- **Post-merge E2E** (needs MCP servers to restart into the new code): `memory_store`/`reference_store` succeed, and `wal_checkpoint(TRUNCATE)` returns busy=0 with the WAL dropping to ~0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Memory storage no longer gets wedged on "database is locked"** — a
+  long-lived MCP database connection left read transactions open after read-only
+  tool calls, which pinned the SQLite write-ahead log (it could grow to
+  gigabytes on disk) and made `memory_store` / `reference_store` fail until a
+  restart. MCP tool calls now release their database snapshot at each call
+  boundary, and the MCP connections are serialized to prevent transaction-state
+  corruption — so writes keep working and the WAL stays bounded.
 - **Outreach emails actually send now** — email (and Discord/voice) outreach was
   being misaddressed to the Telegram forum chat for any category that routes to
   the supergroup, so every such send failed and silently piled up as retries.
--- a/scripts/genesis_mcp_server.py
+++ b/scripts/genesis_mcp_server.py
@@ -141,14 +141,12 @@ def _bootstrap_health(transport_kwargs: dict) -> None:
 
     @asynccontextmanager
     async def _lifespan(server) -> AsyncIterator[None]:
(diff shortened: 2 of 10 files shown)


WingedGuardian:

---

PR #645, opened 2026-06-16
fix(hooks): resolve worktree venv without SIGPIPE in genesis-hook launcher

WingedGuardian:
## Problem
In **worktree** Claude Code sessions, every Genesis hook was failing silently — "Failed with non-blocking status code: No stderr output" (exit **141 = SIGPIPE**). The `.claude/hooks/genesis-hook` launcher locates the Python venv; a worktree has no local `.venv`, so it took a fallback that ran, under `set -euo pipefail`:
```bash
MAIN_ROOT="$(cd "$GENESIS_ROOT" && git worktree list --porcelain 2>/dev/null | head -1 | sed 's/^worktree //')"
```
`head -1` closes the pipe after one line, so with **many worktrees** (this install had 40) git's output exceeds the 64 KB pipe buffer and git is still writing when the reader closes → **SIGPIPE** → `pipefail` + `set -e` silently kill the launcher → **every hook in that worktree fails** (session activity capture, file/edit audit logging, etc.). It read as "resume-specific," but the real determinant is *running from a worktree*.

## Fix
Replace the fragile pipeline with a no-pipe lookup of the main worktree root:
```bash
_common_dir="$(cd "$GENESIS_ROOT" 2>/dev/null && git rev-parse --git-common-dir 2>/dev/null)" || _common_dir=""
MAIN_ROOT=""
if [[ -n "$_common_dir" ]]; then
    MAIN_ROOT="$(cd "$GENESIS_ROOT" 2>/dev/null && cd "$(dirname "$_common_dir")" 2>/dev/null && pwd)" || MAIN_ROOT=""
fi
```
`git rev-parse --git-common-dir` returns the main repo's `.git` from any linked worktree; its parent is the main worktree root (where `.venv` lives). No early-closing pipe → no SIGPIPE. The `|| VAR=""` fallbacks keep it safe under `set -euo pipefail` and degrade to the existing clear "venv not found" error. The fast path (main repo with `.venv`) is untouched.

## Verification
- Reproduced the bug **39/40** iterations with the old pipeline; the fix is **0/40** SIGPIPE and resolves to the **executable main venv** 40/40.
- Before/after via the real launcher (invoked as CC does, from a worktree): old **20/20 fail** → fixed **0/20**; full real-hook run end-to-end → exit 0.
- `bash -n` + `shellcheck` clean.
- New regression-guard test (`tests/test_scripts/test_genesis_hook_wrapper.py`): asserts no `| head` pipeline in code, that `--git-common-dir` is used, and the launcher never exits 141 — would fail on the old code, passes on the fix.

## Deployment note
`genesis-hook` is a committed, per-worktree file, so this fixes the main repo and **future** worktrees; existing stale worktrees keep the old launcher until rebased.

### DIFF
--- a/.claude/hooks/genesis-hook
+++ b/.claude/hooks/genesis-hook
@@ -17,8 +17,17 @@ GENESIS_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
 # Find the Python venv. In a worktree, .venv lives in the main repo, not here.
 PYTHON="$GENESIS_ROOT/.venv/bin/python"
 if [[ ! -x "$PYTHON" ]]; then
-    # Worktree detection: resolve the main worktree via git
-    MAIN_ROOT="$(cd "$GENESIS_ROOT" && git worktree list --porcelain 2>/dev/null | head -1 | sed 's/^worktree //')"
+    # Worktree: .venv lives in the MAIN repo. Resolve the main worktree root via
+    # `git rev-parse --git-common-dir` (the main repo's .git; its parent is the
+    # main worktree root). Do NOT use `git worktree list | head` here: head closes
+    # the pipe after one line, so with many worktrees git gets SIGPIPE, and under
+    # `set -euo pipefail` that silently kills this launcher (exit 141, no stderr),
+    # breaking every hook in the worktree.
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

PR #630, opened 2026-06-14
feat(cc): interactive consoles use auto permission mode, not bypass (WS-20)

WingedGuardian:
## What & why

Interactive Claude Code consoles — the dashboard **web terminal** and the **SSH/tmux dev-console slot** — are driven by a human, so they now launch `claude --permission-mode auto` instead of `--dangerously-skip-permissions`. Auto mode auto-approves common operations (Bash allowlist, edits, sandbox) but still prompts on `deny`/`ask` rules and forced safety checks — which the operator is right there to answer — and keeps deny-rule safety.

**Headless/autonomous sessions are unchanged.** Every path through `CCInvoker` hardcodes `-p` (no TTY to answer a prompt), so they intentionally keep `--dangerously-skip-permissions` (with the existing `disallowed_tools` blacklists + PreToolUse hooks as guardrails). Only the two genuinely interactive entry points change. No `acceptEdits` (it still prompts on edits-vs-bash unevenly and is the wrong tradeoff here).

WS-20 — security-on-merit, emergent from the 2026-06-13 audit's D1 discussion.

## Changes
- `src/genesis/dashboard/routes/terminal.py` — prefill `--permission-mode auto`
- `scripts/cc-slot.sh` — `exec claude --permission-mode auto`
- `tests/test_dashboard/test_terminal_permission_mode.py` — regression guard (asserts the launched command, not comment text)

## Tests
`tests/test_dashboard/` — 116 passed (incl. 2 new). `bash -n scripts/cc-slot.sh` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #646, opened 2026-06-16
feat(cc): opt-in bypass permission mode for interactive consoles

WingedGuardian:
## What

WS-20 (#630) switched the two **interactive** Claude Code launchers — the SSH/tmux dev-console slot (`cc-slot.sh`) and the dashboard web terminal (`terminal.py`) — from `--dangerously-skip-permissions` to `--permission-mode auto`. Auto mode still prompts on deny/ask rules, which adds friction for an operator actively driving the session.

This keeps `auto` as the default (preserving WS-20's safety posture) and adds an **opt-in** to launch friction-free again:

- Set `GENESIS_CC_PERMISSION_MODE=bypass` → launches with `--dangerously-skip-permissions`.
- `cc-slot.sh` also sources an optional `~/.genesis/cc-slot.env`, because SSH `RemoteCommand` does not read `.bashrc`, so a plain env var alone is not operator-settable. A single line in that file (`GENESIS_CC_PERMISSION_MODE=bypass`) enables it for all slots.
- `terminal.py` reads the same var from the dashboard process environment and prefills the command via Jinja.

Unknown/empty values fall back to `auto`. Headless/autonomous sessions (every `CCInvoker -p` path) are intentionally **unchanged**.

## Why

Auto mode's prompting is disruptive for day-to-day interactive work, but a hard revert would undo WS-20's deny-rule safety net. A per-environment opt-in gives operators the choice without changing the secure default.

## Testing

- Unit tests (`tests/test_dashboard/test_terminal_permission_mode.py`): default→auto, bypass opt-in, unknown→auto fallback, templated prefill, cc-slot.sh branches.
- Verified Jinja `render_template_string` renders both modes correctly (the HTML/CSS/JS template body is unaffected).
- `cc-slot.sh` exercised end-to-end via a tmux shim across all modes, including the regression case where a malformed `cc-slot.env` exits non-zero: it now warns and continues instead of silently dropping the SSH session under `set -e`.

## Review notes

Independent review surfaced two issues, both fixed here:
- The `cc-slot.env` source was guarded by `[ -f ] && . file`, making the source the final command of an `&&` list — under `set -euo pipefail` a bad file would abort the launcher silently. Now isolated with a warning-and-continue path.
- Added `{{ launch_cmd | e }}` as defensive escaping (current values are fixed literals; hardens against future dynamic values since `render_template_string` autoescape is filename-dependent).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,15 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Changed
 
+- **Interactive Claude Code consoles can run friction-free again, when you want
+  them to** — the SSH/tmux dev-console slot and the dashboard web terminal still
+  default to `--permission-mode auto` (auto-approves common operations, but still
+  prompts you on deny/ask rules), but you can now opt a session back into
+  `--dangerously-skip-permissions` by setting `GENESIS_CC_PERMISSION_MODE=bypass`.
+  For the SSH slot, put that line in `~/.genesis/cc-slot.env` (SSH sessions don't
+  read your shell profile); for the dashboard terminal, set it in the dashboard's
+  environment. Headless autonomous sessions are unaffected.
+
 - **`update.sh` now keeps Claude Code in sync on your host VM too** — if you run
   Genesis with a Guardian on a separate host VM, updates previously only touched
   the container's Claude Code, letting the host drift behind. `update.sh` now
--- a/scripts/cc-slot.sh
+++ b/scripts/cc-slot.sh
@@ -83,13 +83,30 @@ chmod 700 "$TMPDIR"
 # Without this, intermittent ENOENT failures on /tmp break the Bash tool.
 export CLAUDE_CODE_TMPDIR="$HOME/.genesis/cc-tmp"
 
+# Permission mode for this interactive dev console. Default: auto — auto-approves
+# common ops but still prompts on deny/ask rules, which the operator answers in
+# the tmux session (keeps deny-rule safety). To launch friction-free with
+# --dangerously-skip-permissions, set GENESIS_CC_PERMISSION_MODE=bypass. SSH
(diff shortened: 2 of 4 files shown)


WingedGuardian:

Checked: `scripts/cc-slot.sh`, `src/genesis/dashboard/routes/terminal.py`, `tests/test_dashboard/test_terminal_permission_mode.py`, `CHANGELOG.md`.

No import hygiene issues, no missing test coverage (all new logic in `_cc_launch_command()` is tested across default/bypass/unknown paths), docstring present on the new helper, conventional commit prefix correct. No credentials, private IPs, absolute paths with usernames, or wording concerns in the diff or PR metadata.

---

WingedGuardian:
Merging via admin squash: all CI green (lint, test, CodeQL, leak-detector); Genesis structural + PII/secrets review CLEAN; no inline findings. Codex cloud review was quota-limited (not a blocking finding) — recorded per the pre-merge gate.

WingedGuardian:
**Security review acknowledgment** — the `security-guidance` automated review flagged the bypass case arm as [HIGH] Agent/Subprocess Permission Bypass. **Accepted by design, no change:**

- Default remains `--permission-mode auto` (deny-rule safety preserved). `--dangerously-skip-permissions` is strictly **opt-in**, activated only when the operator deliberately creates `~/.genesis/cc-slot.env` — the explicitly-scoped mechanism the review asks for, not a default toggle.
- This launcher is the operator's own SSH/tmux dev console (operator-only, Tailscale-gated SSH). The operator already holds shell access, so bypass removes interactive prompts but grants no new privilege and adds no remote attack surface. An actor able to set the env var / write the file already has code-exec as the operator.
- Claude Code's own UI shows `⏵⏵ bypass permissions on` when active, so it is never silent.
- A per-launch confirmation prompt was rejected: it reintroduces exactly the friction this opt-in exists to remove, on every connect.

Headless/autonomous paths (CCInvoker `-p`) are unchanged and out of scope.

PR #647, opened 2026-06-16
fix(db): cancellation-safe reads + WAL detection/bounding (database is locked / WAL bloat)

WingedGuardian:
## Problem

Under heavy load or a provider outage, saving to memory (`reference_store` /
`memory_store`) could fail with `OperationalError: database is locked`, and
SQLite's write-ahead log could balloon (it reached ~2 GB in the Jun-13 incident,
freezing checkpoints for ~2 days).

**Root cause (empirically reproduced):** a read coroutine on a long-lived MCP
connection, if cancelled between `cursor = await db.execute(SELECT…)` and the
cursor closing, leaves a pinned SQLite read snapshot. In WAL mode that (a) pins
the checkpoint so the WAL can't truncate, and (b) makes that connection's next
write fail `SQLITE_BUSY` permanently. An earlier `isolation_level=None`
(autocommit) hypothesis was **falsified** — it does not release a held snapshot
and breaks multi-statement atomic writes.

## Fix

1. **Cursor cancellation-safety** — memory-subsystem reads route through
   `execute_fetchall` (returns `list[Row]`, no dangling asyncio cursor) or
   `async with db.execute(...) as cursor:` for `cursor.description` sites. Writes
   are untouched; deferred-isolation atomicity is preserved.
2. **Detection** — `_check_wal_health` raises a high/critical observation
   (surfaced to Telegram + the morning report) when the WAL file is abnormally
   large, so a stuck reader is caught in minutes, not days.
3. **Bounding** — `PRAGMA journal_size_limit=64MB` on the connection helpers and
   the four MCP bootstraps; an hourly `wal_checkpoint(TRUNCATE)` reclaims WAL
   file space.
4. **Bug fix** — the manual WAL checkpoint was a silent no-op: it ran sqlite3
   from the event-loop thread (`db._conn._conn.execute`), which raises a
   thread-bound `ProgrammingError` that a bare `except` swallowed. It now goes
   through the async aiosqlite API so it actually runs.

## Scope

The conversion targets the **memory subsystem** — the connection that pinned the
WAL in production. The remaining ~520 CRUD read sites (mostly short-lived
`get_raw_db` connections that auto-close) are a tracked follow-up.

## Verification

- Empirical repro confirms the cancellation leak **and** that `execute_fetchall`
  is cancellation-safe (including cancelled in-flight).
- New guard tests: `test_db/test_cursor_cancellation.py`,
  `test_awareness/test_wal_health.py`, `test_db/test_connection_pragmas.py`
  (incl. a regression test proving the checkpoint now actually runs).
- `test_db` + `test_memory` + `test_awareness` green (1253 + 553 passing);
  ruff clean; an independent code-review pass was applied (registered the
  `infrastructure_alert` TTL; hardened `count()`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,22 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ## [Unreleased]
 
+### Added
+
+- **Genesis now watches its own database journal size** — if SQLite's
+  write-ahead log grows abnormally large (the sign of a stuck database reader
+  holding the file open), Genesis raises a high/critical alert on Telegram and in
+  the morning report, instead of letting it balloon silently for days.
+
+### Fixed
+
+- **Memory writes no longer get stuck behind a database lock** — under heavy load
+  or a provider outage, a cancelled database read could leave a stale lock that
+  made saving to memory (`reference_store` / `memory_store`) fail with "database
+  is locked" while the write-ahead log ballooned (it reached ~2 GB). Database
+  reads are now cancellation-safe and the journal is size-bounded, so the lock
+  can't get stuck and the file can't run away.
+
 ### Changed
 
 - **`update.sh` now keeps Claude Code in sync on your host VM too** — if you run
--- a/scripts/genesis_mcp_server.py
+++ b/scripts/genesis_mcp_server.py
@@ -149,6 +149,7 @@ def _bootstrap_health(transport_kwargs: dict) -> None:
         db.row_factory = aiosqlite.Row
         await db.execute("PRAGMA journal_mode=WAL")
         await db.execute(f"PRAGMA busy_timeout={BUSY_TIMEOUT_MS}")
+        await db.execute("PRAGMA journal_size_limit=67108864")  # 64 MB WAL file cap
         try:
 
             # Bootstrap standalone router for LLM-dependent tools
@@ -237,6 +238,7 @@ def _bootstrap_memory(transport_kwargs: dict) -> None:
         db.row_factory = aiosqlite.Row
         await db.execute("PRAGMA journal_mode=WAL")
         await db.execute(f"PRAGMA busy_timeout={BUSY_TIMEOUT_MS}")
+        await db.execute("PRAGMA journal_size_limit=67108864")  # 64 MB WAL file cap
 
         try:
 
@@ -279,6 +281,7 @@ def _bootstrap_recon(transport_kwargs: dict) -> None:
         db.row_factory = aiosqlite.Row
         await db.execute("PRAGMA journal_mode=WAL")
         await db.execute(f"PRAGMA busy_timeout={BUSY_TIMEOUT_MS}")
+        await db.execute("PRAGMA journal_size_limit=67108864")  # 64 MB WAL file cap
 
         try:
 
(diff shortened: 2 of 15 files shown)


WingedGuardian (commented):

---

**Structural pass (15 files, 508 additions)**

- No `nohup`, no `/tmp/`, no `task_states` direct inserts, no destructive git ops, no bare `except:` clauses, no mutable defaults.
- All three new async WAL helpers (`_sqlite_wal_checkpoint`, `_sqlite_wal_truncate`, `_check_wal_health`) use `except Exception:` (typed) or `contextlib.suppress(Exception)` — not bare.
- Import hygiene: `import time` slots correctly among stdlib imports; lazy `from pathlib import Path` / `from genesis.env import genesis_db_path` inside `_check_wal_health`'s try-block is intentional (avoiding a top-level import of a heavy module at loop startup).
- `get_unsurfaced` correctly uses `async with db.execute(...) as cursor:` instead of `execute_fetchall` because it needs `cursor.description` — the one case in the PR where that distinction matters.
- New source files (`loop.py` additions, `connection.py`, `crud/*`) all have corresponding new test files (`test_wal_health.py`, `test_connection_pragmas.py`, `test_cursor_cancellation.py`) plus updated mocks in existing suites.
- Conventional commit prefix `fix(db):` present. ✓

**PII/Secrets/Wording pass**

- No secrets, tokens, credentials, private IPs, SSH keys, or connection strings in added lines.
- PR title and body describe a technical WAL/locking incident without language that advertises a credentials leak, data exposure, or named private infrastructure.
- CHANGELOG additions describe functional behavior only — no sensitive wording.

---

WingedGuardian:
### Update: rebased onto WS-15 (`1ae3af7b`) — complementary, combined

WS-15's fix (`fix(db): release MCP read snapshots + serialize connections`) landed on `main` while this was in review. It fixes the same WAL-pin bug at a **different layer**, so this branch now **merges and combines** both (defense-in-depth, not competing):

- **WS-15:** MCP connections → `SerializedConnection`; per-tool-call commit/rollback middleware releases a trailing read snapshot at the boundary (covers all 4 MCPs).
- **This PR:** read-level cancellation-safety (`execute_fetchall`) so the read never leaks — covering memory reads *outside* the MCP boundary (awareness loop, background tasks) too — plus WAL-size **detection alert**, `journal_size_limit` + periodic `TRUNCATE`, and a fix for the awareness-loop manual checkpoint which was a **silent no-op** (it ran sqlite3 from the event-loop thread → thread-bound error swallowed).
- Empirically confirmed WS-15's middleware *does* release on cancellation; my `journal_size_limit` is folded into `get_db`'s `_configure`, so the MCP `get_db(foreign_keys=False)` connections get it too (runtime-verified).

> **⚠️ Note on the WS-15 follow-up:** its commit message names a planned **Phase 2 = `isolation_level=None` autocommit**. I empirically **falsified** that approach (it does *not* release a held snapshot and breaks atomic writes). This PR's `execute_fetchall` conversion already achieves Phase-2's intent safely — **Phase 2 autocommit should be cancelled.**

### Verification

- Merged-tree suite green: **1407 passed** (`test_db` + `test_memory` + `test_awareness` + `test_observability`) + focused interaction tests + import smoke. Runtime-verified the MCP-connection pragma path.
- **Adversarial review:** Codex (usage quota, resets Jul 1) and OpenCode/GLM5 (billing) were **both unavailable**, so the independent-model review was skipped — a Claude `code-reviewer` subagent reviewed as fallback. It found 2 items (a `count_links` empty-guard inconsistency and a `_check_wal_health` cooldown-on-write-failure gap), **both fixed** in this branch. Flagging for a human eye on the read conversions + the awareness-loop changes given the independent reviewer was down.

PR #650, opened 2026-06-16
feat(install): D16 CLAUDE.md inheritance split + set_secret atomicity (WS-13 PR1)

WingedGuardian:
## Summary

**WS-13 PR1 (surgical):** implements the audit's **D16 three-file CLAUDE.md inheritance split** on the host VM, plus a confirmed `set_secret` atomicity fix. PR2 (Guardian stop-reliability fixes — `cc-loop-01`, `GUARD-R2-01`) follows separately.

### D16 — three-file CLAUDE.md split

A `claude` session loads `~/.claude/CLAUDE.md` (user-level) plus a project `CLAUDE.md` discovered from its cwd. This PR wires that inheritance on the host:

- **Shared user-level `~/.claude/CLAUDE.md`** (`host-setup.sh`) — common "this is a Genesis install" facts (host / container / guardian) in managed `<!-- begin:SECTION -->` blocks; the hand-editable Personal Notes section is preserved across re-runs.
- **Guardian-project CLAUDE.md → the diagnosis CC work_dir** (`install_guardian.sh`) — the diagnostic CC runs with `cwd = cc.work_dir`, so the install-dir copy was **never auto-loaded**. It's now symlinked into the work_dir (tracking regenerations of the install-dir copy).
- **Operator-project CLAUDE.md** (new `config/operator-claude.md`) — orients a human who SSHes into the host cold: how to pause/stop the Guardian (`.guardian-maintenance` + `systemctl --user`), how to enter the container. Guarded so it never clobbers an unrelated `~/CLAUDE.md`.

### `set_secret` atomicity (`install.sh`)

The old `set_secret` could **lose** a key (interrupted between the `mv` and the `echo`) or **duplicate** it (on a `grep` short-circuit). It now builds the full new content in a co-located `mktemp` (0600) and does a single atomic rename, preserving the original perms.

### Also

- `install_guardian.sh`: warn instead of silently skipping `skip-worktree` when the install dir isn't a git checkout; fix a self-contradicting `"CLAUDE.md generated"` log that printed even on the template-missing path.

## Verification

- `shellcheck` clean on all modified scripts.
- `set_secret` unit test: overwrite produces no duplicate, values containing `&`/`$`/`\` survive, perms stay `0600`, no temp-file leak on failure.
- **Live end-to-end on a host:** `claude --print` in the Guardian work_dir loads the Guardian CLAUDE.md; a `claude` session in the operator home dir loads **both** the shared user-level and the operator project file; the Guardian stayed healthy throughout.
- Real-code isolation test of the `host-setup.sh` blocks: managed blocks fill correctly, idempotent re-run preserves Personal Notes (no duplication, no placeholder leakage), and the operator clobber-guard works.

## Review

Codex (primary adversarial reviewer) was quota-exhausted; the fallback reviewer flagged 5 issues — **3 applied** (temp-file leak on disk-full, operator-doc UID placeholder, self-contradicting log line) and **2 rejected with rationale** (the parent-dir `chown` is required for the operator-runnable maintenance-flag mechanism; a suggested `grep -Fv` change was itself buggy — `-F` makes the `^` anchor literal and would *cause* duplicate keys).

## Scope note

`install.sh` / `host-setup.sh` are the public installer and can't be fresh-tested in this environment, so changes are deliberately **surgical**. Lower-value install-script polish (tool-version centralization, container-specs population, etc.) is intentionally deferred to a follow-up.

### DIFF
--- a/config/operator-claude.md
+++ b/config/operator-claude.md
@@ -1,0 +1,30 @@
+# Operator — You're on a Genesis Host VM
+
+You (a human) are SSHed into the **host VM** of a Genesis install. You are **not** inside Genesis,
+and you are **not** the Guardian. Genesis (the autonomous agent) runs inside an Incus container; a
+host-side **Guardian** systemd timer watches that container and can auto-recover it.
+
+Shared install facts (host/container identity, IPs, dashboard URL) are in `~/.claude/CLAUDE.md`,
+auto-loaded alongside this file.
+
+## Stop / pause the Guardian (e.g. before host maintenance)
+
+The Guardian checks for a maintenance flag on every tick and stands down if it is present:
+
+- **Pause**:  `touch /var/lib/guardian-snapshots/.guardian-maintenance`
+- **Resume**: `rm /var/lib/guardian-snapshots/.guardian-maintenance`
+- **Hard stop**: `systemctl --user stop genesis-guardian.timer genesis-guardian-watchman.timer`
+  (stop the watchman too — it re-arms the guardian timer)
+- **Status / logs**: `systemctl --user status genesis-guardian.timer` ·
+  `journalctl --user -u genesis-guardian -n 50`
+
+## Work inside Genesis (the container)
+
+- `genesis`  — alias added to your `~/.bashrc`, or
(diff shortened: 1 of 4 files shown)


WingedGuardian:

---

PR #652, opened 2026-06-16
fix(db,scheduler): record conflict-recovery updates + repair gitnexus reindex

WingedGuardian:
## Summary

- **Migration 0024** extends the `update_history` CHECK constraint to allow `'conflicts_pending'` as a valid status. The merge-conflict recovery path in `update.sh` writes this status, but the original constraint `('success','failed','rolled_back')` caused every conflict-recovery update to be silently dropped with an `IntegrityError` — leaving the update history blank after any conflict-resolved sync.

- **`gitnexus analyze --index-only`** is not a valid flag in current GitNexus (exits with `unknown option`). The Monday/Thursday reindex job has been silently failing on every run. The flag is dropped; `gitnexus analyze` is incremental by default since GitNexus 1.6.5.

- **2h timeout** added to `proc.communicate()` in the reindex job. A cold full-repo analysis can take minutes; without a bound the job slot stays occupied permanently if the process hangs.

## Migration details

SQLite cannot `ALTER` CHECK constraints, so migration 0024 uses the standard table-recreate pattern: idempotency guard → drop orphaned temp → create new table → explicit-column `INSERT … SELECT` → drop old → rename. All existing rows survive. No indexes to recreate (only the PK autoindex).

## Test plan

- [ ] `pytest tests/test_db/test_migration_0024.py -v` — 4 tests: new status accepted, bogus rejected, existing rows survive recreate, idempotent double-apply
- [ ] After merge: restart genesis-server and confirm migration 0024 applied; `sqlite3 data/genesis.db "SELECT sql FROM sqlite_master WHERE name='update_history'"` shows `conflicts_pending` in the constraint
- [ ] Trigger a conflict-recovery scenario and confirm a row appears in update history with status `conflicts_pending`


WingedGuardian:
## Code Review — PR #652

**Structural review:** 1 WARNING · 1 NOTE

---

### WARNING — Missing test coverage for new `TimeoutError` path (`src/genesis/surplus/scheduler.py`)

The new `TimeoutError` branch in `run_gitnexus_reindex` — `proc.kill()` → `record_job_failure()` → `return` — is real behavior change with no test exercising it. The happy path (returncode == 0) also now goes through `asyncio.wait_for`, which changes the call shape. The existing scheduler tests (if any cover this function) would only exercise the non-timeout path.

Suggest adding a test that mocks `proc.communicate()` to hang past the timeout and asserts the job slot is released (i.e., the function returns without blocking).

---

### NOTE — `up()` missing docstring (`src/genesis/db/migrations/0024_update_history_conflicts_status.py`)

`down()` has a docstring; `up()` does not. The module-level docstring covers intent, but the asymmetry is worth evening out — especially since `up()` is the primary entry point called by the migration runner. One line is enough: `"""Expand update_history CHECK constraint to include 'conflicts_pending'."""`

---

Structural review: no ERRORs found — migration pattern is correct (table-recreate required for SQLite CHECK change, idempotency guard present, explicit-column INSERT avoids schema drift, orphaned-temp cleanup is there). The `--index-only` removal and 2h timeout justify the CLAUDE.md timeout policy reference in the comment. No pattern violations.

PII/Secrets/Wording scan: CLEAN — no secrets, private IPs, credentials, or sensitive wording in code diff or PR metadata.

---

PR #655, opened 2026-06-16
fix(guardian): alert once per down-episode + restored ping (GUARD-R2-01)

WingedGuardian:
## What & why

When Genesis goes down and the Guardian's diagnosis can't reach Claude Code, the host Guardian (a 30s systemd oneshot) re-ran a full Opus diagnosis **and re-sent a CRITICAL/EMERGENCY alert every 30 seconds** until recovery — an alert storm. Separately, when Genesis recovered **on its own**, the Guardian said nothing — there was no "restored" notification.

This makes the Guardian send **one "down" alert per outage (no repeats, however long it lasts) and one "restored" notification when it comes back** — two messages per episode, regardless of duration.

## Root cause

In `CONFIRMED_DEAD`, the state machine returns `action_needed=True` every tick → `_handle_confirmed_dead` re-diagnoses + re-alerts. The approvable-recovery path is already throttled (it blocks on approval, and the oneshot + 30s timer can't overlap), but the **CC-unavailable / ESCALATE path early-returns without blocking**, so each fresh oneshot tick re-fires. Because each tick is a *new process*, the only robust throttle is **persisted state**. Autonomous recovery (`CONFIRMED_DEAD → HEALTHY`) routed to `_handle_healthy`, which only prunes/heartbeats — no alert.

## Change (minimal scope)

- Persisted `StateData.down_alert_sent` episode flag (survives ticks via `state.json`). `_handle_confirmed_dead` returns early once set — no re-diagnosis (the expensive Opus run), no re-alert — until genuine recovery.
- One `INFO` "Genesis recovered" ping on the previously-silent autonomous `CONFIRMED_DEAD → HEALTHY`, gated on `snapshot.all_alive` so an **auto-reset** (which routes to HEALTHY while *still down*) neither pings nor clears the flag — the storm stays suppressed.
- Flag cleared at every recovery-confirmed point (autonomous, CC-resolved, approval-time, recovery-engine success); **never** in `_reset_to_healthy` (auto-reset also calls it — clearing there would re-enable the storm). Un-marked on a diagnosis crash so the episode isn't silently muted.

**Intended behavior change:** after the first handling, the Guardian no longer silently *retries* recovery on later ticks. The first attempt + its result alert still fire; re-attempting (with a fresh prompt) every 30s was part of the storm being removed.

The interactive approval flow and the existing first-pass alerts are deliberately left untouched (minimal scope).

## Verification

- 13 new tests: storm suppression across ticks, restored ping, the auto-reset trap, restored-ping dedup, diagnosis-crash retry, and a full-episode "exactly two pings regardless of duration" acceptance test.
- Full guardian suite green (360 passed); ruff clean.
- Adversarial code review (Codex quota-exhausted → Claude reviewer): two hardening findings (recovery-path flag-leak, mark-then-crash) applied + covered by tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **The Guardian alerts once when Genesis goes down — and once when it's back** —
+  previously, if Genesis went down and its diagnosis couldn't reach Claude Code,
+  the host Guardian re-ran a full investigation and re-sent a critical Telegram
+  alert every 30 seconds until recovery — an alert storm. It now sends a single
+  "down" alert per outage (no repeats, however long it lasts), and when Genesis
+  comes back on its own it sends a single "restored" notification — which it
+  never did before.
+
 - **Memory writes no longer get stuck behind a database lock** — under heavy load
   or a provider outage, a cancelled database read could leave a stale lock that
   made saving to memory (`reference_store` / `memory_store`) fail with "database
--- a/src/genesis/guardian/check.py
+++ b/src/genesis/guardian/check.py
@@ -236,6 +236,23 @@ async def _check_cycle(
 
     # Step 3: Act based on state
     if state == GuardianState.HEALTHY:
+        # GUARD-R2-01: on GENUINE recovery from a down-episode we alerted about,
+        # send ONE "restored" ping and clear the flag. Gate on snapshot.all_alive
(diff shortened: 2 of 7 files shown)


WingedGuardian (commented):

---

NOTE (`tests/test_guardian/test_check.py`) — inline stdlib imports in test helpers:

- `from datetime import UTC, datetime` inside `_now_iso()` (module-level function)
- `import json` inside `_seed_state()` and several test methods

These are stdlib imports that belong at the module top alongside the existing imports. No functional impact, but they break the stdlib → third-party → local grouping convention.

---

PR #656, opened 2026-06-16
fix(cc): per-session interrupt registry so /stop targets the right subprocess (cc-loop-01)

WingedGuardian:
## What & why

The shared `CCInvoker` (one instance for the whole runtime) tracked the running Claude Code subprocess in a single `_active_proc` slot. When a background task (reflection, inbox, an ego session, mail, sentinel) ran a CC subprocess at the same time as a foreground Telegram conversation, the second to start **overwrote** that slot. Telegram `/stop` then sent SIGINT to whichever subprocess was last spawned — frequently a background task instead of the user's generation. (`DirectSessionRunner` already worked around this with a dedicated invoker; the shared invoker's concurrent callers did not.)

`/stop` now reliably interrupts **your** session.

## Change

- **Per-session registry.** Replace the single `_active_proc` slot with `dict[key, Process]`. `run()`/`run_streaming()` register the spawned proc under `invocation.session_key or "pid:<pid>"` and unregister on cleanup. `interrupt(key=None)` targets the keyed proc; with no key it targets the most-recently-registered live proc (back-compat for existing no-arg callers + the protocol).
- **Stable Telegram key.** A helper `interrupt_key(user_id, chat_id)` → `tg:<user>:<chat>` is derived identically at stream-start (threaded through `handle_message_streaming(session_key=...)` → `CCInvocation.session_key`) and at `/stop`, so no session-id resolution is needed. Background procs key by pid and are never cross-fired.
- **Retry safety.** `_build_fresh_invocation` forwards `session_key`, so a post-timeout fresh retry stays interruptible. `AgentProvider.interrupt` signature updated in lockstep.
- **Proc-lifecycle hardening (from review).** Register before the stdin feed (so the proc is interruptible from spawn) and guard the feed so a broken pipe / task cancellation reaps the subprocess instead of leaking it; `run_streaming` unregisters in a `finally`; both run paths kill a still-live proc on abnormal exit.

## Verification

- `gitnexus impact` on `CCInvoker.interrupt` = **LOW** (the signature change is an additive optional param; existing callers unaffected).
- New tests: keyed interrupt hits the right proc and spares others; no-key → most-recent live; unknown key no-ops; dead-entry pruning; `run()`/`run_streaming()` register under the session key during execution and clear in `finally`; and an **end-to-end test with two real subprocesses + real SIGINT** confirming only the targeted one is stopped.
- 418 cc + Telegram-handler tests pass; ruff clean.
- Reviewed by a genesis-architect pass (design: approve-with-changes, all applied) and a code-reviewer pass (implementation; the proc-leak hardening above came from it). Codex was quota-limited, so review used the Claude reviewer.


WingedGuardian:

---

PR #657, opened 2026-06-16
feat(skills): deliverable-builder — gated pipeline for send-ready deliverables

WingedGuardian:
## What

A new Tier-1 skill, `deliverable-builder`: a gated pipeline that produces send-ready
deliverables (job take-homes, client reports, one-pagers, decks) in the **correct file format —
not raw markdown**, in the user's voice, leading with the strongest material, and calibrated to
a declared authenticity target rather than a one-size "look human" default.

Pipeline: **intake (robust Gate-1 interview) → draft → structure/altitude → voice → anti-slop →
render → fresh-context Gate-2 verification → user sign-off.** A fail-open Stop-hook won't let a
session end with a rendered deliverable that hasn't passed Gate 2.

## Key pieces

- `SKILL.md` + `references/` (intake, structure-altitude, render-guide, qa-protocol, approval-gates).
- `scripts/hooks/deliverable_gate_guard.py` — the Stop-hook gate (fail-open, session-scoped, in
  its own Stop group), registered in `.claude/settings.json`.
- `tests/test_hooks/test_deliverable_gate_guard.py` — 21 tests (state matrix, fail-open,
  session-scoping, path-safety).

## What makes the process non-skippable

- **Gate 1** is a mandatory `AskUserQuestion` interview that pins down format, authenticity
  target (fully human-made vs AI-assisted-OK), visual style, and substance — never guessed.
- **Gate 2** is a fresh-context adversarial reviewer that re-reads the original requirements,
  byte-checks the artifact, and returns per-criterion PASS/FAIL with evidence; failures loop
  back, bounded.
- The **Stop-hook** blocks session-end while a rendered deliverable is unverified.

## Verified

- Dogfooded end-to-end on two deliverables (a synthetic executive memo and a real take-home
  packet). Both passed a fresh-context Gate-2 (per-criterion, fabrication-checked — every figure
  traced to source).
- Code-reviewed; findings fixed (strict marker ownership, null-byte guard, independent Stop group).
- 21 hook tests green; no regression across the repo's existing hook tests.
- Two render/verify bugs caught by dogfooding and fixed: default to a Unicode-capable PDF engine
  (`xelatex`); confirm artifact type by byte-check rather than the `file` command.

## Scope

v1 is foreground-only. Autonomous / task-executor mode, designed decks beyond basic, and XLSX
are deferred to v2. The user's voice exemplars and the adopted humanizer skill live outside the
repo (user overlay), so no private data is included here.


WingedGuardian (commented):

**PII/Secrets/Wording scan: 2 soft warnings, 1 note**

---

### SOFT WARNING — Hardcoded path fragment reveals server username

`references/intake.md` (bash example, fallback session-id resolution):
```bash
SID=$(ls -t ~/.claude/projects/-home-ubuntu-genesis/*.jsonl 2>/dev/null \
```

`-home-ubuntu-genesis` is Claude Code's encoding of `/home/ubuntu/genesis`, which exposes that the server's login user is `ubuntu`. This is committed to a public skill file. Consider replacing with a dynamic derivation or a placeholder comment, e.g.:

```bash
# Replace -home-ubuntu-genesis with Claude Code's encoding of your project path:
# e.g. -home-$(whoami)-genesis  (adjust if your path differs)
SID=$(ls -t ~/.claude/projects/-home-$(whoami)-genesis/*.jsonl 2>/dev/null \
```

---

### SOFT WARNING — Company name records job application history in a public repo

`phData` appears 3× across the public skill docs:
- `SKILL.md`: *"Turn my phData analysis into the submission packet it should have been."*
- `references/intake.md`: *"The phData packet was correctly polished because AI-assisted was fine there…"*
- `references/structure-altitude.md`: *"The phData packet's biggest tell wasn't the prose…"*

This is a real company name committed to a public repo, implicitly recording that you applied there and used Genesis for the submission. The examples work just as well with a placeholder like `<TechCo>` or `AcmeCorp`. Not blocking, but worth a pass before merge.

---

### NOTE — `main()` missing docstring

`scripts/hooks/deliverable_gate_guard.py`: `_decide()` has an inline docstring; `main()` (public function, no leading underscore) does not. The module-level docstring is thorough, but a one-liner on `main()` would be consistent:
```python
def main() -> int:
    """Read hook stdin, run the gate decision, and return the exit code."""
```

---

PR #658, opened 2026-06-16
fix: restore ego proposal delivery + correct Telegram notification routing

WingedGuardian:
## What & why

Investigating misrouted Telegram messages surfaced four connected bugs in the ego → outreach delivery path. This PR is the Phase A set — the verified, low-risk fixes.

### Fixes
- **Ego proposals silently stopped delivering.** The proposal "quality gate" is an LLM judge. When its providers were exhausted, the scorer returned a `judge_call_fail` (score 0.0) and the gate marked *every* proposal `quality_hold` — which `send_digest` silently drops. The gate's docstring promised "fail open," but it failed *closed*. Now a judge-infrastructure error (`judge_call_fail` / `judge_parse_fail`) passes the proposal through; only a genuine below-threshold score holds it.
- **The silent drop is no longer silent.** When `send_digest` returns `None`, we log a warning (proposals stored but undelivered) instead of nothing.
- **Ego notifications now go to the DM, not the Surplus topic.** `notification` had no `delivery_routing` entry, so it fell through to `supergroup` and then the `surplus` catch-all. Added `notification: dm`. Also changed the unmapped-category fallback from `surplus` → `conversation` with a warning, so a missing topic mapping surfaces instead of vanishing into the debug dump.
- **Notification text no longer arrives wrapped in literal quotes.** Strip one layer of matched wrapping quotes (straight or smart) the ego sometimes adds around its message.

### Tests
9 new/updated tests: quality-gate fail-open vs genuine-hold cases, a quote-strip matrix (including a non-wrapped `She said "hi"` case that must be left alone), the DM-routing config assertion, and the updated unmapped-category default. 32 pass locally; `ruff` clean.

### Not in this PR (deliberate)
- Releasing the proposals already stuck in `quality_hold` — a live-DB operation, handled separately.
- Surplus-topic JSON formatting cleanup (verify-first).
- Guardian recovery-approval rework (separate, architect-gated).

Independent code review (Claude reviewer; Codex was rate-limited): clean, no blocking findings.

### DIFF
--- a/config/outreach.yaml
+++ b/config/outreach.yaml
@@ -57,6 +57,10 @@ delivery_routing:
   approval: supergroup
   # Content review goes to dedicated topic — never DM.
   content: supergroup
+  # Ego notifications are the ego "talking to" the user — informational,
+  # not an official proposal. They belong in the private DM, never a
+  # supergroup topic. (Official proposals go to the Ego Proposals topic.)
+  notification: dm
 
 health_alerts:
   # Only these alert IDs go to Telegram (CRITICAL severity only).
--- a/src/genesis/channels/telegram/topics.py
+++ b/src/genesis/channels/telegram/topics.py
@@ -292,4 +292,16 @@ class TopicManager:
             "approval": "approvals",
             "content": "content_review",
         }
-        return mapping.get(outreach_category, "surplus")
+        topic = mapping.get(outreach_category)
+        if topic is None:
+            # Unmapped category — route to the visible conversation topic,
+            # NOT surplus (the debug dump where the Jun 2026 ego_notification
+            # misroute silently landed). Log so a missing mapping surfaces
+            # instead of vanishing.
+            logger.warning(
+                "No topic mapping for outreach category %r — routing to "
+                "'conversation'. Add an explicit mapping if this is real.",
+                outreach_category,
+            )
+            return "conversation"
+        return topic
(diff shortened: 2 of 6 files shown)


WingedGuardian (commented):

### Structural: one NOTE

**NOTE — private symbols imported cross-module** (`src/genesis/ego/session.py`)

```python
from genesis.eval.scorers import (
    _JUDGE_CALL_FAIL,
    _JUDGE_PARSE_FAIL,
    get_scorer,
)
```

`_JUDGE_CALL_FAIL` and `_JUDGE_PARSE_FAIL` are underscore-prefixed, signalling "internal to this module, don't import me." The real risk here isn't stylistic: if those names are ever renamed inside `genesis.eval.scorers`, the import won't error — it'll silently set `judge_error` to a value that never matches, and every judge-infrastructure failure will hold proposals instead of passing them through. That's exactly the bug this PR is fixing. Consider making the constants public (`JUDGE_CALL_FAIL`) or adding them to `__all__` to make the cross-module contract explicit.

No other structural findings — no bare `except:`, no `/tmp/`, no `task_states` direct inserts, no `nohup`, no mutable defaults. Import grouping is clean. All modified source paths have test coverage.

---

No credentials, private IPs, SSH keys, personal emails, or private hostnames in the diff. PR title and body describe a feature regression fix — no language that advertises a security incident or exposes infrastructure details.

---

PR #659, opened 2026-06-16
fix(deps): declare tzlocal (apscheduler local-tz path) for clean CI installs

WingedGuardian:
**Root cause.** `apscheduler>=3.11` lists `tzlocal`, but a clean/cached CI build can satisfy apscheduler without pulling tzlocal. `ego/cadence.py` uses the local-timezone path (`user_timezone()` → APScheduler), so a fresh-cache run hits **83 `No module named 'tzlocal'` collection errors in `tests/ego`**. `main` is green only because its runner cache still has tzlocal; any fresh PR (e.g. #657) trips it.

**Fix.** Declare `tzlocal>=5` directly in `pyproject.toml` (same pattern as the existing `wsproto` transitive pin). Makes the dep explicit for code that genuinely needs local-tz resolution, and the pyproject change busts the stale cache so the install resolves it.

No behavior change; one dependency line.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,6 +6,7 @@ dependencies = [
     "aiosqlite",
     "aiohttp<3.14",  # 3.14 adds required stream_writer kwarg to ClientResponse; aioresponses 0.7.8 doesn't pass it
     "apscheduler>=3.11,<4",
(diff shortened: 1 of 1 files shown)


WingedGuardian:

---

PR #662, opened 2026-06-16
fix(scheduler): strip GitNexus block from CLAUDE.md after reindex (keep in AGENTS.md)

WingedGuardian:
GitNexus's `analyze` step injects a `<!-- gitnexus:start --> … <!-- gitnexus:end -->` block into **both** CLAUDE.md and AGENTS.md — there's no per-file flag (`--skip-agents-md` skips both; `--index-only` skips all injection). We want it in **AGENTS.md** (read by cross-tool agents) but **not** in Claude Code's CLAUDE.md.

This adds a post-reindex strip: after a successful `gitnexus analyze`, the scheduled reindex removes the gitnexus block from CLAUDE.md only (AGENTS.md keeps it). Marker-delimited, idempotent, preserves surrounding content. 5 unit tests; ruff clean.

Note: this is also why #652's drop of `--index-only` is fine — we *want* analyze to keep injecting (so AGENTS.md stays fresh); we just post-strip CLAUDE.md.

### DIFF
--- a/src/genesis/surplus/scheduler.py
+++ b/src/genesis/surplus/scheduler.py
@@ -6,6 +6,7 @@ import asyncio
 import contextlib
 import hashlib
 import logging
+import re
 from datetime import UTC, datetime, timedelta
 from pathlib import Path
 from typing import TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


WingedGuardian (commented):

---

PR #663, opened 2026-06-16
refactor(ego): remove LLM quality gate from proposal pipeline (eval-only judge)

WingedGuardian:
## What & why

Removes the LLM "quality gate" from the ego **proposal** pipeline. The `OUTPUT_QUALITY` judge is an eval-harness primitive that had been mis-applied as a second gate on live proposals (`EgoSession._quality_gate`). It was:
- **redundant** — the Opus realist (`_filter_proposals`) is already the proposal gate;
- **mis-wired** — `quality_hold` only filtered digest *visibility*, never execution, despite a docstring claiming otherwise;
- **fragile** — it ran on a DeepSeek-only `judge` chain; when that chain exhausted (Jun 2026), it stamped *every* proposal `quality_hold` and silently dropped them from the digest (the proposal-silence outage), overwriting the realist's verdict in a column named for the realist.

After this, the **Opus realist + domain-boundary checks are the sole proposal gates**.

## Changes
- Delete `_quality_gate` + its single call in `_filter_proposals` (session.py) → `return filtered` directly.
- Remove both `quality_hold` filters in `send_digest` (proposals.py).
- Update the `output_quality` rubric docstring → it's an eval-harness scorer, not a proposal gate.
- Tests: delete the obsolete `test_quality_gate.py` (fail-open tests) and `TestQualityGateIntegration` (inline filter tests); **add** a regression test that `send_digest` delivers regardless of `realist_verdict` (guards against the gate creeping back).

## Preserved (untouched)
The `OUTPUT_QUALITY` scorer/rubric, `ScorerType`, and the `judge` router profile — still used by the eval harness. The unrelated autonomy-executor `record_quality_gate` (task-deliverable verification) is intentionally left alone.

## Verification
- `ruff` clean; targeted + full blast-radius suite (ego, eval, executor, outreach) = **1177 passed, 1 skipped**.
- Independent code review (Claude reviewer; Codex rate-limited): clean, no findings.
- Confirmed no residual `quality_hold`/`_quality_gate` references in `src/genesis/ego/`; eval harness imports intact.
- Live E2E (a real ego cycle delivering proposals gated only by the realist) to be observed post-deploy.

Follows #658 (which shipped the interim quality-gate fail-open this supersedes).


WingedGuardian:
## First-pass code review — PR #663

**Structural review:** CLEAN — no pattern violations (ERROR) found.

---

### NOTE — regression test key name worth a sanity check (`tests/ego/test_proposals.py:352`)

The new regression test populates proposals with `p["_realist_verdict"] = "quality_hold"` (underscore-prefixed, the transient in-memory key), but the old `send_digest` filter read from `realist_verdict` (no underscore — the DB column):

```python
# old filter (now deleted):
proposals = [p for p in proposals if p.get("realist_verdict") != "quality_hold"]
```

If `create_batch` maps `_realist_verdict` → `realist_verdict` in DB storage (the standard Genesis underscore-strip convention), then `send_digest` loads proposals with `realist_verdict = "quality_hold"` and the test correctly guards against the exact original bug. Likely fine given the convention and the passing suite.

If `create_batch` doesn't strip the underscore prefix, the test only guards against a `_realist_verdict`-keyed filter reappearing — not the `realist_verdict`-column-keyed one that caused the Jun 2026 silence. A belt-and-suspenders fix would be to also (or only) set `p["realist_verdict"] = "quality_hold"` to be unambiguous, independent of DB round-trip behavior.

Not a blocker — just worth a quick confirm that the DB mapping makes the test faithful to the original bug path.

---

PR #664, opened 2026-06-17
fix(guardian): gateway-safe host CLAUDE.md regen, drop empty network block

WingedGuardian:
## What

The host-side Guardian keeps a generated `CLAUDE.md` (from `config/guardian-claude.md`, symlinked into the diagnostic CC work_dir) that the diagnostic Claude Code session loads as project context. Two bugs in how it's regenerated:

1. **skip-worktree wedged gateway updates.** `install_guardian.sh` marked `CLAUDE.md` with `git update-index --skip-worktree`. With that bit set, the gateway `update` verb's `git checkout -- CLAUDE.md` errors and `git pull --ff-only` aborts ("local changes would be overwritten") the moment upstream touches the tracked `CLAUDE.md` — silently stalling every host Guardian update (returns `{"ok":false,"git pull failed"}`).

2. **Half-empty, redundant network block.** Both gateway verbs (`update`, `redeploy`) and the installer appended a `## Network` block built from `guardian.yaml`, but `container_ip` is empty by design (auto-detected) and `host_ip` isn't a key there, so it always produced `- **Container**: <name> at ` / `http://:5000`. It's also redundant: shared host/container facts live in the user-level `~/.claude/CLAUDE.md` (the D16 inheritance design), which the diagnostic CC also loads.

## Fix

- `install_guardian.sh`: stop setting `--skip-worktree`; drop the network-append. (Also removes a `cd` side-effect — verified nothing later in the script depends on cwd.)
- `guardian-gateway.sh` `update`: clear skip-worktree before the checkout so legacy installs self-heal; drop the network-append.
- `guardian-gateway.sh` `redeploy`: drop the network-append.
- The regenerated file is now exactly `config/guardian-claude.md`.

## Tests

`tests/test_scripts/test_guardian_gateway_update.py` runs the **real** gateway `update`/`redeploy` against a throwaway git clone (`systemctl`/`sudo` stubbed), parametrized over skip-worktree set/unset. Confirmed **RED** on the unfixed gateway (pull abort + spurious block), **GREEN** after. Existing `update-cc` gateway tests still pass (15/15); shellcheck + `bash -n` clean.

## Review

Code-reviewer pass: core fix verified correct; the `cd`-removal and consumer-grep claims confirmed. One pre-existing latent issue (failure-path `git stash` restore in the `update` else-branch) tracked as a follow-up — out of scope here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,12 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Guardian host updates no longer silently stall** — on hosts where the
+  Guardian's `CLAUDE.md` had been pinned with git's skip-worktree flag, the
+  Guardian's self-update (`git pull`) would abort the moment that file changed
+  upstream, quietly leaving the host Guardian stuck on old code. The update now
+  clears the flag first, so existing installs self-heal and stay current.
+
 - **Telegram `/stop` now stops your session, not a background task** — when a
   background task (reflection, inbox, an ego session, etc.) was running at the
   same time as your chat, `/stop` could interrupt the wrong one. Each session's
--- a/scripts/guardian-gateway.sh
+++ b/scripts/guardian-gateway.sh
@@ -100,6 +100,10 @@ PYEOF
             # repo's version would overwrite with the wrong content (container
             # dev instructions vs Guardian diagnostic instructions). We
             # regenerate it from config/guardian-claude.md after every pull.
+            # Clear skip-worktree first: legacy installs marked CLAUDE.md
+            # --skip-worktree, which wedges `git pull` ("local changes would be
+            # overwritten") the moment upstream touches the tracked CLAUDE.md.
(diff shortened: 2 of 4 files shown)


WingedGuardian:
## First-pass code review — PR #664

### Structural review

**ERROR (must fix):**

`git add -A` in `tests/test_scripts/test_guardian_gateway_update.py`, inside `_seed_repo()` (~line 107):

```python
_git("add", "-A", cwd=seed)
```

Pattern rule: `git add -A` is prohibited — must stage specific files. In this test-fixture context the temp repo only contains the three files explicitly written (`CLAUDE.md`, `config/guardian-claude.md`, `config/guardian.yaml`), so there is no actual risk — but the pattern is an unconditional ERROR regardless of context.

Fix (trivial):
```python
_git("add", "CLAUDE.md", "config/guardian-claude.md", "config/guardian.yaml", cwd=seed)
```

---

**CLEAN** — no credentials, tokens, private IPs, private hostnames, or problematic wording found in added lines, PR title, body, or CHANGELOG.

---

PR #665, opened 2026-06-17
fix(surplus): keep quotes raw in Telegram surplus posts

WingedGuardian:
## What & why

Surplus-topic messages rendered literal `&quot;` / `&#x27;` instead of `"` / `'`.

`surplus/executor.py` `_post_to_telegram` escaped the body with `html.escape(content)` (default `quote=True`) and sent it `parse_mode="HTML"`. Telegram's HTML parser only unescapes `&lt;`/`&gt;`/`&amp;` in text content, so `&quot;`/`&#x27;` showed verbatim — surplus JSON/findings looked like raw entity soup.

## Change
- `escape(content[:2000])` → `escape(content[:2000], quote=False)`: still escapes `&`/`<`/`>` (needed so the `<b>Surplus: …</b>` label parses + tag-injection safety preserved), but leaves `"`/`'` raw so they render correctly.
- Adds a regression test (`test_post_to_telegram_keeps_quotes_raw`) that fails on the pre-fix code.

Single active surplus delivery path (the OutreachScheduler `_surplus_outreach_job` is unregistered). ruff clean; 32 surplus-executor tests pass. Independent review: clean.

### DIFF
--- a/src/genesis/surplus/executor.py
+++ b/src/genesis/surplus/executor.py
@@ -751,9 +751,13 @@ class SurplusLLMExecutor:
             from html import escape
 
             label = task.task_type.replace("_", " ").title()
+            # quote=False: keep &/</> escaped (needed so the <b> label parses under
+            # parse_mode=HTML) but leave " and ' raw. With quote=True, html.escape turns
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

PR #666, opened 2026-06-17
fix(guardian): keyword-reply recovery approval (replace localhost link)

WingedGuardian:
## What & why

The Guardian's "click to approve recovery" Telegram alert pointed at a localhost HTTP `ApprovalServer` (`http://<ip>:8888/...`) — **not reachable from a phone**, so recovery approval was effectively dead text. This replaces it with a **blocking, two-gate, keyword-reply** approval the host-side Guardian reads via the Telegram `getUpdates` API. No separate bot — it reuses the main Genesis bot token.

## Design

- **Gate everything.** Recovery requires an explicit `APPROVE` reply to **both** gates: Gate 1 authorizes investigation; Gate 2 authorizes the specific proposed action. `DENY` at either gate → no action. Nothing auto-recovers.
- **No fixed timeout.** The wait self-cancels **only** when health returns on its own (the down-episode is moot). Safe because the Guardian is a `Type=oneshot` service — no concurrent instance can exist while it blocks.
- **Shared-token safe.** `_poll_for_keyword_sync` sends **no `offset`** to `getUpdates` (advancing it would consume updates the main bot needs) and filters on `reply_to_message.message_id == gate_message_id`, so it only honours replies to its own prompt. A 409 (main bot alive on the same token) surfaces as a "can't confirm Genesis is down" alert, not silence.
- **Propose-only CC.** The diagnosis LLM investigates and proposes via `recommended_action`; it never executes. `_handle_cc_resolved` is a firewall — any CC "resolved" claim is routed back through the human gate, never auto-confirmed.
- Removes the localhost `ApprovalServer` (`approval.py`, `ApprovalConfig`, the `approval:` block in `guardian.yaml` + `install_guardian.sh`).
- `--test-approval` + a gateway `test-approval` command for host E2E.

## Review

- **Codex** attempted first (adversarial) — rate-limited until Jul 1, so fell back to the Claude code-reviewer.
- Reviewer verdict: **no P1** — core two-gate mechanism, no-offset stale-reply safety, 409 handling, and the CC firewall all verified sound. Three P2s found and **fixed in this PR**: (1) a failed gate-prompt send now clears `down_alert_sent` so a transient Telegram failure can't permanently mute the Guardian; (2) explicit `APPROVE` guards so an unexpected gate reply can never fall through to `execute()`; (3) added the missing flag-state test assertion.

## Tests

`tests/test_guardian/` — **366 passing** locally (`PYTHONPATH=src pytest tests/test_guardian/`). New coverage: keyword-gate poll (reply-to filter, 409 sentinel, no-offset safety), two-gate flow (approve→approve→execute, deny@1, deny@2, self-recover, escalate-no-gate, no-telegram, gate-1/2 send-failure), and propose-only diagnosis. Updated `test_cc_resolved_*` to assert the new firewall contract.

## Verification status

Code review ✅, ruff ✅, targeted tests ✅. **Host E2E pending** — will run `--test-approval` over SSH to the Guardian host (send prompt → reply APPROVE/DENY → confirm the keyword is read) before merge.

### Trade-off noted
Clearing `down_alert_sent` on a failed gate-prompt send means that during a *persistent* Telegram outage (with Genesis also down), the Guardian re-attempts (incl. re-diagnosis) every 30s rather than going silent. This is intentional — a silent watchdog is worse than retrying to reach the user — but flagging it for awareness.


WingedGuardian (commented):

---

PR #667, opened 2026-06-17
fix(memory): stop reference auto-classifier fabricating refs from prose

WingedGuardian:
## Problem
The 2h `memory_extraction` job runs `classify_as_reference()` (regex) over LLM-generated extraction prose and was fabricating credential/network reference-store entries from ordinary text — e.g. a credential label + a bare space + the next word ("can pin checkpoints" → a credential valued "checkpoints"), and an incidental IP inside code-analysis prose → a network reference. Several junk rows reached the live reference store.

## Fix
Two precision gates in `reference_extraction.py`:
- **Credential labels require an explicit separator** (`:`, `=`, or `is `) on the value side of the pair/email/account/single patterns and the labeled-token pattern — not a bare space. Bare-word prose no longer fabricates a value. The user/email label sides stay permissive (they cannot fabricate a value on their own — the value side gates them).
- **Network matches require a context word within 40 chars** of the IP/env-var match, matched on word boundaries (not a whole-content substring scan). This also retires the old env-var "strip the match" hack.

Genuine captures are unaffected: explicit-separator credentials, adjacent-context IPs, known key prefixes, and URLs all still classify.

## Tests
Adds `test_reference_extraction_precision.py` (10 tests): regression cases for every false-positive class (bare-space credential, labeled-token, pair-prose, incidental-IP) driven RED→GREEN; guards that genuine refs still capture; and an end-to-end mixed-chunk test (4 junk + 2 genuine → only the 2 genuine ingested). Full `test_memory` extraction suite stays green; ruff clean.

## Known residual
Colon-prose like `pin: word` can still match (rare, pre-existing — the `pin` label is inherently ambiguous). Tracked as a follow-up.

### DIFF
--- a/src/genesis/memory/reference_extraction.py
+++ b/src/genesis/memory/reference_extraction.py
@@ -10,10 +10,12 @@ the reference store via :func:`ingest_knowledge_unit`.
 This is the SILENT AUTO-CAPTURE path — no user prompt, no explicit flag.
 Runs on the existing extraction cadence with zero new LLM calls.
 
-Classification errs on the side of capturing more:
-- False positives are harmless (unused entries sit idle). False negatives
-  lose credentials permanently. Multiple overlapping patterns are
-  intentional — if any one fires, we capture.
+Classification balances recall against precision:
+- Genuine references are captured broadly, but the credential/network kinds
+  require precise signals (an explicit separator after a credential label;
+  a network context word adjacent to the IP). They run over LLM-generated
+  prose, where a loose match fabricates a misleading value rather than
+  sitting idle harmlessly.
 - The primary capture path is the LLM calling ``reference_store`` in
   real-time. This background path is a safety net that catches things
   the primary path missed.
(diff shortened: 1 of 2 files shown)


WingedGuardian:
## Code Review — PR #667

Import order in `test_reference_extraction_precision.py` is correct (stdlib → third-party → local). No bare `except:`, no mutable defaults, no raw SQL against genesis.db, no `/tmp/` usage, no systemd/nohup violations. The new test file provides coverage for the modified `reference_extraction.py`. Conventional commit prefix `fix(memory):` present ✓

---

**PII/Secrets/Wording scan:**

**SOFT WARNING** — Private IP address in test fixtures

`tests/test_memory/test_reference_extraction_precision.py`:
- Line ~132: `"Server runs on host 192.168.1.50 listening for connections"` (in `test_ip_with_adjacent_context_still_network`)
- Line ~206: `"The Genesis API server runs on host 192.168.1.50 for embeddings"` (in `test_mixed_chunk_ingests_only_genuine_references`)

`192.168.1.50` falls in the private IP range (192.168.x.x). In context these are clearly synthetic strings used to test that the classifier correctly accepts genuine network references — not real infrastructure IPs. Flagging per policy; no action required if this is known-safe example data.

All other test credential strings (`Hunter2!xyz`, `verysecretword`, `Sup3r!secret`, `buckeye614`) are obviously synthetic classifier test fixtures. The `203.0.113.45` addresses are RFC 5737 TEST-NET-3 documentation IPs — clean.

**Wording scan: CLEAN** — the PR title and description describe a classifier precision fix without language that signals a credential/PII exposure incident (e.g. "fixed leaked credentials", "removed exposed IP"). The phrase "junk rows reached the live reference store" describes fabricated-by-regex entries, not a genuine data leak — no git-history wording concern.

---

PR #669, opened 2026-06-17
fix(guardian): harden gateway self-update + add recovery verb

WingedGuardian:
## Problem

The host Guardian's gateway `update` verb ran best-effort host tuning (sudo sysctl/udev) **unguarded** under `set -euo pipefail`. On a host with **passwordless sudo**, an errant `sudo tee` returned non-zero and `set -e` aborted the script **before the final `{"ok":true}` printf** — so `update` exited 1, swallowed its JSON result, and (because the deployed-gateway self-update lived in that same branch) the deployed `~/.local/bin/guardian-gateway.sh` **silently froze**. The install dir kept pulling current via the git-pull step, while the deployed gateway went stale for months — masked entirely because every update *looked* like a failure and nothing monitored deployed-gateway version.

Reproduced in a scratch harness: `sudo -n true` passes + a `sudo` tuning command fails → "Already up to date." + exit 1 + no JSON (matches the live host exactly); with sudo unavailable (block skipped) → exit 0 + clean JSON.

## Fix

- **`update`:** self-update the deployed gateway **first** after a successful pull (so no later best-effort step can leave it stale), and guard every best-effort step (`|| true`) so tuning failures can't abort the update or suppress the success result. The `_HOST_RAM_KB` arithmetic is guarded against empty/odd `/proc/meminfo`.
- **`update`:** record `deployed_commit` in `deploy_state.json` — previously only `redeploy` wrote it, so the watchdog's drift detection silently skipped (`deployed_commit == "unknown"`) for pull-based installs.
- **`version`:** report `gateway_sha` (sha256 of the deployed gateway) so a stale/frozen gateway is detectable from the container.
- **New `sync-gateway` verb:** redeploy the gateway from the (already-pulled) install dir without a git pull — a clean recovery lever when the self-update path is unavailable. No git, no sudo; idempotent copy of a repo-tracked file.
- **`install_guardian.sh`:** same DRY guards on its best-effort sudo tuning; clamp lines converted to `if`-form to remove `set -e` ambiguity.

## Testing

- TDD: 4 new tests, RED on pre-fix code, GREEN after. The key one (`test_update_self_updates_and_succeeds_when_sudo_tuning_fails`) exercises the **passwordless-sudo-failure path that the prior test suite skipped** (its `sudo` stub made `sudo -n true` fail, bypassing the whole block — which is why the bug survived review). Plus `deployed_commit` / `sync-gateway` / `gateway_sha` coverage. `_seed_repo` now seeds the gateway script + `~/.local/bin` so self-update is actually exercised.
- `bash -n` clean on both scripts; `ruff` clean; 7/7 tests pass.
- Adversarial review: Codex (quota-exhausted → Claude code-reviewer fallback). One flagged "P1" (clamp lines aborting under `set -e`) was empirically falsified (the compound's non-zero from an `-e`-ignored `&&` left-operand does not exit; verified on a 32 GiB host) — clamp lines converted to `if`-form anyway for clarity; two nits fixed.

## Follow-up (PR2)

Container-side watchdog `gateway_sha` staleness monitoring + `GuardianRemote.sync_gateway()` wiring for autonomous recovery (consumes the signals this PR emits).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   upstream, quietly leaving the host Guardian stuck on old code. The update now
   clears the flag first, so existing installs self-heal and stay current.
 
+- **Guardian host self-updates are now reliable on hosts with passwordless sudo**
+  — an unguarded step while refreshing kernel tuning could make the Guardian's
+  self-update abort partway, so it reported a failure (and could leave its own
+  updater script frozen on old code) even though the code pull had already
+  succeeded. Kernel tuning is now strictly best-effort and can't derail the
+  update; the update reliably refreshes the updater first and records what it
+  deployed; and there's a new one-step recovery path to refresh a stalled updater.
+
 - **Telegram `/stop` now stops your session, not a background task** — when a
   background task (reflection, inbox, an ego session, etc.) was running at the
   same time as your chat, `/stop` could interrupt the wrong one. Each session's
--- a/scripts/guardian-gateway.sh
+++ b/scripts/guardian-gateway.sh
@@ -9,6 +9,7 @@
 #   reset-state    — reset stuck state machine to healthy
 #   version        — report CC, code, and Node versions
 #   update         — pull latest code + self-update gateway script
+#   sync-gateway    — redeploy gateway script from install dir (no pull; recovery)
 #   redeploy <hash> — receive tar archive on stdin, deploy to install dir
 #   update-cc <ver> — install a pinned Claude Code version (validated semver)
 #   test-approval   — E2E test the keyword-reply approval gate (no recovery)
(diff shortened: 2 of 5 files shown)


WingedGuardian (commented):

---

WingedGuardian:
**Review status:** Automated Codex review is quota-exhausted (see the connector comment above), so adversarial review was performed locally with the Claude code-reviewer subagent (Codex-first attempted, fallback per protocol). Findings: one flagged "P1" (clamp lines aborting under \`set -e\`) was empirically falsified — the compound's non-zero from an \`-e\`-ignored \`&&\` left-operand does not exit (verified by the bash manual + a direct micro-test on a 32 GiB host); converted to \`if\`-form anyway for clarity. Two nits fixed (\`sync-gateway\` JSON-on-failure, \`NEW\` fallback). CI: all 8 checks green (\`test\` incl. the new passwordless-sudo regression test).

PR #671, opened 2026-06-17
fix(restore): make SQLite restore corruption-safe

WingedGuardian:
## Problem

`scripts/restore.sh` rehydrates the live SQLite DB during disaster recovery — the **highest-stakes path in the system** (a bug here destroys data *on recovery*). Three latent corruption risks:
- It `rm`'d **only** the DB file, leaving stale `-wal`/`-shm` sidecars → SQLite would replay the leftover WAL onto the freshly-restored DB and corrupt it.
- It ran **no integrity check** after `.read` → a corrupt restore was silently accepted.
- It **never stopped `genesis-server`** → restoring under a live writer holding the DB could corrupt the result.

## Fix (WS-2, part 1 of the DR-safety track)

- **Quiesce the writer:** stop `genesis-server` before swapping the DB. Guarded (`command -v systemctl` + `is-active --quiet`) so a fresh-box DR with no `systemctl` / inactive unit is a clean no-op; only records "stopped" if the stop actually succeeded. **No auto-restart** — the operator verifies the restore first (an end-of-run note prints the start command).
- **Clear stale WAL/SHM:** `rm -f "$DB_FILE" "$DB_FILE-wal" "$DB_FILE-shm"`.
- **Integrity-check the result:** `PRAGMA integrity_check` after `.read`; `warn` loudly + record a failure (→ non-zero exit) if not sound.

The pre-existing `.pre-restore.<ts>` backup of the old DB is preserved.

## Testing

TDD, fully sandboxed (`HOME` + `GENESIS_DIR` both overridden → the live `~/genesis/data/genesis.db` is unreachable; real `sqlite3`, stubbed `systemctl`). 6 tests, RED→GREEN:
- server stopped before restore + **not** restarted; stale `-wal`/`-shm` cleared; `integrity_check ok` on the happy path;
- **failed-stop** proceeds with a warning and **no** misleading "left stopped" note; **inactive** server skips the stop; **corrupt** restore (intercepted `integrity_check`) → warn + recorded failure + non-zero exit.
- `bash -n` + `shellcheck -S warning` clean; `ruff` clean. An E2E sandbox run shows the full behavior end-to-end.

## Review

[...]

PR #673, opened 2026-06-18
feat(dr): off-site recovery — dated NAS snapshots + restore-from-NAS (2a)

WingedGuardian:
## Problem

The two-tier backup stores the large data (SQLite DB, Qdrant snapshots, transcripts) **only** on the off-site NAS — gitignored from Tier-1. But `restore.sh` had **no NAS code**, so on a fresh DR box it could recover memory/config/secrets but **not the database or vector memory**. The off-site tier was effectively *write-only* — the most important state was stranded.

## Fix (DR upgrade 2a)

- **`backup.sh`** uploads Tier-2 to per-run **dated snapshot dirs** `Genesis/<host>/<UTC-stamp>/{data,qdrant,transcripts}/…` (was fixed filenames overwritten each run), now including **transcripts** (previously local-only). After every upload succeeds it writes a **`COMPLETE`** marker.
- **`restore.sh`** gains a NAS-pull (after the git-pull, before restore): finds the **latest `COMPLETE`** snapshot (`grep` the stamp pattern out of `smbclient ls` — robust to format — newest first, skipping any half-uploaded snapshot), and pulls its `data`/`qdrant`/`transcripts` into `BACKUP_DIR`. The integrity-check from #671 then guards the pulled DB. **This closes the off-site-recovery gap.**

## Safety / robustness
- **No half-restores:** a crashed-mid-upload snapshot lacks `COMPLETE`, so restore falls back to the previous good one (never restores partial data); a failed SQL pull `warn`s → non-zero exit (loud, not silent).
- **`set -e`/`pipefail`:** grep-over-empty-listing pipelines guarded (a bug found + fixed during the build); staging dirs created only when a file is pulled (an empty Qdrant set must not trigger the Qdrant probe).
- **Creds:** NAS creds file (chmod 600, never on cmdline) cleaned up via a `RETURN` trap on any function exit.
- **Same-box re-run:** NAS pull skipped if a local payload already exists unless `--force`.

## Testing

TDD, fully sandboxed (`HOME`+`GENESIS_DIR` → tmp; real `sqlite3`/`gpg`/`git`, stubbed `smbclient`/`systemctl`). 18 DR-script tests:
- **Real E2E:** a genuine GPG-encrypted dump served by the smbclient stub is pulled → decrypted → restored (DB content verified), latest-of-two snapshots selected; qdrant file staged via the pull loop.
- **Incomplete-latest snapshot → falls back** to the older COMPLETE one; no-complete → skips; no-NAS → skips cleanly.
- backup: dated paths + snapshot-tree mkdir; plus the existing restore-safety (6) + DR-04 (5) suites still green.
- `bash -n` + `shellcheck -S warning` + `ruff` clean.

## Review

Codex quota-exhausted → Claude code-reviewer (fallback). Two P1s it found (partial-snapshot data correctness; creds leak) are fixed via the COMPLETE sentinel + `RETURN` trap; the P2s (header docs, warn-on-miss, local-payload skip) and test gaps (incomplete fallback, qdrant loop) are addressed above.

Next: GFS retention + local prune (2b), cloud Tier-2 via rclone (2c).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Off-site backups can now actually be restored** — the large data (the
+  database, vector memory, and transcripts) is stored only on your off-site
+  (NAS) target, but the restore tool had no way to fetch it — so a from-scratch
+  recovery silently couldn't bring back your database or memory. Restore now
+  pulls the latest off-site snapshot before restoring, and backups are written
+  as dated point-in-time snapshots (so you can recover a *specific* run, not just
+  the last one) with transcripts included off-site too.
+
 - **A backup that can't reach off-site storage no longer fails silently** — if
   you've configured an off-site (NAS) backup target and a run captures your data
   locally but can't replicate it off-site, Genesis now sends a distinct alert
--- a/scripts/backup.sh
+++ b/scripts/backup.sh
@@ -315,6 +315,7 @@ if [ -n "$_NAS_TARGET" ]; then
         _SMB_CREDS=$(mktemp)
         chmod 600 "$_SMB_CREDS"
         printf 'username=%s\npassword=%s\n' "$_NAS_USER" "$_NAS_PASS" > "$_SMB_CREDS"
+        _SMB_COMPLETE=$(mktemp)  # empty marker, uploaded only after a full snapshot
 
         _smb() { smbclient "$_NAS_TARGET" -A "$_SMB_CREDS" "$@"; }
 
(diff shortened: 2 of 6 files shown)


WingedGuardian:
Thanks @codex — both were real and important:

- **P1 (source-host NAS path):** fixed in `68dd0eab`. restore.sh was using the *restore* box's `$(hostname)` for the NAS path, which breaks the exact fresh-box DR case. Now honours `GENESIS_BACKUP_NAS_HOST` (documented in the header) **and** auto-detects when there's a single host dir on the NAS, so fresh-box DR works without manual config. Tests: explicit-override + sole-host auto-detect.
- **P2 (COMPLETE upload failure):** fixed. A failed COMPLETE marker upload (payloads OK) now flips `_T2_OK=false` → `tier2_status=partial` + `offsite_confirmed=false` + the off-site alert, since restore would skip a snapshot without that marker. Test added.

`shellcheck`/`bash -n`/`ruff` clean; 21 DR-script tests green.

PR #674, opened 2026-06-18
fix(memory): gate reference auto-capture to user-shared values + mark provenance

WingedGuardian:
## Problem
The 2h `memory_extraction` job auto-captures references by running the regex classifier (`classify_as_reference`) over LLM-extracted prose from **both** user and Genesis turns. Most of the resulting junk is Genesis's *own* analysis prose (e.g. a sentence about `HybridRetriever.recall()` producing a fabricated `password: rerank`), stored indistinguishably from references the user actually shared.

## Fix (root cause)
1. **Gate auto-capture to the user's own words.** `classify_as_reference` and `extract_references_from_chunk` accept `user_text` (the chunk's user-authored content). A match is captured only if its **raw matched token** (password / IP / URL / key) appears in `user_text` (case-insensitive substring — concrete strings the LLM preserves verbatim). `extraction_job` builds `user_text` from the chunk's `role == "user"` messages. `user_text=None` disables the gate (back-compat for direct callers/tests).
2. **Mark auto-captured entries distinctly.** They get the `auto_captured` tag and `confidence=0.60` (vs `0.85` for real-time `reference_store` captures), so manual/verified vs auto/unverified is separable everywhere downstream. Adds a `confidence` param to `ingest_knowledge_unit` (default `0.85`; other callers unaffected).

Scoped to **references only** — general knowledge / procedure / episodic extraction is untouched (the gate is applied at the reference-classification step, not the chunk level).

## Tests
`test_reference_extraction_gating.py` (12 tests): per-kind gating (credential/url/network captured when user-shared, dropped when only in Genesis prose), case-insensitivity, `auto_captured` + 0.60 marking, gate-blocks-ingest, and a mixed-chunk E2E. Existing #667 precision suite + `test_extract_references_from_chunk_counts` stay green; ruff clean.

## Notes
- API additions are backward-compatible (new params default to prior behavior).
- Pre-existing, out of scope: the MCP `_ingest_knowledge_unit` wrapper doesn't yet expose `confidence` (no current caller needs it).

### DIFF
--- a/src/genesis/memory/extraction_job.py
+++ b/src/genesis/memory/extraction_job.py
@@ -212,6 +212,12 @@ async def run_extraction_cycle(
             # modes — this is the dominant path that populates the
             # reference store from historical conversations.
             try:
+                # Gate reference auto-capture to the USER's own words — the
+                # classifier must not mine Genesis's analysis prose (the
+                # dominant source of junk refs) for credentials/IPs/URLs.
+                chunk_user_text = "\n".join(
+                    m.text for m in chunk if m.role == "user"
+                )
                 ref_count = await extract_references_from_chunk(
                     result.extractions,
                     store=store,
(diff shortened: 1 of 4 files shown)


WingedGuardian (commented):
PII/Secrets/Wording scan: CLEAN — test fixture passwords ("Hunter2!xyz" etc.) and IPs (203.0.113.x TEST-NET-3 range) are fabricated test data; PR wording describes a classifier precision fix, not a credential exposure.

---

PR #676, opened 2026-06-18
feat(dashboard): reference store browser, retire known-to-genesis.md mirror

WingedGuardian:
## Summary

Replaces the `~/.genesis/known-to-genesis.md` markdown mirror with a live, access-controlled **References** tab on the dashboard. The flat-file mirror was a denormalized cache: prone to going stale (regen triggers were easy to miss) and to leaking plaintext secrets into backups. This makes the reference store itself the single source of truth, rendered behind auth with an explicit reveal step for secret values.

## What changed

- **New `memory/reference_ops.py`** — neutral domain module shared by the MCP tool layer and the dashboard:
  - `parse_reference_body()` — inverse of the body formatters; the security boundary. Returns only `{description, value}` and **fails closed** (a body without the canonical marker yields empty fields, never the raw body).
  - `delete_reference_entry()` — one delete path (SQLite row + FTS + Qdrant across both collections); the `reference_delete` MCP tool now delegates to it.
- **New `dashboard/routes/references.py`** — `list` / `search` / `<id>` / `<id>/reveal` / `delete` / `stats` / `kinds`. `list`/`search`/`detail` return only the description; the value is reachable solely via the explicit `POST /reveal`.
- **References tab UI** (`genesis_dashboard.html`) — grouped-by-kind browse, kind filter, search, a provenance badge (verified vs auto-captured, by `source_pipeline`), reveal toggle with copy, and confirm-gated delete.
- **Retires the mirror** — deletes `memory/reference_mirror.py`, its boot refresh, and their tests; updates the docs pointer.

## Security

- Every reference route is explicitly gated with `is_authenticated()`. The dashboard's global `before_request` intentionally bypasses auth for `/api/*` (machine callers), so these human-facing routes carry their own gate (a no-op when no dashboard password is set; enforced 403 when one is). The reveal endpoint, which returns plaintext credentials, is gated the same way.
- For `credentials` entries, any occurrence of the secret value inside the free-text description is masked in non-reveal payloads — closing a redaction gap the old mirror had. Non-secret kinds (IPs, URLs) stay readable in their own descriptions.

## Testing

- New unit + route tests (parser contract against both real body formatters; fail-closed; delete guard; auth gate; value-never-in-non-reveal; provenance fields). Full affected-area sweep green (241 tests); ruff clean; the embedded dashboard script passes `node --check`.
- End-to-end data-path check over the entire live reference store: every entry parses, reveal recovers the exact stored value, and **no credential value appears in any non-reveal response**.

## Notes for review

- `db/crud/knowledge.py`: `list_by_domain` rows are now self-describing (carry `domain`, `source_pipeline`, `confidence`); `search_fts` returns `ingested_at`/`confidence` via its existing join. Both additive.
- After merge: a server restart picks up the new routes; the `known-to-genesis.md` file can then be removed (it is no longer written).


WingedGuardian (commented):
## First-pass review — PR #676

### Structural review

**ERRORs: none.**

**WARNING — split imports from the same module** (`src/genesis/mcp/memory/knowledge.py`)

The replacement for the old `reference_mirror` import splits three names from the same source module into three separate `from … import (…)` blocks:

```python
from genesis.memory.reference_ops import (
    REFERENCE_KINDS as _REFERENCE_KINDS,
)
from genesis.memory.reference_ops import (
    REFERENCE_PROJECT as _REFERENCE_PROJECT,
)
from genesis.memory.reference_ops import (
    delete_reference_entry,
)
```

Should be one block:

```python
from genesis.memory.reference_ops import (
    REFERENCE_KINDS as _REFERENCE_KINDS,
    REFERENCE_PROJECT as _REFERENCE_PROJECT,
    delete_reference_entry,
)
```

No other structural findings — no nohup, no task_states direct inserts, no `/tmp/` usage, no bare `except:`, no mutable default arguments, no raw SQL against genesis.db outside the crud layer. New source files (`reference_ops.py`, `references.py`) both have corresponding test files. Public functions carry docstrings.

---

**Hard blocks: none.**

Test fixtures use clearly fake credentials (`admin / Hunter2!xyz`, `pw / secret`) and TEST-NET-3 IPs (`203.0.113.x`, RFC 5737) — not real secrets or private ranges.

**Soft warnings: none.**

No private IP ranges, private hostnames, personal email addresses, or hardware identifiers in added lines.

**Wording check — CLEAN.**

PR title/body and CHANGELOG mention "plaintext secrets leaking into backups" and "secret-leaking flat file" as design-motivation language, not disclosure of a live incident or specific private infrastructure. No hard wording block triggered.

---

WingedGuardian:
Re: the split-imports WARNING — that 3-block form is what this repo's `ruff` isort config produces; combining the names into one block fails `ruff check` (I001 "Import block is un-sorted or un-formatted"), so it's kept as-is to satisfy CI. All other points addressed; no functional change. 7/7 CI checks green; leak-detector + PII scan clean.

PR #678, opened 2026-06-18
feat(backup): include memory/config/secrets in off-site snapshot (A2a)

WingedGuardian:
## What & why

The off-site backup snapshot held only `data/` (SQLite), `qdrant/`, and `transcripts/`. Auto-memory, config overlays, and secrets went to the **git Tier-1** backup *only*. A no-git fresh-box disaster recovery (Tier-1 repo unreachable) therefore could not rehydrate memory or API keys — the off-site copy wasn't actually complete.

This makes the dated off-site snapshot a **complete copy** so a destination-only restore works.

## What changed

- **`backup.sh`** — after the data/qdrant/transcripts uploads (and before the `COMPLETE` gate), also upload:
  - `memory/*.gpg` (flat — the memory dir has no subdirs)
  - `config_overrides/*.local.yaml` (plaintext, mirroring the existing Tier-1 posture)
  - `secrets/secrets.env.gpg` (encrypted blob)
  - A failed upload of a *present* payload flips `_T2_OK`, so `COMPLETE` is gated on these landing too. A *missing* secrets payload (no passphrase) is best-effort and does not fail the snapshot.
- **`restore.sh`** — `_pull_from_offsite` now also pulls those three subdirs into the staging dir, so the existing memory/config/secrets restore sections work without the git clone. Failed pulls `warn` (→ non-zero restore) rather than failing silently.
- **No behaviour change to the git-data push** — this is the additive, no-gap step; retiring the git-data push is a later staged change.
- **No `backup_status.json` field or value-string changes** (dashboard/health contract untouched).

## Testing

- TDD (RED→GREEN). Extended the real-filesystem `local`-backend anchors: backup writes memory/config/secrets into the snapshot + `COMPLETE`; restore rehydrates them from an empty (no-git) source via real GPG decrypt; plus a regression test that a failed pull warns and fails the restore.
- `tests/test_scripts/` — 87 passed. `bash -n` + `shellcheck -S warning` clean.

## Review

Adversarial review via the Claude code-reviewer agent (Codex CLI quota-limited this cycle). It confirmed pipefail safety, `COMPLETE` gating, smb mkdir-before-put ordering, source paths, and the unchanged status contract. It found two observability gaps (silent failed pulls) — both fixed: failed pulls now `warn`, and the loops use process substitution so the warn registers in the parent shell.

No user-visible change yet, so no CHANGELOG entry.

### DIFF
--- a/scripts/backup.sh
+++ b/scripts/backup.sh
@@ -369,6 +369,49 @@ else
         fi
     done
 
+    # Upload memory / config overlays / secrets — previously git-Tier-1 only. Including
+    # them here makes the off-site snapshot a COMPLETE copy, so a no-git fresh-box DR can
+    # rehydrate everything (restore.sh §4/§6/§7 read them from the pulled snapshot). Memory
+    # is flat (MEMORY_DIR has no subdirs); config overlays ship as-is (plaintext, mirroring
+    # the existing Tier-1 + private-repo posture); secrets is the encrypted blob. A failed
+    # upload of a present file flips _T2_OK so COMPLETE is gated on these landing too.
+    backend_mkdir "${_T2_DIR}/memory"
+    for f in memory/*.gpg; do
+        [ -f "$f" ] || continue
+        fname=$(basename "$f")
+        if backend_put "$f" "${_T2_DIR}/memory/$fname"; then
+            log "  off-site: uploaded memory/$fname"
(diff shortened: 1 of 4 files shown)


WingedGuardian (commented):

---

PR #679, opened 2026-06-18
fix(backup): mirror dotfiles to off-site snapshot (find, not shell glob)

WingedGuardian:
## What & why

Follow-up to #678. The off-site memory/config upload enumerated with a `*.gpg` **shell glob**, which silently skips files beginning with `.`. Tier-1 staging (backup.sh §4/§6) enumerates with `find`, which **includes** dotfiles — so the off-site snapshot diverged from the Tier-1 copy by exactly the dotfiles. In practice a transient `.consolidate-lock.gpg` was dropped (300 files staged to Tier-1, 299 reached the off-site snapshot — caught during the live-NAS E2E).

Harmless today (the git Tier-1 backup still has it), but it breaks the "off-site is a *complete* copy" contract and would silently lose files once the git-data push is retired in a later step.

## Fix

- `backup.sh` — enumerate the memory and config_overrides off-site uploads with `find -maxdepth 1 -type f … -print0` + `while … read -d ''`, mirroring exactly how §4/§6 stage them (dotfiles included). Process substitution (not `find | while`) keeps the loop in the parent shell so the `_T2_OK` failure flip still gates the `COMPLETE` marker.
- Restore needs no change: it already lists dotfiles (`ls -1A` / smb) and the filename grep class matches leading dots, so the round-trip is symmetric.

## Testing

- TDD: a new regression test stages a dotfile (`.consolidate-lock`) in the memory dir and asserts `.consolidate-lock.gpg` reaches the off-site snapshot — RED on the glob, GREEN with `find`.
- `tests/test_scripts/` — 88 passed. `bash -n` + `shellcheck -S warning` clean.

## Review

Claude code-reviewer agent (Codex quota-limited this cycle): clean across all six dimensions (parent-shell `_T2_OK`, NUL-safe iteration, safe removal of the `[ -f ]` glob guard, correct `$BACKUP_DIR`-relative paths, no `set -e` abort, correct dotfile basename).

No CHANGELOG (no user-visible change).

### DIFF
--- a/scripts/backup.sh
+++ b/scripts/backup.sh
@@ -375,9 +375,13 @@ else
     # is flat (MEMORY_DIR has no subdirs); config overlays ship as-is (plaintext, mirroring
     # the existing Tier-1 + private-repo posture); secrets is the encrypted blob. A failed
     # upload of a present file flips _T2_OK so COMPLETE is gated on these landing too.
+    #
+    # Enumerate with `find` (not a `*.gpg` shell glob): §4/§6 stage these via `find`, which
+    # includes DOTFILES (e.g. a transient .consolidate-lock), so a glob would silently drop
+    # leading-dot names and leave the off-site copy short of Tier-1. Process substitution
+    # (not `find | while`) keeps the loop in THIS shell so the _T2_OK flip survives.
     backend_mkdir "${_T2_DIR}/memory"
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

PR #680, opened 2026-06-18
feat(observability): ambient capture health monitor + Telegram alert

WingedGuardian:
## What
A Genesis-side monitor that alerts the user (Telegram) when the edge ambient-capture bridge goes dark — the companion to the bridge-side ping fix that just landed.

A periodic `OutreachScheduler` job (every 10 min) SSH-reads the edge bridge's `ambient_health.json` and evaluates it:
- **heartbeat `ts` stale (>5 min)** → bridge process dead/hung
- **`active_connections == 0`** → device disconnected (a reliable negative)
- **`diar_worker_alive` false while enabled** → degraded

It alerts **once per state transition** (with a recovery note when it comes back), and re-alerts a persistent outage at most hourly. A quiet room (old `last_ts`, i.e. no recent utterance) is deliberately **not** treated as a fault.

## Design notes
- The edge host/connection lives ONLY in an install-local `~/.genesis/ambient_remote.yaml` (mirroring `guardian_remote.yaml`); **no host/IP in committed source**. The monitor is a **silent no-op** where that file is absent, so other installs are unaffected.
- The SSH read mirrors `GuardianRemote._ssh_command` (BatchMode, ConnectTimeout, orphan-kill on timeout). Read-only (`cat`).
- `ambient_health` signals are added to the governance no-dedup list (like `cli_approval`) so the monitor's own state machine is the sole dedup authority — otherwise the 24h default dedup would swallow the recovery note + hourly re-alerts.
- Alert delivery reuses `OutreachPipeline.submit_raw` + `OutreachRequest(category=BLOCKER)`, matching the existing `_critical_observations_job`.

## Verification
- ruff clean; 8 unit tests for the pure evaluator (pass).
- Live data-flow run: real config → SSH-read the real edge → correct `ok` verdict.
- Independent code review (pass-with-notes; the recovery-dedup finding fixed).
- On-device E2E (job scheduled + fires; deliberate down→alert) confirmed on deploy.

Companion follow-up: device-side firmware ambient auto-reconnect (defense-in-depth).

### DIFF
--- a/src/genesis/observability/ambient_health.py
+++ b/src/genesis/observability/ambient_health.py
@@ -1,0 +1,174 @@
+"""Ambient-capture health monitor — read the edge bridge's health file + evaluate.
+
+The ambient bridge runs on a separate edge VM and writes ``ambient_health.json``
+every 60s. This module SSH-reads that file (connection in
+``~/.genesis/ambient_remote.yaml``, mirroring ``guardian_remote.yaml``) and turns
+a snapshot into an alert verdict. ``OutreachScheduler._ambient_health_job`` runs
+the read + evaluate on a cadence and alerts the user on a bad-state transition.
+
+No config file -> the monitor is a silent no-op, so installs without an ambient
+edge are unaffected. The edge host/IP lives ONLY in the install-local config,
+never in committed source.
+"""
+from __future__ import annotations
+
+import asyncio
+import json
+import logging
+from dataclasses import dataclass
+from datetime import UTC, datetime
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+_CONFIG_PATH = Path("~/.genesis/ambient_remote.yaml").expanduser()
+_DEFAULT_HEALTH_PATH = "~/ambient_health.json"
+_DEFAULT_KEY = "~/.ssh/id_ed25519"
+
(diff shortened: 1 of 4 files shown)


WingedGuardian (commented):
## First-pass review — PR #680

### Structural review

**WARNING — `import time` at function scope** (`src/genesis/outreach/scheduler.py`, `_ambient_health_job`)

`import time` is placed inside the `try` block of `_ambient_health_job`. `time` is stdlib with zero import cost and no circularity risk; it should be in the module-level import block at the top of `scheduler.py`. (The deferred `from genesis.observability.ambient_health import …` is reasonable as a lazy import to avoid potential circularity or to keep a no-op path truly zero-cost, so that one is fine where it is.)

**NOTE — Missing class docstrings** (`src/genesis/observability/ambient_health.py`)

`AmbientRemoteConfig` (~line 32) and `AmbientVerdict` (~line 107) are public dataclasses with no class-level docstring. Inline field comments are helpful but a one-liner "what this represents / when it's used" docstring would complete the public API documentation expected for new public classes.

---

No API keys, tokens, private IPs, credentials, or sensitive identifiers in added lines. PR title and description contain no language advertising a credential leak or security incident. Paths use `~` expansion rather than `/home/<username>/` literals.

---

Overall: no structural ERRORs or hard blocks. Two minor quality findings above.

---

PR #681, opened 2026-06-18
fix(backup): route large temp files to ~/tmp, not cc-tmp/RAM

WingedGuardian:
## What & why

`backup.sh` and `restore.sh` created their **large** intermediate files — a ~269MB SQLite `.dump` and decrypted Qdrant snapshots — via bare `mktemp`, which inherits `TMPDIR`. That has two bad failure modes:

- **From a Claude Code session**, `TMPDIR` points at `~/.genesis/cc-tmp` — a 500MB budget-policed working-temp folder watched by `genesis-tmp-watchgod`, which **kills idle CC sessions** when it fills. A single 269MB dump is >half the budget; running backups/restores from a session filled it and the watchdog reaped detached terminals.
- **From the 6h cron**, bare `mktemp` lands in `/tmp`, which is **tmpfs (RAM)** — a recurring ~269MB RAM spike.

An existing seeded procedure (`tmp_filesystem_limit`) already mandates "use `~/tmp/` for large temporary files" and "never override `TMPDIR` in scripts/service files." The code just wasn't following it.

## Fix (surgical — does not touch `TMPDIR`)

Route **only the large** `mktemp` calls to a dedicated on-disk dir via `mktemp -p`:
- `GENESIS_BIG_TMP="${GENESIS_BACKUP_TMPDIR:-$HOME/tmp}"` (created with `mkdir -p`, logged).
- `backup.sh`: the SQL dump. `restore.sh`: the SQL dump + the Qdrant snapshot decrypt.
- `TMPDIR` is **not** exported/overridden — Claude Code and all other temp keep their normal location, honoring the procedure and CC's documented `TMPDIR`/`CLAUDE_CODE_TMPDIR` coupling. Tiny temp (COMPLETE marker, SMB creds) stays as-is.

Also strengthens the `CLAUDE.md` "Temp files" rule to name `cc-tmp` explicitly (it previously warned only about `/tmp`).

## Testing

- New regression test: runs `backup.sh` with `TMPDIR`=a cc-tmp stand-in + `GENESIS_BACKUP_TMPDIR`=a dedicated dir, asserts the large dump routes to the dedicated dir (RED before the fix, GREEN after).
- `tests/test_scripts/` (backup + restore files): 15 passed. `bash -n` + `shellcheck -S warning` clean.

## Review

Claude code-reviewer agent (Codex quota-limited this cycle): **clean** — verified variable ordering (after secrets sourcing), GNU `mktemp -p … --suffix` validity, `set -e`/EXIT-trap coverage, that `GENESIS_BIG_TMP` is not exported, absolute-path safety across `cd`, and no remaining large bare `mktemp`.

Follow-ups (separate PRs): runtime Python large-producers (yt-dlp/STT/pr_opener) pass explicit `dir=`; instrument `tmp_watchgod` to log measured size + top consumers. No CHANGELOG (no user-visible behavior change).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -28,9 +28,16 @@ reflection) → Services (routing, memory, outreach, autonomy, surplus) → Data
 - **Env scrub**: `CLAUDE_CODE_SUBPROCESS_ENV_SCRUB=1` is NOT used — Genesis
   hooks and MCP servers require inherited API keys (DeepInfra, Qwen, etc.).
 - **Setup**: `./scripts/bootstrap.sh` (venv, config, services, memory)
-- **Temp files**: `~/tmp/` for transient downloads (media, audio, exports).
-  NEVER use `/tmp/` — it shares the root filesystem and competes with all
-  CC sessions.  Clean up after use.
+- **Temp files**: `~/tmp/` for transient files and any LARGE temp (downloads,
+  media, DB dumps, exports). NEVER write large files to `/tmp/` (a small
+  tmpfs/RAM) or `~/.genesis/cc-tmp/` — the latter is Claude Code's working temp
+  ("oxygen"), policed by the `genesis-tmp-watchgod` service, which **kills CC
+  sessions** when it fills. A CC session's `TMPDIR` points at `cc-tmp` by design;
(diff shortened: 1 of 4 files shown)


WingedGuardian (commented):

---

PR #682, opened 2026-06-18
fix(runtime): route large runtime temp to ~/tmp, not cc-tmp

WingedGuardian:
## What & why

Follow-up to #681 (the bash side). The `genesis-server` systemd unit sets `TMPDIR=~/.genesis/cc-tmp` — the budget-policed working-temp folder the `genesis-tmp-watchgod` service reclaims by **killing idle CC sessions** when it fills. Runtime code that creates a **large** temp file via the default temp location lands it there, so the runtime can fill the oxygen folder and take out CC sessions (the runtime side of the 2026-06-18 incident class).

## Fix

New helper `genesis.util.tmp.big_tmp_dir()` — returns `~/tmp` (honors `GENESIS_BIG_TMP`, creates it), mirroring the `util/atomic.py` pattern. Passed as `dir=` at the confirmed large producers:
- `knowledge/processors/youtube.py` — yt-dlp audio download (and subtitle temp) `TemporaryDirectory`
- `channels/stt.py` — Groq audio-upload `NamedTemporaryFile`
- `contribution/pr_opener.py` — git-worktree-scratch `mkdtemp`
- `eval/promptfoo.py` — eval-artifacts `TemporaryDirectory`

`TMPDIR` is **not** overridden — only these explicit calls move — honoring the `tmp_filesystem_limit` procedure and Claude Code's documented `TMPDIR`/`CLAUDE_CODE_TMPDIR` coupling. (For yt-dlp, the download target is `{tmpdir}/...`, so `dir=` on the `TemporaryDirectory` does steer the download location.)

## Testing

- New `tests/test_util/test_tmp.py` (3): default `~/tmp`, `GENESIS_BIG_TMP` override, and a `NamedTemporaryFile(dir=big_tmp_dir())` lands under it.
- `tests/test_contribution/test_pr_opener.py` (27) pass — exercises the changed `mkdtemp`.
- All 5 modules import cleanly (no circular import); ruff clean.

## Review

Claude code-reviewer agent (Codex quota-limited): **clean** — verified `dir=` validity per call, no circular import, `big_tmp_dir` scope in youtube's locally-`import tempfile` methods, stt cleanup keys on `f.name`/`finally` (unaffected), empty-string env fallback, and no behavior change beyond temp location.

**Takes effect on the next `genesis-server` restart** (a code change to the running service). No CHANGELOG (no user-visible behavior change).

PR #683, opened 2026-06-18
feat(watchgod): log cc-tmp pressure + top consumers for diagnosability

WingedGuardian:
## What & why

Final piece of the temp-hygiene remediation (after #681, #682). When `~/.genesis/cc-tmp` filled (2026-06-18 incident), `tmp_watchgod` **correctly** reclaimed it by killing idle CC sessions — but it logged only the *tier name*, and the nuclear cleanup erased the evidence, so it was impossible to see *what* had filled the folder after the fact.

## Change (observability only — cleanup/kill thresholds & logic unchanged)

- `_log_cc_pressure()`: on ORANGE/RED, **before** cleanup, log measured `used/free/budget` and write a persisted top-8-consumers snapshot to `~/.genesis/logs/cc_tmp_top_<stamp>.txt` (under the log dir, so it survives the nuclear cleanup of cc-tmp).
- Log **which** sessions the RED cleanup kills (previously silent) + log YELLOW's measured numbers.
- **Sourcing guard** (`if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then main "$@"; fi`) so the functions are unit-testable without starting the daemon (the systemd `ExecStart` runs it directly, so `main` still runs in production).

## Review fixes (Claude code-reviewer)

The reviewer passed 5/6 checks and caught one real Important issue + I fixed a related one it surfaced:
- **Unbounded snapshots** — one file per ORANGE/RED poll would accumulate without limit on the very filesystem being protected → keep only the 20 most recent.
- **`set -e` abort on empty cc-tmp** — the top-consumers `du` ran an unguarded command substitution; on the sacred-ground RED path (disk-full but cc-tmp *empty*) the glob is literal, `du` fails, and `set -e` would have aborted the whole daemon → tolerate with `|| true`.

## Testing

- `tests/test_scripts/test_watchgod_instrumentation.py` (3): snapshot+logline written; `check_cc_tmp` at RED captures pressure before cleanup; snapshots stay bounded (also exercises the empty-cc-tmp path). **tmux is stubbed** so the test can never touch a real CC session.
- `bash -n` + `shellcheck -S warning` clean.

Takes effect on the next `genesis-tmp-watchgod` restart. No CHANGELOG (internal observability).

### DIFF
--- a/scripts/tmp_watchgod.sh
+++ b/scripts/tmp_watchgod.sh
@@ -195,12 +195,18 @@ _log_cc_pressure() {
     local stamp snap top
     stamp=$(date -u +%Y%m%dT%H%M%SZ)
     snap="$(dirname "$LOG_FILE")/cc_tmp_top_${stamp}.txt"
-    top=$(du -sm "$CC_TMP_DIR"/* 2>/dev/null | sort -rn | head -8)
+    # `|| true` inside the substitution: an empty cc-tmp (e.g. the sacred-ground RED path,
+    # disk-full but cc-tmp empty) makes the glob literal → du fails → set -e would abort the
+    # whole daemon. Tolerate it; `top` is just empty then.
+    top=$(du -sm "$CC_TMP_DIR"/* 2>/dev/null | sort -rn | head -8 || true)
     {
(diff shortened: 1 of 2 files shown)


WingedGuardian (commented):

---

PR #684, opened 2026-06-19
fix(memory): show full session-trail arc in proactive hook (cap 50, was 8)

WingedGuardian:
## What & why

The per-prompt `[Session trail]` breadcrumb (from `proactive_memory_hook.py`) showed only the **last 8** pivots (`_MAX_TRAIL_DISPLAY = 8`). On a long multi-phase session, early phases scroll off — e.g. an early audit phase became invisible once later phases pushed past 8 pivots, so after compaction only the recent tail was surfaced. The full pivot list is always persisted in `~/.genesis/sessions/<id>/intent_trail.json`; this just widens what's shown.

## Change

- `_MAX_TRAIL_DISPLAY`: **8 → 50** — shows the full arc for typical sessions; the `… →` truncation prefix still applies beyond 50. Display-only constant (used only at the format step), no logic change.
- Tests: replaced the old "last-8" truncation test with `test_full_arc_shown_under_cap` (30 pivots → full arc, no prefix) + `test_long_trail_truncated_at_cap` (reads the actual cap, asserts the window + prefix beyond it).

## Testing

`tests/test_intent_trail.py` — 21 passed. ruff clean. Takes effect on the next prompt after merge (the hook re-reads its script per invocation).

No CHANGELOG (internal observability tweak).

### DIFF
--- a/scripts/proactive_memory_hook.py
+++ b/scripts/proactive_memory_hook.py
@@ -104,7 +104,10 @@ _STOP_WORDS = frozenset({
 _TRAIL_DIR = Path.home() / ".genesis" / "sessions"
 _PIVOT_SIMILARITY_THRESHOLD = 0.3
 _PIVOT_DEBOUNCE_MSGS = 3
-_MAX_TRAIL_DISPLAY = 8  # Show at most this many pivots in injected line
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

PR #685, opened 2026-06-19
fix(observability): ambient-health alerts only on software failure, no nag

WingedGuardian:
## Problem
The ambient-capture health monitor (added in #680) alerted whenever the **device** was absent (`active_connections==0`) and re-alerted hourly. A device being offline — unplugged, idle, or its firmware crashed — is **not a software bug**, so those alerts are noise.

## Policy (user-mandated)
Alert **only when the software/infra itself breaks**, and **never nag**.

## Fix
- `observability/ambient_health.py`: removed the `active_connections==0 → down` branch. The verdict is now `down` only on a **missing/stale heartbeat** (the bridge process is dead/hung — it stopped writing the health file), `degraded` if the **diarization worker crashed** (`diar_worker_alive` false while enabled), `unknown` on a transient read failure (no alert), else `ok` — *including* when the device is simply offline.
- `outreach/scheduler.py`: dropped the hourly re-alert — now alerts **once** on entering down/degraded, with **one** recovery note. Removed the now-unused re-alert timing state (`_ambient_health_last_alert`, `import time`). Hint text now points at the bridge service.
- Tests: `test_device_offline_is_not_a_fault` (offline + healthy → `ok`) and `test_device_offline_does_not_mask_software_failure` (offline + dead worker → still `degraded`).

This is the only Telegram alert in the voice pipeline.

## Verification
- ruff clean; 9/9 unit tests pass; py_compile OK.
- Independent code review: pass (no findings).
- On-device E2E (after deploy): device disconnected but bridge running → no Telegram; bridge process down >5min → one alert; recovery once; no nag across cycles.

### DIFF
--- a/src/genesis/observability/ambient_health.py
+++ b/src/genesis/observability/ambient_health.py
@@ -129,17 +129,19 @@ def evaluate_ambient_health(
 ) -> AmbientVerdict:
     """Turn an ``ambient_health.json`` snapshot into a verdict. Pure + testable.
 
+    Policy: alert ONLY on a SOFTWARE failure, NEVER on the device being absent.
     - ``data is None`` (read failed / edge unreachable) -> ``unknown`` (transient;
       the caller should not flip alert state on a single unknown).
     - heartbeat ``ts`` missing or older than ``_HEARTBEAT_STALE_S`` -> ``down``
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

PR #687, opened 2026-06-19
fix(learning): skill refiner no longer overwrites SKILL.md from a truncated view (+ dashboard WAL/banner)

WingedGuardian:
## Summary

Bundles one correctness fix and two small dashboard observability improvements (different subsystems, all low-risk).

### WS-10 — skill refiner overwrites `SKILL.md` from a truncated view (LRN-01)

The weekly skill-evolution pipeline auto-applies **MINOR** proposals (`autonomy_level=2`) directly onto real `SKILL.md` files via `path.write_text()`. Two compounding bugs let it silently truncate a large skill:

1. `refiner.py` built the LLM prompt with `current_content[:3000]`, so a skill larger than 3000 chars was refined from a **partial view** (the largest real skill is ~17.7K chars).
2. The validator's consistency check was never threaded the original content, and even when reached it only **warned** — it never blocked. So a too-short proposal would auto-apply.

**Fix:**
- Refiner now prompts with the **full** skill content.
- The pipeline threads `current_content` to **both** `applicator.apply()` call sites (`run()` and `propose_for_skill()`).
- The validator now **blocks** (routes to staging for human review) any MINOR proposal under 50% of the original size, instead of only warning. The line-based size floor doesn't catch this — a short proposal above the line floor can still drop most of the content.

### Dashboard

- **Resilience banner wrap** — the resilience-degradation banner now wraps long provider summaries instead of overflowing on narrow widths (CSS only).
- **WAL-size health indicator** — a new `probe_wal()` surfaces the SQLite `-wal` sidecar size under the `genesis.db` health probe (green/yellow/red at 100 MB / 500 MB, mirroring the awareness loop's WAL thresholds). A bloated WAL is an early signal of DB-lock pressure. Skipped when there's no DB connection so a "0 MB" readout never sits next to a DB-error row.

## Testing

- `pytest tests/test_learning/test_skill_evolution.py tests/test_observability/test_health.py` — 62 passed (TDD: RED → GREEN).
- Adjacent consumers (`test_guardian/test_container_side.py`, `test_awareness/test_wal_health.py`) still green.
- `ruff check` clean on the full diff.
- Reviewed by a code-review agent; both flagged items addressed (gate WAL on DB presence; clarify the char-vs-line threshold comment).

## Risk

Low. WS-10 makes the auto-apply guard stricter (worst case: a borderline-but-legitimate proposal routes to human staging instead of auto-applying — non-destructive). The WAL probe and banner change are additive observability.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,23 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   holding the file open), Genesis raises a high/critical alert on Telegram and in
   the morning report, instead of letting it balloon silently for days.
 
+- **The dashboard shows your database journal (WAL) size at a glance** — the
+  Infrastructure health panel now displays the SQLite WAL size next to the
+  database probe, colored green / amber / red, so you can spot DB-lock pressure
+  building before it ever trips an alert.
+
 ### Fixed
+
+- **The skill auto-tuner can no longer truncate a large skill** — Genesis's
+  weekly skill-refinement pass reviewed long skill files from a clipped
+  3,000-character view and could auto-apply a much shorter rewrite, silently
+  dropping most of the content. It now reviews the full skill, and any
+  auto-applied edit that would shrink a skill below half its size is held for
+  review instead of overwriting the file.
+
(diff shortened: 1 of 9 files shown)


WingedGuardian (commented):

---

PR #688, opened 2026-06-19
fix(outreach): dedup window never expired within a UTC day

WingedGuardian:
## What

Outreach dedup (`GovernanceGate._is_duplicate`) compared a stored ISO-8601
`delivered_at` (with a `T` separator) against SQLite's space-separated
`datetime('now', ?)` as **raw strings**. `T` (0x54) sorts after space (0x20),
so a row delivered earlier on the cutoff's own UTC date was treated as *inside*
the window — the dedup window never expired within a calendar day (a 24h window
behaved like ~41h). Outreach deliveries were suppressed for hours, then flushed
together in a single batch (the "10 critical observations at once" symptom).

## Fix

Wrap `delivered_at` in `datetime()` in both dedup queries so SQLite parses both
operands before comparing. The rate-limit and `*_available` quota checks compare
against `date('now')` (no time component, so no `T`-vs-space mismatch) and are
unaffected — left untouched.

## Verification

- New regression test (`test_iso_t_delivered_at_outside_window_not_deduped`)
  uses the production ISO-`T` format at the 24h boundary: **RED** on old code,
  **GREEN** after the fix. Full `test_outreach` suite: **117 passed**; ruff clean.
- Query plan unchanged — still a covering-index search on `idx_outreach_dedup`.
- Verified against real history: of rows genuinely 24–48h old, the old
  comparison wrongly kept **3** inside the window; the fixed comparison keeps
  **0**.

### DIFF
--- a/src/genesis/outreach/governance.py
+++ b/src/genesis/outreach/governance.py
@@ -192,12 +192,17 @@ class GovernanceGate:
             return False
         window_spec = f"-{window_hours} hours"
 
-        # Primary key: (signal_type, topic, category) within the window
+        # Primary key: (signal_type, topic, category) within the window.
+        # NB: wrap delivered_at in datetime(). Stored values are ISO-8601 with a
+        # 'T' separator (datetime.now(UTC).isoformat()); a raw string compare
+        # against SQLite's space-separated datetime('now', ?) is wrong because
+        # 'T' (0x54) > ' ' (0x20), which made the window never expire within a
+        # UTC day. datetime() parses both sides before comparing.
         cursor = await self._db.execute(
             "SELECT COUNT(*) FROM outreach_history "
             "WHERE signal_type = ? AND topic = ? AND category = ? "
             "AND delivered_at IS NOT NULL "
-            "AND delivered_at >= datetime('now', ?)",
+            "AND datetime(delivered_at) >= datetime('now', ?)",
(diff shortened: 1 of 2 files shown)


WingedGuardian (commented):
## First-pass review — PR #688

### Structural review

**WARNING — Raw SQL against genesis.db (pre-existing)**
`src/genesis/outreach/governance.py` — both `_is_duplicate` queries call `self._db.execute()` directly rather than going through the `outreach_crud` module. The pattern rule requires using `genesis.db.crud` modules for production queries (migrations excepted). This violation pre-dates this PR — the diff only modifies the WHERE clause, it doesn't introduce the raw SQL access. The fix itself is correct and surgical. Recommend a follow-up to push this dedup query into `outreach_crud`, where the `datetime()` normalisation would be co-located with the data access and could be validated in one place.

No other structural findings. No bare excepts, no mutable defaults, no nohup, no task_states direct inserts, no /tmp/ usage. Import grouping is correct. The new regression test covers the changed function. Commit prefix (`fix(outreach):`) is conventional.

---

**CLEAN** — No secrets, API keys, private IPs, personal emails, private hostnames, or credential strings found in added lines. PR title and body describe a datetime comparison bug with no security-incident language, no infrastructure references, and no wording that would be problematic in git history.

---

PR #689, opened 2026-06-19
fix(resilience): recovery no longer drops un-executed work or blocks reflections (+ real email Message-ID)

WingedGuardian:
## Summary

Two verified-live correctness fixes from the 2026-06-13 audit (P3).

### WS-6 — recovery orchestrator (`resilience/recovery.py`, `deferred_work.py`, `awareness/loop.py`, `state.py`)

The 30-minute `RecoveryOrchestrator` had three bugs:

1. **Silent data loss** — it drained pending deferred-work items and marked them **completed without executing them** (the code comment admitted it was a Phase 8/9 stub). The drain is removed; deferred work is left **pending**. Reflection items are already re-dispatched by the awareness loop; outreach items await a dedicated consumer (tracked follow-up).
2. **Head-of-line blocking** — the deferred queue carries reflection (priority 30) and `outreach_delivery` (priority 20) items. The awareness consumer pulled the single highest-priority item and bailed if it wasn't a reflection, so a higher-priority outreach item **starved every reflection retry** — and the lossy drain was the only thing clearing it. `next_pending` now takes a `work_type` filter (already supported by the query layer), and the awareness loop filters to reflections.
3. **Unconditional run** — re-dispatch (embedding drain + dead-letter replay) is now **gated** on a stable system (`is_any_degraded`, excluding cc-tmp pressure, which doesn't impair Qdrant/providers); housekeeping (expire stuck/stale, reset-failed-embeddings, overflow check) still runs every cycle. The always-zero `items_drained` report field is replaced by an honest `items_pending`.

### WS-9a — email Message-ID (`channels/email_adapter.py`)

Outbound email set **no Message-ID**, so mail shipped without one and the returned id was a fabricated placeholder — replies couldn't thread and our reply-poller couldn't match them. The adapter now sets a real RFC `Message-ID` via `email.utils.make_msgid` before send and returns it; the thread tracker already registers that returned id, so inbound reply-matching works.

(Outbound `In-Reply-To`/`References` so *our* replies thread in the recipient's client — WS-9b — is a separate follow-up; it needs deeper plumbing through the outreach pipeline.)

## Testing

- `pytest tests/test_resilience/ tests/test_channels/test_email_adapter.py tests/test_awareness/test_loop_resilience.py` — 78 passed (TDD: RED → GREEN).
- Covers: head-of-line filter, data-loss prevention (items left pending), re-dispatch gated when degraded, cc-tmp pressure does **not** gate, real RFC Message-ID.
- `ruff check` clean. Signature change to `next_pending` verified safe (no positional callers).
- Reviewed by a code-review agent; both findings (over-broad gate, misleading dead report field) addressed.

## Risk

Low-medium. The recovery change alters re-dispatch semantics — the deliberate trade-off is that deferred work now accumulates **pending** (surfaced by the existing overflow alert) instead of being silently dropped. `drain_by_priority` remains as an independently-tested queue utility, now unused by recovery.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,19 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Genesis can now detect replies to the emails it sends** — outbound email
+  was going out without a real Message-ID header, so mail clients couldn't thread
+  it and Genesis couldn't match incoming replies back to the original message.
+  Outbound mail now carries a proper Message-ID, so replies are recognized and
+  routed to the right conversation.
+
+- **Background work deferred during an outage is no longer silently dropped** —
+  when the system was degraded, the recovery pass marked queued reflection and
+  outreach work "done" without ever running it, and a stuck outreach item could
+  block reflection retries entirely. Deferred work is now kept until it actually
+  runs, reflections are no longer blocked behind it, and recovery holds off
+  re-trying until the system is genuinely stable.
+
 - **The skill auto-tuner can no longer truncate a large skill** — Genesis's
   weekly skill-refinement pass reviewed long skill files from a clipped
   3,000-character view and could auto-apply a much shorter rewrite, silently
--- a/src/genesis/awareness/loop.py
+++ b/src/genesis/awareness/loop.py
@@ -1196,8 +1196,8 @@ class AwarenessLoop:
             logger.debug("Skipping deferred reflection retry — loop is stopping")
             return
 
-        item = await self._deferred_queue.next_pending(max_priority=40)
-        if not item or item.get("work_type") != "reflection":
+        item = await self._deferred_queue.next_pending(work_type="reflection", max_priority=40)
+        if not item:
             return
 
         item_id = item["id"]
--- a/src/genesis/channels/email_adapter.py
+++ b/src/genesis/channels/email_adapter.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import email.message
+import email.utils
 import logging
 import smtplib
 from typing import Any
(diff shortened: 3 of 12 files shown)


WingedGuardian (commented):
**Structural review: 1 WARNING · 1 NOTE**

---

### WARNING — Missing direct coverage: `src/genesis/resilience/state.py`

`is_any_degraded` gained a new `include_tmp_pressure: bool = True` keyword argument (with keyword-only enforcement via `*`). The new branch is exercised indirectly through `test_recovery.py::test_tmp_pressure_does_not_gate_redispatch` (via the orchestrator), but there is no direct unit test for `ResilienceStateMachine.is_any_degraded(include_tmp_pressure=False)`. A dedicated state-level test would pin the contract more precisely and survive future refactors of the orchestrator fixture.

---

### NOTE — Minor logging gap in `src/genesis/runtime/init/learning.py`

The `learning.py` secondary log fires only when `report.dead_letters_replayed or report.embeddings_recovered` — both are 0 when re-dispatch is skipped (system degraded). In that case `items_pending` doesn't appear in `learning.py`'s log, even if the backlog is growing. `recovery.py`'s own unconditional `"Recovery completed: pending=…"` line still fires, so observability isn't lost — but the `items_pending` column in `learning.py` is only visible during healthy-system cycles, which is the least interesting time to see it.

---

### SOFT WARNING — `tests/test_channels/test_email_adapter.py` (test_send_message_sets_real_message_id)

```python
assert mid.rstrip(">").endswith("@gmail.com"), \
    f"Message-ID domain not from from_address: {mid}"
```

The assertion hardcodes `@gmail.com` as the expected domain, which reveals that the test adapter fixture is configured against a Gmail account. No personal address appears in the diff itself, so this is not a hard block. The test would be more portable and less provider-revealing if it derived the domain dynamically from the fixture (e.g. `adapter._from_address.rsplit("@", 1)[-1]`) rather than hardcoding it.

---

### Everything else: CLEAN

- No bare `except:`, mutable defaults, raw SQL, `/tmp/` misuse, `nohup`, or direct `task_states` inserts.
- No API keys, tokens, private IPs, SSH keys, or secret strings in added lines.
- PR title/body contains no wording describing PII leaks, credential exposure, or private infrastructure names.
- Conventional commit prefix (`fix(resilience):`) ✓
- `email.utils` import correctly placed within the stdlib group in alphabetical order ✓
- All modified `src/` files have corresponding updated test files, with the `state.py` caveat above ✓

---

PR #690, opened 2026-06-19
feat(feedback): Self-Improvement Outcome Bus — foundation (dark capture)

WingedGuardian:
## Self-Improvement Outcome Bus — foundation (PR-A, dark capture)

Genesis has the *organs* of self-improvement (calibration, capability model, per-subsystem quality scoring) but no closed feedback loop — the loop is open at **capture**. The single most valuable signal, post-execution "did it actually work" (T1), is written today as a `|completed:`/`|failed:` string suffix on `ego_proposals.user_response` and is otherwise unsurfaced: of 63 proposals that actually ran, only **1** has a structured outcome anywhere a learning loop can read.

This PR lays the foundation: an append-only, attributed, quality-weighted `outcome_events` ledger plus the machinery to populate it. It is **DARK** — nothing in production emits to it and nothing consumes it yet (those are follow-up PRs). The only live wiring is a scheduled, behavior-neutral harvester.

### What's here (Steps 1–3)
- **`outcome_events` table** — migration `0025` (+ `down`) and fresh-install registration in `_tables.py`; strict CRUD (`record`/`exists`/`count`/`aggregate_by_domain`/`calibration_by_domain`/`recent`).
- **`feedback/bus.py`** — fire-and-forget `record_outcome` (never raises) + the signal taxonomy: a controlled `signal_type` vocabulary and a tier policy (T1 ground truth › T2 rationale-bearing › T3 coverage). The controlled vocabulary guarantees that two signals on the same ref (e.g. a user decision and a later execution outcome on one proposal) never collide on the unique key — so the bus can never silently drop ground truth.
- **`feedback/harvest.py`** — `OutcomeHarvester` folds existing siloed signals (`ego_proposals`, `outreach_history`) into the ledger without mutating them; incremental `run()` + one-shot `run_backfill()` guarded by an `ego_state` marker. Per-source isolation; idempotent on the unique key.
- **Scheduler wiring** — `CronTrigger(8:45/20:45)`, clock-time before the capability-map refresh, off the loaded overnight windows. Behavior-neutral (populates a table nothing reads).

### Signal-quality discipline
Tiering measures signal *quality*, not desirability of proposing — nothing here teaches the ego to propose less. A timeout/expiry is recorded as coverage (T3), never as disapproval, and stays out of any quality denominator. A bare approve/reject is T3; only a rationale-bearing decision is T2; post-execution outcome is T1 and outranks user approval.

### Verification
- 50 targeted tests (migration up/down/idempotency, fresh-install path, DB CHECK, CRUD, taxonomy, harvester mapping, backfill guard, regression test for the unique-key non-collision). Ruff clean.
- **End-to-end against a copy of the live DB:** T1 `execution_outcome` capture **1 → 47**; calibratable rows (T1 + confidence + value) **~1 → 47**; 764 rows correctly tiered (47 T1 / 58 T2 / 659 T3); backfill idempotent (guarded re-run is a no-op).
- Reviewed by an architecture pass (pre-implementation) and a code-review pass; findings fixed and re-verified.

### Deploy note
On merge + deploy the table is created at startup and the harvester backfills on its first scheduled tick — there is no separate manual step. Still dark: no behavior change until the emit-hooks and consumer PRs land.

### Deferred (next PRs)
- **PR-B:** emit hooks at the deep CRUD chokepoints (behind a flag) + capability-map rewire to read the ledger.
- **PR-C:** advisory deficiency consumer, after a 1–2 week dark-capture soak.

### DIFF
--- a/src/genesis/db/crud/outcome_events.py
+++ b/src/genesis/db/crud/outcome_events.py
@@ -1,0 +1,201 @@
+"""CRUD for ``outcome_events`` — the Self-Improvement Outcome Bus ledger.
+
+Thin, strict data layer. ``record`` raises ``ValueError`` on malformed input
+(invalid tier/class/polarity, empty key fields) so bugs surface in tests; the
+fire-and-forget leniency that protects production paths lives one layer up in
+``feedback/bus.py``. Writes are idempotent on the unique key
+``(source, ref_type, ref_id, signal_type)``.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import uuid
+from datetime import UTC, datetime, timedelta
+
+import aiosqlite
+
+logger = logging.getLogger(__name__)
+
+# Enum domains enforced by the table CHECK constraints — validated here too so
+# we raise a clear ValueError instead of relying on INSERT OR IGNORE silently
+# swallowing a CHECK violation (OR IGNORE skips BOTH unique AND check failures).
+_VALID_TIERS = frozenset({1, 2, 3})
+_VALID_CLASSES = frozenset({"implicit", "explicit"})
+_VALID_POLARITIES = frozenset({"positive", "negative", "neutral"})
+
+_COLUMNS = (
+    "id", "source", "ref_type", "ref_id", "domain", "signal_type",
+    "signal_class", "signal_tier", "polarity", "value", "stated_confidence",
+    "prediction_error", "reason", "reason_text", "metadata", "harvested_from",
+    "occurred_at",
+)
+
+
+def _rows_to_dicts(cur: aiosqlite.Cursor, rows: list) -> list[dict]:
+    cols = [d[0] for d in cur.description]
+    return [dict(zip(cols, r, strict=False)) for r in rows]
+
+
+async def record(
+    db: aiosqlite.Connection,
+    *,
+    source: str,
+    ref_type: str,
+    ref_id: str,
+    signal_type: str,
+    signal_tier: int,
+    domain: str | None = None,
+    signal_class: str = "implicit",
+    polarity: str | None = None,
+    value: float | None = None,
+    stated_confidence: float | None = None,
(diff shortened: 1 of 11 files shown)


WingedGuardian (commented):
## First-pass review — PR #690

### Structural: 1 WARNING

**WARNING — Raw SQL outside CRUD layer: `src/genesis/feedback/harvest.py`**

`_harvest_proposals()` and `_harvest_outreach()` execute raw SQL `SELECT` statements directly against `ego_proposals` and `outreach_history` rather than routing through `genesis.db.crud` modules. The convention exists to keep data access consistent and auditable across the codebase.

Mitigating context that keeps this a WARNING rather than ERROR: (a) both methods are read-only — they never mutate source tables; (b) `ego_crud` is already imported and used correctly for the `ego_state` backfill marker; (c) the harvest queries are intentionally custom (optional time-window parameter, multi-column bulk scans) with no existing CRUD equivalents.

Suggested path forward: either add bulk-read functions to `ego.crud` / `outreach.crud` for these queries, or add a short inline comment in `harvest.py` documenting why raw SQL is used here (e.g. "no CRUD equivalent for bulk time-windowed scan") so future reviewers don't flag it as an oversight.

---

No secrets, tokens, private IP addresses, personal email addresses, private hostnames, or AWS/GCP identifiers found in the added lines. PR title and description contain no language that reveals security incidents, credential leaks, or private infrastructure details.

---

WingedGuardian:
Note: the automated Codex review is **quota-limited** on this PR (no Codex comment posted). Per the pre-merge policy, an inline code-review was performed pre-push instead — it caught and fixed two real bugs (backfill marker set on source failure → would lock the gate; `expired` status unhandled), both now covered by tests. 50 targeted tests pass locally + E2E verified against a copy of prod (T1 capture 1 → 47). Merge can proceed on CI alone.

PR #691, opened 2026-06-19
fix(observability): resolve dead-letter alerts on drain + band-dedup the writer

WingedGuardian:
## What

The observation pipeline was **write-only**. `_alert_dead_letter_accumulation`
(`observability/snapshots/queues.py`) created a `critical` `infrastructure_alert`
observation when the dead-letter queue grew past 50, but **nothing resolved it
when the queue drained** — so stale "DLQ at 319" alerts lingered for the 3-day
TTL and poisoned the morning report (which surfaces unresolved observations).
The writer also embedded the raw count in the content, so each hourly tick
produced a new `content_hash` and ~10 near-duplicate rows accumulated per spike.

## Fix

- **Band the count** (`50-99 / 100-199 / 200-499 / 500+`) and derive a stable
  `content_hash` from the band + pass `skip_if_duplicate` — a drifting spike is
  one row per band (the exact count still shows in the content). A **band change
  bypasses the 1h cooldown** so a worsening still escalates promptly.
- **`_resolve_dead_letter_alerts()`** — called from `queues()` when the queue
  drains below threshold. The compensating recovery path. Intentionally
  unconditional (an in-memory guard would be lost on restart and reintroduce
  the staleness).
- **`MORNING_REPORT.md`** staleness guard: cross-check stored infra/provider/
  security observations against the live System Health section and treat
  contradicted ones as resolved rather than repeating a stale alarm.

## Verification

- 7 new tests (`tests/test_observability/test_queues_dead_letter.py`): band
  bucketing, same-band dedup, band-escalation-within-cooldown, resolve-on-drain,
  resolve-then-respike re-alerts, `queues()` end-to-end. Full `test_observability`
  + `test_health_data`: **202 passed**; ruff clean.
- Code review caught and fixed a missed-escalation bug (a band-crossing
  worsening was being swallowed by the cooldown).
- Wiring: `queues()` is the same path (`health_data.py`) that created the
  original alerts, so the resolve rides a proven-live call site.

### DIFF
--- a/src/genesis/identity/MORNING_REPORT.md
+++ b/src/genesis/identity/MORNING_REPORT.md
@@ -57,6 +57,23 @@ delete it. See VOICE.md (advisory tone) for full reference.
 - Observation/finding counts without content — raw numbers are noise
 - Background task completion counts — unless something failed
 
+### Staleness check — cross-check stored alerts against live state
+
+Observations are point-in-time snapshots and may have resolved since they were
+written. Before surfacing ANY infrastructure / provider / security observation
+as urgent, check whether the live **System Health** data in this report
+corroborates it:
+- "Dead letter queue at 319" is STALE if System Health shows dead_letters at 0.
+- "Provider chain exhausted / circuit breaker open" is STALE if providers are
+  currently healthy.
+- "DASHBOARD_PASSWORD not set / endpoint unauthenticated" is STALE if the live
+  config shows it set.
+
+If a stored observation contradicts live System Health, treat it as resolved:
+note it at most as "earlier <condition>, now recovered", or omit it. Do NOT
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

WingedGuardian:
Heads-up on overlap: #695 originally also fixed the dead-letter resolve-on-drain alarm, but this PR's design is superior (band-based DB-backed dedup that also fixes the duplicate-rows-per-spike accumulation, unconditional restart-safe resolve, and the MORNING_REPORT.md staleness cross-check). #695 has dropped its DLQ changes and ceded the dead-letter fix to this PR — they no longer conflict on `queues.py`. #695 is now reflection-only (idle alive-pulse for the false 'reflection dark' alarm).

PR #693, opened 2026-06-19
fix(reflection): resume approved light reflections + drop dead bridge

WingedGuardian:
## What & why

Light reflection's gated Claude Code fallback was half-wired. When light's free
model chain (gemini → groq → mistral) is fully exhausted during a provider
outage, `4_light_reflection` has no `dispatch:` field, so it defaults to `dual`
and escalates to an **approval-gated** CC fallback — which creates a *pending*
approval and returns `success=False`. A user could approve it, but nothing ever
resumed light: `_resume_approved_reflections` only iterated `("deep",
"strategic")`. So approving a light reflection did nothing, and the dispatch
result was silently discarded.

## Changes

- **Resume light:** add `light` to `_resume_approved_reflections` via an explicit
  `{name: Depth}` map — replacing the prior binary
  (`Depth.DEEP if name == "deep" else Depth.STRATEGIC`) that would have
  mis-dispatched light as an expensive STRATEGIC reflection.
- **Observability:** `perform_tick` now captures the `reflect()` result and logs
  a deferred reflection instead of a silent no-op.
- **Dead code:** delete the shadowed `src/genesis/cc/reflection_bridge.py`
  (41 KB). It is superseded by the `reflection_bridge/` package and never
  imported; the `GROUNDWORK(v4-executor)` tag is preserved in `_bridge.py`.
- Docs: CHANGELOG entry + two stale file-path references fixed.

## Verification

- ruff clean; **330 tests pass** across `tests/test_awareness/`,
  `tests/test_reflection/`, `tests/test_cc/test_reflection_bridge*`.
- New unit tests: light resumes as `Depth.LIGHT`; depth-map no-collapse guard.
- Real-component integration test (no mocks): `ApprovalManager` + crud + full
  schema — a `reflection_light` request goes pending → approved → found →
  atomically consumed (double-consume returns False).
- Approval invariant preserved: `skip_approval=True` is used only on the resume
  path, after `mark_consumed` succeeds on a request the user already approved.

## ⚠️ DO NOT MERGE YET

Holding merge intentionally: multiple concurrent sessions are working on closely
related systems (routing / reflection / resilience). Merge only after
coordination, to avoid collisions.


WingedGuardian (commented):

---

PR #695, opened 2026-06-19
fix(awareness): reflection idle heartbeat so quiet periods aren't 'dark'

WingedGuardian:
## Reflection idle alive-pulse (quiet periods aren't "dark")

> **Reconciled with #691:** this PR originally also fixed the dead-letter resolve-on-drain alarm, but #691 (`fix/observation-pipeline-staleness`) independently fixed the *same* bug with a superior design (band-based DB-backed dedup that also kills the duplicate-rows-per-spike problem, an unconditional restart-safe resolve, and a `MORNING_REPORT.md` staleness cross-check). Those DLQ changes were dropped here — **#691 owns the DLQ fix**. This PR is now reflection-only.

### The bug
The `reflection` subsystem heartbeat only fired on a **successful** reflection, and the overdue threshold was 20 min. Reflections only fire when signals warrant — calm periods can legitimately be quiet for hours — so `subsystem_heartbeats` falsely reported reflection "dark" (the morning-report "10.6h dark" alarm).

### The fix
- A calm Micro tick (the silent branch — no critical signals, no reflection attempted) now emits an idle alive-pulse heartbeat. It fires **only on the silent path**, so a tick that *attempts and fails* a reflection still doesn't pulse → a real outage still ages out and alarms.
- Overdue threshold raised 20min → 4h (matches ego). Genuine outages surface faster via the awareness heartbeat (6 min), the `cc` resilience axis, and the deferred-work backlog.

### Verification (E2E, read-only against live data)
Confirmed the bug is **live**: the awareness loop is healthy (260 events/last hour; surplus/ego heartbeats ~1 min fresh) but the reflection heartbeat is **7.4h stale** → falsely "dark". The idle-pulse (firing on every silent Micro tick) keeps it fresh; the threshold raise alone would not have covered the 7.4h gap. Targeted tests pass (idle-pulse fires on silent tick but NOT on critical success/failure; 30-min heartbeat reads "alive"). Reviewed by code-reviewer agent.


WingedGuardian (commented):
## First-pass review — PR #695

### Structural review

**WARNING — Import hygiene** `src/genesis/observability/snapshots/queues.py`

`from genesis.db.crud import observations as obs_crud` is a deferred local import placed inside the body of `_resolve_dead_letter_alert()`. If there's no confirmed circular-import cycle requiring it, move it to the module-level imports block (stdlib → third-party → local). Local imports in function bodies are only warranted when they're needed to break a circular dependency.

---

**WARNING — Raw SQL in test** `tests/test_mcp/test_standalone_health.py` — `test_reflection_alive_at_30_minutes`

```python
await db.execute(
    "INSERT INTO events (id, timestamp, subsystem, severity, event_type, "
    "message, created_at) VALUES (?, ?, ?, ?, ?, ?, ?)",
    ...
)
```

This inserts directly into the `events` table via raw SQL, bypassing the CRUD layer. The parallel new test file (`test_queues_snapshot.py`) correctly uses `obs.create()` / `obs.query()`. If there's no `events` CRUD insert method exposed, a small test helper (`insert_test_event(db, ...)`) would keep the pattern consistent and prevent drift as the schema evolves.

---

**NOTE — Broad suppress** `src/genesis/awareness/loop.py`

```python
with contextlib.suppress(Exception):
    await self._event_bus.emit(...)
```

`contextlib.suppress(Exception)` catches all exceptions from the idle-pulse emit, including programming errors (wrong argument signature, type mismatches) that would never surface in logs. The best-effort / never-interrupt-the-loop intent is clearly documented in the comment, which is good. Consider narrowing to the specific exception types the event bus actually raises on transport failure (e.g. `asyncio.InvalidStateError`, `RuntimeError`), or falling back to a `try/except Exception: logger.debug(..., exc_info=True)` so swallowed errors at least leave a trace.

---

No credentials, API keys, private IP addresses, SSH material, or private hostnames found in any added lines. PR title, description, and CHANGELOG entry describe the observable fix without referencing private infrastructure, security incidents, or credential cleanup.

---

PR #697, opened 2026-06-19
feat: observation resolution lifecycle v1 — dashboard stages, ego read-receipts, provider recovery-resolve

WingedGuardian:
## What & why

Stale Genesis `observations` accumulate on the dashboard and in the ego's reasoning context with
no per-stage visibility — you couldn't tell what had been read vs acted on vs resolved, and the
only resolver was a time-based TTL. This is **observation lifecycle v1**: deterministic, with no
per-observation LLM deliberation (modelled on email read-receipts).

A real-DB audit reshaped the original plan. The TTL sweep turned out healthy, and the
ego-escalation path that prompted this had already self-healed — so the `subject_key` **supersession**
mechanism was **deferred** as having no current adopter. The shipped v1 is the data-backed subset:
de-noise the surfaces, add read/acted/resolved visibility, and fix a genuine write-only-recovery bug.

## Changes

- **`fix(routing)` — `provider_failure` resolves on breaker recovery.** `ProviderEscalation.record_recovery()`
  cleared in-memory tracking on recovery but never resolved the observation, so a recovered provider's
  "failing" alert lingered until TTL. It now resolves that provider's row, keyed on the deterministic
  per-provider `content_hash`, so **only the recovered provider clears** (a still-down one is untouched).
  The resolve is deferred + loop-guarded (the caller `CircuitBreaker.record_success` is sync). Adds a
  `resolve_by_content_hash` CRUD helper and a shared hash helper so create/resolve can't diverge.

- **`feat(dashboard)` — lifecycle stage badges.** The Observations panel now shows a
  `new / read / acted / resolved` badge per item, derived in the route from existing fields
  (`resolved` / `influenced_action` / `retrieved_count` / `surfaced_at`) — no schema change. Replaces
  the old binary "unsurfaced" dot; mark-read flips `new → read` optimistically.

- **`feat(ego)` — read-receipts + unread-first demotion.** Both ego context readers
  (`genesis_context._observations_section`, `user_context._genesis_escalations_section`) pulled
  unresolved observations into the prompt every cycle but recorded no read, so already-seen items
  re-surfaced as new (the "thematic rut"). They now record a **non-fatal** read-receipt and add an
  unread-first tiebreaker **after** the priority sort — so equal-priority seen items demote below
  unread ones while a critical is never hidden.

- **`fix(observations)` — `conversation_pivot` reclassified internal.** Per-session conversation
  telemetry was inflating the user-facing surfaces; it's now excluded from the dashboard / morning
  report / recurring-patterns / proactive outreach (still consumed by L1 essential-knowledge, which
  queries it by type directly, so unaffected). Plus an explicit `provider_failure` TTL backstop.

## Verification

- Targeted tests per piece + a broad regression sweep (1600+ green); every piece has new test coverage.
- **Code review** caught one off-by-one I introduced (a column-index shift after adding `id` to a
  SELECT broke the genesis-ego redirect-investigation signal) — fixed + regression-tested.
- **End-to-end:** the provider recovery-resolve verified through the *real* `CircuitBreaker` +
  `ProviderEscalation` chain (recovered clears, still-down untouched); the de-noise verified against
  real data (roughly halved the dashboard's unsurfaced count).

## Deferred (separate follow-ups)

- `subject_key` supersession machinery (no current adopter, per the audit).
- Part 2: Genesis-memory + `MEMORY.md` currency (reuses the same verifiable-recheck principle).

## Residuals

Live-deploy confirmations only — the badge rendering in-browser and the ego de-noise observed over
real cycles (both require merge + deploy). No built-not-wired paths; every changed path has a live caller.


WingedGuardian:

---

### WARNING — Raw SQL modified outside CRUD module

`src/genesis/ego/genesis_context.py:284` · `src/genesis/ego/user_context.py:764`

Both ego context readers execute `await self._db.execute("SELECT ...")` directly against `genesis.db` outside of `genesis.db.crud` modules. This PR adds `id` to the `SELECT` column list and `(retrieved_count > 0)` to the `ORDER BY` in both — extending pre-existing raw SQL, not introducing it.

The write-path (read-receipts via `increment_retrieved_batch`) correctly goes through CRUD. Flagging as a tracking item rather than a hard blocker since the violation pre-dates this PR and the changes here are minimal column additions. Next time these queries are meaningfully changed, consider pulling them into dedicated CRUD functions (e.g. `fetch_observations_for_ego_context` / `fetch_escalations_for_user_ego`).

---

**Everything else clean:**

- No bare `except:` clauses — all catch `Exception` or `RuntimeError` explicitly. ✓
- `resolve_by_content_hash` and `_observation_status` are new CRUD/module-level functions with full docstrings and tests. ✓
- All 4 changed source files (`observations.py`, `escalation.py`, `genesis_context.py`, `user_context.py`) have corresponding test additions; `test_observations_status.py` is a new file covering the new derivation function. ✓
- No `nohup`, no direct `task_states` inserts, no `/tmp/` usage, no force-push patterns, no mutable default arguments, no star imports. ✓
- Commit prefix `feat:` present in PR title. ✓
- No PII, secrets, private IPs, or wording issues in code diff, PR metadata, or CHANGELOG additions. ✓

---

PR #699, opened 2026-06-19
fix(update): don't block deploy on known-ephemeral local files (settings.local.json, AGENTS.md)

WingedGuardian:
update.sh's dirty-tree guard aborted on any modified tracked file. Two files are routinely dirty on a normal install and safe to leave in place — `.claude/settings.local.json` (CC rewrites local perms) and top-level `AGENTS.md` (GitNexus auto-stat block) — both untouched by the fast-forward merge and self-regenerating, yet they blocked every update. The guard now excuses exactly these top-level paths (`EPHEMERAL_DIRTY_RE`); real tracked changes (incl. `src/AGENTS.md`) still abort. Verified: filter logic + `bash -n`.

Note: takes effect on the deploy AFTER it lands (the guard runs the pre-pull copy of update.sh).


WingedGuardian (commented):

---

PR #700, opened 2026-06-20
fix: deploy + CI reliability — update.sh restart race + harvest test date-bomb

WingedGuardian:
Two reliability fixes surfaced by tonight's deploy verification.

## 1. update.sh ran stale code after a "successful" deploy (`scripts/update.sh`)

- update.sh stops genesis-server, merges new code, applies DB migrations, then "restarts". The systemd unit has `Restart=on-failure` + `RestartUSec=10s`. The stop falls back to `kill -TERM` when systemctl is flaky, which systemd treats as a **failure** → it **auto-restarted the server ~1s before the merge completed**, so the process imported pre-merge Python and cached it.
- Migrations still applied (the runner scans migration **files on disk**, not loaded modules) — masking the problem.
- The final restart used `systemctl start` — a **no-op on the already-running process** — so the new code never loaded. Net: **new code on disk + new DB schema, but old code in the live process.** Confirmed empirically: a cron added in new code never registered while an older sibling cron fired on time.

**Fix:** `_start_genesis_server` uses `systemctl restart` (not `start`) so the final step always reloads current on-disk code; new `_ensure_server_down()` does an explicit `systemctl stop` (disarms the on-failure restart timer) + confirms the process is gone before merge/migrations, aborting the update if it can't. Wired into the main path and `_do_rollback`.

## 2. Harvest test was a date-bomb blocking all CI (`tests/feedback/test_harvest.py`)

`_add_proposal` hardcoded `resolved_at=2026-06-18T01:00`, which falls outside `run(window_days=2)` once wall-clock passes 2026-06-20 01:00 UTC — `test_incremental_run_is_idempotent_on_unique_key` started failing (`assert 0==1`) on main, **blocking every PR's CI**. The harvester is correct (it windows properly); the test was time-brittle. Defaulted the helper timestamps relative to `now` so the suite is time-independent (tests asserting window *exclusion* still override with explicit old dates).

## Verification
- `bash -n` + `shellcheck -S warning` clean on update.sh; `_ensure_server_down` logic-tested with mocked systemctl/pgrep (returns 0 when down, aborts when stuck up). The `restart` mechanism was confirmed live tonight (manual restart correctly reloaded the deployed commit).
- `tests/feedback/test_harvest.py`: 23 passed, ruff clean.
- Reviewed by genesis-architect (design) + code-reviewer (implementation); findings applied.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,15 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Updates now reliably load the new code** — an update could finish "successfully"
+  while the running Genesis process kept executing the *old* code: when the updater
+  stopped the server, systemd's auto-restart could bring it back on the pre-update
+  code before the new code was even pulled, and the updater's final restart was a
+  no-op on the already-running process. The database migrated but the live process
+  didn't, leaving new code on disk and old code in memory. The updater now forces a
+  true restart at the end and makes sure the server stays down during the upgrade, so
+  an update always activates the version it just installed.
+
 - **Recovered providers stop alarming once they come back** — when a model provider's
   circuit breaker reopens after an outage, Genesis now clears that provider's "failing"
   alert instead of leaving it lingering for days until it expired. Per-session conversation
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

PR #706, opened 2026-06-20
fix(campaigns): record tick job_health so crashes are observable

WingedGuardian:
## Problem
The campaign runner owns a private APScheduler with no error listener, and `_tick_wrapper` swallowed all exceptions with only `logger.exception`. A tick crash on the live discord-engagement campaign was invisible to job_health -> JobHealthCollector -> ego/dashboard (verified: zero `campaign_*` rows in the live job_health table despite weeks of active runs).

## Fix
`_tick_wrapper` now records job health via `GenesisRuntime.instance().record_job_success/failure(job_id)`, mirroring the surplus scheduler's established pattern. The recording is best-effort (suppressed) and never propagates to APScheduler, so the tick-level swallow contract is preserved. The `job_id` (`campaign_<name>`) matches the scheduler job id so dashboard/reconcile align.

## Verification
- Unit tests: failure recorded + not propagated; success recorded; both suppress paths (success and failure) covered.
- Full `tests/test_campaigns/` suite green (38).
- Real-runtime e2e: a failing tick records `consecutive_failures: 1` + `last_error`; a clean tick resets to `0` with `last_success`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,11 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **A campaign that crashes mid-tick no longer fails silently** — when a scheduled campaign
+  tick raises an error, Genesis now records it in job-health tracking, so the failure surfaces
+  in the dashboard and to the ego instead of vanishing into the server log. Campaign
+  reliability problems become visible instead of going unnoticed.
+
 - **Surplus brainstorm messages read like prose, not raw JSON** — Genesis's background
   brainstorm ideas posted to the Telegram "Surplus" topic now render as clean bulleted text
   (idea, detail, and why it matters) instead of the raw ```json``` code block the model
--- a/src/genesis/campaigns/runner.py
+++ b/src/genesis/campaigns/runner.py
@@ -73,22 +73,37 @@ class CampaignRunner:
             )
             return
 
+        job_id = f"campaign_{camp['name']}"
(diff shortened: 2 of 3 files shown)


WingedGuardian:
## First-pass review — PR #706

### Structural review

**WARNING — Deferred import suggests circular dependency**

`src/genesis/campaigns/runner.py`, inside `_tick_wrapper`:

```python
from genesis.runtime import GenesisRuntime
```

This import lives inside the method body rather than at the module top level. That pattern is almost always a workaround for a circular import between `genesis.campaigns.runner` and `genesis.runtime`. The code works, but it hides the dependency from static analysis and IDE navigation, and it re-executes the module lookup on every tick (minor overhead, but avoidable).

Worth checking: can `GenesisRuntime` be imported at the top of `runner.py`? If not, that confirms a real circular dependency worth resolving at the module boundary (e.g. moving the shared type to a third module), even if it's out of scope for this PR.

---

Everything else is clean:

- No pattern violations (nohup, bare `except:`, mutable defaults, `/tmp/`, raw SQL, direct `task_states` inserts, etc.)
- Import grouping, conventional commit prefix (`fix(campaigns):`), docstring on the modified function — all correct.
- New behaviour is fully covered by four targeted tests in `tests/test_campaigns/test_runner.py`.
- CHANGELOG entry describes the functional fix clearly with no wording concerns.

**CLEAN** — no API keys, tokens, private IPs, credentials, personal addresses, private hostnames, or security-incident language found in added lines, PR title, or PR description.

---

PR #707, opened 2026-06-20
fix(surplus): humanize brainstorm findings in Telegram (strip JSON fences)

WingedGuardian:
## Problem
Surplus `BRAINSTORM_USER`/`BRAINSTORM_SELF` tasks are prompted to return a fenced `{"findings":[{title,content,relevance,...}]}` JSON object. `_post_to_telegram` posted that verbatim, so the user saw a raw ```json{...}``` code block in the Telegram "surplus" topic on every brainstorm.

## Fix
New `_humanize_surplus_content` helper, called once at the top of `_post_to_telegram`:
- Unwraps a whole-message ```json fence and, when the payload is a findings list, renders plain-text bullets (title / body / `Why:` relevance).
- Non-findings JSON, plain text, and non-JSON code fences (```python, bare ```) pass through untouched.
- Never raises. Only the Telegram post is humanized; the stored insight content is left raw for downstream consumers.

Complements #665 (which kept quotes raw); this strips the fences.

## Verification
- Unit + integration tests (humanizer edge cases + `_post_to_telegram` integration), incl. a regression test that the existing keeps-quotes-raw behavior is unchanged.
- Full `tests/test_surplus/` suite green (199).
- In-process e2e: a realistic multi-finding payload renders as clean bulleted prose with no fences or JSON keys.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,11 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Surplus brainstorm messages read like prose, not raw JSON** — Genesis's background
+  brainstorm ideas posted to the Telegram "Surplus" topic now render as clean bulleted text
+  (idea, detail, and why it matters) instead of the raw ```json``` code block the model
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

PR #710, opened 2026-06-20
fix(routing): third free fallback for procedure-extraction chain

WingedGuardian:
## What
Add `mistral-large-free` as a third fallback to the `38_procedure_extraction` routing chain (previously `[openrouter-deepseek-v4, groq-free]`).

## Why
With only two free providers, a simultaneous OpenRouter + Groq outage exhausted the chain — live errors `"All providers exhausted for 38_procedure_extraction"` (2026-06-18) — silently halting the self-learning routine that captures reusable procedures from struggle detection.

`mistral-large-free` is an independent vendor (survives an OpenRouter+Groq outage), verified healthy in live routing (0 self-failures, actively rescuing other call sites).

## Verification
- Loaded the edited config through the real `genesis.routing.config.load_config` parser: chain resolves to **3 registered providers** (the parser raises on any unknown provider, so a clean load is a real reference check).
- Fallback-only: `mistral-large-free` fires only when the first two providers fail — can only add resilience, never changes primary routing.

## Risk
Minimal — single config line; no code paths changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,11 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Procedure learning survives a two-provider outage** — the routine that captures reusable
+  procedures from Genesis's own struggles ran on only two free model providers; when both were
+  down at once it exhausted its chain and silently stopped learning. A third independent free
+  fallback now keeps it working through overlapping provider outages.
+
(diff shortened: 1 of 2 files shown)


WingedGuardian:
Structural review: SKIPPED — config/changelog-only PR (no `.py` files changed).
PII/Secrets/Wording scan: CLEAN — no sensitive data, private IPs, credentials, or incident-disclosure language found in diff or PR metadata.

---

PR #712, opened 2026-06-20
fix(observability): de-noise L1 (ghost sessions, harness blobs, wings)

WingedGuardian:
Three small "cognitive-state hygiene" fixes — each targets noise that was showing up on the dashboard and in Genesis's own always-on L1 context. All verified against current `main`.

## Fixes

**RC-1 — ghost zero-message sessions** (`scripts/genesis_session_end.py`)
A session that ended before any messages were exchanged (e.g. mid server-restart) still wrote a patch into `session_patches.json`, producing ghost "no topic · 0 msgs" rows in the dashboard recent-sessions list. Now gated on `if messages:`. (`_LAST_SESSION_FILE` — a separate single-value temporal-awareness file — is intentionally left as-is; it does not feed the recent-sessions list.)

**#16 — harness noise in L1 "Active Work"** (`scripts/proactive_memory_hook.py`)
The UserPromptSubmit hook recorded a `conversation_pivot` for every prompt, including harness-injected turns (`<task-notification>`, `<system-reminder>`, slash-command and local-command envelopes). Those raw blobs then surfaced in L1 "Active Work" and the session-trail line. New `_is_harness_envelope()` early-returns for those turns, so only genuine user prompts become pivots.

**#17 — wing prefix leak + vocab pollution** (`src/genesis/memory/store.py`, `essential_knowledge.py`)
Synthesis prompts model tags as `wing={wing}`; the LLM occasionally echoed the literal `wing=channels` value, stored verbatim. New `_strip_kv_prefix` sanitizes `wing`/`room` at the top of `store()` — before the value fans out to the FTS5 tag, the Qdrant payload, and `memory_metadata`. The L1 `_wing_stats` view now filters to the controlled `taxonomy.WINGS` vocabulary (display-only; no DB mutation).

## Verification
- 50 targeted tests pass (RC-1 `main()` end-to-end both sides; harness-envelope detection incl. `<command-flag>`; store strip across all three sinks; `_wing_stats` canonical filtering). Lint clean.
- Wiring confirmed: hook passes the real prompt into the trail fn; `_strip_kv_prefix` precedes all three store sinks.
- Inline code-review pass (one finding applied: collapsed `<command-*>` entries to a single `<command-` prefix to catch `<command-flag>` and future variants).

## Risk
Low. RC-1 and #16 only *suppress* writes that shouldn't happen; #17's strip only fires on a leading `key=`/`key:` prefix (real values pass through), and the wings filter is display-only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   lane. They're now recorded at low priority — still tracked for trend deltas, just no longer
   treated as urgent.
 
+- **Genesis's at-a-glance state views stopped showing internal noise** — three cleanups to the
+  dashboard and to Genesis's own always-on context: empty sessions (ended before any messages were
+  exchanged) no longer appear as ghost "0 msgs" rows in the recent-sessions list; the "Active Work"
+  summary no longer ingests raw harness notifications (task-completion blobs, system reminders,
+  slash-command metadata) as if they were your prompts; and the memory "Wings" breakdown shows only
+  real, controlled-vocabulary domains instead of malformed or one-off tags. What you — and Genesis —
+  see reflects genuine activity, not plumbing.
+
 - **A campaign that crashes mid-tick no longer fails silently** — when a scheduled campaign
   tick raises an error, Genesis now records it in job-health tracking, so the failure surfaces
   in the dashboard and to the ego instead of vanishing into the server log. Campaign
--- a/scripts/genesis_session_end.py
+++ b/scripts/genesis_session_end.py
@@ -163,14 +163,19 @@ def main() -> None:
         except OSError as exc:
             print(f"SessionEnd hook: failed to write pending bookmark: {exc}", file=sys.stderr)
 
-    # 3. Append session patch for cognitive state freshness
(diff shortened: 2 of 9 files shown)


WingedGuardian:
## First-pass review — PR #712

### Structural review

**WARNING — Import hygiene** (`src/genesis/memory/essential_knowledge.py`)

`from genesis.memory.taxonomy import WINGS` is placed inside the body of `_wing_stats` rather than at module top-level:

```python
async def _wing_stats(db: aiosqlite.Connection) -> dict[str, int]:
    """..."""
    from genesis.memory.taxonomy import WINGS   # ← function-level import
    try:
        ...
```

Function-level imports are typically a signal of a circular import being worked around. If that's the case here, add a brief comment (`# avoid circular import`) so the intent is clear to the next reader. If there's no circular dependency, move it to the top-level import block with the other local imports.

Everything else is clean: no bare `except:`, no `/tmp/` usage, no `nohup`, no raw SQL writes, no mutable defaults, no star imports, no task_states direct inserts. New test file `test_wing_stats.py` covers the new behavior; all other changed paths have corresponding test coverage.

---

**CLEAN** — No API keys, tokens, credentials, private IPs, MAC addresses, private hostnames, or personal email addresses in any added line. PR title, body, and CHANGELOG addition all use functional/UX framing with no language implying a security incident or credential/PII cleanup.

---

PR #715, opened 2026-06-20
feat(autonomy): evidence-gated earn-back promotion proposals

WingedGuardian:
Wires an **evidence-gated, user-approved "earn-back"** path so a demoted autonomy category isn't stuck forever. `promote()`/`restore_earned_level()` previously had zero non-test callers and `record_success` no longer auto-promotes, so once a category regressed it could never climb back.

## Design (user-chosen: evidence-gated)
- **Detection** — `AutonomyManager.detect_earnback_candidates()`: a category is a candidate iff `current_level < earned_level` AND `bayesian_level(successes, corrections) >= earned_level`. The Bayesian gate keeps it **silent while the lower level is genuinely correct** (e.g. the live `direct_session` at 33S/14C → posterior 0.694 → L3 < earned L4 does **not** fire) and only proposes once the category re-earns the level. `consecutive_corrections` is deliberately excluded (it resets to 0 on regression).
- **Proposal** — `EgoCadenceManager._check_earnback_opportunities()` runs in the existing 30-min `_sweep_with_expiry`, **user-ego only**, raising an `autonomy_earnback` proposal. Anti-spam: skip a category with a pending earn-back proposal (+ create_batch content-hash dedup) or an active **7-day reject cooldown**.
- **Approval → promote** — `ego/earnback.py::handle_earnback_resolution` is a narrow, shared hook called from **all four** resolution paths (Telegram `resolve_proposals`, the `ego_proposal_resolve` MCP tool, and the two dashboard routes). On approval it re-reads state, **skips the promote if a regression postdates the proposal**, otherwise calls `manager.promote()` and marks the proposal `executed`. On reject it sets the cooldown. It is a no-op for any non-earn-back proposal and never changes the other paths' behaviour.
- **Guard B** — the approved-proposal sweep (`_sweep_approved_inner`) skips `autonomy_earnback` so it can never be dispatched as a session.

**Never auto-promotes** — every restoration is the user's explicit decision, and it respects a legitimate demotion (no nagging while the evidence agrees the lower level belongs).

## Architecture review
Ran a `genesis-architect` pass **before** coding (it steered me off a risky "re-route all four paths through `resolve_proposals`" approach toward the narrow shared hook) and a `code-reviewer` pass after, which found 3 issues — all fixed: (1) wrap the Telegram-path hook in try/except so a DB error can't abort sibling-proposal resolution; (2) parse ISO timestamps instead of comparing strings in the staleness guard; (3) mark the proposal `executed` even when `promote()` declines, so it can't linger `approved` or block re-proposing.

## Verification
- **24 new tests** + no regression in 129 existing ego tests. Includes an **E2E test through the real resolution path with a real `AutonomyManager`**: seed a demoted-but-recovered category → detect → proposal → approve → real `promote()` fires → `current_level` 2→4, proposal `executed`. All runs against an in-memory test DB — **no live autonomy state touched** (`direct_session` stays at L3, which is correct).
- No DB migration (the `autonomy_state` regression-timestamp columns already exist). Lint clean.
- Docs: `autonomy-deep-dive.md` updated (its "promotion caps at +1 per success" line was stale — promotion is approval-gated now) + earn-back documented.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **Earned autonomy can be restored after a regression** — when Genesis loses a level of
+  autonomy in a category (e.g. after a correction), that demotion is no longer a dead end. Once
+  the category's track record recovers enough that the evidence again supports the earned level,
+  Genesis proposes restoring it and asks you to approve — it never silently re-grants authority,
+  and it won't nag while the lower level is genuinely warranted. Previously a demoted category
+  had no path back up.
+
 - **Genesis self-corrects a bad memory recall instead of running with it** — on high-stakes
   lookups (the explicit memory and knowledge recall tools), Genesis now grades whether the
   recalled results are actually on-topic, and when a recall comes back clearly irrelevant it
--- a/docs/architecture/autonomy-deep-dive.md
+++ b/docs/architecture/autonomy-deep-dive.md
@@ -91,9 +91,19 @@ reasonable inference from sparse data.
 | 10 | 5 | 0.73 | L3 |
 | 50 | 5 | 0.90 | L4 |
 
-Promotion caps at +1 level per success — prevents overfitting to lucky streaks.
-Regression has no such cap. A single correction can drop multiple levels if the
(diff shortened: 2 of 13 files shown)


WingedGuardian (commented):

---

PR #711, opened 2026-06-20
feat(memory): selective corrective retrieval (CRAG) on high-stakes recall

WingedGuardian:
## What & why

Genesis was expected to have CRAG (corrective retrieval) wired; only the calibrated
rubric existed. **W-CRAG v1** closes that: when an explicit `memory_recall` /
`knowledge_recall` returns clearly off-topic results, Genesis grades them and
self-corrects — broadening the search, drawing on the knowledge base (and, for
knowledge queries, the web) — instead of feeding itself junk context. Follows
canonical CRAG (arXiv 2401.15884): lightweight evaluator → confidence → 3-bucket policy.

## Design (decisions locked with the user)

- **Grader:** new `crag_grade` call site (fast cross-vendor: `openrouter-haiku` →
  free fallbacks), reusing a new `memory_recall_grounding_runtime` rubric variant
  (the calibrated relevance rubric, minus the eval-only `{expected}` placeholder so
  runtime traces don't pollute eval calibration). Added to degradation skip sets.
- **Conservative rollout:** latency pre-gate (skip grading when the top score is
  already confident); **ACT only on a clear `Incorrect` verdict** (relaxed
  re-retrieve + raw KB; **web fallback on the knowledge path only** — episodic
  memory isn't on the web); **`Ambiguous` is DARK** (logged, no action) so the
  grader's thresholds can be calibrated on real traffic before widening.
- **Best-effort & latency-safe:** per-grade `asyncio.wait_for` timeout + graceful
  pass-through on any failure → a working recall is never degraded or stalled, even
  if the grader is down. Augmentation uses the **raw retriever** (no self-nesting).
- **Wiring:** default-ON on `memory_recall` (non-compact path) + `knowledge_recall`.
  Latency-sensitive paths (proactive hook, voice, CC context injection, ego) use the
  raw retriever and are **unaffected**. `emit_recall_corrected` logs every decision
  for calibration.

## Verification

- ruff clean; **139 tests pass** — 21 new `test_corrective.py` unit tests + regression
  across `test_mcp/test_memory_mcp`, `test_memory/test_mcp_integration`,
  `test_dashboard/test_tool_api`, `test_eval/*` (no regressions).
- **E2E against the real grader** (`openrouter-haiku`): relevant memory → score **1.0**
  → Correct (kept, no augmentation); irrelevant → **0.0** → Incorrect → augmentation
  fires (`relaxed` + raw `source=knowledge`), no web on the memory path; calibration
[...]

PR #716, opened 2026-06-20
feat(memory): knowledge-base provenance — external-world vs first-party (WS-7, D12)

WingedGuardian:
## WS-7 — Knowledge-base provenance (audit D12)

**Problem:** external-world knowledge (the `knowledge_base` collection — ingested docs/APIs/papers, and now CRAG web results) could masquerade as first-party ground truth once recalled into an LLM context. The self-model couldn't tell what it *learned from the world* from what it *observed itself*, so it over-trusted ingested/scraped content.

**Fix:** draw the first-party-vs-external line at LLM-context entry.

- **Discriminator:** `RetrievalResult.collection` (new, defaulted, set at all 3 construction sites from the authoritative Qdrant collection — reliable where the per-item `source` string and `payload["scope"]` were not).
- **Shared labeller:** `genesis.memory.provenance` — `is_external`, `provenance_descriptor` ("first-party memory" vs "external-world knowledge (source: …)"), `short_source`, and `label_result_dicts` (a final MCP-return pass).
- **Surfaced everywhere KB enters context:** `memory_recall` (compact inline + full post-pass), `memory_expand` (builds a mid→collection map since it bypasses `RetrievalResult`), `knowledge_recall` (all-external), the proactive hook (`[KB·<tier>]`), the ego user-corrections recall (pinned to `source="episodic"` so KB can't pose as a correction), the voice handler (prefixes external snippets), and the dashboard memory-search API.
- **Reliability fix:** the `memory_recall` source="both" KB score floor now keys on the collection discriminator instead of `payload["scope"]` (absent on FTS5-only rows, so the floor previously slipped).
- **CRAG (#711) interaction:** corrective retrieval augments `memory_recall`/`knowledge_recall` with relaxed/raw-KB re-retrieval and a web fallback whose reduced-shape dicts dropped the discriminator. Labeling runs as a **final pass after** `maybe_correct_recall`, so original, augmented, **and web** items are all labeled; `_retrieval_to_dict` now preserves `collection`+`source_pipeline`; `crag_web` → "external-world knowledge (source: web)".

No schema migration (provenance already stored; collection always known at retrieval).

## Verification
- 1776 tests pass across memory / MCP / hooks / ego / voice / dashboard / routing (incl. CRAG's own `test_corrective.py` — #711 intact).
- Focused adversarial code review (CRAG interaction): no findings.
- **E2E on real Qdrant + real embeddings:** invariant held — every `knowledge_base` hit labeled external (with correct source tier), every episodic hit first-party; the MCP tools (`memory_recall`/`knowledge_recall`/`memory_expand`) surface correct provenance on live data.

### DIFF
--- a/.claude/docs/mcp-tools-guide.md
+++ b/.claude/docs/mcp-tools-guide.md
@@ -37,6 +37,14 @@ If it's a repeatable process → `procedure_store`. If it's deferred work →
 4. `reference_lookup` if it's a credential/URL/IP
 5. `procedure_recall` if it's a known workflow
 
+**Provenance on recall results (first-party vs external-world):** `memory_recall`,
+`knowledge_recall`, and `memory_expand` tag every result with a `provenance` field —
+`first-party memory` (Genesis's own observations/decisions/conversations) vs
+`external-world knowledge (source: …)` (the knowledge base, ingested docs, or corrective
+web results). Treat external-world content as information *about* the world to weigh, not
+Genesis's own ground truth. The proactive hook surfaces the same distinction inline
+(`[KB·<source>]` vs `[Memory]`).
+
 ## Health Debugging — Escalation Path
 
 Start broad, drill into specifics:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   Genesis proposes restoring it and asks you to approve — it never silently re-grants authority,
   and it won't nag while the lower level is genuinely warranted. Previously a demoted category
   had no path back up.
+- **Genesis tells its own memories apart from what it read on the world** — every recalled
+  knowledge-base item (ingested docs, and the new corrective web results) is now labeled
+  "external-world knowledge (source: …)" wherever it reaches Genesis's context: explicit
+  recall, the proactive memory hook, voice, and the dashboard memory search. First-party
+  memories (Genesis's own observations and your conversations) stay labeled as such, so the
+  model never mistakes an ingested document — or a web snippet — for its own ground truth.
+  The knowledge-base relevance floor that keeps low-quality bulk content out of answers now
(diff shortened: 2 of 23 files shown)


WingedGuardian (commented):
## First-pass review — WS-7 / D12 KB provenance

**Structural review:** 3 WARNINGs (no ERRORs)

---

### WARNING — Import inside function body (×2)

Imports belong at the top of the file, not inside function bodies. Two new lazy imports were added in production code:

**`src/genesis/channels/voice/handler.py`** — inside `handle()`:
```python
from genesis.memory.provenance import is_external
```

**`src/genesis/dashboard/routes/memory.py`** — inside `memory_search()`:
```python
from genesis.memory.provenance import provenance_descriptor
```

`genesis.memory.provenance` has no heavy transitive dependencies (`from __future__ import annotations` only), so there's no startup cost or circular-import concern that would justify the lazy placement. Move both to file top-level with the rest of the imports.

---

### WARNING — Hook script importing from genesis package

**`scripts/proactive_memory_hook.py`** — inside `_format_results()`:
```python
from genesis.memory.provenance import short_source
```

Genesis convention: hook scripts must stay stdlib-only or `genesis.env`. Importing from `genesis.memory.*` at function-call time means the venv must be fully importable every time this hook fires. The module itself is lightweight, but the convention exists to keep hooks resilient to partial-startup states. If the hook already depends on genesis packages elsewhere, this is an existing violation that predates this PR — but it's worth tracking either way. Consider exposing the `short_source` lookup as a pure-dict mapping (inline in the hook) to avoid the dependency, or document the pre-existing hook/genesis coupling so the rule can be updated to reflect reality.

---

Everything else looks solid: the new `provenance.py` module is well-factored with complete docstrings, the `collection` field defaults safely to `"episodic_memory"` (never mislabels external), the FTS5-only path fix is covered by a targeted test, all new `src/` files have corresponding tests, and the CRAG interaction (post-`maybe_correct_recall` labeling pass) is both documented and tested end-to-end.

---

PR #717, opened 2026-06-20
feat(learning): cognitive self-modification rollback ledger

WingedGuardian:
## Problem

Genesis autonomously **overwrites its own cognitive config files at runtime with no backup and no undo** — skill `SKILL.md` refinement (weekly), `TRIAGE_CALIBRATION.md` regen (daily), and `USER_KNOWLEDGE.md` synthesis (daily). If a self-edit degrades cognition, there is no way to revert it. Two of these files are gitignored, so git is not a rollback path.

## What this adds

An append-only **cognitive self-modification ledger** that captures the pre-image before each overwrite, plus a drift-guarded rollback and operator MCP tools.

- **`cognitive_file_modifications` table** (migration `0027`) — stores `prior_content` + `applied_content` inline (files <8 KB; covered by the DB backup). Named to avoid the pre-existing dead `file_modifications` table (CC-tool audit, zero callers — removal tracked as a follow-up).
- **`learning/cognitive_ledger.py`** — `record_file_modification` (read pre-image → unique-temp same-dir atomic write → best-effort ledger insert → prune-on-write, keep 30/target), `record_existing` (record a write the caller already did), `rollback(mod_id, force=False)` (drift guard: refuse if the on-disk file changed since, unless forced; restore the pre-image or delete the file; emits an observation on every outcome), `recent`.
- **3 wired chokepoints** — the skill applicator and triage calibration route their write through the ledger; the user-knowledge synthesis brackets the existing loader call (no loader change).
- **Operator MCP tools** — `cognitive_modification_status` (read) and `cognitive_modification_rollback` (revert).

## Design

- **The write is primary; the ledger is best-effort** — a failing ledger insert is logged and swallowed, never blocking cognition; a real file-write failure still propagates.
- **Operator surface only** — no automated cognitive path reads the ledger (grep-verified), same discipline as `ego_calibration_snapshots`. Auto-rollback on a degradation signal is a deliberate, soak-gated follow-up.

## Verification

- 90 targeted tests (CRUD, ledger record/rollback/drift/prune, wiring, MCP surface, migration up/down/idempotency/fresh-install); ruff clean.
- genesis-architect design review + code-reviewer (a concurrent-temp race was caught and fixed: unique temp name + cleanup-on-failure, with a regression test).
- End-to-end against a copy of the live DB: migration applies cleanly, record → rollback restores, drift refused, force overrides, observations emitted, no orphaned temp files.

## Follow-ups (tracked)

- Remove the dead `file_modifications` table (separate schema-deletion change).
- Auto-rollback consumer wired to the Outcome Bus degradation signal (soak-gated).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **You can undo a change Genesis made to its own skills or calibration** — when Genesis
+  autonomously refines a skill, retunes its triage calibration, or re-synthesizes its user
+  knowledge, it now keeps a recoverable snapshot of the previous version. If one of those
+  self-edits turns out worse, you can list the recent self-modifications and roll any of them
+  back to its prior contents — with a safety check that refuses to overwrite a file that has
+  changed since (unless you force it).
+
 - **Genesis self-corrects a bad memory recall instead of running with it** — on high-stakes
   lookups (the explicit memory and knowledge recall tools), Genesis now grades whether the
   recalled results are actually on-topic, and when a recall comes back clearly irrelevant it
--- a/src/genesis/db/crud/cognitive_file_modifications.py
+++ b/src/genesis/db/crud/cognitive_file_modifications.py
@@ -1,0 +1,149 @@
+"""CRUD for ``cognitive_file_modifications`` — the cognitive self-mod ledger.
+
+Thin, strict data layer. Append-only ledger of cognitive-config file overwrites
+(skill ``SKILL.md`` / ``TRIAGE_CALIBRATION.md`` / ``USER_KNOWLEDGE.md``), each row
+capturing the PRE-IMAGE so the write can be rolled back. ``record`` raises
+``ValueError`` on malformed input; the best-effort leniency that protects the live
+cognitive write paths lives one layer up in ``learning/cognitive_ledger.py``.
+
+NOT the (dead, CC-tool-audit) ``file_modifications`` table — different scope, and
+this one stores ``prior_content``/``applied_content``.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import json
+import logging
+import uuid
+from datetime import UTC, datetime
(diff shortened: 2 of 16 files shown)


WingedGuardian (commented):
## First-pass review — structural + PII/wording

### Structural: 1 WARNING

**WARNING — deferred import inside function body (`src/genesis/learning/skills/applicator.py`)**

The import added to `apply()`:

```python
from genesis.learning.cognitive_ledger import record_file_modification
```

is placed inside the method body rather than at the module top level. The other wired chokepoint in this same PR (`calibration.py`) correctly places the equivalent import at the module level. There is no circular-import path that would require deferring it here (`cognitive_ledger` → `cfm_crud` → no path back to `applicator`), so this looks like an accidental placement. Move to the module-level import block.

The deferred import in `runtime/init/learning.py` is intentional (inside a best-effort `try/except` block — if the import itself fails, the whole ledger capture is swallowed cleanly), so that one is fine.

---

- No API keys, tokens, secrets, or credentials in added lines.
- No private IPs, IPv6 ULA addresses, SSH keys, or DB connection strings.
- No personal emails, username-bearing absolute paths, hardware identifiers, or private hostnames (test-fixture paths like `"/skills/x/SKILL.md"` are excluded per policy).
- PR title, body, and CHANGELOG addition contain no incident language ("fixed leak", "removed credentials", "cleaned up exposure", etc.).

---

Everything else is clean: no `nohup`, no raw `task_states` insert, no `/tmp/` temp usage (same-dir atomic write confirmed in code), no bare `except:`, no mutable default arguments, all new public functions documented, all four new `src/` modules have corresponding test files, conventional commit prefix correct.

---

WingedGuardian:
Merge status: CI all green (8/8 — lint, leak-detector, migration-check, test [8,567 passing], CodeQL ×4). Codex automated review was **quota-limited** (no findings). First-pass structural review left 1 WARNING (deferred import in `applicator.py`) — addressed in `56983773` (hoisted to module level). An inline code-review during development also caught and fixed a concurrent-temp race. Squash-merging.

PR #718, opened 2026-06-20
feat(observability): tracing/spans backbone — see what the agents do

WingedGuardian:
## What

A homegrown, OpenTelemetry-convention **tracing backbone**: a *trace* is one logical Genesis operation (a reflection cycle, an ego cycle, a dispatched Claude Code session, an LLM call) decomposed into parent/child **spans** with timing, status, and attributes, persisted to a new `otel_spans` table. This is the data layer for inspecting **what Genesis actually does**, end to end, instead of piecing it together from logs — the foundation a trace-waterfall dashboard will render.

No OpenTelemetry SDK — the field semantics follow OTel conventions so a `to_otel()` export adapter is trivial later, but nothing depends on the SDK.

## Why

A live demo surfaced two gaps; this closes the second ("can't show what its agents do"). Routing/health honesty was the first (already shipped).

## How (6 phases, each its own commit)

- **A0** — `otel_spans` schema (migration 0027 + fresh-install DDL). Dual clocks: wall-clock epoch-µs for cross-process ordering, monotonic-derived duration.
- **A1** — span primitives: `Span` + `_current_span` ContextVar + `start_span()` context manager; batched `SpanWriter` cloned from the `ProviderActivityTracker` transport (sync record → in-memory batch → tracked-task flush → `INSERT OR IGNORE`). Best-effort and transparent — only the caller's own exception ever propagates.
- **A2** — every LLM call traced at the single `route_call` chokepoint (logic moved verbatim into `_route_call_inner`). Denormalized provider/model/tokens/cost written from the same `RoutingResult` that feeds `cost_events` (no value drift; `cost_events` stays authoritative for billing).
- **A3** — operation root spans (`reflection.cycle`, `ego.cycle`, `cc.session`) so LLM calls nest meaningfully; cross-process trace context (`GENESIS_TRACE_ID`/`GENESIS_PARENT_SPAN_ID`) injected into dispatched CC sessions.
- **A4** — CC tool spans: a PostToolUse hook drops JSONL records (fires only for dispatched, traced sessions) and a server-side ingest drains them into `otel_spans`. Flat-file + async ingest **deliberately not a cross-process DB writer** — a `.*` hook fires far more often than the existing Write/Edit audit hook, and a high-frequency cross-process WAL writer risks lock contention.
- **A5** — read APIs (`get_trace`, `list_recent_traces`, `get_session_spans`), daily retention prune, and a config/dashboard gate (`observability` settings domain) + the `GENESIS_SPANS_DISABLED` kill switch.

## Safety

- **Hot-path**: span capture never blocks or breaks routing — proven by a test where the writer *raises* and `route_call` still succeeds. No `await` on the span path; the flush is a tracked background task.
- **WAL**: span writes are batched and best-effort; the CC hook writes a flat file, never the DB.
- **Kill switch**: capture is config- and env-gated, read once at startup.

## Verification

- ruff clean; **914 pass / 2 skip / 0 fail** in CI-order across routing + observability + cc; ~100 new tests.
- Two independent code reviews — hot-path safety, ContextVar isolation, and migration↔fresh-install schema parity confirmed; all real findings fixed.
- **Full E2E (20/20)** through the real production classes: migration on a copy of the live DB, the real bootstrap `init()` wiring, a real `Router` producing a nested trace, the real hook+ingest stitching a CC tool span cross-process, and the kill switch.

## Activation & scope

- Capture turns on at the **next server deploy** (pull + restart); off-by-config/env until then.
- **Deferred** (reserved span kinds, no callers yet): MCP-recall spans, executor task/step spans.
- A pre-existing, unrelated test-isolation bug (a logging-config test that doesn't restore state) was found and tracked separately — not touched here.


WingedGuardian:
## Structural Review

### ERRORs
None found.

Checked for: `nohup`, `INSERT INTO task_states`, `/tmp/` paths, `git add .` / `git add -A`, `git push --force` / `git reset --hard`, `pip install -e`, bare `except:` (all exceptions typed as `except Exception`), mutable default arguments, `os.killpg`, raw SQL against `genesis.db` outside migration files. None present.

### WARNINGs

**W1 — Import of a private symbol across module boundary (`span_ingest.py`)**
`span_ingest.py` does `from genesis.observability.span_writer import _INSERT as _SPAN_INSERT`. The rationale ("single source of column order") is sound, but importing an underscore-prefixed private from a sibling module creates fragile coupling: any internal refactor of `span_writer` (rename, format string change) silently breaks ingest. Consider promoting `_INSERT` to a public name like `SPAN_INSERT_SQL` exported from `span_writer.py`.

**W2 — Deferred `aiosqlite` import inside each reader function (`span_reader.py`)**
`list_recent_traces`, `get_trace`, and `get_session_spans` each do `import aiosqlite as _sql` at call time solely to set `db.row_factory`. `aiosqlite` is already a declared dependency and the `db` parameter is already an `aiosqlite.Connection` — the pattern hides an always-present dependency and makes static analysis harder. Consider moving to the module top or using `TYPE_CHECKING` consistently.

**W3 — Streaming span path has no coverage gap test (`test_invoker_spans.py`)**
`run_streaming` delegates to `_run_streaming_inner` via the same span-capture pattern as `_run_inner`, but no test patches `_run_streaming_inner` to verify the streaming path correctly captures output cost/tokens in the span. Not a correctness bug, but the parallel to the existing `run_records_cc_session_span` test should be closed.

**W4 — `cost_known: bool = True` default may silently misclassify error/unknown-cost paths (`routing/types.py`)**
The new field on `RoutingResult` defaults to `True`. Any call site that constructs `RoutingResult` manually for an error or zero-cost scenario without setting `cost_known=False` will have the span incorrectly record "cost is measured." The PR's own code appears consistent; downstream manual constructors in pre-existing code were not audited here — worth a grep before merge.

### NOTEs

**N1 — New public functions without docstrings**
- `span_reader.py`: `flatten_tree`, `get_session_spans`
- `spans.py`: `current_trace_context` (cross-process contract surface, worth documenting)
- `span_ingest.py`: `_incoming_dir`, `_incoming_files`

**N2 — `_NullSpan.trace_id = None` vs `Span.trace_id: str` type asymmetry (`spans.py`)**
`_NullSpan` sets `trace_id = None` while `Span.trace_id` is typed `str`. The `current_trace_context()` gate handles this correctly today, but any future call site that dereferences `.trace_id` directly on a `Span | _NullSpan` return must handle `None`. Worth a comment at the definition site for future readers.

**N3 — `span_config.py` resolves repo root via `parents[3]` from `__file__`**
`Path(__file__).resolve().parents[3] / "config" / "observability.yaml"` silently breaks if the module moves. An inline comment already acknowledges this — flagging it as a maintenance consideration only.

---

## PII / Secrets / Wording Scan

### Hard Blocks
None found.

Scanned all added lines for API keys, tokens, passwords, private IP ranges (10.x, 192.168.x, 172.16–31.x, 100.64–127.x), IPv6 ULA (fd00::/8), SSH key material, and DB connection strings with credentials. None present.

### Soft Warnings
None found.

Scanned for personal email addresses, hardcoded `/home/<username>/` paths, AWS/GCP account IDs, hardware identifiers, private hostnames, and timezone identifiers in non-config files. None present. All `~/.genesis/` paths use the tilde form throughout.

### Wording Issues
None found.

PR title, body, and any CHANGELOG additions were scanned for language referencing PII/secrets leaks, credential exposure, security incidents, or private infrastructure names. None present.

---

## Overall Verdict

**COMMENT ONLY** — no structural violations, no secrets or PII. Two findings worth addressing before or shortly after merge: the cross-module private-symbol import (W1) and the missing streaming span test coverage (W3). W4 is a low-risk gap worth a quick grep. All other findings are informational.

---

PR #719, opened 2026-06-20
fix(memory): allow multiple link types between a memory pair (DLI-04)

WingedGuardian:
## What & why (audit DLI-04 / decision D15)

Genesis's `memory_links` table had a primary key of `(source_id, target_id)`, so only **one** link could exist between any two memories. A second link of a *different* type between the same pair — e.g. marking a pair `contradicts` when it was already linked `supports`, or `succeeded_by` when one memory replaces another — raised a UNIQUE violation that the writers (linker / connection_pass / dream_cycle) catch and swallow. The edge was **silently dropped**, so the memory graph quietly lost contradicting and superseding relationships.

This makes `link_type` part of the primary key — `(source_id, target_id, link_type)` — so distinct relationship types between the same pair coexist.

## Changes
- **`migrations/0029_memory_links_link_type_pk.py`** (new) — rebuilds the table with the new PK. No dedup needed (the old PK was *stricter*, so existing rows are already unique under the new one); idempotent (`PRAGMA table_info` PK check); obeys the runner's no-commit contract; ships a `down()`.
- **`_tables.py`** — canonical PK updated so fresh installs match.
- **`crud/memory_links.py`** — `inter_candidate_links` now `SELECT DISTINCT` so a multi-type pair doesn't inflate the binary adjacency boost in retrieval; `delete` gains an optional `link_type` to remove a single typed edge instead of every edge for the pair.

## Verification
- 21 new/updated unit tests (CRUD behavior + a full migration matrix: 2nd-type persists, rows survive, same-triplet still rejected, indexes recreated, idempotent, missing-table no-op).
- 751-test blast-radius sweep across the `memory_links` consumers (linker, graph, dream-cycle, dream-link-repair, supersession, health, retrieval) — all green.
- **Real `MigrationRunner` applied `0029` to a copy of the production DB**: 125,170 rows preserved, `integrity_check: ok`, 2.85s, idempotent, multi-type insert confirmed working.
- Code review: clean (no P1/P2).

Readers verified compatible. `graph.py`'s `DiGraph` collapses multi-type pairs to one edge — a documented, pre-existing limitation (true multi-relational traversal would need `MultiDiGraph`); out of scope here.

Part of audit work-stream **WS-5** (memory retrieval correctness). Held for review + merge approval.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,6 +97,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Genesis no longer loses contradicting or superseding links between memories** — its memory
+  graph could only hold one relationship between any two memories, so recording a second kind
+  (for example marking a pair as "contradicts" when they were already linked as "supports", or
+  "succeeded_by" when one memory replaces another) was silently dropped. Different relationship
+  types between the same two memories are now all kept, so Genesis reasons over a fuller, more
+  honest picture of how its memories relate.
+
 - **Procedure learning survives a two-provider outage** — the routine that captures reusable
   procedures from Genesis's own struggles ran on only two free model providers; when both were
   down at once it exhausted its chain and silently stopped learning. A third independent free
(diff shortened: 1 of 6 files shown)


WingedGuardian (commented):

---

PR #720, opened 2026-06-20
feat(ego): goal-pursuit upgrade — stuck detection + recommend-and-apply status changes (Phase 6)

WingedGuardian:
## What

Phase 6 of the self-improvement roadmap: Genesis engages actively with stalled goals while keeping the user in full control — **recommend-only, no autonomous goal writes**.

### Feature A — stuck detection
A stale goal that has been *worked on* (≥2 executed proposals) but still isn't advancing is now classified **stuck** rather than merely **stale**. Stuck goals get a higher-priority `goal_review` signal, and the goal deep-dive context prompts the ego to diagnose *why* prior effort didn't move the goal (wrong strategy / blocker / needs decomposing) — framed as a hypothesis, explicitly **not** "just close it" (anti-timidity).

`updated_at` is not a reliable progress proxy (bumped by progress notes / edits, not by proposal execution), so the executed-proposal count is the signal.

### Feature B — recommend-and-apply
The ego's existing `pause`/`deprioritize` goal recommendation stops being a passive observation and becomes an **approvable proposal**. It is applied **only on user approval**, never autonomously, and without dispatching a session. `close` stays a passive observation (achieve-vs-abandon is the user's call).

Implementation clones the proven autonomy earn-back pattern: build a `goal_status_change` proposal out-of-band (`create_batch` + `send_digest`, bypassing the realist gate) and apply it on approval via a shared handler (`ego/goal_actions.py`) wired into **all four** proposal-resolution entry points (Telegram, MCP tool, two dashboard routes). A regression test fails if any entry point drops the hook.

## Safety / invariants
- **Recommend-only:** the only goal-state write in the diff is the approval-gated handler (grep-proven). No path changes goal status/priority without user approval.
- Validated transitions only (`paused`, priority notch) — `achieved`/`abandoned` can never be applied via this path.
- A `goal_status_change` proposal is never dispatched as a session (sweep skip-guard).
- No migration (`ego_proposals.action_type` has no CHECK; `goal_id`/`expected_outputs` already thread through `create_batch`).

## Tests
- 28 new unit tests (stuck classification, handler apply/reject/invalid-value, session routing, four-path wiring guard).
- Full ego suite + dashboard + MCP green; ruff clean.
- E2E against a copy of the production database (real schema): 13/13 — `create_batch` threads `goal_id`/`expected_outputs`, pause & priority apply on approval, reject = no change, terminal value refused.

## Deferred (tracked follow-ups)
- Blocker/replan memory (cross-cycle continuity on what was tried).
- Sub-goal auto-decompose (ego-proposed child goals; reuses the parent/child + completion-cascade model already present).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **Genesis spots goals it's stuck on and asks before easing off them** — when a goal you
+  set has been worked on (several dispatched sessions) but still isn't moving, Genesis now
+  recognizes it as *stuck* rather than merely idle, bumps it up for review, and digs into
+  *why* it stalled instead of nudging it again. If it concludes the goal should be paused or
+  deprioritized, that becomes a proposal you approve or reject — nothing about your goals
+  changes without your say-so.
+
 - **Genesis records traces of what it does** — reflections, ego cycles, every LLM
   call, and the tools its dispatched Claude Code sessions run are now captured as
   nested trace spans (one trace per operation), so its activity can be inspected
--- a/src/genesis/dashboard/routes/comms.py
+++ b/src/genesis/dashboard/routes/comms.py
@@ -197,27 +197,29 @@ async def comms_resolve_proposal(proposal_id: str):
     except Exception:
         logger.warning("Journal resolve failed for proposal %s", proposal_id)
 
-    # Autonomy earn-back: promote on approval / cooldown on reject.
+    # Resolution side-effects (earn-back + goal status change). Fetch the
+    # resolved proposal once and run both hooks against it; each is wrapped so
+    # one failing never aborts the other.
+    prop = None
     try:
-        from genesis.ego.earnback import handle_earnback_resolution
-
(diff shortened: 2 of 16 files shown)


WingedGuardian:
## Structural review

### ERROR — Raw SQL in production code

**`src/genesis/ego/session.py` — `_propose_goal_status_change`**

The anti-spam deduplication check issues raw SQL directly against `ego_proposals`:

```python
    "SELECT 1 FROM ego_proposals WHERE goal_id = ? "
    "AND action_type = 'goal_status_change' "
    "AND status IN ('pending', 'approved') LIMIT 1",
    (goal_id,),
)
if await cursor.fetchone():
```

Rule: *Raw SQL against genesis.db must go through `genesis.db.crud` modules, except in migrations.* This is production business logic, not a migration.

Fix: extract to a CRUD helper in `genesis.db.crud.ego` — something like `has_open_goal_status_change(db, goal_id) -> bool` — and call that instead. The call site in `session.py` should then be a single awaited CRUD call.

---

No credentials, private IPs, personal emails, hardcoded home paths, or private hostnames on any added line. PR title, description, and CHANGELOG addition contain no security-incident or infrastructure-exposure language.

---

PR #722, opened 2026-06-20
fix(observability): capture CC tool spans in dispatched sessions via --settings

WingedGuardian:
## Problem

The tracing/spans backbone (#718) records server-side spans (reflection / ego / `llm.call`), but `otel_spans` had **0 `kind='tool'` rows** despite ~2,400 dispatched CC sessions. The `cc_span_hook` PostToolUse hook never fired in dispatched sessions.

**Root cause:** dispatched sessions run `claude -p` with a working directory outside any git repo (`~/.genesis/background-sessions`). Claude Code discovers project `.claude/settings.json` via **git-root detection**, so it never loaded the repo's hooks there — the hook was registered only in the repo settings. Confirmed three ways: official CC docs, a controlled probe, and a proxy hook (`file_modification_audit_hook`, same launcher) that produced 0 rows across all dispatched sessions.

## Fix

1. **`cc_span_settings_path()`** (`cc/invoker.py`) — lazily + idempotently generates a minimal CC settings file (`~/.genesis/cc-span-settings.json`) registering **only** the span hook, with an **absolute** `genesis-hook` launcher path (cwd-independent; not `${CLAUDE_PROJECT_DIR}`, which CC leaves unset in dispatched sessions). Atomic write, content-compare to avoid churn.
2. **`_build_args`** passes `--settings <that file>` on every dispatch — the supported headless mechanism (CC merges it with user settings). Safe universally: the hook no-ops unless `GENESIS_TRACE_ID` is set.
3. **Removed** the now-redundant repo-level `cc_span_hook` registration from `.claude/settings.json`. Some dispatches (executor) run with a git-**worktree** cwd, which *does* load repo settings — keeping both would fire the hook twice → duplicate spans.
4. **Hardening:** `_run_streaming` now targets the output-format value via the `--output-format` flag instead of a bare `args.index("json")` scan (the new `--settings .../cc-span-settings.json` arg contains the substring "json").

The hook, ingest, and reader are unchanged — they were correct, just never loading.

## Verification

- `ruff check` clean.
- **627 tests pass** (`test_cc/` + `test_observability/` + otel schema), incl. 6 new tests + an isolation conftest.
- **E2E (×2): real dispatched `claude -p` in a non-git cwd → hook fires → JSONL with the injected `trace_id`/`parent_span_id` → ingest → `otel_spans kind='tool'` nested under `cc.session`.** This is the cross-process chain the original A4 E2E never exercised (it invoked the hook directly).

## Notes
- Codex review was **quota-limited**; the Claude code-reviewer pass was clean (one brittleness finding, addressed in #4 above).
- Separate finding filed for review: background-cwd dispatched sessions load none of the repo PreToolUse safety guards (only worktree-cwd dispatches do) — likely intentional (gated via allowed-tools + the autonomy gate), flagged for a decision.
- Activates on the next server deploy (restart regenerates the settings file with the correct install path).

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -348,11 +348,6 @@
                         "type": "command",
                         "command": "${CLAUDE_PROJECT_DIR}/.claude/hooks/genesis-hook hooks/session_observer_hook.py",
                         "timeout": 500
-                    },
-                    {
-                        "type": "command",
-                        "command": "${CLAUDE_PROJECT_DIR}/.claude/hooks/genesis-hook hooks/cc_span_hook.py",
-                        "timeout": 500
                     }
                 ]
             },
--- a/scripts/hooks/cc_span_hook.py
+++ b/scripts/hooks/cc_span_hook.py
@@ -15,6 +15,11 @@ write the DB, but only on Write/Edit; a ``.*`` matcher fires far more often).
 
 Budget: <50ms (JSON parse + locked file append). No LLM, no network, no SQLite.
 Honors ``GENESIS_SPANS_INCOMING_DIR`` (test override).
+
+Registration: NOT via the repo ``.claude/settings.json`` — dispatched sessions
+run with a cwd outside any git repo, so CC never discovers project settings
+there. ``CCInvoker`` injects this hook per-dispatch via ``--settings`` pointing
+at a generated minimal file (``cc_span_settings_path`` in ``genesis.cc.invoker``).
 """
 
 from __future__ import annotations
--- a/src/genesis/cc/invoker.py
+++ b/src/genesis/cc/invoker.py
@@ -74,6 +74,78 @@ def _get_scope_args() -> list[str]:
     return _SCOPE_ARGS
 
 
+# A minimal, runtime-generated CC settings file that registers ONLY the span
+# PostToolUse hook. Lives outside the repo so it is install-local and never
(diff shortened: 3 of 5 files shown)


WingedGuardian:

**Reviewed:** `.claude/settings.json`, `scripts/hooks/cc_span_hook.py`, `src/genesis/cc/invoker.py`, `tests/test_cc/conftest.py`, `tests/test_cc/test_invoker.py`

Spot-checks that passed:
- `except OSError:` — typed, not bare ✓
- `/tmp/cc-span-settings.json` in `test_invoker.py` is a monkeypatch string, not actual temp file creation; production path writes to `~/.genesis/cc-span-settings.json` ✓
- No `${CLAUDE_PROJECT_DIR}` in generated settings file (correctly uses absolute launcher path) ✓
- `args.index("--output-format") + 1` fix is sound — eliminates the substring collision with the new `--settings` arg ✓
- `os.replace` atomic write with pid-suffixed temp name — correct pattern ✓
- 6 new tests cover `cc_span_settings_path` (generate, None-when-missing, idempotent, stale-rewrite) and both `_build_args` branches ✓
- `conftest.py` autouse fixture properly isolates the module-level `_CC_SPAN_SETTINGS_PATH` across all test_cc tests ✓
- Conventional commit prefix present (`fix:`) ✓
- PR wording: no PII/credential/security-incident language ✓

---

PR #723, opened 2026-06-20
fix(awareness): emit reflection idle heartbeat on quiet (depth=None) ticks

WingedGuardian:
## What

Two fixes that stop the daily morning/health report from false-alarming about reflection.

### Fix A — reflection no longer falsely reported "dark" during quiet periods
The reflection heartbeat (read by `subsystem_heartbeats`, overdue threshold 4h) was only refreshed when a reflection actually ran. Quiet ticks classify to `depth=None` (~93% of ticks — nothing triggered, or a ceiling/floor throttle) and dispatch no reflection, so during calm periods (especially overnight) the heartbeat aged past 4h and the report warned reflection had gone "dark" while the loop was perfectly healthy. Verified against live data: real 8–13h overnight heartbeat gaps.

Emit a lightweight reflection heartbeat on the `depth=None` dispatch branch of `_on_tick`, via a new `_emit_reflection_idle_heartbeat` helper. Gated on `result.db_available` so a DEGRADED tick (DB unavailable → classification skipped) does **not** pulse — a genuine reflection outage still ages out and alarms. Extends the existing Micro-silent idle pulse (#695), which only covered `depth=MICRO` ticks and never reached the dominant `depth=None` path.

### Fix B — correct a stale call-site comment
`outreach_fallback` was documented as "consumer NOT BUILT … CRITICAL BUG". That is false: `OutreachRecoveryWorker` (`resilience/outreach_recovery.py`) polls the deferred-work queue, retries with exponential backoff, and is wired in `runtime/init/outreach.py`. Marked wired.

## Tests
- `_emit_reflection_idle_heartbeat` gate, both directions (healthy quiet tick pulses; degraded tick does not).
- Wiring test through the real `_on_tick` (no signals → `depth=None` → exactly one REFLECTION heartbeat) — guards against re-introducing a dead-code placement inside `_dispatch_reflection`.
- Full `tests/test_awareness/` package green (126 tests); ruff clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,6 +120,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   signal — whether its ranking merely favors the memories it retrieves most often — so
   entrenchment of stale-but-popular memories can be watched over time (measured, not acted on).
 
+- **Reflection is no longer falsely reported as "dark" during quiet periods** — the health and
+  morning reports track each subsystem by a periodic heartbeat, but the reflection loop only emitted
+  one when it actually ran a reflection. During legitimately calm stretches (most overnight ticks do
+  no reflection by design), that heartbeat could age past its overdue threshold and the report would
+  warn that reflection had gone silent or "dark" even though the loop was perfectly healthy. Reflection
+  now emits a lightweight idle heartbeat on those quiet ticks, so it only reads as overdue during a
+  genuine outage.
+
 - **Knowledge-base searches stopped silently returning nothing** — the relevance floor that
   trims low-quality knowledge results was a fixed absolute cutoff that, on the score scale recall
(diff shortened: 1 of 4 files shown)


WingedGuardian (commented):

---

WingedGuardian:
Review note: Codex automated review is quota-exhausted on this PR, so per the merge gate this relies on CI plus a local code-reviewer pass. That pass caught a real bug in the first cut — the idle pulse was placed inside `_dispatch_reflection`, which only runs when `classified_depth is not None`, so it was dead code for the `depth=None` quiet ticks it targets. Fixed by moving the pulse to the `_on_tick` dispatch branch via `_emit_reflection_idle_heartbeat`, with a wiring test (`test_on_tick_pulses_reflection_heartbeat_when_idle`) that drives the real `_on_tick` path and would fail if the pulse were mis-placed again. Re-review: resolved/ship.

PR #726, opened 2026-06-21
feat(dashboard): trace-waterfall "Traces" tab over otel_spans

WingedGuardian:
## What

Adds a read-only **Traces** tab to the dashboard that renders each recorded operation as a nested **waterfall**: a reflection / ego cycle / dispatched session → its LLM calls → sub-sessions → the tools that session ran, laid out as bars on a shared timeline with click-through span detail. It consumes the `span_reader` read APIs from the spans backbone (#718) — closing the "see what the agents do" loop in the UI.

## Changes

- **`routes/spans.py`** (new): `GET /api/genesis/spans/recent` (root spans newest-first) and `GET /api/genesis/spans/trace/<id>` (tree flattened server-side via the tested `flatten_tree`). Read-only; auth-open like the other `/api` routes; `limit` floored at 1 / capped at 200. Registered in `routes/__init__.py`.
- **`genesis_dashboard.html`**: Alpine store state + fetches + `computeWaterfall` (offset/width as % of the trace timeline; point-span ticks; clock-skew clamp; min-width floor with shift-left-to-fit), the list → inline waterfall → span-detail panel, a nav button, and `.wf-*` CSS. Span detail surfaces provider/model/call_site/tokens/cost for LLM spans plus pretty-printed attributes.
- **`test_spans_api.py`** (new): 7 route tests, including the real `flatten_tree` contract.
- CHANGELOG entry (user-visible).

## Verification

- ruff clean; 156 `test_dashboard/` tests pass (7 new, 0 regression); `node --check` clean on the full inline module; added HTML tag-balanced.
- **E2E against the live DB (read-only) via a faithful local harness + a real browser**: list renders with correct durations/relative-times/status and a 30s live poll; the waterfall renders a real 32-span `cc.session` trace as 1 bar + 31 point-span ticks at the expected offsets; span detail shows the LLM block + attributes. Backend endpoints verified over HTTP (recent / trace / 404 / limit guard).

## Review

Claude code-reviewer (Codex quota-exhausted): 6 findings, 5 valid and fixed (waterfall min-width/overflow ordering, polling-interval cleanup, negative-limit guard, zero-duration tick detection, stale-fetch request token); 1 was a non-issue.

## Notes

- Read-only, additive: no migration, no backend logic change, no writes.
- Activates after merge + a server restart (the running server serves main-tree code via the editable install). Activation is config-free.


WingedGuardian:

---

PR #728, opened 2026-06-21
fix(reflection): supersede + soften stale quality/regression alarms

WingedGuardian:
## What

Final PR of the morning-report false-alarm batch: stop stale `quality_drift` / `learning_regression` alarms from persisting and from surfacing as fresh. (PR-1 fixed the reflection "dark" heartbeat; PR-2 fixed the age-diluted quality metric; this closes the loop on the *alarm lifecycle*.)

These observations were written weekly but never resolved, so the morning report kept repeating a days-old flag as current, and the regression alarm fired on any 2 consecutive declines in a noisy LLM-assigned self-score.

### Changes
- **Supersession (resolve-before-write).** `route_calibration` resolves any prior unresolved `quality_drift` before writing this cycle's result; `emit_regression_signal` does the same for `learning_regression`. Ordering matters: `resolve_by_source_and_type` resolves *all* unresolved matches, so resolving after the write would bury the new row. Resolving regardless of the new result also clears a standing alarm when the metric recovers.
- **Resolve-on-recovery.** New `LearningStabilityMonitor.resolve_regression_if_standing`, called from the calibration scheduler in the `else` of the regression check — clears a standing regression once effectiveness recovers, instead of letting it linger until its TTL.
- **Softened `check_regression`.** Still requires consecutive decline, but now also a meaningful total drop (≥ 0.10) **and** a latest score below a healthy floor (< 0.60), so small wobbles and high-but-dipping trajectories don't latch a persistent CRITICAL.
- **Morning-report staleness grounding.** Observations older than 3 days are shown with their displayed severity demoted one level and tagged `[aged]` (stored priority unchanged); `MORNING_REPORT.md` gains observation age-tier guidance so the LLM treats aged signals as historical.

## Tests
- `test_stability.py`: small-dip / high-baseline → no regression; supersession; resolve-on-recovery (+ no-op).
- `test_output_router.py`: clean cycle clears a standing drift; fresh drift supersedes prior.
- `test_morning_report.py`: a 6-day-old critical is demoted to high and tagged `[aged]`; a 2-hour-old one is untouched.
- 307 tests green across `tests/test_reflection/` + `tests/test_outreach/`; ruff clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,6 +119,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Stale "quality drift" and "learning regression" alarms no longer linger or cry wolf** — when
+  Genesis's weekly self-checks flagged a quality drift or a learning regression, that flag stayed
+  active until it expired days later, so the morning report kept repeating a days-old warning as if
+  it were current. Now each weekly check supersedes the previous flag and clears it outright once the
+  underlying metric recovers, the regression alarm only fires on a meaningful sustained drop (not a
+  small wobble in a noisy self-score), and any observation older than three days is shown in the
+  report demoted and tagged as historical rather than as a fresh alarm.
+
 - **Genesis stops false-alarming about its own reflection quality declining** — its weekly
   self-assessment scored "reflection quality" from how often its recent reflections had been
   retrieved, but counted a running total over the 50 newest reflections. Because brand-new
--- a/src/genesis/identity/MORNING_REPORT.md
+++ b/src/genesis/identity/MORNING_REPORT.md
@@ -76,6 +76,17 @@ note it at most as "earlier <condition>, now recovered", or omit it. Do NOT
 repeat the stale alarm as if it were current. When in doubt, trust the live
 System Health numbers over an older observation's text.
 
+**Observation age tiers.** Every observation in "What I Noticed" is tagged with
+its age; treat age as a strong prior on relevance:
+- Under 24h: current — surface normally.
+- 1-3 days old: surface only if still actionable; lead with the age.
+- Over 3 days old (shown demoted and tagged "[aged]"): a historical signal, not
+  a current condition. Do NOT present it as a fresh alarm — mention it at most
+  as "earlier <condition>", or omit it if a newer signal supersedes it.
+This applies to ALL observation types, including cognitive-quality and learning
+signals: a days-old "quality drift" or "learning regression" note is not, by
+itself, evidence of a current problem.
+
 ## What to Include
 
 **Primary focus (this is what the report is FOR):**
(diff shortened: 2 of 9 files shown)


WingedGuardian:
Review note: Codex automated review is quota-exhausted; relies on CI plus a local code-reviewer pass (verdict: ship). The pass verified the resolve-before-write ordering in both route_calibration and emit_regression_signal (resolve_by_source_and_type only touches resolved=0 rows, so it can't bury the just-written row), the newest-first magnitude/baseline gate arithmetic in check_regression (existing TestCheckRegression cases still pass), the resolve-on-recovery scheduler wiring + safe no-op, the display-only morning-report demotion, and that no consumer depends on these observations staying unresolved (quality_drift already L1-excluded; learning_regression is intentionally user-facing and a fresh superseding row is exactly what's wanted).

WingedGuardian (commented):
**First-pass review — structural + PII/wording scan**

---

### Structural review

#### ⚠️ WARNING — Import hygiene

**`tests/test_outreach/test_morning_report.py`**

`from datetime import UTC, datetime, timedelta` is placed inside the body of `test_observation_insights_demotes_aged` rather than at module top. The symbols are already used in other tests in this file (or are stdlib), so a top-level import would be the right fix.

```python
# Move to top of file with other stdlib imports
from datetime import UTC, datetime, timedelta
```

---

#### ⚠️ WARNING — Missing scheduler-level integration test

**`src/genesis/reflection/scheduler.py`** (`_run_calibration`, the new `else` branch)

The new orchestration contract is: *regression detected → `emit_regression_signal`; no regression → `resolve_regression_if_standing`*. The callee (`resolve_regression_if_standing`) is well-tested in `test_stability.py`, but the scheduler's own branching logic — that the right arm is called in the right case — has no coverage at the scheduler level. If a `test_scheduler.py` exists, it may be worth adding a case that patches `check_regression` to return `False` and asserts `resolve_regression_if_standing` is invoked.

---

#### 📝 NOTE — Inverted constant name

**`src/genesis/reflection/stability.py:29`**

```python
_MAX_HEALTHY_SCORE = 0.60
```

The usage is `scores[0] < _MAX_HEALTHY_SCORE` — i.e., the alarm fires when the latest score is *below* this value. That makes 0.60 the *minimum* healthy score, not the maximum. The name reads backwards: a reader expects `_MAX_HEALTHY_SCORE` to be an upper bound on health, not a lower floor. Suggest `_HEALTHY_SCORE_FLOOR` or `_MIN_HEALTHY_SCORE`.

---

**Code diff (added lines):** CLEAN — no credentials, secrets, private IPs, hardcoded paths with usernames, or hardware identifiers.

**PR title, description, CHANGELOG additions:** CLEAN — no language referencing data exposure, credential cleanup, or infrastructure details that would be visible in git history.

---

Structural review: 2 WARNINGs, 1 NOTE — no ERRORs.

---

PR #725, opened 2026-06-20
fix(reflection): age-fair reflection-quality metric (no false decline from pool aging)

WingedGuardian:
## What

Stop the weekly self-assessment from inventing a reflection-quality decline that the morning report then alarms on.

### The bug
`reflection_quality` was scored from how many of the **50 most-recent** `cc_reflection_deep` observations had `retrieved_count > 0`. `retrieved_count` is a cumulative lifetime counter and the pool is recency-windowed, so the rate mechanically fell as new (not-yet-retrieved) reflections entered the window — a **measurement artifact**, not real degradation. It produced a false "0.70 → 0.35 quality crater" that fed baseless warnings into the morning report. (Verified: the reflection pipeline is healthy; the newest reflections are the *most*-retrieved cohort. Full-corpus rate ≠ the windowed rate; retrieval rate is monotonic in observation age.)

### The fix
- **Age-fair maturity cohort:** count over `cc_reflection_deep` observations created **3–10 days ago** — old enough to have had a fair chance to be retrieved, recent enough to reflect current quality. Removes the age-composition artifact.
- **Influence ⊆ retrieved:** count a reflection as influential only if it was also retrieved. An observation can be flagged influenced via a memory-op target without being retrieved (nothing enforced the invariant), which produced impossible `influenced > retrieved` ratios that dragged the score down.
- **Honest abstention:** report `data_available: false` when the cohort is below 5 rows. Deep-reflection volume is low (~one cohort's worth/week), so the dimension honestly abstains rather than scoring on noise. `SELF_ASSESSMENT.md` dimension 1 is updated to match and to forbid inferring a decline from a small/empty cohort.

The whole gather dict is `json.dumps`-ed into the assessment prompt, so the new `data_available`/`cohort_size` fields surface to the assessor directly.

## Tests
- New `TestReflectionQualityCohort`: locks the 3–10d window (1d and 20d excluded), the `<5` abstention, the influence-among-retrieved rule (4-combination matrix), and the dict contract.
- Full `tests/test_reflection/` + `test_cc/test_reflection_bridge.py` green (207); ruff clean.

## Notes
- No consumer reads the dropped `total_observations` input key (`OutputRouter` parses the LLM's *output*; `check_regression` reads a different dimension). Confirmed in review.
- Stale-alarm lifecycle (superseding prior `quality_drift`/`learning_regression`, softening the regression alarm, morning-report staleness grounding) follows in a separate PR.

PR #730, opened 2026-06-21
fix(reflection): fresh calibration row per cycle + honest str return

WingedGuardian:
## What

A small, focused fix to the weekly quality-calibration routing (`reflection/output_router.py`), follow-up to the #728 alarm-lifecycle work.

**Two coupled issues, both pre-existing:**

1. **Contract bug.** `route_calibration` and `route_assessment` are typed `-> str` with docstrings promising "observation ID or empty string on failure," but `_write_observation` returns `None` on a content-hash dedup hit — so both could leak `None`.
2. **The dedup quirk.** On the non-drift path, an unchanged week's byte-identical JSON deduped to no new row, so `scheduler._already_ran_this_week('quality_calibration')` (which looks for a *this-week* row) could miss the run and re-trigger the weekly calibration on a restart-catchup.

## The fix

- `route_calibration` / `route_assessment`: `return obs_id or ""` — honor the `-> str` contract (the docstrings were already promising this; they were just untrue).
- `route_calibration`: extend #728's supersede-before-write to **both** `quality_drift` *and* `quality_calibration`. The prior identical row is resolved first, so the new write is never deduped against an unresolved row → a fresh this-week row is always written and the scheduler guard sees the run.
- **Not** applied to `route_assessment`: `self_assessment` is a historical ledger the #725 maturity-cohort metric reads, so resolving old rows would corrupt the trend record.

## Why it's safe

- `_write_observation`'s shared semantics are unchanged; the only two non-test callers (in the reflection bridge) discard the return value.
- `quality_calibration` carries a 14-day TTL (state, not a ledger) and is excluded from L1 essential-knowledge; the only recency-reader is the scheduler guard this repairs. The new row is written *after* the resolve, so it still surfaces normally.

## Verification

- 45 targeted tests pass (3 new: assessment dedup→`""`; unchanged calibration writes a fresh row; calibration contract guard). Existing #728 supersession tests unaffected.
- **E2E against the real scheduler guard:** seeded a prior-week unresolved row + ran an identical clean cycle → prior row superseded, fresh this-week row written, `_already_ran_this_week('quality_calibration')` returns `True` (no false re-trigger).

## Deferred (follow-up)

A deeper idempotency-guard hardening — the drift-week type mismatch (guard checks `quality_calibration` but a drift week writes `quality_drift`) and switching the guard to the authoritative `job_health.last_run` — is tracked separately to keep this PR surgical.

No CHANGELOG entry: internal robustness/contract fix with no observable behavior change in normal operation.

### DIFF
--- a/src/genesis/reflection/output_router.py
+++ b/src/genesis/reflection/output_router.py
@@ -566,7 +566,12 @@ class OutputRouter:
             "self-assessment", content, output.overall_score,
         )
 
-        return obs_id
+        # `_write_observation` returns None when this week's content hashes
+        # identically to an existing unresolved row (content-hash dedup). Honor
+        # the `-> str` contract / "empty string on failure" docstring rather than
+        # leaking None to callers. (No supersession here: self_assessment is a
+        # historical ledger the maturity-cohort metric reads — see #725.)
+        return obs_id or ""
 
     async def route_calibration(
         self, output: QualityCalibrationOutput, db: aiosqlite.Connection,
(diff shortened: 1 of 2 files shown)


WingedGuardian (commented):

---

WingedGuardian:
**Review note:** Codex automated review hit its usage limit on this PR (quota-exhausted, not a pass/fail signal). Per the pre-merge gate, merging on green CI alone. This PR was additionally reviewed by an inline code-reviewer agent (supersession ordering, #728 non-regression, `or ""` caller impact, test quality) and verified E2E against the real `scheduler._already_ran_this_week` guard. All CI checks green.

PR #732, opened 2026-06-21
fix(ego): emit ego heartbeat on a dedicated 5-min liveness pulse

WingedGuardian:
## Problem

The `ego` subsystem reads **`overdue`** in `subsystem_heartbeats` even when the ego is healthy and actively cycling (cycles every 15–60 min). It's a false alarm from a monitoring-wiring gap, not an ego outage.

## Root cause

The ego's health heartbeat was emitted **only** from `_emit_heartbeat("tick")` at the top of `EgoCadenceManager._on_tick` — i.e. bolted to the proactive interval tick, which is unreliable *by design*:

- **Stretches:** `_update_interval` backs the interval off geometrically up to `_recency_max_interval()` (`_RECENCY_TIERS`, up to 4320 min / 72h) during quiet periods.
- **Perpetually deferred:** every completed cycle calls `reschedule_job("ego_cycle", now+interval)`, so frequent reactive cycles keep resetting the proactive tick's next-run before it fires.
- **Resets on restart:** the `IntervalTrigger` restarts from zero on each server restart.

So a healthy ego routinely goes >4h between `_on_tick` firings — past the ego overdue threshold (14400s) — while cycles keep running via the signal-consumer loop (reactive/escalation/morning/sweep). The patient is fine; the wire to the monitor is loose.

## Fix

Add a dedicated `ego_heartbeat` APScheduler job (`IntervalTrigger(minutes=5)`, **never rescheduled**) that emits the liveness pulse independent of work cadence — mirroring the fixed-interval heartbeats of awareness/surplus/dashboard, and the recent reflection-idle-heartbeat normalization (#723). Registered in `start()`, so both the user-ego and genesis-ego cadence managers get one (each in its own scheduler). An initial pulse is emitted at `start()` so a fresh restart isn't briefly stale. The existing `_on_tick` heartbeat is kept (harmless extra context).

No change to ego cognition, cadence, or gating — purely an observability/liveness wiring fix.

Also: `mcp/health/manifest.py` ego heartbeat metadata/comment corrected to the real 5-min cadence (the value is display-only — bound to the unused `_interval_s`; only `overdue_s` drives the check).

## Verification

- `ruff` clean; 106 targeted tests pass (`tests/ego/test_cadence.py`, `tests/ego/test_init.py`, `tests/test_mcp/test_health_mcp.py`), including new heartbeat tests (registration, handler→emit, pause-still-emits, no-bus no-op, exact `Subsystem.EGO`/`heartbeat` contract, cold-start initial pulse).
- **Real-component E2E** (no prod server touched): real `EgoCadenceManager.start()` + `_on_heartbeat()` → real `GenesisEventBus` persistence → real `events` table → real `_impl_subsystem_heartbeats()` returns ego **`alive`** (age 0.5s).
- Reviewed by code-reviewer + genesis-architect (both: sound); all accepted findings applied.

## Notes

- Codex review unavailable (quota-exhausted); Claude code-reviewer + genesis-architect used.
- Follow-up (separate, not this PR): the proactive `_on_tick` itself is starved by the reschedule-deferral + restart trap, so the ego rarely runs its idle proactive cycle / deep-think upgrade — a behavioral question to address on its own.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,6 +119,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **The ego subsystem no longer shows a false "overdue" health status** — the health view marks
+  a background subsystem "overdue" if it hasn't checked in within a window, and the ego's check-in
+  was tied to its proactive thinking timer. That timer deliberately slows down during quiet
+  periods and gets pushed back every time the ego does other work, so it could legitimately go
+  hours between ticks — long enough to trip the 4-hour "overdue" alarm even while the ego was
+  healthy and actively running cycles. The ego now sends a steady "alive" check-in every few
+  minutes, independent of its work pace, so its health status reflects reality.
+
 - **Stale "quality drift" and "learning regression" alarms no longer linger or cry wolf** — when
   Genesis's weekly self-checks flagged a quality drift or a learning regression, that flag stayed
   active until it expired days later, so the morning report kept repeating a days-old warning as if
--- a/src/genesis/ego/cadence.py
+++ b/src/genesis/ego/cadence.py
@@ -163,6 +163,21 @@ class EgoCadenceManager:
             max_instances=1,
             misfire_grace_time=300,
         )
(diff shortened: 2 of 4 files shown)


WingedGuardian (commented):

---

PR #735, opened 2026-06-21
feat(observability): surface ambient edge health in the infrastructure snapshot

WingedGuardian:
## What
Stage D of the ambient-capture roadmap: make the ambient edge bridge's health **queryable** via the infrastructure snapshot / `health_status` MCP — instead of only a transient Telegram alert (#680/#685 added the alert, but nothing surfaced the verdict anywhere you could query it).

## How
- New `probe_ambient_health()` in `observability/health.py` — reuses the existing `ambient_health` module (config load → SSH read → `evaluate_ambient_health`) and maps the verdict to a `ProbeResult`:
  - `ok→HEALTHY`, `degraded→DEGRADED`, `down→DOWN`, `unknown→DEGRADED` (edge unreachable is neither healthy nor a confirmed-dead bridge).
  - Returns `None` when no ambient edge is configured → the snapshot omits ambient entirely (an absent edge is not a fault, unlike guardian).
  - Mirrors `probe_guardian`'s SSH-with-TTL-cache (60s) so repeated snapshot reads don't re-SSH.
- Wired into `infrastructure()` after the guardian block → flows into `health_status` automatically.

## Scope
Observability only. Deliberately does NOT touch `_KNOWN_SUBSYSTEMS` or the dream cycle (those belong to the deferred Stage C memory-graduation work).

## Tests / verification
- 9 new unit + integration tests (verdict mapping, unconfigured→None, unreachable→DEGRADED, TTL cache, snapshot attach/omit/error-guard); full `test_observability/test_health.py` + `test_ambient_health.py` green (38 tests).
- Verified **live** against the running ambient edge: the probe SSH-reads the bridge and `infra["ambient"]` surfaces a real healthy verdict.

### DIFF
--- a/src/genesis/observability/health.py
+++ b/src/genesis/observability/health.py
@@ -33,6 +33,11 @@ _guardian_remote_config_checked: bool = False
 _guardian_ssh_cache: dict[str, tuple[float, ProbeResult]] = {}
 _GUARDIAN_SSH_TTL = 60.0
 
+# Ambient-edge SSH probe cache: {host_ip: (monotonic_timestamp, ProbeResult)}.
+# Mirrors the guardian TTL so repeated snapshot reads don't re-SSH the edge.
+_ambient_ssh_cache: dict[str, tuple[float, ProbeResult]] = {}
+_AMBIENT_SSH_TTL = 60.0
+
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

PR #736, opened 2026-06-21
fix(inbox): defer cooldown modifications + dedup re-pasted URLs + tz hardening

WingedGuardian:
## Context

Came out of an inbox investigation ("why didn't my pasted link get evaluated?"). The original symptom self-resolved (the item evaluated on the next interval — no permanent-drop bug there), but the dig surfaced three genuine defects. Scope was narrowed by due diligence: the duplication-*prevention* layer first considered was dropped because the monotonic baseline guard (#473) already dedups seen URLs.

## Fixes

1. **Cooldown-consume** — when a file was modified within `evaluation_cooldown_seconds` of its last eval, the cooldown branch wrote a `completed` `inbox_items` row carrying the *new* content hash with no evaluation. `get_all_known` then returned that hash, so the change was never re-detected — new content added shortly after an eval was stranded until the next edit (and lost if it was the last edit). Now the cooldown path **defers without consuming**: it leaves the known hash unchanged so the next check past the cooldown re-detects and evaluates it.

2. **URL re-paste dedup** — new `normalize_url_line()` strips volatile tracking/share query params (`utm_*`, `mc_*`, `rcm`, `fbclid`, `gclid`, `igshid`, …; URL **paths** left intact). Applied in `_compute_new_content`'s line comparison, so the same article re-pasted with different tracking params (e.g. a post shared from phone vs. desktop) yields an empty delta — no re-evaluation, no duplicate follow-up. The original line is preserved for evaluation. Complements #473's byte-identical dedup.

3. **Timezone hardening** — new `parse_utc_iso()` (treats a naive stored timestamp as UTC, returns `None` on bad input). The two `except (ValueError, TypeError): pass` swallows in the monitor (approval-staleness + cooldown guards) — which silently *disabled* those guards on a tz error — now log a warning instead.

## Testing

- TDD: each behavioral fix has a failing-test-first regression (`test_cooldown_defers_without_dropping_modification`, the `_compute_new_content` dedup tests).
- True E2E through `check_once`: `test_e2e_url_repaste_different_tracking_not_reevaluated`.
- Full `tests/test_inbox/` + `tests/test_util/` green locally (264 passed); `ruff check` clean on the diff.

## Review

Adversarial code-review pass: no P1/P2 issues; cooldown timing, re-detection churn-safety, tz subtraction, and URL false-dedup all explicitly cleared. One nit applied (P3-2: exclude `<>` from the URL match for markdown `<url>` form).

## Notes

- No DB schema change.
- Not auto-deployed (the server was restart-thrashing during the session); deploy is a separate gated step.
- Cleanup of pre-existing duplicate inbox follow-ups was handled separately as a data operation, not in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,6 +119,18 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Inbox items added soon after an evaluation are no longer silently skipped** — if you added a
+  link or note to an inbox file within the cool-down window just after Genesis had evaluated that
+  file, the new item could be marked as seen without ever being evaluated, and it stayed stranded
+  until you edited the file again. Genesis now defers those additions and picks them up on the
+  next pass once the cool-down clears, so nothing you add gets lost.
+
+- **Re-sharing an article you already added no longer creates a duplicate evaluation** — links
+  often carry per-share tracking parameters (for example, the same LinkedIn post shared from your
+  phone vs. your desktop produces different URLs), which used to make a re-paste look brand new.
+  Genesis now ignores those tracking parameters when deciding what's new, so the same article
+  isn't evaluated twice or spawn a duplicate follow-up.
+
 - **The ego subsystem no longer shows a false "overdue" health status** — the health view marks
   a background subsystem "overdue" if it hasn't checked in within a window, and the ego's check-in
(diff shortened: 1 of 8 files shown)


WingedGuardian (commented):

---

WingedGuardian:
Merging on CI + structural review (CLEAN) + inline adversarial code-review (no P1/P2). Note: the Codex review bot was quota-limited and did not run, so this merges per the quota-exhausted path of the pre-merge gate.

PR #739, opened 2026-06-21
chore(transcripts): local retention 180d + backups keep forever

WingedGuardian:
Two halves of one transcript-retention policy, since transcripts are Genesis's long-term conversational memory (the last-resort grep path + `cc_sessions` read the `.jsonl` files):

**1. Local — 180 days.** Sets `cleanupPeriodDays: 180` in the project `.claude/settings.json` (up from Claude Code's 30-day default). Project-level settings are honored by CC and ship with the repo, so it's an install-time default. Not retroactive — future cleanup passes only.

**2. Backups — keep forever (by default).** The encrypted 6h backup is the permanent archive: `transcripts/*.gpg` already accumulate locally and in every off-site snapshot, and are never auto-pruned. Adds an explicit policy note in `backup.sh §3` so the planned DR retention/prune work (GFS snapshot pruning + local-staging prune) must EXEMPT `transcripts/`. A user can opt into expiry, but the system never auto-expires backed-up transcripts the way the local store does.

Net effect: local disk stays modest (a half-year window for fast grep), while the full history is preserved durably in encrypted backups (on-box + off-site).

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -3,7 +3,7 @@
         "DISABLE_AUTOUPDATER": "1",
         "DISABLE_UPDATES": "1"
     },
-    "cleanupPeriodDays": 365,
+    "cleanupPeriodDays": 180,
     "hooks": {
         "SessionStart": [
             {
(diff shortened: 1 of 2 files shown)


WingedGuardian:
Config-only change — structural review skipped.

PII/Secrets/Wording scan: CLEAN — no secrets, private IPs, credentials, usernames in paths, or sensitive wording in the diff, PR title, or description.

---

PR #741, opened 2026-06-22
feat(follow-ups): kind/domain/goal_id schema foundation (PR1)

WingedGuardian:
## PR1 of the follow-up system redesign — schema foundation

The `follow_ups` table had become a graveyard because one overloaded `strategy`
enum fused *who executes* and *when* into a single field, leaving the long tail
with no path to action or closure. The redesign decomposes that into orthogonal
axes; this PR lands the **schema foundation** for two of them plus a goal link.
It is strictly **additive** — `strategy` and the surplus dispatcher are untouched.

### New columns (migration `0034`)
- **`kind`** — `follow_up` | `tabled`. Tabled = tracked but never dispatched or
  surfaced as action (the honest someday/maybe lane, reached by judgment).
- **`domain`** — `internal` | `user_world` (nullable). Genesis-COO vs user-CEO
  jurisdiction. NULL = not yet classified.
- **`goal_id`** — link target for a future follow-up → goal promotion.

`kind` backfills to `follow_up` via the column default. `domain` backfill is
**NULL-on-uncertainty**: a correct-by-construction source (model-benchmarking) →
`internal`; inbox items reuse the evaluator's own per-item genesis/user judgment;
an internal-keyword hit → `internal`; otherwise left NULL (never a `user_world`
guess — a wrong domain would hide a row from the ego that needs it, whereas NULL
stays visible to both and is reclassifiable later).

### Other changes
- Extract a shared `domain_classifier` (internal-only detector; returns
  `internal` or `None`, never guesses `user_world`).
- `crud.create()` gains `kind`/`domain`/`goal_id`; the action/pending readers
  gain `include_tabled=False`, so a tabled item can never be dispatched or
  surfaced once tabling exists.
- Data-layer functions for the upcoming management UI (delete / set_kind /
  set_domain / paginated filtered query / counts / distinct sources).
- The follow-up MCP tool and the four producers now set `domain`.

### Not in scope (follow-on PRs)
- The dashboard management cockpit (PR2).
- Scoping existing ego/report/signal consumers by `domain` (PR3 — behavior change).

### Verification
- Migration applied to a copy of the live DB: all three columns present,
  `kind` = `follow_up` across the board, domain split internal/user_world/NULL =
  184 / 4 / 180, and re-running is a no-op (idempotent).
- `ruff` clean; targeted suites green (schema CHECK constraints, CRUD round-trips
  + tabled exclusion + filtered query, dispatcher, ego context, inbox digest).
- Fixed four test fixtures that hand-rolled a partial `follow_ups` schema.

### DIFF
--- a/src/genesis/autonomy/executor/engine.py
+++ b/src/genesis/autonomy/executor/engine.py
@@ -896,6 +896,7 @@ class CCSessionExecutor:
                 source="task_executor",
                 strategy="ego_judgment",
                 reason=content,
+                domain="internal",
             )
         except Exception:
             logger.exception("Failed to create challenge follow-up")
--- a/src/genesis/autonomy/executor/trace.py
+++ b/src/genesis/autonomy/executor/trace.py
@@ -643,6 +643,7 @@ class ExecutionTracer:
                 strategy="ego_judgment",
                 reason="Retrospective learning extraction failed — needs investigation",
                 priority="medium",
+                domain="internal",
             )
         except Exception:
             # Last resort — at least the log warning above was emitted
--- a/src/genesis/db/crud/follow_ups.py
+++ b/src/genesis/db/crud/follow_ups.py
@@ -36,18 +36,27 @@ async def create(
     scheduled_at: str | None = None,
     priority: str = "medium",
     pinned: bool = False,
+    kind: str = "follow_up",
+    domain: str | None = None,
+    goal_id: str | None = None,
     id: str | None = None,
 ) -> str:
-    """Create a follow-up and return its ID."""
+    """Create a follow-up and return its ID.
+
+    kind:    'follow_up' (intended for action) or 'tabled' (tracked, not for action).
+    domain:  'internal' | 'user_world' | None (None = not yet classified).
+    goal_id: optional link to a unified goal (user_goals.id) for future promotion.
+    """
     fid = id or _new_id()
     await db.execute(
         """INSERT INTO follow_ups
            (id, source, source_session, content, reason, strategy,
-            scheduled_at, status, priority, pinned, created_at)
-           VALUES (?, ?, ?, ?, ?, ?, ?, 'pending', ?, ?, ?)""",
+            scheduled_at, status, priority, pinned, kind, domain, goal_id,
+            created_at)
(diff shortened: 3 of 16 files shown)


WingedGuardian (commented):
## First-pass review — PR #741

### Structural review

**WARNING — Missing test file for new module**

`src/genesis/ego/domain_classifier.py` is a new public module in `src/genesis/ego/` with two exported functions (`is_genesis_internal`, `classify_domain`), but no corresponding `tests/ego/test_domain_classifier.py` appears in this PR.

The functions are exercised indirectly — `is_genesis_internal` via the existing `test_user_context.py` fixtures (which import through `user_context.py`), and `classify_domain` via the MCP tool and follow-up CRUD. But there's no unit coverage for the classifier's own keyword-match logic (e.g. confirming that a known internal keyword returns `'internal'`, that an empty string returns `None`, that `classify_domain` never returns `'user_world'`). Given this module is now the single source of truth for domain classification across the system, a targeted unit test file would be worth adding — either here or as a follow-on.

Everything else is structurally clean:
- No `nohup`, no direct `task_states` inserts, no `/tmp/` usage, no `git add .`, no force-push
- No bare `except:` clauses (all caught exceptions are typed)
- No mutable defaults
- SQL injection surface is clean: `_build_filter_where` uses `?` params throughout; `ORDER BY` is gated by the `_SORT_MAP` allowlist in `query_page`; the f-string `kind_clause` in the reader functions is a static literal, never user input
- Import hygiene: no star imports; `domain_classifier.py` has zero genesis-internal imports so no circular risk
- Migration is self-contained (no `genesis.ego` import; keyword snapshot is intentional and documented)
- Conventional commit prefix: ✓

---

**CLEAN** — no secrets, credentials, private IPs, or private hostnames in added lines. PR title and description contain no security-incident language or infrastructure-exposure wording.</body>
</invoke>

---

WingedGuardian:
Merging on green CI (verified for the exact head SHA `74735154`: test/lint/migration-check/leak-detector/CodeQL all pass) plus a clean inline code-review pass. The Codex automated review was quota-exhausted, so per the pre-merge gate this merges on CI alone.

PR #742, opened 2026-06-22
feat(security,learning): skill-security scan + skill-evolution provenance (inbox Phase A)

WingedGuardian:
### 1. Skill-security scan (`src/genesis/security/skill_scan.py`)
Wrap NVIDIA SkillSpector to audit installed skills and file findings as recon
observations (`source=recon`, `type=finding`, `category=skill-security`),
surfaced via `recon_findings(job_type="skill-security")` + the dashboard.

- **Trusted-source allowlist (Option C):** a raw scan flags ~68 CRITICAL — nearly
  every curated, capable skill (gstack suite, Genesis's own) trips SkillSpector's
  excessive-agency / dangerous-code patterns by design. So we scan everything for
  visibility but **file findings only for untrusted/unknown skills.** First-party
  repo roots auto-trusted; `--seed-trusted` blesses the current install;
  `--no-trust` restores the file-everything mode.
- SkillSpector exits **non-zero to signal risk, not failure** — the wrapper trusts
  the JSON whenever it parses (else it would silently drop the highest-risk skills).
- Per-skill timeout + graceful skip for outliers (a skill can bundle thousands of
  files); issue `location` normalized (str or dict); `skip_if_duplicate` so
  re-scans don't pile up.
- **SkillSpector is an external dependency** (not vendored): deploy needs it
  installed + a one-time `--seed-trusted` run.

### 2. Skill-evolution provenance (`src/genesis/learning/skills/*`)
Record *why* a skill mutation happened: an optional `provenance_trace` on
`SkillProposal`, derived from the `SkillReport` (trend / usage / success rate /
failure patterns), is written (with `failure_patterns_addressed`) into the
`cognitive_file_modifications` ledger metadata. No migration (JSON metadata);
the optional `_parse_response` kwarg leaves existing callers unaffected.

### Verification
148 targeted tests · ruff clean · E2E: full 97-skill sweep, live-store →
`recon_findings` → cleanup against the canonical DB, trust-filter ON → filed 0 /
`--no-trust` → all filed.

**Review note:** Codex was quota-exhausted and the code-reviewer subagent hit
repeated 529s, so this carries a rigorous self-review (which already caught + fixed
an rc-handling bug that would have dropped every CRITICAL skill, plus the
false-positive flood). Please run the PR-stage review before merge.

### Follow-ups (not in this PR)
- Recurring recon job for the skill scan.
- Baseline-and-diff drift detection (flag only new/changed skills).

### DIFF
--- a/src/genesis/learning/skills/applicator.py
+++ b/src/genesis/learning/skills/applicator.py
@@ -28,6 +28,22 @@ class SkillApplicator:
         self._autonomy_level = autonomy_level
         self._validator = SkillValidator()
 
+    @staticmethod
+    def _build_modification_metadata(proposal: SkillProposal) -> dict:
+        """Metadata recorded with a skill mutation in the cognitive-mod ledger.
+
+        Captures *why* the skill changed — the triggering signals
+        (failure_patterns_addressed + the SkillReport-derived provenance trace) —
+        not just the diff, so a degrading edit can be understood and rolled back.
+        """
+        return {
+            "skill_name": proposal.skill_name,
+            "change_size": proposal.change_size.value,
+            "confidence": proposal.confidence,
+            "failure_patterns_addressed": proposal.failure_patterns_addressed,
+            "provenance_trace": proposal.provenance_trace,
+        }
+
     async def apply(
         self,
         proposal: SkillProposal,
@@ -79,11 +95,7 @@ class SkillApplicator:
                 path=path,
                 new_content=proposal.proposed_content,
                 summary=f"MINOR auto-apply: {proposal.rationale[:200]}",
-                metadata={
-                    "skill_name": proposal.skill_name,
-                    "change_size": proposal.change_size.value,
-                    "confidence": proposal.confidence,
-                },
+                metadata=self._build_modification_metadata(proposal),
             )
 
             # Log observation (include validation warnings if any)
--- a/src/genesis/learning/skills/refiner.py
+++ b/src/genesis/learning/skills/refiner.py
@@ -51,7 +51,31 @@ class SkillRefiner:
             logger.exception("Skill refiner LLM call failed")
             return None
 
-        return self._parse_response(report.skill_name, result.content)
+        return self._parse_response(
+            report.skill_name,
+            result.content,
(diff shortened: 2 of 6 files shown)


WingedGuardian (commented):
## First-pass review — PR #742

**Structural review:** 2 WARNINGs, no ERRORs

---

### WARNING 1 — Trust-check mismatch in `main()` summary counts (`src/genesis/security/skill_scan.py`)

The `filed` and `skipped` counts printed at the end of `main()` re-evaluate trust using `Path(r.source)` (the `source` field from the SkillSpector JSON report):

```python
filed   = sum(1 for r in results if r.score >= args.min_score and not trusted(Path(r.source)))
skipped = sum(1 for r in results if trusted(Path(r.source)))
```

But `scan_and_store` made the actual filing decision using the original `skill_dir` passed into the loop — not `r.source`. If SkillSpector populates `source` with a git remote URL or a resolved absolute path that differs from the input directory, the printed counts will disagree with what was actually filed to recon. (The filing logic in `scan_and_store` itself is correct; only the display is affected.)

Suggestion: derive the counts from the `scan_and_store` return value or pass the original `skill_dir` list alongside `results` so the summary reflects real behaviour.

---

### WARNING 2 — `apply()` call site not integration-tested (`tests/test_learning/test_skill_provenance.py`)

`test_modification_metadata_includes_provenance_and_failure_patterns` tests `_build_modification_metadata` in isolation (correct), but no test exercises the `apply()` path end-to-end to confirm the new metadata dict (`failure_patterns_addressed`, `provenance_trace`) actually reaches `cognitive_file_modify`. If an existing `tests/test_learning/test_applicator.py` (or similar) covers `apply()`, worth adding an assertion there that the new fields are present in the `metadata` kwarg. Otherwise a future caller change could silently drop them without a test failure.

---

*No raw-SQL violations, no mutable defaults, no bare `except:`, no `/tmp/` usage — temp file correctly lands in `~/tmp`. New files both have corresponding test coverage. Import grouping clean (lazy imports inside functions are intentional for the optional SkillSpector dependency). Conventional commit prefix present.*

---

PR #744, opened 2026-06-22
fix(db,scheduler): migration false-failure reconciliation + decouple GitNexus CLAUDE.md strip

WingedGuardian:
### 1. Migration false-failure reconciliation (deploy-critical)

Deploying migration `0034` failed with `database is locked` **yet the migration had
actually committed** — in WAL mode a `COMMIT` can return `SQLITE_BUSY` from a
post-commit autocheckpoint *after* the commit frame is durably written, when
concurrent writers hold the WAL. The runner treated that as a failure, so the update
rolled the **code** back while the **DB** advanced (new-schema + old-code). Server
startup has the same exposure — it raises and fails init on a "failed" migration.

**Fix:** in `run_pending()`'s error path, roll back first (a no-op if the commit
already landed; undoes a genuinely-open transaction), then reconcile against
`schema_migrations`. If the migration is durably recorded it committed → treat it as
applied; only a genuinely-absent row is a real failure. Mechanism-agnostic, the happy
path is unchanged, and it fixes both callers (deploy + server startup). Migrations
also get a dedicated 60s busy-timeout (the app stays at 5s) so the lock and the
COMMIT-checkpoint can wait out contention. A regression test reproduces the exact
incident and asserts the reconciliation path actually fired.

### 2. Decouple the GitNexus CLAUDE.md strip

The strip that keeps GitNexus's auto-injected block out of `CLAUDE.md` only ran after
the scheduled Mon/Thu reindex, so out-of-band reindexes (GitNexus's own staleness
`analyze`) re-injected the block and it persisted for days. Added an hourly job plus
a one-shot strip on scheduler startup, reusing the existing idempotent strip
(`AGENTS.md` intentionally keeps its block).

### Verification

- The reproduction test fails on the old runner and passes on the new one
  (committed-but-busy → reconciled to applied; a genuine failure still rolls back).
  Full `test_db/` + `test_surplus/` suites green (904 tests).
- The migration entrypoint `--status` / `--dry-run` run clean against a real database
  with the new timeout. **No new migration in this PR**, so deploying it is a no-op on
  the migration step.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ## [Unreleased]
 
+### Fixed
+
+- **Updates no longer abort when a schema migration actually succeeded** — if the
+  database was busy during an update (for example a background task writing at the same
+  time), a migration could commit successfully yet still surface a transient "database
+  is locked" error, which made the update roll the code back while the database had
+  already moved forward. Updates now confirm whether the migration was truly applied
+  before treating it as a failure, and give migrations more room to wait out a busy
+  database.
+
 ## [v3.0b16] - 2026-06-21
 
 ### Added
--- a/src/genesis/db/connection.py
+++ b/src/genesis/db/connection.py
@@ -26,6 +26,14 @@ DEFAULT_DB_PATH = genesis_db_path()
 
 BUSY_TIMEOUT_MS = 5000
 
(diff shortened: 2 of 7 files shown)


WingedGuardian (commented):

---

WingedGuardian:
Merging on green CI (verified for the exact head SHA `c9146fb3`: test/lint/migration-check/leak-detector/CodeQL all pass) plus a clean inline code-review. Codex automated review was quota-exhausted, so per the pre-merge gate this merges on CI alone. This PR adds no migration, so the deploy is a no-op on the migration step.

PR #745, opened 2026-06-22
fix(tests): kill shared-fixture contaminant + feat(recon): operationalize skill-security scan

WingedGuardian:
## Summary

Two related pieces of work:

1. **Fix a latent test-suite contaminant** that silently poisoned the shared `db` fixture (it forced an earlier PR's storage test to go deterministic and cost a multi-session hunt).
2. **Operationalize the merged skill-security scan** (NVIDIA SkillSpector → recon findings) so it actually runs in production — it was code-complete but deploy-pending.

A third item — a **GitHub Discovery** recon engine — was scoped this round (design written to the internal spec doc) and is deliberately **not** in this PR; it's tracked as a follow-up.

## Task 1 — test contaminant + guard

Root cause: `tests/test_resilience/test_outreach_recovery.py` did `obs_crud.create = AsyncMock()` as a bare module-attribute assignment, never restored. It leaked process-wide — every later real insert returned a truthy Mock but wrote nothing, so a distant victim read 0 rows on its own connection and failed mysteriously (only in the full suite).

- Restore the patch via `with patch(...)` (matches the file's sibling tests).
- Add an **autouse conftest guard** that fails the *offending* test at teardown when it leaks a `unittest.mock` onto the real `genesis.db.crud.observations` module — so this class fails loudly at the source, not in a baffled victim downstream.

Verified RED→GREEN (12 failed → 123 passed on the minimal repro; the guard pins the leaker by name) and **0 false-positives across 4,464 tests**, including dirs that legitimately monkeypatch the guarded attribute (autouse teardown ordering restores those first).

Also fixes an unrelated, environment-specific test-isolation gap found along the way: `test_load_autonomous_cli_policy_from_yaml` didn't isolate the user-config `.local.yaml` overlay, so it failed on any live install (CI was green — no overlay there). Now pins the overlay dir to `tmp_path`.

## Task 2 — operationalize the skill-security scan

Wires the merged `genesis.security.skill_scan` for production, mirroring `ModelIntelligenceJob`:

- **`SkillSecurityScanJob`** (`recon/skill_security_scan_job.py`): resolves the binary (explicit → `SKILLSPECTOR_BIN` → install path → `PATH`), scans installed skills, files findings for **untrusted** skills only, and no-ops cleanly when the binary is absent.
- **SurplusScheduler**: `set_skill_security_scan_job` + `run_skill_security_scan` wrapper + a weekly `CronTrigger` (Mon 2am — never `IntervalTrigger`, which resets on restart). Wired in `runtime/init/surplus.py`.
- **`recon_run_skill_scan()`** MCP tool for on-demand runs (auto-exposed, like `recon_run_model_intelligence`).
- **`bootstrap.sh`**: installs NVIDIA SkillSpector to a stable path + a one-time `--seed-trusted` to bless the current install.

### Review fixes (also fix the Phase-A CLI, since `scan_and_store` is shared)
- The blocking SkillSpector subprocess now runs via `asyncio.to_thread`, so the weekly sweep can't stall the scheduler event loop.
- Each result's `source` is stamped with the actual scanned dir (not SkillSpector's self-reported field), so re-derived filed/trusted counts can't disagree with what was actually filed.

## Verification

- **Unit/integration**: job tests, Phase-A scan tests, and a new `test_surplus` test that locks job registration on `start()` (Level-4 wiring) — all green.
- **E2E against real SkillSpector v2.2.3**: `Discovered 97 skills; scanned 96 (1 timed out, skipped gracefully); filed 59 untrusted, 29 trusted-source skipped.` The CRITICAL/100 scorers are all first-party skills, correctly trust-skipped — confirming the "scan all, file untrusted only" design keeps the recon lane clean. Real JSON output parses correctly; `source` is the dir path.
- Reviewed by two independent agents; findings addressed.

## Notes / follow-ups (filed separately)
- GitHub Discovery recon engine (scoped, build next).
- Promote 4 private helpers in `skill_scan.py` to public (cross-package import smell).
- Add a `recon` capability-manifest entry for CC-session discoverability.
- Very large skills (e.g. tool-bundle skills) hit the 120s timeout and are skipped — safe today, a coverage gap worth future tuning.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -350,6 +350,26 @@ fi
 if command -v serena &>/dev/null; then
     echo "  Serena: $(serena --version 2>/dev/null || echo 'installed')"
 fi
+
+# SkillSpector (NVIDIA — skill-security scanner; external dep, not vendored).
+# The scheduled skill-security scan resolves the binary from this stable path.
+SKILLSPECTOR_DIR="$HOME/.genesis/deps/skillspector"
+if [[ ! -x "$SKILLSPECTOR_DIR/.venv/bin/skillspector" ]]; then
+    echo "  SkillSpector not found — installing..."
+    mkdir -p "$HOME/.genesis/deps"
+    if [[ ! -d "$SKILLSPECTOR_DIR/.git" ]]; then
+        git clone --depth 1 https://github.com/NVIDIA/SkillSpector.git "$SKILLSPECTOR_DIR" 2>/dev/null \
+            || echo "  WARNING: SkillSpector clone failed (skill-security scan will no-op until installed)"
+    fi
+    if [[ -d "$SKILLSPECTOR_DIR" ]]; then
+        "$PYTHON_BIN" -m venv "$SKILLSPECTOR_DIR/.venv" 2>/dev/null \
+            && "$SKILLSPECTOR_DIR/.venv/bin/pip" install -q "$SKILLSPECTOR_DIR" 2>/dev/null \
+            || echo "  WARNING: SkillSpector install failed (non-critical)"
+    fi
+fi
+if [[ -x "$SKILLSPECTOR_DIR/.venv/bin/skillspector" ]]; then
+    echo "  SkillSpector: installed at $SKILLSPECTOR_DIR/.venv/bin/skillspector"
+fi
 echo
 
 # --- Python venv ---
@@ -367,6 +387,13 @@ if ! "$VENV_DIR/bin/python" -c "from genesis.runtime import GenesisRuntime" 2>/d
     echo "  Re-run: $VENV_DIR/bin/pip install -e $GENESIS_ROOT --verbose"
     exit 1
 fi
+echo
+
+# --- Skill-security trusted baseline ---
+# Bless the currently-installed skills so the scheduled scan files findings only
+# for NEW/untrusted skills (a raw scan flags ~every capable skill as CRITICAL).
+echo "--- Seeding skill-security trusted allowlist ---"
+"$VENV_DIR/bin/python" -m genesis.security.skill_scan --seed-trusted 2>&1 | tail -1 || true
 echo
 
 # --- Secrets ---
--- a/src/genesis/mcp/recon_mcp.py
+++ b/src/genesis/mcp/recon_mcp.py
@@ -350,3 +350,22 @@ async def recon_run_model_intelligence() -> dict:
         db=_db, profile_registry=profile_registry, surplus_queue=_surplus_queue,
     )
     return await job.run()
+
+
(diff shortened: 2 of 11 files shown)


WingedGuardian:
## First-pass review — PR #745

### Structural review

**WARNING — Private cross-package imports**
`src/genesis/recon/skill_security_scan_job.py` imports 4 private symbols from `genesis.security.skill_scan`: `_default_roots`, `_default_trusted_file`, `_load_trusted_names`, `_repo_skill_roots`. These underscored names are internal to that module; importing them across package boundaries is a code smell. The PR description already tracks the fix ("Promote 4 private helpers in `skill_scan.py` to public"), so this is a tracking note — make sure the follow-up is filed before this gets stale.

**WARNING — Async tests missing `@pytest.mark.asyncio`**
`tests/test_recon/test_skill_security_scan_job.py` — `test_run_skips_gracefully_when_binary_missing` and `test_run_delegates_and_summarizes` are `async def` but carry no `@pytest.mark.asyncio` decorator. Sibling tests in the suite (e.g. `tests/test_resilience/test_outreach_recovery.py`) use explicit marks. Without `asyncio_mode = "auto"` set globally in pytest config, these functions execute as plain sync callables — the coroutine is instantiated but never awaited, and the test passes trivially, covering nothing. Please confirm `asyncio_mode = "auto"` is in `pyproject.toml`/`pytest.ini`, or add the decorator.

**WARNING — `total_findings` threshold disagrees with what was actually stored**
`src/genesis/recon/skill_security_scan_job.py:62`:
```python
filed = sum(1 for r in results if r.score >= 1 and not trusted(Path(r.source)))
```
`scan_and_store()` uses its `min_score` parameter (a default defined in `skill_scan.py`, not passed here) to decide what actually reaches `storer`. The `run()` summary recalculates with `score >= 1`, so if the default `min_score` is anything other than 1, `total_findings` overstates the number of stored DB findings. The metric and the storage should use the same threshold — either pass `min_score=1` explicitly to `scan_and_store`, or read the threshold from the same constant.

**NOTE — `CronTrigger` used before its in-function import**
`src/genesis/surplus/scheduler.py` — the new skill-security scan block calls `CronTrigger(...)` at a point in `start()` that precedes the existing `from apscheduler.triggers.cron import CronTrigger` line (visible in the unchanged context immediately after the new block). Tests pass, so `CronTrigger` must be in scope from an earlier module-level or function-level import — but the placement is confusing and could trip a future reader. Consider hoisting the import to module level or to the top of `start()`.

---

**CLEAN** — No API keys, tokens, private IPs, private hostnames, credentials, or SSH material found in added lines. PR title and description contain no language advertising a security incident or credential/PII cleanup.

---

PR #750, opened 2026-06-22
fix(observability): malformed ambient config now visible, not silent

WingedGuardian:
## Problem
A present-but-malformed `~/.genesis/ambient_remote.yaml` (unparseable, or missing `host_ip`/`host_user`) used to return `None` from `load_ambient_remote_config()` — **identical to "no ambient edge configured."** So a config typo silently disabled BOTH the dashboard ambient probe (#735) **and** the Telegram down-alert monitor (#680/#685), with only a log warning nobody reads. (Found while verifying the dashboard relabel — *this* install's config is fine, but the silent-disable is a real footgun.)

## Fix
- `load_ambient_remote_config()` now **raises `AmbientRemoteConfigError`** when the file is present AND enabled but malformed. Absent / `enabled: false` still return `None` (legit no-ops, no raise).
- `probe_ambient_health` catches it → returns a **DEGRADED `ambient` card** (`details.verdict = "misconfigured"`) carrying the reason → visible on the dashboard.
- `_ambient_health_job` catches it → `logger.error(...)` (loud) + records a clean job run (no phantom job failure).

A misconfigured ambient monitor is now **visible** instead of invisible.

## Tests
- Loader: absent→None, `enabled:false`→None, valid→config, missing `host_ip`→raise, missing `host_user`→raise, unparseable YAML→raise.
- Probe: misconfigured→DEGRADED.
- `tests/test_observability/` green; ruff clean. Code-reviewed (2 findings fixed: doc-IP in a fixture, missing `host_ip`-branch test).

### DIFF
--- a/src/genesis/observability/ambient_health.py
+++ b/src/genesis/observability/ambient_health.py
@@ -39,31 +39,47 @@ class AmbientRemoteConfig:
     health_path: str = _DEFAULT_HEALTH_PATH
 
 
+class AmbientRemoteConfigError(Exception):
+    """Raised when ``~/.genesis/ambient_remote.yaml`` is present AND enabled but
+    malformed (unparseable, or missing ``host_ip``/``host_user``).
+
+    Distinct from absent/disabled (which return ``None``) so callers can surface a
+    VISIBLE "misconfigured" state — instead of a config typo silently looking
+    identical to "no ambient edge configured" and disabling monitoring unnoticed."""
+
+
 def load_ambient_remote_config() -> AmbientRemoteConfig | None:
-    """Load ``~/.genesis/ambient_remote.yaml``; return None if absent, disabled,
-    or invalid (monitor then no-ops). Never raises."""
+    """Load ``~/.genesis/ambient_remote.yaml``.
+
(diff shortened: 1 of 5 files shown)


WingedGuardian (commented):
**Structural review: 1 WARNING — no ERRORs**

---

### WARNING — Missing test coverage for `_ambient_health_job` error branch

`src/genesis/outreach/scheduler.py` — the new `except AmbientRemoteConfigError` block in `_ambient_health_job` (inner try, ~line 510) calls `logger.error(...)` + `await self._record_job_result("ambient_health")` + returns. This is a new code path with observable side effects but no corresponding test was added.

The `probe_ambient_health` misconfigured path is covered in `test_health.py`. Suggest adding a parallel test for `_ambient_health_job` (mock `load_ambient_remote_config` to raise `AmbientRemoteConfigError`, assert `_record_job_result` is called and the job returns cleanly without propagating).

---

Everything else is clean: no pattern violations, imports correctly grouped, new exception class and updated loader docstring are both properly documented, conventional commit prefix is valid, no PII/secrets/wording issues in code or PR metadata.

---

PR #746, opened 2026-06-22
feat(follow-ups): cockpit dashboard tab — PR2 (manage/drain the backlog)

WingedGuardian:
## Follow-ups cockpit — PR2 of the follow-up subsystem redesign

A dedicated **Follow-ups** dashboard tab so the backlog can actually be drained: server-side filter/search/sort/paginate the full table, flip items between the follow-up and tabled lanes, override a misclassified domain, edit priority/pin, mark Done, or delete — individually or in batches.

PR1 (#741) landed the schema foundation (`kind`/`domain`/`goal_id` + the cockpit data layer). **This PR adds no migration** — it builds on that live schema.

### What's in it

**Data layer** (`db/crud/follow_ups.py`)
- `set_priority` (the one cockpit mutator PR1 didn't add) — validated against the `priority` CHECK set.
- Atomic batch helpers `delete_batch` / `set_kind_batch` / `update_status_batch` — single-statement `WHERE id IN (...)` so a multi-row action is one transaction (no silent partial failure on a large selection). `update_status_batch` validates status and stamps `completed_at` on terminal states, mirroring `update_status`.

**Routes** (`dashboard/routes/follow_ups.py`)
- `GET  /api/genesis/follow-ups/cockpit` — paginated/sorted/filtered list.
- `GET  /api/genesis/follow-ups/filters` — distinct sources + the fixed status set.
- `POST /api/genesis/follow-ups/<id>/{done,delete,pin,priority,kind,domain}`.
- `POST /api/genesis/follow-ups/batch` — done/delete/tabled/follow_up (max 200).
- Reads return an empty 200 when unbootstrapped; mutations return 503/404/400; the batch dispatch is exception-guarded.

**UI** (`templates/genesis_dashboard.html`)
- A `Follow-ups | Tabled | All` tier toggle (default = follow-ups; tabled is a deliberate click-away).
- Domain / Status / Source / Sort filters + keyword search + true page-N-of-M pagination.
- Per-row actions in the expanded detail: **Done/Dismiss** (soft — status→completed, row kept) is primary; **Delete forever** is secondary and gated behind a JS confirm; plus Pin, Priority, Move-lane, and Domain-override.
- Batch actions over a checkbox selection. Row body mirrors the existing read-only Work-tab follow-ups panel; filter/batch/pagination chrome mirrors the Observations tab. The existing Work-tab panel is untouched (additive).

### Design notes
[...]

PR #751, opened 2026-06-22
feat(ego,outreach,learning): domain-scope follow-up consumers (PR3)

WingedGuardian:
Third piece of the follow-up subsystem redesign (after #741 schema foundation and #746 the cockpit). Scopes the user-facing follow-up consumers to the user's domain so they surface the user's items instead of Genesis's internal-engineering backlog. **No migration** — the `domain` column already exists (#741).

## Why
Every consumer read `follow_ups` domain-blind. Because `user_input_needed` is effectively a junk-drawer strategy (the dominant pending class), the user ego, the morning report, and the `user_goal_staleness` signal were dominated by internal-dev backlog. Two concrete effects, both verified against live data:
- **Genuine user items were invisible.** `get_actionable` caps at the top 50 by `(priority, created_at)`; the real user-world items are recent, so they rank past the cap and never reach the user ego at all.
- **The staleness signal was mis-measuring.** Named/rendered as *user*-goal staleness, it counted internal items too — pinning it near the top of its range off internal backlog.

## What
- **CRUD (commit 1):** an optional exact-match `domain` param on `get_pending` / `get_by_status` / `get_actionable` / `get_recently_completed` (via a small `_domain_eq` helper). When set, scopes in SQL *before* the LIMIT; `domain=None` is a byte-identical no-op, so every other caller is untouched. Cockpit-consistent exact-match — deliberately not a second NULL-handling idiom.
- **Consumers (commit 2):** the user ego's open-threads section, the "awaiting user input" counter, the `user_goal_staleness` signal, and the morning report are scoped to the user's domain (signal + counter also exclude tabled rows). The genesis ego is **deferred** — its existing strategy/source post-filter already rejects the relevant rows, so scoping it is a near-no-op.

## Deliberate trade-offs (accepted)
- **Pinned / high-priority cross-domain items re-home to the cockpit.** Strict scoping drops them from the user (CEO) ego — by design: they're internal/COO-domain work, fully visible in the cockpit (#746). Nothing is lost, just correctly homed.
- **The signal change softens Light-reflection cadence.** The signal feeds the Light-reflection scorer, so correcting its value (a de-biasing) reduces how often Light reflections fire. Intended.

## Testing
- ~200 targeted tests green; ruff clean. New tests cover the domain filter, the `domain=None` backward-compat no-op per reader, the signal scoping, the ego-level internal-exclusion, and the pinned-cross-domain re-homing.
- **E2E against a consistent copy of the live DB, running this branch's code:** the unscoped top-50 contains zero user-world items (the burial), while the scoped query returns exactly the genuine user items; the counter and signal de-skew as expected. The full live verification (ego cycle, reflection cadence, morning report on the running server) is post-deploy.

CHANGELOG: no entry here — consistent with #741/#746; a consolidated redesign entry belongs at the next release.

### DIFF
--- a/src/genesis/db/crud/follow_ups.py
+++ b/src/genesis/db/crud/follow_ups.py
@@ -25,6 +25,20 @@ def _normalize_scheduled_at(iso_str: str | None) -> str | None:
     return dt.isoformat()
 
 
+def _domain_eq(domain: str | None) -> tuple[str, list[str]]:
+    """Build an exact-match domain WHERE fragment for the reader queries.
+
+    Returns ``("", [])`` when ``domain`` is None (no-op — the caller behaves
+    byte-identically to before), else ``(" AND domain = ?", [domain])``. Exact
+    match only: when a domain is given, NULL-domain rows are excluded. This
+    mirrors the cockpit's ``_build_filter_where`` exact-match semantics — it is
+    deliberately NOT a second NULL-handling idiom (no "domain OR NULL" union).
+    """
+    if domain is None:
+        return "", []
+    return " AND domain = ?", [domain]
+
+
 async def create(
     db: aiosqlite.Connection,
     *,
@@ -74,11 +88,13 @@ async def get_pending(
     source: str | None = None,
     strategy: str | None = None,
     include_tabled: bool = False,
+    domain: str | None = None,
 ) -> list[dict]:
-    """Get pending follow-ups, optionally filtered by source/strategy.
+    """Get pending follow-ups, optionally filtered by source/strategy/domain.
 
     Tabled follow-ups (kind='tabled') are excluded unless include_tabled=True —
     tabled items are tracked but never dispatched or surfaced as action.
+    domain (exact match) scopes to that domain only; None = all domains (no-op).
     """
     query = "SELECT * FROM follow_ups WHERE status = 'pending'"
     params: list[str] = []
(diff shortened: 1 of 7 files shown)


WingedGuardian:
## First-pass review

### Structural review

**NOTE (pre-existing, not introduced by this PR):** Two call sites execute inline SQL directly against `follow_ups` rather than going through the CRUD layer:

- `src/genesis/ego/user_context.py` (~line 688): `_backlog_summary_section` runs `SELECT COUNT(*), MIN(created_at) FROM follow_ups …` inline. PR3 correctly adds `AND domain = 'user_world' AND kind = 'follow_up'` to scope it — but the raw-SQL site predates this PR. No CRUD equivalent exists for this aggregate shape yet.
- `src/genesis/learning/signals/user_goal_staleness.py` (~line 71): `_check_follow_ups` runs `SELECT created_at FROM follow_ups … LIMIT 1` inline. Same pattern — new domain conditions are correct, but the site bypasses CRUD. Could eventually use `get_pending(db, strategy="user_input_needed", domain="user_world")` + sort/limit, but that's a future refactor.

Not a blocker; flagging for awareness since both files were touched and the pattern is an ERROR-class rule.

Everything else is clean:
- No bare `except:`, mutable defaults, `/tmp/` usage, `git add .`/`git add -A`, direct `task_states` inserts, `nohup`, or `os.killpg`.
- Import hygiene clean in the new test file.
- Test coverage: all four CRUD readers have domain-scoping tests; new `tests/test_learning/test_user_goal_staleness.py` covers the signal; `test_user_context.py` updated with the internal-exclusion case. ✓
- Docstrings updated on every modified public function. ✓
- Commit prefix `feat(...)` follows convention. ✓

No credentials, private IPs, private hostnames, hardware identifiers, or wording that signals a security incident or infrastructure exposure in the diff or PR metadata.

---

PR #758, opened 2026-06-23
fix(memory): evidence-gate dream-cycle auto-merge + survivor fix

WingedGuardian:
## Problem

The dream-cycle entity-resolution phase auto-merges near-duplicate episodic
memories whenever cosine ≥ 0.95 — **on cosine alone, with no corroboration**.
The LLM path (0.85–0.95) is already double-gated by an adversarial second
opinion, so the auto-merge branch is the only place a memory is permanently
deprecated on a single number.

A replay over **1,749 historical auto-merges** (audit → live Qdrant payloads)
showed this is not hypothetical: real false-merges happen at 0.95–0.96 cosine —
different PR numbers ("Merged PR #260" vs "#480"), plan-vs-implementation facts
("Planned LLM reranking" vs "Default rerank=True"), and distinct queries merged
on shared boilerplate. A false-merge permanently deprecates a distinct memory
and corrupts episodic retrieval.

## Change

**1. Evidence gate (auto-merge branch only).** `compute_evidence_strength`
combines four signals — cosine (dominant) + temporal closeness + confidence +
a load-bearing bar-raiser — into a `[0,1]` score. Below `EVIDENCE_THRESHOLD`
(0.30) the pair is logged as a `flagged` audit row (`low_evidence`) for review
instead of being deprecated. Coincident-weakness design: near-identical text
(high cosine) always clears the gate; no single weak signal can block. The LLM
path is untouched. **No migration** — reuses the existing `flagged` action.

**2. Survivor fix.** For confirmed duplicates (auto_merge + llm_merge),
`pick_duplicate_survivor` keeps the more-retrieved (load-bearing) memory as
survivor instead of always-newest. Live data showed a `retrieved_count=5`
memory being deprecated in favor of a `retrieved_count=0` one. The
contradiction / `succeeded_by` path intentionally keeps temporal survivorship.

**3. Drive-by safety fix.** The pre-existing `contradicts` link handler bound
`src`/`tgt` inside the `try`; an import/create failure would hit the `except`
with an `UnboundLocalError` and crash the whole resolution phase. Direction is
now resolved before the `try`.

## Calibration (how the threshold was chosen)

The `EVIDENCE_THRESHOLD` and signal weights were calibrated by replaying the
scorer over all 1,749 historical merges, not guessed. At **0.30** the gate:

- blocks **~5%** of historical merges, **all within the 0.95–0.96 suspect band**;
- **never** blocks a ≥0.98 near-identical pair (0 of 353);
- the blocked rows skew to genuine false-merges (different PR numbers,
  plan-vs-impl, distinct sessions far apart in time).

This is a conservative correction: a missed merge is recoverable on the next
weekly run; a false-merge is not.

## Verification

- **TDD**: 9 unit tests (evidence scorer invariants + survivor) + 3 integration
  tests exercising the full `run_entity_resolution` path against a real
  in-memory audit DB. 60 targeted memory/dream tests green, `ruff` clean.
- **E2E on real data**: replayed the *shipped* `compute_evidence_strength` over
  1,727 real merges — 0 raises, 5.0% blocked (matches calibration), 0 high-cosine
  blocked.
- **Review**: genesis-architect (design) + two code-review passes (first
  found+fixed a critical pre-existing `UnboundLocalError`; second on the final
  commit is clean, no findings). External-model second opinion (Codex / OpenCode)
  was quota-blocked at review time — merge relies on CI + the Claude-family
  passes per the review-quota policy.

## Deferred (tracked, not in this PR)

- `succeeded_by` link direction is unreliable when one memory lacks a timestamp
  (`ts_a != ts_b` on `"" vs real`) — pre-existing, in the contradiction path.
- Survivor-aware behavior in the contradiction path (out of scope; duplicates
  only here).

### DIFF
--- a/src/genesis/memory/dream_entity_scan.py
+++ b/src/genesis/memory/dream_entity_scan.py
@@ -13,6 +13,7 @@ Contradictions create ``contradicts`` or ``succeeded_by`` links in the graph.
 
 from __future__ import annotations
 
+import json
 import logging
 from datetime import UTC, datetime
 from typing import TYPE_CHECKING, Any
@@ -59,17 +60,21 @@ async def run_entity_resolution(
     """
     from genesis.memory.entity_resolution import (
         AUTO_MERGE_THRESHOLD,
+        EVIDENCE_THRESHOLD,
         LLM_CHECK_FLOOR,
         MAX_ENTITY_CHECKS_PER_RUN,
         check_semantic_overlap,
+        compute_evidence_strength,
         find_dedup_candidates,
         log_resolution,
+        pick_duplicate_survivor,
     )
     from genesis.qdrant import collections as qdrant_ops
 
     report: dict[str, Any] = {
         "candidates_found": 0,
         "auto_merged": 0,
+        "low_evidence_skipped": 0,
         "llm_checked": 0,
         "llm_merged": 0,
         "contradictions": 0,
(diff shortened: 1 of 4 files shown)


WingedGuardian:

---

PR #759, opened 2026-06-23
feat(memory): locate self-retrieval MCP tool (Track 5.1)

WingedGuardian:
## Summary

Adds `locate` — a deterministic, **read-only** MCP tool on the `genesis-memory` server
(`mcp__genesis-memory__locate`) that surfaces recently-changed files by mtime across known roots,
filtered by time window / file type / scope / name-glob / optional content match, ranked by recency.

It's a Genesis-owned alternative to Claude Code's `Grep`/`Glob` (unreliable under the open-agents A/B
issue) — a dependable way to answer "what changed recently / where is the doc about X?" instead of
falling back to reconstructing state from git.

## Behavior

- **Scopes:** `docs` (default = plans + output), `plans`, `output`, `repo`, `worktrees`, `all`.
  `repo` / `worktrees` / `all` are opt-in (heavier).
- **Filters:** `within` (e.g. `12h` / `24h` / `7d` / `0` = no limit), `file_type` (any/code/noncode),
  `name` (case-insensitive glob), `contains` (+ `regex`), `limit` (default 50, hard cap 200).
- **Output:** recency-ranked results with `mtime`, human `age`, size, scope label, and per-file
  content-match snippets. Honest truncation signals — `truncated`, `content_pool_capped`,
  `scan_truncated` — so coverage is never silently capped.

## Implementation

- Pure-Python `os.scandir` walk (symlink-safe, permission-tolerant) with directory pruning
  (`.git` / `.venv` / `node_modules` / build caches / nested worktrees / etc.).
- Content matching via `rg --json` with a pure-Python `re` fallback (identical result shape). Binary
  and oversize files are pre-filtered before either backend (ripgrep skips its own binary/size
  filtering when passed explicit path args, so the tool gates candidates itself).
- Worktrees discovered via `git worktree list --porcelain`.
- New `env.py` path helpers (`genesis_home` / `claude_home` / `plans_dir` / `output_dir`) following the
  existing env-override + default pattern.

## Testing

- 26 unit tests (`tests/test_mcp/test_locate.py`): recency boundaries, type/scope/glob filters,
  pruning, rg + fallback parity, binary/oversize skip, truncation signals, ordering, symlink and
  permission safety, MCP registration.
- Existing memory-MCP tests unaffected.
- E2E-verified against a live filesystem (default `docs` query returns in single-digit milliseconds;
  pruning keeps a full-repo scan fast).

## Notes

- Strictly read-only; no migration. New MCP tools become available to Claude Code on the next session.
- First slice of the self-retrieval track of the omnipresence layer.

### DIFF
--- a/src/genesis/env.py
+++ b/src/genesis/env.py
@@ -93,6 +93,30 @@ def genesis_db_path() -> Path:
     return repo_root() / "data" / "genesis.db"
 
 
+def genesis_home() -> Path:
+    """Resolve the Genesis runtime home (~/.genesis): output, sessions, config."""
+    value = os.environ.get("GENESIS_HOME")
+    return Path(value).expanduser() if value else Path.home() / ".genesis"
+
+
+def claude_home() -> Path:
+    """Resolve the Claude Code home (~/.claude): plans, skills, projects."""
+    value = os.environ.get("CLAUDE_HOME")
+    return Path(value).expanduser() if value else Path.home() / ".claude"
+
+
+def plans_dir() -> Path:
+    """Working plan/roadmap docs (~/.claude/plans)."""
+    value = os.environ.get("GENESIS_PLANS_DIR")
+    return Path(value).expanduser() if value else claude_home() / "plans"
+
+
+def output_dir() -> Path:
+    """Genesis report/spec/content output (~/.genesis/output)."""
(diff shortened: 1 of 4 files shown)


WingedGuardian:

---

PR #760, opened 2026-06-23
fix(guardian): wire watchdog escalations to Telegram (11569455)

WingedGuardian:
## The gap (follow-up 11569455)
The Guardian watchdog was constructed with no outreach wiring (`init/guardian.py` passed only `remote` + `event_bus`), so its **three** user-facing escalations only ever hit the event bus, never the user:
1. **recovery-failed** — Guardian DOWN *and* the SSH restart failed (the last line of defense is down);
2. **code-drift** — the host is running stale Guardian code;
3. **gateway-staleness** — the deployed gateway script is frozen after a failed self-heal.

A genuine Guardian degradation went unnoticed unless someone watched the event stream. The pre-existing `self._outreach_queue.enqueue(msg, priority=…, source=…)` call was also dead *and* matched no real API.

## The fix
Inject the outreach pipeline (`rt._outreach_pipeline`, initialized before guardian init) and deliver via a shared `_alert_user(topic, context, source_id)` helper using `submit_raw(text, OutreachRequest(category=BLOCKER, signal_type="guardian_alert", source_id=…))` — the idiomatic urgent-infrastructure path that **delivers to Telegram immediately, skips the LLM drafter, and applies built-in dedup** (the same path the health alerts use). `_alert_user` is a no-op when the pipeline is unwired and `try/except`-wrapped so an outreach failure never breaks a watchdog tick.

**Cadence — one alert per episode**, re-armed on resolution, so a sustained problem pings once (not every ~15 min). Gateway already had `_gateway_escalated`; code-drift gains `_drift_escalated` and recovery-failed gains `_recovery_failed_escalated` (both reset when the condition clears). Each site uses a **distinct topic** so the pipeline's `(signal_type, topic, category)` dedup never cross-suppresses one Guardian alert with another.

**Why direct delivery, not the health-alert whitelist:** the health-alert list is built by a hardcoded probe pipeline that never reads the event bus, so guardian events can't become whitelist-able alerts. Direct delivery is the correct path.

## Tests (393 `test_guardian` green)
Once-per-episode + re-arm for all three sites (code-drift ticked past threshold → one alert; recovery DOWN×3 → one alert, recover→DOWN → second alert; gateway escalate-once), the recovery/gateway alert envelopes, and direct `_alert_user` coverage (real BLOCKER envelope, None-graceful no-op, failure-swallow). No migration.

### DIFF
--- a/src/genesis/guardian/watchdog.py
+++ b/src/genesis/guardian/watchdog.py
@@ -52,21 +52,55 @@ class GuardianWatchdog:
         self,
         remote,  # GuardianRemote — import avoided for loose coupling
         event_bus=None,
-        outreach_queue=None,
+        outreach_pipeline=None,  # OutreachPipeline — for user-facing alerts
     ) -> None:
         self._remote = remote
         self._event_bus = event_bus
-        self._outreach_queue = outreach_queue
+        self._outreach_pipeline = outreach_pipeline
         self._sentinel = None
         self._last_recovery_at: datetime | None = None
         self._last_reset_at: datetime | None = None
         self._consecutive_stuck: int = 0
+        self._recovery_failed_escalated: bool = False  # one alert per DOWN episode
         self._drift_count: int = 0
+        self._drift_escalated: bool = False  # one user alert per drift episode
         # Deployed-gateway staleness: the install-dir code can be current while
         # ~/.local/bin/guardian-gateway.sh lags (the bug that froze it ~2 months).
         self._gateway_drift_count: int = 0
         self._gateway_resync_attempted: bool = False
         self._gateway_escalated: bool = False
+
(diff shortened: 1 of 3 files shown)


WingedGuardian (commented):

---

**NOTE** `src/genesis/guardian/watchdog.py` — `_alert_user()` contains a deferred `from genesis.outreach.types import OutreachCategory, OutreachRequest` inside the function body, which breaks the stdlib → third-party → local import grouping convention. Clearly intentional — it avoids a circular import and is consistent with the file's existing "import avoided for loose coupling" pattern on `remote`. Fine as-is; noting for future readers.

**NOTE** `src/genesis/runtime/init/guardian.py:61–63` — the inline comment cites specific line numbers in `_core.py` (`:413`, `:432`). These will silently go stale as the file shifts. Consider referencing function or section names instead of line numbers.

---

PR #762, opened 2026-06-23
fix(guardian): gate code-drift detection to last deploy, not live HEAD

WingedGuardian:
## Problem

The container-side Guardian watchdog raises false **\"Guardian code drift\"** Telegram alerts:

```
level=ERROR genesis.guardian.watchdog "Guardian code drift detected for 3 ticks: container=<a> host=<b>"
```

## Root cause

`_check_code_drift_inner` compared the host's `deployed_commit` against the last Guardian-path commit reachable from the container's **live `HEAD`**. But the two move on different cadences:

- The **host Guardian** is only redeployed when `scripts/update.sh` runs (it archives HEAD and pushes a `redeploy` to the host *if* a Guardian-path changed). It is not on a timer.
- The container's **`main`** advances on **every PR merge**.

So `HEAD` races ahead of the host between deploys, and any Guardian-path commit landed in that window trips a false alarm ~15 min later (3 ticks) — even though the host is exactly where the last deploy put it. Comparing two commits alone can't distinguish *\"behind because no deploy ran yet\"* (benign) from *\"behind because a deploy failed/froze\"* (the #669/#670 case this check exists for).

## Fix

Measure drift against **what `update.sh` last successfully deployed**, not live `HEAD`:

- New `_last_deployed_commit()` reads the latest `status='success'` row's `new_commit` from `update_history` (the container's own deploy ledger). Function-level imports keep `genesis.db` off the host's import path (this module is container-side — see `guardian/DEPLOYMENT.md`).
- `container_hash` is now derived from that deploy baseline (`git log -1 --format=%h <deploy_ref> -- <GUARDIAN_PATHS>`).
- No baseline recorded → skip (never fall back to HEAD). `deploy_ref` unresolvable / no Guardian-path commit reachable → skip, with a debug log so it's observable.
- Everything downstream (ancestry check, drift counting, once-per-episode Telegram alert) is unchanged.

Result: benign between-deploy lag goes silent; a genuinely stale/frozen host (deploy attempted but not applied) **still alerts**.

Review follow-ups included: observable empty-skip log; `ORDER BY datetime(started_at)` so mixed timezone offsets order by true instant.

## Verification

- Targeted: `tests/test_guardian/` — **400 passed** (7 new/changed cases: deploy-baseline reference, no-baseline skip, empty-reachable skip, `_last_deployed_commit` success/none/missing-db, mixed-timezone ordering).
- End-to-end against live state:
  - host = latest successful deploy commit → **no alert** (false alarm gone).
  - host frozen at an older commit the deploy baseline contains → **still alerts** (signal preserved).

### DIFF
--- a/src/genesis/db/crud/update_history.py
+++ b/src/genesis/db/crud/update_history.py
@@ -1,0 +1,25 @@
+"""CRUD reader for the ``update_history`` ledger (Genesis self-update attempts).
+
+Writes are owned by ``scripts/update.sh`` (the deploy script records each run);
+this module provides read access through the CRUD layer so callers never issue
+raw SQL against ``genesis.db``.
+"""
+
+from __future__ import annotations
+
+import aiosqlite
+
+
+async def last_successful_deploy_commit(db: aiosqlite.Connection) -> str | None:
+    """Return ``new_commit`` of the most recent successful update, else None.
+
+    Ordered by ``datetime(started_at)`` so ISO timestamps with differing
+    timezone offsets compare by true instant rather than lexicographically.
+    """
+    cur = await db.execute(
+        "SELECT new_commit FROM update_history "
+        "WHERE status = 'success' AND new_commit IS NOT NULL "
+        "ORDER BY datetime(started_at) DESC LIMIT 1",
+    )
+    row = await cur.fetchone()
+    return str(row[0]).strip() if row and row[0] else None
--- a/src/genesis/guardian/watchdog.py
+++ b/src/genesis/guardian/watchdog.py
@@ -309,6 +309,13 @@ class GuardianWatchdog:
         if branch_result.returncode == 0 and branch_result.stdout.strip() not in ("main",):
             return
         if not container_hash:
+            # deploy_ref resolved but no Guardian-path commit is reachable from
+            # it — nothing to compare. Log so this silent-skip is observable
+            # (mirrors the no-baseline debug above) rather than going dark.
+            logger.debug(
+                "Guardian drift check: no Guardian-path commit reachable from "
+                "deploy_ref=%s — skipping", deploy_ref,
+            )
             return
 
         # Get host's deployed hash via SSH version command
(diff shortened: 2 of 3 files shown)


WingedGuardian:
## First-pass Code Review — PR #762

### Structural Review

**ERROR — Raw SQL against `genesis.db` in `_last_deployed_commit()`**

**File:** `src/genesis/guardian/watchdog.py` — new `_last_deployed_commit` method

`_last_deployed_commit` opens a direct `aiosqlite` connection to `genesis.db` and issues raw SQL against `update_history`:

```python
cur = await db.execute(
    "SELECT new_commit FROM update_history "
    "WHERE status = 'success' AND new_commit IS NOT NULL "
    "ORDER BY datetime(started_at) DESC LIMIT 1",
)
```

Genesis pattern: **raw SQL against `genesis.db` must go through `genesis.db.crud` modules** (migrations are the only exception). This isn't a migration.

The docstring explains the function-level imports are to keep `genesis.db` off the host's import path — that's a valid architectural concern, but it doesn't require bypassing the CRUD layer. A CRUD helper can also be imported at function level. If an `update_history` CRUD reader doesn't exist yet, create one (or add a query helper to the closest existing module) and import it inside the function body the same way `genesis.db.connection.BUSY_TIMEOUT_MS` and `genesis.env.genesis_db_path` are already imported here.

If importing even a CRUD module at function level would genuinely pull in the full genesis.db stack and break the host-side constraint, document that explicitly in the docstring as a justified exception — currently the comment explains *why function-level imports*, not *why raw SQL*.

---

**CLEAN** — No secrets, API keys, private IPs, personal emails, private hostnames, or wording issues found in the code diff or PR metadata.

---

WingedGuardian:
Addressed the **ERROR** (raw SQL against `genesis.db`) in `0f16ea21`:

- Added `genesis.db.crud.update_history.last_successful_deploy_commit(db)` — the query now lives in the CRUD layer.
- `_last_deployed_commit()` retains only connection management + the function-level imports that keep `genesis.db` off the host's import path (container-side constraint), and delegates the query to the CRUD reader.

Behavior unchanged: 44 `test_guardian` cases + live e2e (`deploy_ref=053f5998`) still pass. The PII/secrets scan was already CLEAN.

PR #766, opened 2026-06-23
fix(dashboard): count LLM totals from activity_log, not cost_events

WingedGuardian:
## Problem
The Internals tab's LLM **totals** (calls/1h, calls/24h) read from `cost_events`, but the router writes a `cost_events` row only when cost is non-zero/unknown — so every **free-tier** call was invisible in the totals. Live data: **615** LLM calls in `activity_log` vs **141** `cost_events` rows in 24h — a ~75% undercount, which made the tab "look broken."

## Change
Count `calls_1h`/`calls_24h` from `activity_log WHERE provider LIKE 'llm.%'`; keep the dollar cost (`cost_1h`/`cost_24h`) sourced from `cost_events`. Per-provider rows already used `activity_log` and were correct — only the totals query was wrong.

## Not changed (verified, not bugs)
- Per-provider cards correctly show no per-call cost for free providers (guarded on `cost_24h > 0`); the near-zero **total** cost is accurate (mostly free tiers).
- `is_free` flags left as-is — whether `glm51`/`deepseek-v4-flash` are truly free is a real-world pricing decision, not a code bug.

## Tests
`tests/test_dashboard/test_vitals_llm_totals.py` — real-DB regression test proving totals count from `activity_log` (7) not `cost_events` (would be 2).

### DIFF
--- a/src/genesis/dashboard/routes/vitals.py
+++ b/src/genesis/dashboard/routes/vitals.py
@@ -468,21 +468,31 @@ async def _build_llm_section(rt, routing_ctx: dict) -> dict:
             providers_list.sort(key=lambda p: (-p["calls_24h"], p["name"]))
             section["providers"] = providers_list
 
-            # Totals
+            # Totals — call counts come from activity_log (EVERY LLM call is
+            # logged there, including free-tier providers that emit no
+            # cost_events row); dollar cost comes from cost_events. Counting
+            # calls from cost_events undercounted by ~75% because free calls
+            # are gated out of cost_events (router writes only when cost > 0).
             cursor = await rt.db.execute(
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

**NOTE** — `src/genesis/dashboard/routes/vitals.py`: the new `rt.db.execute(... FROM activity_log ...)` is raw SQL against genesis.db. This is consistent with the pre-existing pattern throughout `_build_llm_section` (the whole function uses raw SQL), so not a new violation, but the module hasn't migrated to CRUD modules. Worth a follow-up task if the CRUD-module convention is being actively enforced across `dashboard/routes/`.

---

PR #772, opened 2026-06-24
feat(deliberation): deliberate() 'the chorus' — model-panel verdict + dissent

WingedGuardian:
## Summary

Adds `deliberate()` — "the chorus" — a cognitive primitive that consults a panel of models and returns
a synthesized verdict (and, in analysis mode, the explicit disagreement). Exposed as the on-demand
`mcp__genesis-health__deliberate` tool, for genuinely high-stakes or contested decisions where one
narrow model isn't enough. Opt-in, **paid** (OpenRouter Fusion), recursion-blocked, and never raises
(failures come back as `error`). Not wired into any production judgment path.

## Two modes

- **`synthesis`** (default): the `openrouter/fusion` model-slug — a server-side panel + judge produces
  a fast prose verdict (~45–110s); the disagreement is woven into the prose.
- **`analysis`**: a free, tool-capable orchestrator + the Fusion **server-tool** — returns
  **machine-structured** `{consensus, dissent[], blind_spots[], confidence}` (~3 min). Use when you
  need deep, separable disagreement.

## Design

- New package `src/genesis/deliberation/` (types, core, backends/) behind a `Backend` protocol, so
  additional backends drop in without touching existing code.
- The Fusion backend calls `litellm.acompletion` directly (a documented exception to routing through
  the Router) so it can own the request shape, read the real per-call cost from `response.usage.cost`
  (litellm can't price these aggregator model strings), and map every error to a graceful result.
- Explicit `max_tokens` (the provider's default trips a 402); the server-tool config is passed via
  `extra_body` so it isn't dropped; cost/latency captured per call. ContextVar recursion guard.

## Testing

- 26 unit tests (`tests/test_deliberation/`): both modes, the parser (structured JSON + prose fallback +
  fenced/embedded JSON), recursion guard, error/timeout handling, cost extraction, and MCP registration.
  litellm is mocked; real response captures are used as fixtures.
- Existing health-MCP tests unaffected. Both modes verified live end-to-end.

## Notes

- Self-contained — no `model_routing.yaml` / `model_profiles.yaml` changes (the model string is a
  constant; cost is read from the response).
- New MCP tools become available to Claude Code on the next session.
- First on-demand surface of the deliberation track; call-site consumers and additional backends are
  future work.

### DIFF
--- a/src/genesis/deliberation/__init__.py
+++ b/src/genesis/deliberation/__init__.py
@@ -1,0 +1,9 @@
+"""Deliberation — "the chorus" (Track 4, omnipresence layer).
+
+`deliberate()` asks a panel of models and returns a synthesized verdict + the dissent.
+"""
+
+from genesis.deliberation.core import deliberate
+from genesis.deliberation.types import DeliberationResult, PerModel
+
+__all__ = ["DeliberationResult", "PerModel", "deliberate"]
--- a/src/genesis/deliberation/backends/__init__.py
+++ b/src/genesis/deliberation/backends/__init__.py
@@ -1,0 +1,16 @@
+"""Deliberation backends registry.
+
+v1 ships Fusion only. Fugu (`model: fugu`) and a Genesis-orchestrated panel land as new
+files registered here, with no change to `deliberate()` or existing backends.
+"""
+
+from __future__ import annotations
+
+from genesis.deliberation.backends.base import Backend
+from genesis.deliberation.backends.fusion import FusionBackend
+
+BACKENDS: dict[str, Backend] = {"fusion": FusionBackend()}
+
+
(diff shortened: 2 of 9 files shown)


WingedGuardian:

---

**NOTE — Missing docstrings on two public symbols:**

- `FusionBackend.run` in `src/genesis/deliberation/backends/fusion.py` — implements the `Backend` protocol but has no docstring. The protocol's `base.py` module-level docstring documents the contract; a one-liner on the method itself would be consistent with the other public surfaces here.
- `get_backend` in `src/genesis/deliberation/backends/__init__.py:15` — public, no docstring.

Both are minor; the surrounding module-level docstrings carry most of the intent. No blocking issues.</body>
</invoke>

---

PR #773, opened 2026-06-24
fix(eval): truthful procedure-confidence composite + procedure store hygiene

WingedGuardian:
## Problem
The Internals "composite" self-improvement score drifted ~0.66→0.57 over 8 weeks and looked like a cognitive regression. It was **~entirely a measurement artifact**: the composite (and the procedure dimension's `mean_confidence`) averaged the confidence of **all** non-deprecated procedures, and the store had been flooded with **331 speculative L4 candidates** (confidence ≈ 0, never invoked). That dragged the validated mean **0.71 → 0.096**, pulling the composite down ~0.10. Same family as the prior age-unfair "crater" artifact (#725). Underneath sat two real hygiene gaps: no cap on speculative creation, and no decay of never-used candidates.

## Changes
1. **Metric fix** (`eval/j9_aggregator.py`): average **validated** procedures only (`speculative=0`). The composite's standalone `AVG` adds `AND speculative=0`; the procedure dimension uses `AVG(CASE WHEN speculative=0 THEN confidence END)` so `total_procedures` still counts the whole store (consistent with `tier_distribution`). Fix-forward — the weekly snapshot self-heals the 12-week trend.
2. **Decay** (`learning/procedural/promoter.py` + `db/crud/procedural.py`): a decay arm in the hourly `promote_and_demote` soft-deprecates `speculative=1 AND (success_count+failure_count)=0 AND confidence<0.3 AND age>14d` via a **single bulk UPDATE** (`deprecate_stale_speculative`). Keyed on outcome counts (not the dead `invocation_count`); cutoff computed in Python and compared as ISO strings (version-independent, no SQLite `datetime()` dependency).
3. **Throttle** (`memory/extraction_job.py` + `procedure_extraction.py`): per-session cap (`max_procedures_per_session=3`, mirroring the existing `max_extractions_per_session`), shared across the extraction + struggle streams via a `max_new` budget.

## Verification
- Live data: composite procedure-signal `0.095 → 0.708`; projected 06-21 composite `0.566 → 0.685`.
- E2E (VACUUM INTO snapshot of the live DB): decay deprecates exactly the 2 currently-eligible rows, **0 wrongly-touched** (no validated / has-outcome / confident rows).
- 120 targeted tests pass (`test_eval`, `test_learning`, `test_db/test_procedural`). Code-reviewed; 2 findings addressed (bulk UPDATE; version-independent cutoff).

No migrations. The existing 331 age out gradually (~2 weeks); the metric is truthful immediately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **The Internals "composite" self-improvement score is no longer dragged down by draft
+  procedures** — Genesis extracts candidate procedures from its own sessions; these start
+  unvalidated (near-zero confidence) until they prove useful. The weekly composite score was
+  averaging *every* procedure's confidence, so a burst of new drafts made the score crater
+  even though nothing had actually regressed. The score now reflects only validated
+  procedures. Genesis also caps how many drafts a single session can create, so the
+  procedure store stops accumulating dead weight.
 - **Updates no longer abort when a schema migration actually succeeded** — if the
   database was busy during an update (for example a background task writing at the same
   time), a migration could commit successfully yet still surface a transient "database
--- a/docs/architecture/genesis-v3-procedure-activation.md
+++ b/docs/architecture/genesis-v3-procedure-activation.md
@@ -54,6 +54,12 @@ Demotion is **evidence-driven only** — never metric drift:
 - 3+ failure-mode hits AND failure_count >= success_count + 3 → tier - 1
 - confidence < 0.3 AND total samples >= 3 → quarantine (excluded everywhere)
 
+> **Eval note:** the J-9 system composite's procedure-confidence signal and the
+> procedure dimension's `mean_confidence` average **validated** procedures
+> (speculative=0) only. Speculative candidates start at conf≈0 and would measure
+> extraction *volume*, not knowledge *quality* (`total_procedures` still counts
+> the whole store, consistent with `tier_distribution`).
+
 The `_compute_tier` function in `promoter.py` is strict promote-only: it
 returns the highest tier the row's metrics qualify for, but never returns
 a lower rank than the row's current tier. A procedure whose confidence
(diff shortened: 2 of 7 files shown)


WingedGuardian:
Structural review: 2 NOTEs — no ERRORs or WARNINGs.

---

**NOTE 1 — `run_extraction_cycle` docstring not updated for new parameter**
`src/genesis/memory/extraction_job.py` — `max_procedures_per_session` was added to the function signature but the function's docstring doesn't describe it. `extract_procedures_from_chunk`'s docstring was updated for `max_new` (the downstream parameter); consistency would call for the same here.

---

**NOTE 2 — Struggle-stream throttle may not increment `session_procs` after storing**
`src/genesis/memory/extraction_job.py` — the extraction stream correctly does:
```python
session_procs += proc_count
```
after calling `extract_procedures_from_chunk`. The struggle-stream block adds the guard `and session_procs < max_procedures_per_session`, but no `session_procs += 1` appears in the diff after `judge_struggle_procedure` stores a procedure. Since `session_procs` is entirely new to this PR, any increment in the struggle block would have shown as a `+` line — its absence suggests the counter isn't updated there. If so, a session with multiple transcripts could store one struggle procedure per transcript beyond the cap. Worth a quick check; if the increment is in unchanged context further down the block, this is a false positive.

---

WingedGuardian:
**Force-pushed: dropped the age-decay arm per the no-heuristic-demotion principle.**

Per the governing decision that no procedure may be demoted/deprecated by a *heuristic* (only genuine outcome evidence or a user-approved content judgement), this PR's two decay commits were removed:
- ~~`e5ecbf70` feat(learning): decay stale unused speculative procedures~~ — dropped
- ~~`8c42fd20` refactor(learning): bulk decay UPDATE + version-independent age cutoff~~ — dropped

The decay code (`crud/procedural.py` decay fn, `promoter.py` wiring, `test_promoter.py`) is fully reverted to main, and the decay paragraphs were scrubbed from the CHANGELOG entry and `genesis-v3-procedure-activation.md`.

**Kept** — the two non-heuristic changes this PR is now scoped to:
- `a54719bd` exclude speculative procedures from the J-9 confidence signals (validated-only `mean_confidence` — a truthfulness fix, not a demotion)
- `9edb6997` (was `30c78c77`) cap speculative procedure extraction per session (an extraction-volume cap, not a demotion)

Net diff is now 7 files; 44 targeted tests green.

PR #769, opened 2026-06-24
feat(experimentation): Evo loop core + CC-CLI provider (recommend-only)

WingedGuardian:
## Context

On top of the merged Crucible foundation (#763 — the `experiment_run` run button +
the `~/.genesis/config/reflection/` prompt overlay). This adds **Evo**, the
evolution loop: generate N variants of the reflection prompt, A/B each against
the current prompt, and **recommend** the best. **Recommend-only, measure-only** —
it surfaces a recommendation and mutates nothing. The promotion/apply path (the
sensitive piece) is a separate, architect-gated PR-B, not here.

## What

- **`experimentation/evo.py`**
  - `build_directive_variants(base, n)` — variants by **directive-append** (`base +
    a fixed mutation directive`): deterministic, no generation LLM call, no
    truncation on the large prompt. The pattern the foundation E2E proved.
  - `run_evo(...)` — fan-out: A/B each variant vs control via the foundation's
    `run_reflection_experiment`; keep those beating control at **Bonferroni α/N**;
    pick the best survivor; **re-validate on a DISJOINT held-out golden slice**
    (gated `p ≤ α`; `holdout_disjoint` surfaced when the set is too small).
  - `persist_evo_summary(...)` — one `eval_runs` row per run (`kind=evo_summary`).
- **`experimentation/cc_router.py` — `CCCliRouter`**: runs gen/judge through the
  `claude` CLI (the CC subscription) when the litellm Claude providers (OpenRouter)
  are dry. Single-shot `claude -p` with tools, MCP, and hooks **disabled** for a
  clean completion; the timeout branch reaps the subprocess. Drop-in
  `StandaloneRoutingResult`.
- **`evo_run` MCP tool**: evolve the current effective deep-reflection prompt
  (overlay-aware), recommend the winner, persist the summary (surfaced by
  `experiment_status.evo_runs`). Providers `cc-haiku`/`cc-sonnet` route to the CLI.
  `autonomous_action=False`. **No promotion/apply.**

No migration, no new table (reuses `eval_runs`).

## Verification

- **Tests:** 16 hermetic for the new code (α/N gating, best-survivor selection,
  held-out rejection, directive variants, summary persistence + surfacing, the
  cc-router success/failure/timeout-kill paths, provider wiring). 76 experimentation
  tests green, ruff clean.
- **Real-score E2E (live, via CC CLI `cc-haiku`):** generation produces real
  reflections; the judge returns **real scores (0.74, 0.79)**; the full `evo_run`
[...]

PR #763, opened 2026-06-23
feat(crucible): foundation — experiment_run tool + reflection-prompt overlay

WingedGuardian:
## Context

The cognitive A/B experimentation harness ("Crucible", formerly "Phase 7",
`src/genesis/experimentation/`) was **built but unrun**: 197 tests, deployed,
yet **0 real experiments** (`eval_runs WHERE trigger='experiment'` = 0). It was
deliberately a *measurement library with no action surface* — (1) no way to
**run** an experiment, and (2) no clean **apply** target (the live deep-reflection
prompt is a markdown file in the git tree, with no settings/overlay knob).

This PR builds the **foundation** for the self-improvement loop ("Evo", next):
make Crucible runnable, and give it a clean writable knob to promote into. It
does **not** build Evo or any promotion path — recommend-only throughout.

## What

**1. `experiment_run` MCP tool** (`mcp/health/experiment_run.py`) — the run
button. Wires `run_reflection_experiment` so a control-vs-treatment
reflection-prompt A/B can be run + persisted to `eval_runs` (trigger=experiment),
then surfaced by the existing read-only `experiment_status`. Mirrors that tool's
service/DB pattern. **Recommend-only** (`autonomous_action: False`) — measures
and records, never promotes. Defaults gen/judge to a healthy free provider
(`groq-free`), caller-overridable (the runner module's stale `nvidia-nim-deepseek`
judge default is frequently down).

**2. Reflection-prompt overlay** (`cc/reflection_bridge/_prompts.py`) —
`system_prompt_for_depth` now checks a `~/.genesis/config/reflection/` overlay
(env `GENESIS_REFLECTION_PROMPT_DIR`) **before** the repo `prompt_dir`, then the
hardcoded fallback. A writable, out-of-git-tree target for hand-overriding the
reflection prompt (and the future Evo promotion), consistent with Genesis's
CAPS/overlay philosophy. **No overlay file → behavior identical to before** (a
live cognitive path — verified by the regression suite + a reviewer trace).

No migration. No new tables.

## Verification

- **Tests:** hermetic — `experiment_run` (param→variant mapping, db persistence
  pass-through, provider threading, missing-golden-set + no-db guards) +
  reflection-overlay precedence/fallback. 89 targeted tests green, ruff clean.
- **Live E2E:** a real `groq-free` A/B over the actual golden set ran the full
  pipeline (gen → judge → McNemar win-rate) and **persisted 2 `experiment`
  rows**; the overlay returned the override when set and the real 13.5KB repo
[...]

PR #774, opened 2026-06-24
feat(evo): human-gated reflection-prompt promotion (Evo PR-B)

WingedGuardian:
## Summary

Closes the Evo self-improvement loop: a measured reflection-prompt **winner** can now become an **approvable proposal** and, on the user's approval, be **applied to the live deep-reflection overlay — reversibly**. Builds on the Crucible foundation (#763) and the Evo measure-only loop (#769).

**Recommend-only / human-gated throughout.** Genesis recommends; the user decides. Nothing touches live cognition until a human approves a proposal, and any applied change is one rollback away from reverting.

## The loop

`evo_run` measures variants of the **canonical** deep-reflection prompt (Bonferroni α/N + disjoint held-out re-validation). When a winner is confirmed, it is **auto-filed** as a `cognitive_variant_promotion` proposal (dashboard-delivered). On approval, `handle_cognitive_variant_resolution` writes the winning prompt to the `~/.genesis/config/reflection/REFLECTION_DEEP.md` overlay via the rollback-able cognitive ledger.

## Recommend-only = 3 independent defense layers

1. **Apply-at-approval** — the resolution handler applies ONLY on `approved`, and marks the proposal `executed` **unconditionally** (even if the write fails), so it never lingers `approved` for the dispatch sweep to grab.
2. **Dispatch blocklist** — a single-source `_NEVER_DISPATCH_ACTION_TYPES` tuple guards **both** auto-dispatch paths in `ego/session.py` (the approved-sweep and the execution-briefs path, which previously had no action_type guard).
3. **Autonomy domain gate** — the action_type maps to `SELF_MODIFY` in `autonomy/classification.py` (hard-blocked from background dispatch). Without this it fell through to `EXTERNAL_READ` (always allowed) — the dispatch gate would have been a passthrough.

## Design decisions

- **Canonical base, not overlay** — variants are built against the canonical repo prompt and the filed winner is exactly `canonical + winning_directive`, so *what is measured equals what is applied* and repeated promotions never stack directives.
- **Dashboard-delivered** — the proposal is stored and approvable on the dashboard Ego tab (the apply hook is wired into all 4 resolution sites). A Telegram ping belongs with the deferred autonomous cadence.
- **0.75 confidence floor** (`1 − held-out p`), enforced at file time and re-checked in the handler.
- **Self-scoring caveat** — when generator and judge share a provider, the proposal rationale says so, so the user can discount optimistic scores.
- The 13KB winner prompt rides in `expected_outputs` (verified NOT rendered in the digest or the dashboard proposal API); `content`/`rationale` carry a human summary.

## Verification

- **genesis-architect review: recommend-only AIRTIGHT** — confirmed an approved promotion cannot reach session dispatch via any path, and the handler cannot write on a non-approved status.
- **Regression green**: autonomy 801 / ego 711 / experimentation 56 passed; ruff clean; all modules import.
- **End-to-end** (real functions, no apply-path mocks): winner → file → approve → overlay live → rollback reverts. Both no-dispatch paths proven behaviorally.
- **Real-provider E2E** (cc-haiku gen + judge): the pipeline produced real scores (mean ~0.73) and correctly filed nothing at small N where no winner is statistically confirmable.

## Known caveat (pre-existing, not introduced here)

The autonomy `_proposal_gate` is fail-open on exception in both dispatch paths. This is harmless for this action_type because the blocklist fires first, but it is a latent weakness for any future type relying solely on the gate. Out of scope for this PR.

## Deferred follow-on

Autonomous Evo→propose cadence (its own cost model + `CronTrigger`); weight-variant evolution; agentic experiment runner.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,15 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ## [Unreleased]
 
+### Added
+
+- **Genesis can now propose improvements to how it reflects — and apply them only with your approval.**
+  The Evo loop measures variations of the deep-reflection prompt against a golden set
+  (with held-out re-validation), and when one is a confirmed improvement it files a
+  proposal on the dashboard. Approving it updates the live reflection prompt; the change
+  is fully reversible (one click rolls it back). Nothing is ever applied automatically —
+  Genesis recommends, you decide.
+
 ### Changed
 
 - **The dashboard Infrastructure card now labels the ambient-capture bridge as "Voice Bridge"** —
--- a/src/genesis/autonomy/classification.py
+++ b/src/genesis/autonomy/classification.py
@@ -288,6 +288,11 @@ ACTION_TYPE_DOMAIN_MAP: dict[str, ActionDomain] = {
     "payment": ActionDomain.FINANCIAL,
     "code_change": ActionDomain.SELF_MODIFY,
     "refactor": ActionDomain.SELF_MODIFY,
+    # Evo promotion rewrites Genesis's own deep-reflection prompt — a change to
+    # its cognition. SELF_MODIFY is hard-blocked from background dispatch
+    # (ACTION_DOMAIN_MIN_LEVEL = None), so an approved promotion can never be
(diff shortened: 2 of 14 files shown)


WingedGuardian (commented):

---

PR #782, opened 2026-06-25
refactor(learning): rename procedure `speculative` column to `draft`

WingedGuardian:
## What

Renames the `procedural_memory.speculative` column to `draft` — and every coupled Python identifier — in **procedure context only**. A procedure's unproven state isn't "speculative" (the term Genesis reserves for observations and hypothesis-claims); it's an untested **draft** destined for real use.

`observations.speculative` and `speculative_claims.speculative` are a genuinely different concept and are **intentionally left unchanged** (user-confirmed vocabulary divergence).

This is the `speculative→draft` half (PR-C) of the procedure-vocabulary cleanup. Tier rename (CORE/ADVISORY/LIBRARY/DORMANT) shipped earlier; a dashboard procedures viewer (PR-D) follows.

## How

`SELECT *` couples the returned row-dict key to the column name, so the column rename **forces** renaming every `row["speculative"]` read and `speculative=` kwarg in lockstep — a miss is a loud `KeyError`/`OperationalError`, never silent.

- **Migration `0037`** — `ALTER TABLE … RENAME COLUMN` + index rename (`idx_procedural_speculative` → `idx_procedural_draft`). Pure rename, no value rewrite; idempotent (skips if already `draft` or base table absent); leaves the sibling tables untouched.
- Renamed across CRUD, extractor, judge, promoter, executor trace, the J9 VALIDATED filter (`draft = 0`), the seed script, and the `procedure_store` MCP tool.
- Internal counter `despeculated` → `drafts_cleared`; architecture docs updated.

## Verification

- **TDD migration test** (5 cases): rename + data preservation, index rename, leaves observations/claims alone, idempotent re-run, base-table-absent skip.
- **Targeted procedure surface**: 234 pass. **Broad regression** (`test_learning/eval/db/mcp`): **1873 passed, 3 skipped**. `ruff` clean.
- **E2E on a real 437-procedure DB copy** applied via the migration runner: `draft` distribution `{0:33, 1:404}` preserved exactly, index renamed, observations/`speculative_claims` columns untouched, idempotent, recorded in `schema_migrations`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #784, opened 2026-06-25
fix(db): retry migrations on SQLITE_LOCKED so deploys survive concurrent readers

WingedGuardian:
## Problem

Deploying PR #782 (migration `0037`, an `ALTER TABLE RENAME COLUMN`) **failed twice tonight with `database is locked`** and auto-rolled-back — even with all other CC sessions idle. 30+ prior migrations had succeeded with the same concurrency, so concurrency wasn't the differentiator.

Root cause: a schema migration's DDL needs an exclusive schema lock. When a concurrent reader holds the table schema, SQLite raises `database is locked` (`SQLITE_LOCKED`) — a class `busy_timeout` does **not** wait on. So a single-shot apply fails the instant a brief *periodic* writer is mid-transaction. Here that writer is the **outreach scheduler** (every ~5 min, churning a stuck retry backlog). The migration's one shot kept landing on that ~5-min lock window; the lock is otherwise free ~100% of the time.

On a busy multi-session install, *every* future schema migration is exposed to this coin-flip.

## Fix

The runner now **retries the whole atomic apply on a lock error** — safe because a failed migration is fully rolled back (`BEGIN..COMMIT` envelope) — with backoff (10 attempts, 2–10s) until it catches a free window.

- The **reconcile path** (commit landed, then post-commit `SQLITE_BUSY` from a WAL autocheckpoint — the #744 fix) is checked **before** the retry decision, so a committed migration is never re-run.
- **Non-lock errors still fail fast** (no retry) — a real migration bug isn't masked.

## Verification

- **34 runner tests** pass, incl. 4 new: retry-fail-then-succeed, give-up-after-max, no-retry-on-non-lock-error, and reconcile-wins-over-retry. All existing atomicity / proxy / reconcile behavior preserved.
- **E2E**: applied `0037` to a live-DB copy while a background thread held a 4s write lock — the runner logged `retry 1/10 → 2/10 → 3/10` and succeeded the instant the lock released (3.7s elapsed; `draft` column present). Proves it retries past real contention.

Unblocks the PR #782 (`speculative`→`draft`) deploy.

### DIFF
--- a/src/genesis/db/migrations/runner.py
+++ b/src/genesis/db/migrations/runner.py
@@ -26,9 +26,11 @@ state. On failure: rollback, stop, remaining migrations not attempted.
 
 from __future__ import annotations
 
+import asyncio
 import importlib
 import logging
 import re
+import sqlite3
 import time
 from dataclasses import dataclass
 from datetime import UTC, datetime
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

PR #785, opened 2026-06-25
fix(outreach): stop the self-send follow-up email spam loop

WingedGuardian:
## Problem

Background follow-up sessions were flooding the approvals tab with held "cold email to the agent's own address" messages — a new one every ~5 minutes. All were correctly **held** at the WS-8 capability gate (none ever sent), but the queue kept regenerating them, and the churn periodically locked the DB (it broke a recent deploy twice).

## Root cause — a 4-layer self-send loop

1. **Origin:** follow-up/reply sessions run as subprocesses where the outreach MCP has `pipeline=None`. `outreach_send(thread_id, channel="email")` fell back to `pending_outreach.enqueue()`, which **dropped the `thread_id` and recipient** (no such columns) — even though the tool's docstring says `thread_id` routes the recipient.
2. **Default-to-self:** on drain, `_deliver` resolved `recipient = validated_recipient or recipients["email"]`, and `recipients["email"]` defaulted to the agent's **own** address.
3. **No self-guard:** the email gate classified the self-send (`email/send/identity`, ASK) → **HELD** → wrote `pending_email_sends` + a pending approval.
4. **Multiplier:** the drain only marked the queue row delivered on `DELIVERED`/`ENGAGED`; **HELD and REJECTED both retried forever** → a new held self-email every cycle.

## The fix (defense-in-depth)

- **Origin** — Migration `0038` adds nullable `thread_id` + `validated_recipient` to `pending_outreach`; `enqueue()`/`ensure_table()`/canonical schema carry them; the `outreach_send` fallback resolves the thread recipient (hoisted above the pipeline/fallback split) and enqueues both.
- **Chokepoint** — `_deliver` drops an email addressed to the agent's own `from_address`, or with no resolved recipient, as terminal `IGNORED` — **before** the gate (which would HOLD it) and **before** the no-recipient defer (which would retry-loop). Fires on the approved-resume path too. The self-address default is removed; email recipients now come from the thread or `OUTREACH_RECIPIENT_EMAIL`.
- **Multiplier** — the drain treats `HELD` (handed to the approval queue) and `IGNORED` as terminal, and drops any row stuck >24h.
- **Watcher** — an approved hold the pipeline terminally skips is marked rejected + consumed (no busy-loop, no ghost approval).
- **Observability** — governance denial reasons now surface in the result (retry logs were blank).

## Testing

- TDD throughout (every behavioral change watched fail first). New tests across migration, CRUD, MCP fallback, scheduler drain, pipeline guard, and watcher.
- Full consolidated regression green; lint clean on all changed files.
- **Code review** (agent): confirmed correctness on 10 checklist items; 1 P1 + 2 P2 found and all fixed + tested.
- **E2E on a copy of the live DB:** the real migration runner applied only `0038` (no failures), recorded it, preserved data, idempotent re-run — and a round-trip enqueue/drain routed to the real recipient, not self.

The 83 held self-emails + their approvals were already cleared from the live DB (approvals cancelled to preserve the tamper-evident audit chain; the rest hard-deleted).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   even though nothing had actually regressed. The score now reflects only validated
   procedures. Genesis also caps how many drafts a single session can create, so the
   procedure store stops accumulating dead weight.
+- **Genesis no longer floods its own approvals with follow-up emails addressed to itself.**
+  A follow-up drafted in a background session could lose its thread (and therefore its
+  recipient) on the way to the outreach queue, then fall back to Genesis's own email address.
+  The capability gate correctly held each one for approval, but because the queue kept
+  retrying, a new held "email to myself" piled up every few minutes. Genesis now keeps the
+  thread recipient with the queued message, never sends an email to its own address, and
+  treats a held or undeliverable message as resolved instead of retrying it forever. A
+  message that can never be delivered is now dropped after a day rather than looping.
 - **Updates no longer abort when a schema migration actually succeeded** — if the
   database was busy during an update (for example a background task writing at the same
   time), a migration could commit successfully yet still surface a transient "database
--- a/src/genesis/autonomy/email_gate_watcher.py
+++ b/src/genesis/autonomy/email_gate_watcher.py
@@ -99,8 +99,12 @@ async def drain_pending_email_sends(rt: object) -> int:
                 # The pipeline terminally SKIPPED this approved send (self-
                 # addressed / no recipient). It is not deliverable and not an
                 # error — mark the hold rejected so it can't busy-loop every
-                # cycle. No correction (the skip is a guard, not owner intent).
+                # cycle, and consume the approval so it doesn't linger as a ghost
+                # approved/unconsumed row. No correction (a guard, not owner intent).
                 if await pes.mark_rejected(db, row["id"], rejected_at=now):
+                    await approval_crud.mark_consumed(
+                        db, row["request_id"], consumed_at=now,
+                    )
                     resolved += 1
                 logger.warning(
                     "Held email %s skipped by pipeline (IGNORED) — marked terminal",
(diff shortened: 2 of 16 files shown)


WingedGuardian:
## First-pass review — PR #785

### Structural review

**Pattern violations (ERROR):** None found.

**Import hygiene (WARNING):** Clean — no circular imports, star imports, or grouping issues in added code.

**Missing coverage (WARNING):** Clean — new `src/genesis/db/migrations/0038_pending_outreach_thread_recipient.py` has a corresponding `tests/test_db/test_migration_0038_pending_outreach_thread_recipient.py`. All modified source functions have test coverage added.

> `src/genesis/db/migrations/0038_pending_outreach_thread_recipient.py` — `up()` has no `await db.commit()` after the `ALTER TABLE` statements. The CRUD sibling `ensure_table()` (in `pending_outreach.py`) explicitly calls `await db.commit()` after its DDL. If the migration runner doesn't commit externally after calling `up()`, the schema changes won't persist. Worth confirming the runner's commit semantics — or adding an explicit `await db.commit()` at the end of `up()` to make the migration self-contained.

**Genesis conventions (WARNING):** Clean. Commit prefix `fix(outreach):` is correct. No nohup, no IntervalTrigger, no `$CLAUDE_PROJECT_DIR`.

---

**Code diff — hard blocks:** None.

**Code diff — soft warnings:**

> `tests/test_outreach/test_pipeline.py` (three new test functions) — `self_addr = "genesisagiagent@gmail.com"` is hardcoded as the test fixture value for the agent's own address. This appears to be the real operational email address inferred from `gmail_addr` in `runtime/init/outreach.py`. Consider substituting `agent@example.com` (or similar test-domain address) in these stubs to avoid a real operational address appearing in version-controlled test code.

**PR title / body / CHANGELOG wording:** CLEAN — the description explains the mechanism at a functional level (self-send loop, queue retry, approval flood) with no private IP addresses, hostnames, credential details, or wording that would constitute a hard block.

---

PR #788, opened 2026-06-26
fix(hooks): strip auto-injected GitNexus block from staged CLAUDE.md

WingedGuardian:
## Problem

The `post-commit` hook runs `npx gitnexus analyze`, which re-injects a
`<!-- gitnexus:start -->…<!-- gitnexus:end -->` block into **both** `CLAUDE.md`
and `AGENTS.md` on every commit (the tool has no per-file flag). The block is
meant to live only in `AGENTS.md`. The hourly `run_gitnexus_strip` surplus job
cleans `CLAUDE.md`'s working tree, but there's always a window between the
post-commit re-inject and the next strip — and a commit (or a worktree cut) in
that window captures the block. That's how the block has been leaking into
feature branches.

## Fix

Strip the block in `pre-commit`, at the git index, where it can never reach a
commit regardless of timing:

- Operates on the **staged blob** (`git cat-file blob :CLAUDE.md` → strip →
  `git hash-object -w` → `git update-index --cacheinfo`), so unrelated
  working-tree edits to `CLAUDE.md` are untouched.
- **Anchors** the start marker to column 0 (`re.MULTILINE`) so authored prose
  that merely mentions the markers mid-line is never matched.
- Uses `surrogateescape` so a non-UTF8 byte stays a lossless no-op instead of
  blocking an unrelated commit.
- Mirrors the strip to the working tree so the file doesn't re-appear as
  modified right after the commit.
- New hook hash recorded in `.genesis-hook-versions` (enforced by
  `check_hook_versions.sh`).

The hourly `run_gitnexus_strip` job remains the backstop for `--no-verify`
commits and fresh clones before `install.sh` runs.

## Testing

Manually exercised the installed hook against: a normal injected block (stripped
from both index and working tree), a real `CLAUDE.md` edit committed alongside a
block (edit preserved, block stripped), authored prose mentioning the markers
mid-line (preserved — anchoring), and a clean/un-staged `CLAUDE.md` (no-op,
silent). Adversarially reviewed for index-rewrite correctness, regex
over/under-match, and failure-mode signaling.


WingedGuardian:

---

PR #796, opened 2026-06-27
feat(cc): pluggable DirectSession profile overlays

WingedGuardian:
## What

Adds an optional extension point so a deployment can register its own
background-session (DirectSession) profiles — including Bash-scoped ones —
without editing the shipped `direct_session.py`.

If a local `genesis/cc/profile_overlay.py` is present, `_load_profile_overlays()`
hands it a `ProfileOverlayContext` that exposes the same building-block
disallow lists the built-in profiles use, plus an `add_profile()` method.
When the module is absent (the default), the loader is a no-op and the built-in
profiles are unchanged.

## Why

Profiles were a hardcoded dict, so any install-specific profile forced an edit
to tracked code. This mirrors how install-local modules already plug in: the
generic mechanism ships here; the install-specific profile definitions stay
local.

## Behaviour / safety

- `add_profile()` refuses to redefine a built-in profile (an overlay may only
  add), and the loader surfaces such config errors loudly while isolating
  genuine runtime failures so a broken overlay can't block session spawning.
- `_PROFILE_TO_MCP` is hoisted to module scope so an overlay can map its own
  MCP server set.
- Note documented in `background-sessions.md`: allowlisting an interpreter pins
  only a command's first token, so an interpreter-scoped overlay relies on its
  addendum for any "only run module X" restriction — appropriate for trusted
  internal sessions, not untrusted input.

## Also in this PR

- `user_job_create` now validates `profile` against the live `VALID_PROFILES`
  instead of a stale hardcoded `{observe, interact, research}` set — it was
  wrongly rejecting `campaign`/`steward`/`mail`/`discord-monitor` jobs.
- Fix a `campaign_create` docstring that implied `profile` was an enum of those
  three.

## Tests

16 hermetic overlay-mechanism tests (registration, override guard, broken-overlay
resilience, no-op when absent, collision surfacing, invocation flow). Suite
passes both with and without an overlay present.


WingedGuardian:
Review note: independent Codex review was unavailable this cycle (Codex code-review usage limit reached). Reviewed instead via the project's architect reviewer across two passes (overlay mechanism + security posture); all findings addressed (loader surfaces overlay misconfig loudly; user_job profile validation fixed). Merging on green CI per the quota-limited-review path.

PR #797, opened 2026-06-27
feat(learning): track procedure surfaced_count for honest self-learning funnel

WingedGuardian:
## Problem

Genesis surfaces a learned procedure into context two ways without an explicit recall: the proactive memory hook (on a prompt) and the PreToolUse advisor (on a tool call). Neither was tracked. Only an explicit `procedure_recall` writes `invocation_count`, so the self-learning health funnel (`loop_closure_status`) couldn't see contextual surfacing at all — procedures that *were* reaching the model read as a leak ("captured but nothing flowing — loop OPEN"), even when they were being shown every relevant turn.

## Change

- Add a `surfaced_count` column to `procedural_memory` (migration `0039` + canonical schema; retrofits existing installs).
- Both contextual hooks bump it via a fire-and-forget write **after** the advisory/injection is delivered, so there's no added latency on the user prompt or tool call.
- `procedure_funnel` now reports `captured → surfaced → invoked → reached → measured`, where `reached` is the de-duped union (surfaced **or** invoked). A procedure that has neither been surfaced nor invoked is the real leak (`leak_never_reached`).

## Safety: surfacing never promotes

`surfaced_count` is observability **only**. The procedure promoter reads `invocation_count` exclusively, so passively showing an unproven draft can never inflate its standing or promote it a tier. The two writers bump `surfaced_count` alone — never `invocation_count` or `confidence`. This boundary is asserted directly in tests and was verified end-to-end against production-shaped data (a surface bump moved `surfaced_count 0→1` while `invocation_count` stayed `0`).

## Testing

- New + updated unit tests cover: the funnel's surfaced/invoked/reached/leak math and overlap de-dup; the promotion-inert property in both hooks; idempotent re-surface; missing-id no-op; fail-open on a bad DB; and an end-to-end advisor `main()` run (bump only on a match, only for triggers carrying a procedure id).
- Migration retrofit + idempotency verified directly (existing table without the column → added; re-apply is a no-op; bare DB guarded).
- Full `test_db` + `test_learning` suites green; ruff clean.

## Scope

Contextual surfacing only (proactive hook + tool advisor). Blind session-start injection of always-on procedures is intentionally excluded. Explicit `procedure_recall` continues to be tracked as `invoked`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,7 +56,12 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   ever presenting it as settled guidance. Blind session-start injection still stays
   limited to the most-proven, always-on procedures. Genesis also now repairs
   procedures that were missing their embedding, so they stop being silently
-  invisible to this relevance matching.
+  invisible to this relevance matching. Genesis also now counts each time a
+  procedure is surfaced into context this way, so its own self-learning health
+  check reports learned procedures honestly as *reaching* its attention rather
+  than falsely flagging them as lost — and this surfacing count is kept strictly
+  separate from the signals that promote a procedure, so merely showing a draft
+  can never inflate its standing.
 
 - **Procedures you actually use now earn their keep.** How often a learned
   procedure is recalled ("reads") now counts as a dampened usefulness signal:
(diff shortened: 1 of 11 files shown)


WingedGuardian:
Automated Codex review was quota-limited (no review ran). Merging on green CI plus manual architect + adversarial code review (both clean; all findings addressed) and an end-to-end check against a copy of production data.

PR #798, opened 2026-06-28
fix(bootstrap): seed baseline procedures at install; de-track generated trigger cache

WingedGuardian:
## Problem

Genesis ships a set of battle-tested **baseline procedures** (use yt-dlp for YouTube, prefer `PYTHONPATH` over an editable install in worktrees, validate a pgid before `os.killpg`, keep large temp out of `/tmp`, etc.) plus a PreToolUse advisor that surfaces the relevant one as you work. The seeding script that installs them (`scripts/seed_procedures.py`) was **wired into nothing** — not bootstrap, not install, not runtime. So:

- Fresh installs ran **without** their baseline procedures in the database.
- The shipped advisor trigger cache (`config/procedure_triggers.yaml`) referenced procedure ids that didn't exist as DB rows — so any DB write-back keyed on them (usage/outcome tracking) silently did nothing.

## Change

- **bootstrap.sh** now seeds the baseline procedures after the venv is ready (idempotent via deterministic ids), and surfaces a clear warning if seeding doesn't complete instead of failing silently.
- **seed_procedures.py** resolves the database through the app's own path resolver (honoring `$GENESIS_DB_PATH` / `$GENESIS_REPO_ROOT`) rather than a hardcoded `~/genesis` path, so non-standard install layouts seed the database the server actually reads.
- **De-track `config/procedure_triggers.yaml`** — it's install-specific generated state (regenerated per-install from the database by the seed script and the procedure promoter), so it's now gitignored instead of shipped as a stale committed file.
- **update.sh**: discard local edits to known-ephemeral tracked files (the trigger cache, `AGENTS.md`) immediately before merging, so de-tracking a file that's rewritten in place at runtime doesn't abort the self-update with "local changes would be overwritten." This completes the existing `EPHEMERAL_DIRTY_RE` guard, which until now excused these files from the pre-merge check but left the merge itself to fail on them.

## Testing

- Seeding verified idempotent (re-running adds nothing) and correct against production-shaped data; the baseline procedures land at their intended tiers and the advisor cache is regenerated to match the DB.
- The `update.sh` pre-merge discard was verified by simulating the exact failure: a dirty (unstaged, staged, and clean) tracked cache plus an incoming commit that de-tracks it — without the fix the merge aborts; with it the merge is clean in all cases.
- Bootstrap seed step verified for both success (no warning) and failure (warning, non-fatal) paths; `bash -n` clean on both scripts.

## Note for existing installs

Because a self-update script can't patch the *currently running* copy of itself, an existing install pulling this change runs its old `update.sh`, which may hit a one-time merge abort on the de-tracked cache and roll back. Re-running the update after the new `update.sh` is in place (or a one-time `git checkout -- config/procedure_triggers.yaml`) resolves it; every subsequent update is clean.


WingedGuardian:
Automated Codex review was quota-limited (no review ran). Merging on green CI plus two manual adversarial reviews (bootstrap/db-path/visibility and the update.sh self-update path; both clean, all findings addressed) and local + production-shaped verification.

PR #805, opened 2026-06-30
feat(feedback): harvest surplus_tasks into the Outcome Bus as tier-1 ground truth

WingedGuardian:
## What

The Outcome Bus ledger held only tier-3 *coverage* signals — **zero tier-1 execution ground truth** — because only `ego_proposals` and `outreach_history` were harvested into it, and both are currently low-yield. This adds the richest unharvested execution-outcome source, **`surplus_tasks`** (Genesis's background autonomous work queue), as a tier-1 producer, so the self-improvement signal actually accrues real ground truth.

## Changes

- **New `_harvest_surplus_tasks` source** in `feedback/harvest.py`, mirroring the existing harvesters: a terminal `surplus_tasks` status of `completed` → `EXECUTION_OUTCOME` (positive, value 1.0); `failed` → negative (value 0.0) carrying `failure_reason`. `domain = task_type`, `source = "surplus"`, `ref_id = id`. Registered in both `run()` (incremental window) and `run_backfill()` (one-shot history).
- **Versioned backfill marker.** The one-shot backfill is guarded by a single `ego_state` marker. Adding a source after that marker is set would otherwise never backfill the new source's history. The marker now carries a schema version (`v{N}:{iso}`; a legacy bare-ISO value is treated as v1), so a version bump re-runs the backfill exactly once on existing installs. Re-running is safe — every source is idempotent on the `(source, ref_type, ref_id, signal_type)` unique key, so already-harvested rows are `INSERT OR IGNORE` no-ops.

## Behaviour

**Behaviour-neutral — the bus stays read-only/"dark".** `calibration_pairs` is source-filtered (ego only) and `surplus` rows carry a NULL stated-confidence, so they never enter ego calibration; `aggregate_by_domain` is read-only observability. No cognitive path consumes the new rows. This is purely producer-side: it makes the ledger honest, it does not change any decision.

## Tests

`tests/feedback/test_harvest.py` — new `TestHarvestSurplusTasks` (completed→T1 positive, failed→T1 negative + reason, non-terminal/cancelled skipped, idempotent on the unique key, incremental-window vs backfill) and `TestVersionedBackfillMarker` (version parsing incl. malformed-value clamping, legacy marker triggers a one-time re-backfill and re-stamps, current-version skips, and a reliability guard that a surplus failure must not lock the backfill gate). Full `tests/feedback/` suite green.

### DIFF
--- a/src/genesis/feedback/harvest.py
+++ b/src/genesis/feedback/harvest.py
@@ -24,6 +24,11 @@ Source authority (no double-counting):
 
 Each source read is wrapped independently (``_safe``) so one bad source can never
 abort the others.
+
+Note: this module is a read-only ETL pipeline — its per-source ``SELECT``s are
+intentional raw SQL (dynamic, filtered reads that don't map to a standard crud
+API), NOT a crud-layer violation. The only write path is ``record_outcome`` via
+``genesis.feedback.bus``; the authoritative source tables are never mutated.
 """
 
 from __future__ import annotations
--- a/src/genesis/feedback/harvest.py
+++ b/src/genesis/feedback/harvest.py
@@ -39,6 +39,14 @@ from genesis.feedback.bus import SignalType, record_outcome
 logger = logging.getLogger(__name__)
 
 BACKFILL_MARKER = "outcome_backfill_done"
+# Bumped whenever a NEW harvest source is added AFTER installs have already set
+# the marker — forces a one-time re-backfill so the new source's history lands.
+# The stored value is "v{N}:{iso}"; a bare ISO (no "v" prefix) is the legacy v1
+# stamp (proposals + outreach only). Re-running is safe: every source is
+# idempotent on the (source, ref_type, ref_id, signal_type) unique key, so
+# already-harvested rows are INSERT-OR-IGNORE no-ops (read I/O, no WAL write).
+# v2 (2026-06-29): added the surplus_tasks source.
+BACKFILL_VERSION = 2
 
 # Markers update_proposal_outcome appends to ego_proposals.user_response.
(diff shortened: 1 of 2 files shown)


WingedGuardian (commented):

### Structural

**WARNING — Raw SQL pattern in `_harvest_surplus_tasks` is consistent with pre-existing harvesters but lacks a documented crud exemption**

`src/genesis/feedback/harvest.py`

All three harvesters (`_harvest_proposals`, `_harvest_outreach`, and the new `_harvest_surplus_tasks`) use `self._db.execute(sql, params)` directly for their SELECT reads rather than routing through a `genesis.db.crud` module. The new method is internally consistent with the established file pattern, and all raw SQL here is SELECT-only — the only write (`record_outcome`) goes through `genesis.feedback.bus`.

The project rule is "raw SQL against genesis.db should use crud modules, except in migrations." The harvest pipeline is arguably a special case (dynamic filtered SELECTs that don't map to a standard CRUD API), but that exception isn't documented anywhere in the module. Two options:

1. Add a note to the module docstring stating that `harvest.py` is a read-only ETL pipeline with intentional raw SQL reads, so future reviewers don't re-flag it.
2. Track a follow-up to add a `surplus_tasks` read accessor to the crud layer and refactor all three harvesters at once (beyond the scope of this PR).

Not blocking this PR on a pattern that predates it and is shared across all harvesters in the file.

---

### PII / Secrets / Wording — CLEAN

No API keys, tokens, private IP addresses, personal email addresses, private hostnames, hardware identifiers, or other sensitive data found in added lines. PR title and description contain no language referencing security incidents, credential leaks, or private infrastructure details.

---

Structural review: 1 WARNING (raw SQL pattern), no ERRORs.

---

PR #807, opened 2026-06-30
feat(ego): optional Outcome Bus capability feed behind a default-OFF flag

WingedGuardian:
## What

Adds an optional sixth source to the ego's capability map — tier-1 execution
ground truth from the Outcome Bus (`outcome_events`, `source='surplus'`) — behind
a new config flag `outcome_bus_capability_feed` that is **OFF by default**.

While the flag is off (the default), the capability map is computed exactly as
before from its five existing sources, so behaviour is unchanged. When an
operator enables it, recent per-domain execution success rates are folded into
the ego's self-model. The capability map is display-only — it shapes the
performance summary the ego sees about itself in its own context; no control
flow keys off it.

## Why

The Outcome Bus already records real tier-1 execution outcomes, but nothing
consumed them yet. This wires the first, safest consumer behind an explicit
toggle, so the connection can be enabled deliberately rather than implicitly.

## Details

- `EgoConfig.outcome_bus_capability_feed: bool = False` (+ `config/ego.yaml`
  entry and a bool validator) — live-read each capability-map refresh.
- `aggregate_by_domain(..., source=)` — a new optional, backward-compatible
  filter so the consumer reads a single producer's domains and never
  double-counts producers the aggregator already includes by other means.
- The sixth source applies a sample-size floor (n >= 3) consistent with the
  existing session source, and contributes a success rate in [0, 1].

## Tests

New unit tests for the source filter and for the aggregator in both flag states
(off = output identical to the five-source baseline; on = the new domains appear
with correct rates, the noise gate holds, and it composes with existing
sources). Full targeted ego/db suites pass.

### DIFF
--- a/config/ego.yaml
+++ b/config/ego.yaml
@@ -46,3 +46,11 @@ failure_backoff_minutes: 60  # pause after N consecutive failures
 # so it can self-correct (informational only, never a mechanical rescale).
 # Live — takes effect next cycle. Set false to turn off instantly.
 calibration_injection_enabled: true
+
+# LC3-B go-live gate (default OFF): fold Outcome Bus tier-1 execution ground
+# truth (source='surplus') into the capability map as a 6th source. While false,
+# the capability map is unchanged (original 5 sources). capability_map is
+# display-only, so enabling this only adjusts the self-performance numbers the
+# ego sees in its context — no control-flow gate keys off it. Live — takes
+# effect next capability-map refresh.
+outcome_bus_capability_feed: false
--- a/config/ego.yaml
+++ b/config/ego.yaml
@@ -52,5 +52,7 @@ calibration_injection_enabled: true
 # the capability map is unchanged (original 5 sources). capability_map is
 # display-only, so enabling this only adjusts the self-performance numbers the
 # ego sees in its context — no control-flow gate keys off it. Live — takes
-# effect next capability-map refresh.
+# effect at the next capability-map refresh (9:15/21:15), no restart needed.
+# NOTE: settings_update marks the whole ego domain "restart required"; that is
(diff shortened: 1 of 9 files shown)


WingedGuardian (commented):
## Code Review — PR #807

### Structural Review

**WARNING — Missing test for new `validate_ego_config` branch**

`src/genesis/ego/config.py`: `validate_ego_config` was extended with a new `outcome_bus_capability_feed` boolean-type check, but no corresponding test case for this validation path appears in the PR (neither an existing test updated nor a new one added). If `tests/test_ego/test_config.py` or equivalent covers the other validation branches, it should get a case for `"outcome_bus_capability_feed must be a boolean"` — specifically: a non-bool value (e.g. `"yes"`, `1`) should produce that error string. The production code is obviously correct (it mirrors the `calibration_injection_enabled` branch exactly), but the path is untested in this diff.

No other structural violations found. All `except` clauses are typed (`except Exception:`), no mutable default arguments, no raw SQL against `genesis.db`, no `/tmp/` usage, no `nohup`, no `task_states` direct inserts, no star imports, no circular import risks, conventional commit prefix present.

---

**CLEAN** — No API keys, tokens, private IPs, private hostnames, personal email addresses, or hardcoded credential paths found in added lines. PR title and description contain no language referencing security incidents, credential leaks, or private infrastructure.

---

WingedGuardian:
Addressed the review WARNING in `7bcd4244`: added `validate_ego_config` boolean-branch tests — `outcome_bus_capability_feed` reject-non-bool (incl. the `1`/`0` that `isinstance(_, bool)` correctly excludes) + accept-bool, plus the sibling `calibration_injection_enabled` branch which was equally untested. CI green.

PR #808, opened 2026-06-30
feat(cc): model diversification — run non-Anthropic models behind Claude Code

WingedGuardian:
## What

First-class **model diversification**: Genesis can run on a curated roster of models behind the Claude Code harness, with automatic failover during a Claude outage and intentional per-session model selection. Default behavior is unchanged — routing only activates when a roster model is configured.

## How it works

- **Roster policy layer** (`cc/roster.py`, `config/cc_roster.yaml`) — maps a roster name (e.g. `glm-5.2`) to the environment that points a CC subprocess at a provider's native Anthropic-compatible endpoint. Auth tokens are resolved from the environment by name; never stored in config.
- **Chokepoint selection** (`CCInvoker`) — `roster.apply_active` resolves the active model at the single invoker chokepoint, gated by an opt-in `roster_eligible` flag. It never raises: any failure degrades to native Claude.
- **Resume continuity** — a routed session persists its endpoint (auth-env *name* only, never the token) and resumes on the same provider.
- **Conversation failover (sticky)** — on an account-wide Claude rate-limit, the turn runs on a roster peer **with full tools** before the degraded contingency path, and the peer session is resumed across consecutive outage turns. A successful Claude turn auto-recovers. Switch/recovery notifications via the outreach pipeline.
- **DirectSession selection** — background sessions can be dispatched on a chosen roster model (`roster_model`); an unknown or unconfigured model fails loudly rather than silently downgrading.

## Safety

- Default `cc_roster.default: claude` → behavior is byte-identical to today until a roster model is configured.
- Routing removes the Anthropic API key from the subprocess environment so it never leaks to a third-party endpoint.
- Failover is wrapped so any error falls through to the existing contingency path.

## Testing

- Full CC unit suite passes (476 tests); ruff clean.
- Simulated end-to-end: real roster config → verified subprocess routing (endpoint set, Anthropic key removed) → full failover → recovery flow against a real database and state file.
- Live cross-provider failover against a secondary endpoint is not yet validated end-to-end; simulated and unit coverage is in place.

Follow-ups (separate): fallback UX (session header + dashboard banner), an agentic gauntlet to gate roster membership, and a foreground launcher.


WingedGuardian (commented):

---

PR #810, opened 2026-06-30
fix(inbox): URL-level batching, delta-correct resume, follow-up dedup

WingedGuardian:
## What & why

The inbox evaluates URLs dropped into a persistent note (delta vs a per-file `evaluated_content` baseline). Forensics on the live data found four defects — acute during a long rate-limit window but **structural**:

1. **"16 evals at once"** — dispatch batched by *file*, so a 16-URL note became one giant CC call.
2. **Re-evaluation of already-seen URLs** — the approval→resume path re-read the *whole file* and discarded the delta (an approved eval re-chewed every URL).
3. **Duplicate follow-ups** — `follow_ups.create` had zero dedup.
4. **Approval/detection "storm"** — *not fixed here* (root cause unproven); this PR ships read-only instrumentation and leaves the amplifier intact for a separate diagnosis track.

## What changed (PR-1)

- **Segment a file's delta into items** (`scanner.segment_items`: each URL line = 1 item; contiguous prose = 1 note), grouped into batches of **`items_per_eval`** (default 5), one `inbox_items` row per batch under a shared `drop_id`. A 16-URL drop → ~4 evals, each its own `Genesis-N.genesis.md`.
- **One approval per drop** (gate ON): a single `route()` per drop; on approval all batches dispatch directly; the approval is consumed once. Works gate ON and OFF.
- **Delta-correct resume**: parked batches re-dispatch from the persisted `batch_items` slice — never a full-file re-read.
- **Per-batch baseline**: each batch merges only its own lines, so a failed sibling batch's lines stay un-baselined and retry.
- **Follow-up dedup** via `dedup_key` (+ partial-unique index).
- **Gate-error safety**: a `route()` exception fails the drop's rows (retriable) instead of leaving them stuck `processing`.
- **Storm instrumentation** (`STORM_DIAG`, read-only) to root-cause the churn after deploy.

Schema: `inbox_items.drop_id`, `inbox_items.batch_items`, `follow_ups.dedup_key` (+ indexes), dual-written into the canonical DDL, the `create_all_tables` path, and numbered migration `0040`.

## Verification

- Tests: `tests/test_inbox` (261), `tests/test_db` (780, incl. migration-runner applying 0040), `tests/test_autonomy/test_executor` (302) — all green. 26 new tests prove the new behavior.
- `genesis-architect` review: 2 BLOCKER + 3 HIGH addressed; mediums deferred with rationale.
- E2E: migration 0040 applies + is idempotent on a copy of the live DB; the Genesis-84 scenario (16 URLs → 4 evals + 4 files; +1 URL → 1 eval seeing only the new URL); a live dry run against real inbox content (real 73-line note → 7 clean evals, 0 stuck rows, 0 errors).

## Deliberately deferred (tracked as follow-ups)

- **Approval storm** root-cause + fix (PR-2): observe via `STORM_DIAG` after deploy; the restart also clears a stale-process confound.
- **Partial-failure auto-retry**: a *partially* failed drop currently retries its failed batches on the next edit to that note (a *totally* failed drop auto-retries). Fixing auto-retry touches the same detection path as the storm, so it's bundled into PR-2.
- **Semantic follow-up dedup**: the `dedup_key` is deterministic (URL/title + next_step); a semantic layer was evaluated and deferred separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,16 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Inbox evaluations no longer cram a whole batch of links into one giant pass — and stop
+  re-evaluating links they've already covered.** When you drop many URLs into an inbox note,
+  Genesis now evaluates them in small groups (≈5 at a time, configurable via
+  `items_per_eval`), each producing its own `…-N.genesis.md` response file, instead of one
+  sprawling evaluation of everything at once. Crucially, once a link has been evaluated it is
+  not re-evaluated when you add new links to the same note — only the genuinely new links are
+  processed (previously an approved evaluation could re-chew the entire file). When the CLI
+  approval gate is on, you approve a drop once and all its groups run under that single
+  approval. Duplicate follow-up items from the same recommendation are now also prevented.
+
 - **Campaign results no longer sit uncaptured until the next scheduled tick.** Previously a
   campaign that ran every couple of days would finish its background session but not record
   the outcome (or cost, or notify you) until the *following* tick — so a finished run could
--- a/config/inbox_monitor.yaml
+++ b/config/inbox_monitor.yaml
@@ -7,6 +7,13 @@ inbox_monitor:
   # the source (e.g. Untitled.md → Untitled.genesis.md).
   response_dir: "_genesis"
   check_interval_seconds: 1800
+  # items_per_eval: how many items (URLs / prose blocks) go into ONE Claude
+  # evaluation. A file's NEW content is segmented into items and grouped into
+  # batches of this size, so a 16-URL drop becomes ~4 evals (each its own
+  # Genesis-N.genesis.md), not one mega-evaluation. Default 5.
+  items_per_eval: 5
+  # batch_size is LEGACY (files-per-cycle) and no longer governs eval grouping;
+  # items_per_eval replaces it. Kept for backward compatibility.
   batch_size: 1
   model: "sonnet"
   effort: "medium"
--- a/src/genesis/db/crud/follow_ups.py
+++ b/src/genesis/db/crud/follow_ups.py
@@ -53,27 +53,48 @@ async def create(
     kind: str = "follow_up",
     domain: str | None = None,
     goal_id: str | None = None,
+    dedup_key: str | None = None,
     id: str | None = None,
 ) -> str:
     """Create a follow-up and return its ID.
 
-    kind:    'follow_up' (intended for action) or 'tabled' (tracked, not for action).
-    domain:  'internal' | 'user_world' | None (None = not yet classified).
-    goal_id: optional link to a unified goal (user_goals.id) for future promotion.
+    kind:     'follow_up' (intended for action) or 'tabled' (tracked, not for action).
+    domain:   'internal' | 'user_world' | None (None = not yet classified).
+    goal_id:  optional link to a unified goal (user_goals.id) for future promotion.
+    dedup_key: optional idempotency key. Callers that may re-run (e.g. inbox
+              re-evaluation) pass a stable hash so the same recommendation does
+              not create duplicate rows; a partial unique index backstops races.
     """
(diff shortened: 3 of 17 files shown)


WingedGuardian (commented):
## First-pass review — PR #810

**Structural review:** No ERRORs. Three WARNINGs, four NOTEs below.
**PII/Secrets/Wording scan:** CLEAN — no secrets, private IPs, hardcoded credentials, or sensitive wording in PR metadata.

---

### ⚠️ WARNING — Missing type annotations on new dispatch methods

**File:** `src/genesis/inbox/monitor.py`

Five new private async methods — `_set_drop_status`, `_acquire_drop_approval`, `_dispatch_one_batch`, `_complete_batch_baseline`, `_notify_batch` — have untyped `items` / `item` parameters. mypy/pyright coverage is incomplete on the refactored dispatch path as a result.

---

### ⚠️ WARNING — Repeated deferred `inbox_items` imports inside method bodies

**File:** `src/genesis/inbox/monitor.py`

`from genesis.db.crud import inbox_items` is deferred inside five separate method bodies (`_queue_drop`, `_set_drop_status`, `_acquire_drop_approval`, `_dispatch_one_batch`, `_complete_batch_baseline`). The module is already available at process load time. This pattern is inconsistent with the rest of `monitor.py`, obscures import errors until the code path is exercised at runtime, and adds minor overhead per call. Move to module-level.

---

### ⚠️ WARNING — Undocumented kill-window in `_acquire_drop_approval` (claim-before-decide)

**File:** `src/genesis/inbox/monitor.py`

All drop rows are set to `status='processing'` **before** the gate decision. If the monitor process is killed between that write and the gate call, rows stay stuck in `'processing'` until `expire_stuck` fires — the same "invisible to detection" class of bug the PR explicitly addresses for the gate-raises path. The semantics are intentional, but:
1. The narrow kill-window is unmentioned in the surrounding comments.
2. There is no test covering a kill between `set_processing` and the gate.

A brief comment and/or a test covering this window would close the gap.

---

### 📝 NOTE — `import hashlib` deferred inside method body

**File:** `src/genesis/inbox/monitor.py` → `_create_follow_ups_from_eval`

`hashlib` is stdlib with no circular-import risk. Move it to module-level imports for consistency and to surface any import issues at startup rather than at first dedup execution.

---

### 📝 NOTE — `_consume_approval` missing docstring

**File:** `src/genesis/inbox/monitor.py`

All other new methods in this PR have full docstrings; `_consume_approval` has only a short inline comment in the `def` line. Minor consistency gap.

---

### 📝 NOTE — Raw SQL in test assertions (non-blocking)

**File:** `tests/test_inbox/test_drop_dispatch.py`

Several assertions issue raw SQL directly (e.g. `db.execute('SELECT status, drop_id FROM inbox_items WHERE ...')`). Not a rule violation (tests are excluded from the crud-module requirement), but using the already-imported crud helpers where possible would make these assertions more resilient to schema renames.

---

### 📝 NOTE — `decision.mode == 'api'` fallthrough path untested

**File:** `src/genesis/inbox/monitor.py` → `_acquire_drop_approval`

The `mode == 'api'` branch logs an error and falls through to `'approved'`, dispatching via CLI. This is correct given `api_call_site_id=None`, but there is no test covering this path. If the invariant breaks silently the drop dispatches CLI without consuming API output — worth a guard test or assertion.

---

WingedGuardian:
**Review status:** All CI checks pass (lint, leak-detector, migration-check, CodeQL ×3, full test suite 9m38s). The automated Codex review was **quota-limited** (no findings posted). The change was instead reviewed by a `genesis-architect` pass before this PR: 2 BLOCKER + 3 HIGH findings were fixed (gate-error try/except so a failed `route()` doesn't strand rows in `processing`; `_build_invocation` now takes `system_prompt`; hard-reject sets `retry_count=max`; consume-before-dispatch on resume; removed dead `get_drop_rows`/`get_pending_drop_rows` + `resumed_ids`). Mediums (storm `get_all_known` determinism, partial-failure auto-retry, semantic dedup) deferred to PR-2 with rationale in the description.

PR #811, opened 2026-06-30
feat(cc): surface model-fallback state on the dashboard and CLI

WingedGuardian:
## What

When Claude is rate-limited account-wide and Genesis fails the conversation over to a roster peer (e.g. GLM), that degraded condition is now **visible** instead of silent:

- **Dashboard** — a slim page-top banner (`⚠ Running on <peer> — Claude <reason> since <time>`), and the CC Sessions card goes to a `fallback` state with a badge naming the active peer and since-time. Driven by a new `fallback` field on the CC sessions health snapshot (no new route).
- **Interactive CLI session start** — an advisory "Genesis Fallback Active" notice. The `[model]` header stays honest: interactive sessions run on native Claude and don't fail over; the notice reflects the *server's* state.

## Recovery is now turn-independent

Previously the fallback flag only cleared on a successful foreground turn, so the indicators could go stale during idle. Now it clears on **any** successful home-model call — foreground turns and background sessions — plus a slow safety probe for a totally idle system. Recovery is anchored on the actual rate-limited home model, so a success on a different model never clears it prematurely.

## Notes

- Honest precedence: a hard `UNAVAILABLE` (contingency) outranks the fallback indicator, so a real outage is never shown as a softer "fallback".
- Internal-facing surfaces only (dashboard + CLI session context).

## Verification

- Targeted unit tests for the snapshot field, the shared recovery helper, the background-session clear, and the safety probe.
- Session-start hook exercised end-to-end (notice shows when active, absent when cleared).
- Dashboard rendering verified in a real browser: active → top banner + degraded card with badge; cleared → hidden, card back to healthy.

### DIFF
--- a/scripts/genesis_session_context.py
+++ b/scripts/genesis_session_context.py
@@ -232,6 +232,40 @@ def main() -> None:
             except (OSError, ValueError):
                 pass  # Never block session start
 
+        # Fallback NOTICE (advisory): Genesis's server-side ConversationLoop is
+        # currently running on a roster peer (e.g. GLM) because the home model
+        # (Claude) is rate-limited/exhausted account-wide. This interactive CLI
+        # session runs on CC-native pinned Claude and does NOT fail over itself
+        # (failover is server-side, a different process) — so the [model] header
+        # stays honest and we surface the server's degraded condition here as a
+        # separate block. Read the cross-process state file directly (import-free,
+        # fail-open), mirroring the status.json read above.
+        # Writer: src/genesis/cc/fallback_state.py.
+        _fallback_file = Path.home() / ".genesis" / "cc_fallback_state.json"
+        if _fallback_file.exists():
+            try:
+                import json as _json_fb
+                fb = _json_fb.loads(_fallback_file.read_text())
+                if isinstance(fb, dict) and fb.get("is_fallback"):
+                    peer = fb.get("fallback") or "a roster peer"
+                    home = fb.get("original") or "Claude"
+                    reason = (fb.get("reason") or "unknown").replace("_", " ")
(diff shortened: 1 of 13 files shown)


WingedGuardian:
## First-pass review — PR #811

**Structural review:** 2 NOTEs (no ERRORs or WARNINGs)

---

### NOTE — Missing docstrings on public methods

**`src/genesis/resilience/cc_fallback_probe.py`** — `CCFallbackProbeWorker.start()` and `stop()` are public methods without docstrings. The class docstring covers intent, but the methods have no per-method doc. Low priority given how self-evident they are, but worth adding for consistency with the rest of the module.

---

### NOTE — Missing `@pytest.mark.asyncio` on `TestBackgroundFallbackRecovery`

**`tests/test_cc/test_direct_session.py`** — All four async test methods in the new `TestBackgroundFallbackRecovery` class lack `@pytest.mark.asyncio` decorators:

```python
async def test_home_claude_success_clears(...)
async def test_peer_success_with_claude_home_does_not_clear(...)
async def test_home_peer_success_clears(...)
async def test_native_claude_success_with_peer_home_does_not_clear(...)
```

The sibling file added in this same PR (`tests/test_cc/test_fallback_recovery.py`) uses explicit `@pytest.mark.asyncio` on every async test method. If the project does not have `asyncio_mode = "auto"` in the pytest config, these four methods will be collected but never actually awaited — they'll appear to pass while exercising nothing. Worth verifying against `pyproject.toml` / `pytest.ini` and adding the decorators if auto mode is not configured.

---

Structural review complete. No pattern violations found beyond the notes above.

---

PR #812, opened 2026-06-30
fix(inbox): baseline read order by processed_at (reuse breaks created_at sort)

WingedGuardian:
## Live-caught regression in PR #810

While verifying PR #810 on the live system, the first real drop exposed a read-ordering bug:

`_queue_drop` reuses retriable-failed rows via `reuse_as_pending`, which **preserves the reused row's old `created_at`**. But `get_evaluated_content` (and `get_last_completed_at`) ordered by `created_at DESC`. So a freshly-completed *reused* row — which holds the correct, fullest cumulative baseline — could sort **below** an older completed row that merely has a newer `created_at`. The read then returned a **stale, partial baseline**, and the next scan **re-evaluated lines already done** (the exact behavior PR-1 fixes).

**Observed live:** a reused row `created 2026-03-18` but `processed` today held both newly-dropped URLs (`evlen 22271`); a `14:52`-created row (`evlen 21906`, missing one URL) won the `created_at` sort. The next scan would have re-evaluated the dropped URL.

## Fix

Order both baseline reads by **`processed_at DESC`** (completion time = the current baseline), with `created_at`/`rowid` as tiebreaks. Completion time is the correct key because the baseline is built at completion and reuse scrambles `created_at`.

- `get_evaluated_content`: `ORDER BY processed_at DESC, created_at DESC, rowid DESC`
- `get_last_completed_at`: `ORDER BY processed_at DESC, created_at DESC`

2 regression tests added (reused row with old `created_at` + latest `processed_at` must win). Full `tests/test_inbox` green (263).

### DIFF
--- a/src/genesis/db/crud/inbox_items.py
+++ b/src/genesis/db/crud/inbox_items.py
@@ -298,13 +298,17 @@ async def get_last_completed_at(
     Includes both normal evaluations (with response files) and Acknowledged
     items (no response file) so cooldown applies uniformly.
 
-    Orders by processed_at (completion time), not created_at, so a reused row
-    (which keeps its old created_at via reuse_as_pending) doesn't make the
-    cooldown read a stale completion time. See get_evaluated_content.
+    Orders by processed_at (completion time), not created_at, so a row that was
+    detected long ago but completed recently (e.g. parked for approval, then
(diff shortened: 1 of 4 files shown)


WingedGuardian:

---

PR #813, opened 2026-06-30
fix(inbox): follow-up dedup race + consume-failure visibility (dispatch safety, part 1)

WingedGuardian:
## Scope

The two cleanly-separable, low-risk items from the dispatch crash-safety cluster (follow-up `0b68d341`). The third item (#1, resume double-dispatch on crash) is intentionally **not** here — it needs an at-most-once dispatch state machine and is left in `0b68d341` to be done deliberately (with the storm/restart work).

## #3 — follow-up dedup IntegrityError no longer aborts the rec loop
`_create_follow_ups_from_eval` does a non-atomic check-then-insert against the partial-unique `dedup_key` index. A racing duplicate (another writer inserts the same key between the `exists` check and the insert) raised `sqlite3.IntegrityError`, which propagated and **silently aborted every remaining recommendation in that evaluation** (the outer try only logged "non-fatal"). Now the per-rec `create()` catches `IntegrityError` and `continue`s, so the rest of the evaluation's recommendations are still created.

## #2 — consume-failure visibility
`_consume_approval` swallowed all errors at debug and returned nothing. It now **returns bool** and logs a failed / already-consumed `mark_consumed` at **WARNING** (a non-consumed approval can be ridden by a later inbox drop until the gate's 24h staleness window). The resume caller still dispatches regardless — *acting* on the result (skip-and-retry on transient failure vs fail-and-re-detect when the approval was already consumed) requires the at-most-once dispatch state machine, so that part stays in `0b68d341` (#1) to avoid introducing a stranded-parked-drop state here.

## Tests
2 new (IntegrityError mid-loop doesn't abort remaining recs; `_consume_approval` returns True with no gate / False on already-consumed / False on raise). Full `tests/test_inbox` green (269).

### DIFF
--- a/src/genesis/inbox/monitor.py
+++ b/src/genesis/inbox/monitor.py
@@ -1134,20 +1134,39 @@ class InboxMonitor:
             )
         return "approved"
 
-    async def _consume_approval(self, request_id: str) -> None:
+    async def _consume_approval(self, request_id: str) -> bool:
         """Consume a resume drop's approval so the next drop cannot ride the
-        same (content-agnostic, stable-key) approval."""
+        same (content-agnostic, stable-key) approval.
+
+        Returns True if the approval was consumed (or there is no gate to
+        consume against), False if the consume failed or the approval was
+        already consumed. A failure is logged at WARNING (it was previously
+        swallowed at debug) because a non-consumed approval can be ridden by a
+        later inbox drop until the gate's 24h staleness window expires.
+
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

WingedGuardian:
Merge note: Codex code review was quota-limited (usage cap reached). Merging on green CI + CLEAN structural/PII review per the quota-exhausted merge path. Findings, if any surface later, tracked via follow-up `0b68d341`.

PR #815, opened 2026-06-30
fix(dashboard): move system banners to page-top (slim bars, every tab)

WingedGuardian:
## What

The dashboard's stale-data, paused, resilience-degradation, and provider-alert banners lived **inside** the System Health panel. That panel is a flex-row container, so an active banner rendered as a stretched full-height **column** wedged between the panel title and the cards — and the banners only showed on the Overview tab.

This moves all of them to the **page-top block container** (joining the CC model-fallback banner added in #811), so each renders as a **slim full-width bar** and stays visible on **every tab** — which is what you want for global conditions (API unreachable, Genesis paused, resilience degraded, provider credit/key alerts).

Pure relocation: the banner conditions, styles, and bindings are unchanged.

## Verification

Rendered in a real browser with fallback + resilience + provider-critical + provider-warning all active simultaneously: all banners stack as slim full-width bars (~32px tall) above the panel, and remain visible after switching to a non-overview tab.

### DIFF
--- a/src/genesis/dashboard/templates/genesis_dashboard.html
+++ b/src/genesis/dashboard/templates/genesis_dashboard.html
@@ -4544,13 +4544,41 @@
               @click="location.hash = 'backup'">Backup & Updates</button>
     </nav>
 
-    <!-- CC model-fallback banner (page-top, full-width, visible on every tab):
-         the server-side conversation is failed over to a roster peer because the
-         home model (Claude) is rate-limited account-wide. Placed here (block flow,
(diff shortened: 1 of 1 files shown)


WingedGuardian:
Structural review: CLEAN — single HTML template change, no Python files modified, no structural pattern rules apply.
PII/Secrets/Wording scan: CLEAN — no credentials, IPs, personal emails, private hostnames, or sensitive-wording issues found in added lines or PR metadata.

---

PR #816, opened 2026-07-01
feat(feedback): verified-correctness signal for surplus background tasks

WingedGuardian:
## What

Background (\"surplus\") tasks feed the Outcome Bus, but a terminal `status='completed'` only proves the work **ran** — not that its output was **useful**. That left the surplus signal ~99.6% positive and non-discriminative.

This adds a second, orthogonal axis: for insight-producing task types, record whether the intake pipeline actually kept any of the task's output, and surface a discriminative negative when it didn't.

## How

- **`surplus_tasks.outcome_quality`** (`'useful'` | `'hollow'` | `NULL`), with migration `0041`. Set at task completion, reusing the existing intake quality verdict — no new model call:
  - `hollow` = intake ran but routed **everything** to discard (the task ran yet produced nothing of value)
  - `useful` = intake kept at least one finding (knowledge or observation)
  - `NULL` = action tasks, legacy rows, intake failures, and empty/too-short output (ambiguous or not a quality signal — left untouched)
- **Only insight-producing task types are eligible** (`INSIGHT_PRODUCING_TASK_TYPES`). Action tasks, pipeline *intermediate* steps (their output feeds the next step, not the knowledge base), and monitoring/probe types (a \"nothing noteworthy\" pass is a valid success) are deliberately excluded to avoid false negatives.
- **The harvester change is purely additive.** The existing `EXECUTION_OUTCOME` positive is unchanged for every completed task; a `hollow` completion *additionally* emits a tier-1 `VERIFICATION_FAILED`. The two are distinct signal types, so they coexist and remain idempotent across re-harvests. In per-domain rollups: useful = 1.0 > hollow ≈ 0.5 > hard failure = 0.0.

## Verification

- 92 targeted tests (harvester emit, per-branch scheduler verdict, CRUD, migration) + lint clean.
- End-to-end check on a realistic dataset: verdict at completion → harvest → tier-1 negatives → per-domain aggregation, with legacy rows keeping their original positive-only behaviour (no retroactive changes) and re-harvests adding no duplicates.

## Notes

- No backfill-version bump: only tasks completed after deploy can be `hollow`, so there is nothing historical to re-process.
- The capability-map consumer that reads these aggregates remains behind its existing default-off flag; a code comment marks the exact-weighting decision to revisit when that feed is enabled.

### DIFF
--- a/src/genesis/db/crud/outcome_events.py
+++ b/src/genesis/db/crud/outcome_events.py
@@ -163,11 +163,14 @@ async def aggregate_by_domain(
 
     Note on surplus "hollow" tasks: an insight task that ran but produced nothing
     useful emits TWO tier-1 rows — a positive EXECUTION_OUTCOME ("it ran") and a
-    negative VERIFICATION_FAILED ("output was useless") — so it shows up here as
-    ``n=2, positive=1, negative=1, avg_value≈0.5`` by design. A useful task is
-    ``n=1, avg_value=1.0``; a hard failure is ``n=1, avg_value=0.0``. Read
-    ``avg_value`` (or ``positive``/``negative``), not ``positive/n``, to rank
-    domain quality — the latter understates useful tasks relative to hollow ones.
+    negative VERIFICATION_FAILED ("output was useless") — so it contributes ``n=2``
+    (one positive, one negative) to its domain. For the ``source='surplus'`` feed,
+    every ``value`` is 0.0/1.0 and aligned with ``polarity``, so ``avg_value`` and
+    ``positive/n`` are IDENTICAL here; both already fold in the hollow penalty.
+    Ordering still holds — a domain of U useful / H hollow / F failed scores
+    ``(U+H)/(U+2H+F)``, i.e. all-useful=1.0 > hollow-mixed > all-failed=0.0. The
+    n=2 weighting makes a hollow task count as ~1/3 effective credit, not 1/2; a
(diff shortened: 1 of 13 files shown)


WingedGuardian (commented):
Structural review: 1 WARNING found.

---

### WARNING — Import hygiene: duplicate inline imports from the same module

**`src/genesis/surplus/scheduler.py`** inside `dispatch_once()`

The PR adds `INSIGHT_PRODUCING_TASK_TYPES` as a separate lazy import immediately above the pre-existing lazy import of `TaskType`, leaving two consecutive `from genesis.surplus.types import …` lines:

```python
from genesis.surplus.types import INSIGHT_PRODUCING_TASK_TYPES
from genesis.surplus.types import TaskType as _TT
```

These should be a single statement:

```python
from genesis.surplus.types import INSIGHT_PRODUCING_TASK_TYPES, TaskType as _TT
```

---

WingedGuardian:
Addressed the import-hygiene warning in `a68effcc`. The suggested single-line form (`import INSIGHT_PRODUCING_TASK_TYPES, TaskType as _TT`) isn't reachable here — the repo's isort keeps `combine_as_imports` off, so ruff always splits the aliased `TaskType as _TT` back onto its own statement. Instead I hoisted `INSIGHT_PRODUCING_TASK_TYPES` to the existing top-level import alongside `SurplusExecutor` (types.py is a dependency-free leaf → no circular-import risk), which removes the duplicate lazy line entirely. Ruff clean, scheduler tests green.

PR #817, opened 2026-07-01
fix(inbox): stop phantom re-detection storm (deterministic known-hash + delta-before-cooldown)

WingedGuardian:
## What

Fixes the inbox **detection storm** (PR-2a of 3 — detection, then dispatch at-most-once, then partial-failure retry). Root cause proven via live diagnostics + DB forensics.

**Root cause:** `get_all_known()` returned a *stale* whole-file hash for a file during the window when no row carried the file's *current* hash — its query has no `ORDER BY` (arbitrary dict-overwrite winner among a file's many rows), and the cooldown-defer path wrote no row. So the "known" hash lagged the on-disk hash → the file was re-detected as "modified" on every scan. While an approval was pending, that repeatedly cancelled + recreated it (the storm).

## Changes

1. **`get_all_known` determinism** — `ORDER BY created_at ASC, rowid ASC` on both queries, so the newest row's hash wins the per-file overwrite deterministically. A file has many rows (one per batch, plus reused rows), and `reuse_as_pending` resets `created_at` to now while keeping the old (low) rowid, so insertion/rowid order is not recency.
2. **Delta before cooldown** — the modified-files loop computes the delta before the cooldown check. Cooldown now gates only genuinely-new content; on an **empty** delta (bytes changed but no new content vs the baseline — e.g. a re-pasted URL with different tracking params) it always writes a completing row to advance the known hash, so a phantom-modified settles in one scan. Stale-pending supersede is shared across both paths.

The cancel-on-detect amplifier is left untouched — it is downstream of detection and no longer fires spuriously once detection is correct.

## Verification

- genesis-architect reviewed pre-implementation (one tweak incorporated).
- TDD, red-first proven: the storm-kill test **fails on the old monitor**, passes on the fix.
- Multi-agent code review (9 angles + verify + sweep): no bugs; a data-loss candidate was refuted; one regression test added.
- Full inbox suite: **273 passed**; ruff clean.
- Live-DB E2E: fixed `get_all_known` is deterministic and returns the current hash against the real 306-row inbox note.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Inbox notes that change without adding anything new no longer get re-scanned over and over.**
+  Editing an inbox note in a way that changes its bytes but not its actual content — re-pasting a
+  link with different tracking/share parameters, reordering lines, tweaking whitespace — used to
+  leave the note looking "modified" on every scan, which (while a link approval was pending) could
+  repeatedly cancel and recreate that approval. Genesis now recognizes there's no new content and
+  marks the note current in a single scan, so it settles instead of churning.
+
 - **Inbox evaluations no longer cram a whole batch of links into one giant pass — and stop
   re-evaluating links they've already covered.** When you drop many URLs into an inbox note,
   Genesis now evaluates them in small groups (≈5 at a time, configurable via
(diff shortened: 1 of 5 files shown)


WingedGuardian (commented):

---

WingedGuardian:
Merge note: Codex review quota-limited (usage cap). Merging on all-green CI + a clean local multi-agent review (9 finder angles + verify + sweep; one CRITICAL candidate refuted, one regression test added) + genesis-architect pre-review + live-DB E2E. Per the quota-exhausted merge path.

PR #818, opened 2026-07-01
feat(infra): automatic disk hygiene — scheduled worktree reaping + cache reclamation

WingedGuardian:
## Why

The container root filesystem could fill to 100% and stall Genesis — disrupting
the server, its SQLite WAL, and backups. The root cause: a worktree-cleanup
script existed but was **never scheduled by any installer**, so merged git
worktrees (and regenerable caches) accumulated unbounded until the disk filled.

## What this does

**Automatic disk hygiene**, wired to run on its own:

- **New `genesis-disk-hygiene` systemd timer** (daily) that runs the existing
  worktree reaper plus a new cache-reclamation step. Installer-enabled from
  templates, so every install gets it — the piece that was missing before.
- **`scripts/disk_reclaim.py`** — clears only *regenerable* caches on a strict
  allowlist: package caches (npm/uv/pip) every run; heavier reindexable caches
  (code-graph indexes) only when the disk is genuinely under pressure (≥90%).
  Never touches the repo, virtualenv, data, config, secrets, browser profiles,
  or generated output.
- **Pre-100% safety net** — the health system now clears regenerable caches
  automatically at 90% (before the disk fills) and raises the warning threshold
  to 85% for earlier visibility.

Merged worktrees are moved to a 7-day recovery trash bin, never hard-deleted;
in-progress and unmerged work is always skipped.

## Verification

- Ran end-to-end under the real (sandboxed) service: reaped merged worktrees
  (correctly skipping active/recent/unmerged), reclaimed cache space, exited
  cleanly.
- New unit tests for the reclaim allowlist, threshold gating, and the health
  probe; existing suites pass; lint clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **Genesis now keeps its own disk clean automatically, so it won't quietly fill up and stall.**
+  A daily hygiene job reaps git worktrees whose branches have already merged (moving them to a
+  7-day recovery trash bin, never deleting work in progress) and clears regenerable caches that
+  otherwise creep up over time. If the disk still climbs toward full, Genesis clears the heavier
+  reindexable caches automatically at 90% — before the disk hits 100% and disrupts the server,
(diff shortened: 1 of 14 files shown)


WingedGuardian:

---

PR #819, opened 2026-07-01
feat(eval): agentic model-roster gauntlet (earn roster membership via a real CC fix-loop)

WingedGuardian:
## What

Adds a first-class **agentic gauntlet** that validates a model roster member can actually drive Claude Code through a multi-turn coding task — not just answer a single prompt. `genesis eval gauntlet --model <name>` has the model (native Claude, or a routed peer reached via its native Anthropic-compatible endpoint) fix real broken Python projects inside a live Claude Code session, then scores it **objectively by running the project's tests**.

Until now, roster membership was asserted (a `validated` field nothing read); this makes it *earned*.

## How it works

- **Real path, not a mock.** Runs through the actual `CCInvoker` against a throwaway fixture copy, so it exercises the same selection chokepoint + arg/env production uses. Model selection mirrors the background-session idiom (`roster_eligible = bool(overrides)`); the route actually taken is verified via `CCOutput.via_proxy`.
- **Objective, cheat-resistant scoring.** PASS = the suite goes green **and** the protected surface (tests + pytest config, any depth) is unchanged — so editing tests / `conftest.py` / `pyproject.toml` to fake a pass is caught. Generated `__pycache__` is excluded so honest runs aren't flagged.
- **Honest failure taxonomy.** Infra failures (rate-limit, quota, timeout) are *skipped*, excluded from the verdict — a peer being out of balance is not a quality regression.
- **Isolated & safe.** No MCP servers, a dedicated Bash-sandbox tmpdir (new additive `CCInvocation.claude_code_tmpdir`), and a per-model file lock so a manual run and the scheduled job can't interleave.
- **Advisory gating only.** A model that previously passed and now fails triggers an alert + a human-gated proposal — it is **never** auto-removed from selection/failover. Cold-start failures (never-passed models) don't alert.
- **Weekly run, opt-in.** A `CronTrigger` job re-validates each roster model, **off by default** (`gauntlet.scheduled` in the roster config) so it never spends on paid peers without an explicit opt-in. Manual CLI runs anytime.

## Testing

- 16 hermetic unit tests (scorer verdicts, cheat-guard, file lock, regression detection, DB round-trip).
- Full end-to-end against native Claude: **3/3 fixtures pass** (a stats lib, a multi-file intervals package with an implement-from-docstring function, and a lexer→parser→evaluator calc engine with a new operator to implement).
- `ruff` clean; no regressions across the affected CC / eval / ego test suites.

## Notes

- Three committed fixtures live under `src/genesis/eval/gauntlet_fixtures/` (intentionally broken → excluded from lint and test discovery).
- Non-Anthropic peer validation exercises the same code path; validating a specific peer additionally requires that provider's funded key.


WingedGuardian:
## First-pass review — PR #819

**Structural: 1 WARNING, 1 NOTE**

---

### WARNING — Missing coverage: `gauntlet_regression_actions.py`

`src/genesis/ego/gauntlet_regression_actions.py` is a new file in `src/` with no direct unit tests. `tests/test_eval/test_gauntlet.py` thoroughly exercises `gauntlet_regression.py` (the detection side) via `ego_crud` monkeypatches, but `handle_gauntlet_regression_resolution` itself is never imported or called in the test suite. Two paths need coverage:

- **Approved → executed**: the function calls `ego_crud.execute_proposal` and returns `True`.
- **Non-approved status → no-op**: returns `False` without touching the DB.

This function is wired into three separate resolution entry points (dashboard `ego.py`, `ego/proposals.py`, `mcp/health/ego_tools.py`), so a silent failure here would let an approved `gauntlet_regression` proposal linger on the board and potentially be re-dispatched. Direct unit tests for `gauntlet_regression_actions` would match the pattern already set by `j9_regression_actions`.

---

### NOTE — Missing docstring: `GauntletFixture`

`GauntletFixture` in `src/genesis/eval/gauntlet.py` is a public `@dataclass(frozen=True)` with no class-level docstring. Its five fields are clear by name, but a one-liner saying what it represents (a committed fixture project the gauntlet model must fix) would align with the docstrings on all other public types in the module.

---

No API keys, tokens, private IPs, hardcoded paths, personal emails, or private hostnames found in added lines. PR title, body, and CHANGELOG addition contain no wording that references a PII leak, credential exposure, or security incident.


---

PR #821, opened 2026-07-01
fix(inbox): dispatch at-most-once via row-state claim (no duplicate eval on crash)

WingedGuardian:
## What

Makes inbox dispatch **at-most-once** (PR-2b of 3 — follows #817 detection storm; PR-2c partial-failure retry next). Fixes a restart/crash double-dispatch edge found reviewing the batching rework.

**Root cause:** the resume pass consumed the approval *gate* then dispatched, but never moved the row out of its `awaiting_approval:` parked state until completion. A crash between consume and completion left the row re-findable by `get_awaiting_approval`, so the next scan re-resumed and re-dispatched it — duplicate eval + duplicate `…-N.genesis.md`. The consume return value was also ignored, so a failed `mark_consumed` left an unconsumed content-agnostic approval a later drop could ride.

## Changes

- **`claim_for_dispatch(id, reqid)`** — an atomic CAS that flips a parked row `awaiting_approval:<reqid>` → `dispatching:<reqid>` (a new marker prefix), returning whether this call won the claim. The **row state is now the at-most-once authority**.
- **Resume loop** claims each row *before* the CC call, skips losers, then consumes (best-effort) and dispatches only claimed rows.
- A claimed (`dispatching:`) row is invisible to `get_awaiting_approval` (so a crash can't re-resume it) but **is** reaped by `expire_stuck_processing` after its timeout — recovered into the retry path. New-drop dispatch is unchanged (already crash-safe).
- Consume is now best-effort: a failure neither strands the drop nor bypasses at-most-once; it only leaves the approval rideable until the gate's 24h staleness window (WARNING-logged).
- **Cleanup (review-driven):** completion now clears the in-flight marker from `error_message`, so a finished item no longer shows internal dispatch state.

## Verification

- TDD, red-first proven: the double-dispatch test fails without the claim; the marker-clear test fails without the clear.
- Multi-agent code review (line-by-line + SQL, removed-behavior, cross-file + altitude): no correctness bugs; two LOW cosmetic findings fixed at root (marker clear).
- Full inbox suite: **280 passed** (with #817 present); ruff clean.
- Live-schema E2E: awaiting → claim (single-winner) → invisible-to-resume → reap → retriable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,12 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   approval gate is on, you approve a drop once and all its groups run under that single
   approval. Duplicate follow-up items from the same recommendation are now also prevented.
 
+- **An approved inbox evaluation that gets interrupted mid-run can no longer be evaluated
+  twice.** If the server restarted (or crashed) in the narrow window right after you approved a
+  link evaluation but before it finished, the next scan could re-run the same evaluation and
+  write a duplicate `…-N.genesis.md` file. Genesis now claims each evaluation the moment it
+  starts, so an interrupted one is recovered and retried rather than run a second time.
+
 - **Campaign results no longer sit uncaptured until the next scheduled tick.** Previously a
   campaign that ran every couple of days would finish its background session but not record
   the outcome (or cost, or notify you) until the *following* tick — so a finished run could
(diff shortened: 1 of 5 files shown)


WingedGuardian (commented):

---

WingedGuardian:
Merge note: Codex review quota-limited. Merging on all-green CI + clean local multi-agent review (3 finders: no correctness bugs; 2 LOW cosmetic fixed at root) + live-schema E2E. PR-2a (#817) observed clean post-deploy (0 STORM_DIAG). Per the quota-exhausted merge path.

PR #823, opened 2026-07-01
fix: harden STEERING auto-write + raise CC timeout, drop timeout retry

WingedGuardian:
## Problem

Two independent robustness bugs, exposed together by a single conversational message.

**1. STEERING.md corruption.** A benign conversational *status update* was classified
`outcome=approach_failure`, which fires the STEERING.md auto-population path. Its guard was a
naive full-text regex (`never|don't|stop|…`) that matches an *incidental* keyword — e.g. "never"
inside a positive phrase like "it's never too late" — so the raw first 200 chars of the message
were written verbatim as a "hard constraint". The writer then **flattened the file's `##` section
structure** because it parsed rules per physical line and dropped every header and separator.

**2. Conversational timeout double-fault.** The global CC invocation timeout default was 600s
(10 min). A substantive task started in chat hit that cap; and because `CCTimeoutError` is a
`CCError`, the stale-resume recovery retried fresh — burning a **second** full 600s window —
before returning a fallback.

## Fix

This is the **safe subset**; the outcome-classifier root-cause (why a status update was labeled
`approach_failure` at all) is a separate, spike-gated PR.

- **Directive-shape guard** (`learning/pipeline.py`): a fail-closed `_looks_like_directive`
  (≤2 sentences, ≤30 words, imperative opener) now gates the STEERING write. A chatty/status
  message can no longer become a hard constraint even when the outcome is mis-classified.
  Removes the naive keyword regex.
- **Structure-preserving STEERING writer** (`identity/loader.py`): parse and write as
  `---`-delimited blocks, so a rule's `##` header and multi-line body survive
  read→append→write; FIFO eviction drops whole blocks; a separator line appearing inside a
  rule body is neutralized so it can't split the block on a later round-trip.
- **Timeouts** (`cc/types.py`, `cc/conversation.py`, `resilience/cc_fallback_probe.py`): raise
  the global default 600→7200 (2h project floor — short caps fight legitimate long work; the
  subprocess kill still bounds a genuinely hung process); pin the CC fallback liveness probe to
  300s (fast-fail); and stop retrying on `CCTimeoutError` in both `_try_invoke` and
  `_try_invoke_streaming` (a timeout is not a stale-resume failure).
- **CONVERSATION.md**: guide long tasks to break into steps and hand genuinely long work to a
  background session (acknowledging immediately) instead of blocking the turn silently.

## Tests

- Directive-guard unit suite (genuine directives pass; status updates / the incident message
  rejected; layer boundaries).
- STEERING block round-trip + `##`-header preservation + whole-block FIFO eviction +
  separator-in-body.
- Timeout-no-retry on both invoke paths; global-default assertion.
- A pipeline-level regression reproducing the exact mislabel (`APPROACH_FAILURE` + a chatty
  message) → **no** steering write.

`test_cc` (494) and `test_learning` (649) pass locally; `ruff` clean. Verified E2E against a copy
of the live STEERING.md: incident blocked, a genuine directive stored, all sections preserved.


WingedGuardian (commented):

**PII/Secrets/Wording scan: SOFT WARNING**

---

### Private user message verbatim in public test corpus

**`tests/test_learning/test_steering_guard.py` lines 17–25** and **`tests/test_learning/test_pipeline.py`** (added `incident` variable)

The actual text of the private Telegram DM that triggered the 2026-06-30 incident is stored verbatim in a public test file as `_INCIDENT_MESSAGE`. It reproduces identifiable private communication from a real conversation, including references to specific professional activities ("Autonomize is dead", "the agent challenge", "the conference", "arxiv"). Once committed, this string lives in the permanent public git history.

The test itself is well-designed and the structural fix it validates is correct. The concern is that it uses the real DM rather than a synthetic message with the same structural properties. A synthetic replacement preserves every test-meaningful characteristic — multi-sentence, chatty opener, "never" mid-sentence rather than at the start — without committing the actual private message:

```python
# Synthetic stand-in with identical structural properties to the real incident
_INCIDENT_MESSAGE = (
    "Yeah sorry for being quiet lately. The sprint went well, no action needed "
    "there. I skipped the meeting. The paper didn't go out, "
    "but we should keep going -- its never too late"
)
```

This is flagged as a **soft warning** (private communication committed to a public repo), not a hard block — the message contains no credentials or direct PII such as email, IP, or account ID. The author of this PR presumably made this trade-off knowingly; flagging for awareness only.

---

WingedGuardian:
Soft warning (verbatim private message in test corpus) **resolved**: the real DM was replaced with a synthetic stand-in that preserves the test-meaningful structure (across test_steering_guard.py, test_pipeline.py, and a pipeline.py comment). Verified no identifying tokens remain in the tree. Codex review was quota-limited; change was inline-reviewed. Merging on CI + inline review.

PR #824, opened 2026-07-01
fix(inbox): partial-failure auto-retry for stranded batches

WingedGuardian:
## What

Final PR of the inbox series (PR-2c; follows #817 storm fix and #821 dispatch at-most-once). Auto-retries a batch that failed partway through a multi-batch evaluation.

**The gap:** when only SOME batches of a drop succeed and others fail (e.g. a rate-limit), the completed sibling keeps the file's whole-file hash in `get_all_known`, so the file is never re-detected as "modified" and the failed batch is stranded until the user next edits the note. `_queue_drop` (the only path that reuses failed rows) runs only from change detection, so partial failures never auto-retried. (A *total* failure already auto-retries — excluded from `get_all_known` → re-detected.)

## Changes

- **`get_retriable_failure_files`** (new CRUD): files with a retriable failed row (`retry_count < max`, not `approval_invalidated:`) **and no in-flight (`pending`/`processing`) row**.
- **`_check_once_inner`** computes retry candidates (minus files already detected this scan) and merges them into the early-return guard; **`_phase_create_records`** gains a retry loop that re-derives the delta and re-queues via the existing `_queue_drop` (row-reuse) machinery. New `CheckResult.items_retried`.
- If the failed batch's URLs were **removed** from the file (empty delta), **`mark_file_failures_abandoned`** flips the stale rows to `approval_invalidated:` so the file stops being a candidate.

## Design decisions

- **Cooldown-exempt.** A retry is failure recovery, not a re-eval on a user edit, so it fires the next scan (natural ~interval spacing) rather than waiting a full cooldown.
- **Bounded (no retry-storm), proven by test:** `retry_count` is the load-bearing bound (a stranded batch retries in lockstep at most `max_retries` times, then drops out — verified `test_retry_is_bounded_by_retry_count`, incl. no row proliferation); the no-in-flight guard prevents double-queue; the `count_url_failures` storm guard is applied on the retry path too (defense-in-depth).
- **Known tradeoff:** the retry loop and the modified-files loop share some plumbing (read → delta → `_queue_drop`). They're kept separate — different guards, cadence, and empty-delta handling — rather than refactoring the already-deployed modified loop; noted for a possible future dedup.

## Verification

- genesis-architect pre-review (storm guard + cooldown cadence + altitude — incorporated/adjudicated).
- TDD, red-first: 4 CRUD + 4 monitor tests (headline auto-retry-without-edit, cooldown-exempt, storm-guard, retry-count bound).
- Multi-agent review (logic/bounds, signature/cross-file, altitude/SQL): the one "infinite-loop" candidate was refuted empirically (bound test); no missed callers; `CheckResult` field insertion is named-arg-safe.
- Full inbox suite: **281 passed** (with #817); ruff clean.
- Live-schema E2E: candidate-select / in-flight-exclude / abandon all correct against a copy of the real DB.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,12 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   repeatedly cancel and recreate that approval. Genesis now recognizes there's no new content and
   marks the note current in a single scan, so it settles instead of churning.
 
+- **Links that fail partway through an evaluation now retry themselves.** When only some of the
+  links in a note evaluate successfully and the rest fail (for example, a few get rate-limited),
+  the failed links used to sit untouched until you edited the note again. Genesis now
+  automatically re-attempts the stranded links on a later scan on its own — bounded so a link that
+  keeps failing eventually stops retrying rather than looping.
+
 - **Inbox evaluations no longer cram a whole batch of links into one giant pass — and stop
   re-evaluating links they've already covered.** When you drop many URLs into an inbox note,
   Genesis now evaluates them in small groups (≈5 at a time, configurable via
--- a/src/genesis/db/crud/inbox_items.py
+++ b/src/genesis/db/crud/inbox_items.py
@@ -484,6 +484,61 @@ async def get_retriable_failed_rows(
     return [dict(r) for r in await cursor.fetchall()]
 
 
+async def get_retriable_failure_files(
+    db: aiosqlite.Connection, *, max_retries: int = 3,
+) -> list[str]:
+    """Return file_paths that have a stranded retriable-failed batch to retry.
+
+    A file qualifies for partial-failure AUTO-RETRY when it has >=1 failed row
+    that is still retriable (``retry_count < max_retries`` and not
+    ``approval_invalidated:``) AND it has NO row currently ``pending`` or
+    ``processing`` (nothing in flight for it). Such files are otherwise stranded:
+    when only SOME batches of a drop fail, a completed sibling keeps the file's
+    hash in :func:`get_all_known`, so detection never re-surfaces the file and
+    the failed batch would only retry on the next user edit. The monitor uses
+    this to re-queue the failed batches independently of file-change detection.
+    Excludes approval-invalidated rows (those need a fresh approval, not a retry).
+    """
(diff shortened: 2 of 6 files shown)


WingedGuardian (commented):
## First-pass review: PR #824 — `fix(inbox): partial-failure auto-retry for stranded batches`

### Structural review

**WARNING — Raw SQL in test assertions (`tests/test_inbox/test_drop_dispatch.py`)**

Two new tests use raw SQL directly against the DB instead of going through `genesis.db.crud.inbox_items`:

```python
# test_partial_failure_auto_retries_without_edit (~line 174)
assert sorted(r["status"] for r in [dict(x) for x in await (await db.execute(
    "SELECT status FROM inbox_items WHERE file_path=?", (str(fp),))).fetchall()]) == ["completed", "failed"]

# test_retry_is_bounded_by_retry_count (~line 228)
n = (await (await db.execute(
    "SELECT COUNT(*) FROM inbox_items WHERE file_path=?", (str(fp),))).fetchone())[0]
```

The companion `test_crud_drop_batching.py` uses CRUD methods throughout (`get_by_id`, `get_retriable_failure_files`, etc.). If `inbox_items` doesn't already expose a "get all rows for file" or "count rows for file" helper, these would be candidates for a thin CRUD function — or at minimum the raw SQL should be pulled into a local helper inside the test module so the schema dependency is localized.

This is a WARNING, not a block — the queries are read-only assertion aids in test code, but they bind the test directly to table/column names and bypass any future schema evolution in the CRUD layer.

---

**CLEAN** — no API keys, tokens, private IPs, SSH keys, hardcoded credentials, personal email addresses, or home-directory paths with usernames found in any added lines. PR title, description, and CHANGELOG entry contain no incident/leak/credential language.

---

**Summary:** one structural WARNING (raw SQL in test assertions); no errors, no PII/secrets issues, no wording issues. No blocking findings.

---

WingedGuardian:
Merge note: Codex review quota-limited. Merging on all-green CI + genesis-architect pre-review + local multi-agent review (infinite-loop candidate refuted empirically by test_retry_is_bounded_by_retry_count) + live-schema E2E. Rebased onto main-with-#821; 288 inbox tests green with all three PRs. Per the quota-exhausted merge path.

PR #825, opened 2026-07-01
fix(inbox): abandon partial-failure retry candidates that can never be retried

WingedGuardian:
## What

Follow-on to the inbox series (after #817/#821/#824). PR-2c's partial-failure retry loop selects candidates from the DB (files with a retriable `failed` row + no in-flight row). Two **terminal** states left a candidate recurring **every scan forever** — a `WARNING` + wasted query/read each ~30-min scan:

- **source file DELETED** → the read raised `FileNotFoundError` → warn + skip.
- **source file EMPTIED** → `content.strip()` empty → skip.

Observed live post-#824 deploy for two deleted notes (`Untitled.md`, `Verify claims & research.md`) — 27 stranded failed rows re-logging every scan.

## Change

Both terminal states now **abandon** the file's stranded retriable-failed rows via `mark_file_failures_abandoned` (`reason="source file deleted"` / `"source file is empty"`), flipping them to `approval_invalidated:` so the file drops out of `get_retriable_failure_files` — stopping the recurrence. This mirrors the existing empty-delta ("content removed before retry") abandon. `mark_file_failures_abandoned` gains a `reason` kwarg (default unchanged) so all three sites store an accurate reason.

`PermissionError` is deliberately **not** abandoned (possibly transient — a later scan's read may succeed; keep warn + skip). If an abandoned note is ever re-created it's detected as new and evaluated fresh (abandoned rows are excluded from `get_all_known`).

Internal log-noise/hygiene fix — no user-visible behavior change, **no CHANGELOG**. No schema, no migration.

## Verification

- TDD, red-first: `test_retry_candidate_vanished_file_is_abandoned` (proven red — the file stayed a candidate) + `test_retry_candidate_empty_file_is_abandoned`.
- Multi-agent review (line-by-line/SQL + behavior/edges): SQL param order verified; single existing caller confirmed; the empty-file gap the review surfaced is **fixed here** (broader-class completion); the storm-guard skip is correctly left alone (it's a *temporary* 48h guard, not a terminal state).
- Full inbox suite: **290 passed**; ruff clean.
- Live-DB E2E: the two real stale candidates (27 rows) are abandoned and drop out of the retry set against a copy of production.

### DIFF
--- a/src/genesis/db/crud/inbox_items.py
+++ b/src/genesis/db/crud/inbox_items.py
@@ -515,24 +515,25 @@ async def get_retriable_failure_files(
 
 async def mark_file_failures_abandoned(
     db: aiosqlite.Connection, file_path: str, *, max_retries: int = 3,
+    reason: str = "content removed before retry",
 ) -> int:
     """Mark a file's retriable failed rows as approval-invalidated (abandoned).
 
-    Used when a retry candidate's failed content is no longer present in the
-    file (the user removed those URLs before the retry ran): the recomputed
-    delta is empty, so there is nothing to re-queue, but the retriable-failed
-    rows would otherwise keep the file a retry candidate forever. Flipping them
-    to the ``approval_invalidated:`` prefix excludes them from
-    :func:`get_retriable_failure_files` / :func:`get_retriable_failed_rows`.
-    Returns the number of rows updated.
+    Used when a retry candidate can never be retried again — either its failed
+    content was removed from the file (empty delta; the default ``reason``), or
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

WingedGuardian:
Merge note: Codex review quota-limited. Merging on all-green CI + local 2-finder review (SQL param order verified, single caller confirmed; a broader-class empty-file gap surfaced by review was fixed in-PR; storm-guard correctly left alone) + live-DB E2E (27 stale rows self-heal). Per the quota-exhausted merge path.

PR #826, opened 2026-07-01
fix(learning): scope outcome classifier to Genesis-attempted tasks

WingedGuardian:
## What & why

A foreground user status-update on Telegram — the user reporting how their **own external projects and plans** were going, plus a forward-looking intent ("let's keep going… it's never too late") — was classified by `31_outcome_classification` as `approach_failure`. The classifier treated the user's real-world statuses about their own external projects, and their forward intent, as *Genesis's own failed goals*.

On a foreground turn, that single mislabel cascades into four harms:

- the `STEERING.md` auto-write (origin of the STEERING-corruption incident),
- a false `record_correction` autonomy penalty (docks earned autonomy),
- legacy procedure extraction (spurious procedure rows),
- a curiosity drive-weight nudge.

## The fix

Prompt-only change to `_build_prompt` in `src/genesis/learning/classification/outcome.py`. **The `_parse_response` hard gate that guards partial-completion (FM1) is untouched.**

- Scope "goal" to a concrete task the user asked Genesis to do *this turn* **and that Genesis then attempted**.
- Explicitly exclude (a) the user's real-world statuses about their own external projects and (b) forward-looking intentions — neither belongs in `goals_failed`.
- Summarizing, proposing options, or asking a clarifying question *before acting* is `success`, not a failure.
- Two contrastive worked examples anchor the boundary (status → success vs. genuine partial → approach_failure).

## Validation

Spike-gated against the **real router** (glm51 primary, N=5, 8 cases):

| case | before | after |
|---|---|---|
| incident (status + forward intent) | approach_failure ×5 | **success ×5** |
| genuine partial completion | approach_failure ×5 | approach_failure ×5 |
| wrong-tool-then-corrected | workaround_success ×5 | workaround_success ×5 |
| clean status / clean success | success | success |
| different-domain status | success (flaked once) | success ×5 |
| **mixed: status + a failed task** | approach_failure ×5 | **approach_failure ×5** (status half does not short-circuit the failure) |
| mixed: status + a done task | success | success |

- Live E2E on the shipped file against the real interaction: `success` 5/5.
- `ruff` clean; new prompt-guard test added; full `tests/test_learning/` suite green (633 passed, 3 skipped).
- Reviewed by an automated code-reviewer (0 critical); the two "tighten" items it raised (success-class wording for mixed turns, test-guard brittleness) are addressed here.

## Regression markers

After deploy, the `approach_failure` rate should *ease* toward more plausible levels but **not** collapse to ~0 — genuine task failures (including those mixed into a status message) must still register. Falsified if a known real Genesis-task failure starts logging as `success`.


WingedGuardian (commented):
**Structural review: CLEAN — no pattern violations found.**

---

**SOFT WARNING — Verbatim private message fragment in test docstring**

`tests/test_learning/test_classification.py` embeds the actual Telegram message that triggered the incident:

```python
The 2026-06-30 incident: a Telegram status-update ("Autonomize is dead…
but we should continue… its never too late") was mislabeled ...
```

If "Autonomize" is a private project, company, or product name that isn't already public, this becomes part of the permanent public git history on merge. The technical context (why the message mislabeled `approach_failure`) doesn't require the verbatim text — a paraphrase like `"a status update about a project and a forward-looking intent"` would make the same point without disclosing the name.

Nothing to fix if this is intentional or already public — just flagging so it's a conscious choice rather than an oversight.

---

WingedGuardian:
Codex automated review was **quota-limited** (usage limit reached), so no automated findings were posted. The change was reviewed inline by a code-reviewer agent (0 critical; 2 'tighten' items — mixed-turn success-class wording and a brittle test guard — both addressed in this PR) and empirically spike-validated against the real router (8 cases × N=5) plus a live E2E. Merging on CI + inline review per the pre-merge protocol.

PR #827, opened 2026-07-01
feat(surplus): grade autonomous insight tasks with an LLM quality judge

WingedGuardian:
## What

Background "surplus" tasks (brainstorms, audits, anticipatory research) run
autonomously and feed insights into the knowledge base. Until now the system
only recorded whether such a task *ran to completion* — a signal that is almost
always positive and says nothing about whether the produced insight was any good.

This adds a measurement-only quality judge that grades each completed insight
task's output using the existing evaluation LLM-judge (the `output_quality`
rubric) and records the verdict per task:

- **useful** — the judge passed the output
- **hollow** — the judge scored it below threshold
- **null** — no verdict (judge unavailable, non-insight task, or too-short output)

A `hollow` verdict surfaces as a verification-failed signal alongside the usual
execution-outcome, giving the system a genuine two-sided quality signal about its
own autonomous work instead of a near-constant "it ran" positive.

## Design

- **Measurement-only.** The judge does not change how insights are stored or
  routed — it only observes quality. Trusted sources stay trusted.
- **Outage-safe.** A judge-provider failure yields a null verdict, never a false
  negative, so an outage can't flood the signal with fake failures.
- **Never fails the task.** The judge runs after completion and never raises; a
  judge problem cannot fail a task that otherwise completed.

## Changes

- New `surplus/quality_judge.py` with per-task-type grading intent for the 11
  insight-producing task types.
- New `judge_score` / `judge_detail` columns on `surplus_tasks` (migration 0043)
  storing the continuous score + rationale for later calibration and display.
- Insight-task completion grades the full output and records the verdict; the
  harvester's verification-failed path is now reachable for the first time.

## Verification

- Unit tests for the judge (verdict mapping, the outage guard, task-type gating,
  the short-content skip), the migration (idempotent, additive), and the
  scheduler wiring.
- End-to-end with the real judge: two-sided on real and control content, the
  verdict persists, and the harvester emits the expected outcome signals
  (idempotent across re-runs).

## Notes

The per-task-type grading prompts are a first pass, and the signal is not yet
used to drive any behaviour — that stays behind a default-off switch pending
per-domain calibration.

### DIFF
--- a/src/genesis/db/crud/surplus_tasks.py
+++ b/src/genesis/db/crud/surplus_tasks.py
@@ -74,11 +74,15 @@ async def mark_completed(
     completed_at: str,
     result_staging_id: str | None = None,
     outcome_quality: str | None = None,
+    judge_score: float | None = None,
+    judge_detail: str | None = None,
 ) -> bool:
     cursor = await db.execute(
         "UPDATE surplus_tasks SET status = 'completed', completed_at = ?, "
-        "result_staging_id = ?, outcome_quality = ? WHERE id = ?",
-        (completed_at, result_staging_id, outcome_quality, id),
+        "result_staging_id = ?, outcome_quality = ?, judge_score = ?, "
+        "judge_detail = ? WHERE id = ?",
+        (completed_at, result_staging_id, outcome_quality, judge_score,
+         judge_detail, id),
     )
     await db.commit()
     return cursor.rowcount > 0
--- a/src/genesis/db/migrations/0043_surplus_judge_score.py
+++ b/src/genesis/db/migrations/0043_surplus_judge_score.py
@@ -1,0 +1,44 @@
+"""Add judge_score + judge_detail columns to surplus_tasks (quality judge).
+
+The ``outcome_quality`` verdict on insight-producing surplus tasks is now set by
+a measurement-only LLM quality judge (``surplus.quality_judge``) rather than the
+old intake-routing heuristic (which was structurally unreachable — curated
+surplus sources skip scoring and route at a fixed 0.6 confidence, so intake
+never discarded everything, so 'hollow' could never fire). ``outcome_quality``
+keeps its meaning at the *bus* level — ``useful`` / ``hollow`` / NULL — but now
+means "judge passed / judge failed / no verdict" instead of
+"intake kept something / intake discarded everything / n-a".
+
+This migration adds the two continuous, human-readable columns behind that
+binary verdict, for calibration and display (they are NOT read by the Outcome
+Bus harvester, which only branches on ``outcome_quality``):
+
+  - ``judge_score``  REAL : the judge's [0, 1] quality score.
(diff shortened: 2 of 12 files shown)


WingedGuardian:
## First-pass review — PR #827

### Structural review

**ERRORs:** none

**WARNING — `src/genesis/feedback/harvest.py`**

The `reason_text` string in `_harvest_surplus_tasks` changed:

```
- reason_text="intake routed all findings to discard (hollow)"
+ reason_text="quality judge scored output below threshold (hollow)"
```

If `tests/test_surplus/test_harvest.py` or any other harvest test file has a string assertion against the old value, that test will silently pass here but break if the assertion is pattern-matched. The diff doesn't include any harvest test updates. Please check whether any existing harvest tests need to be updated to reflect the new reason string.

**NOTEs:** none

---

**CLEAN** — no API keys, private IPs, credentials, personal emails, absolute paths with usernames, private hostnames, or sensitive wording in PR title/body found on any added lines.

---

WingedGuardian (commented):
**Structural review: 2 NOTEs — no ERRORs or WARNINGs**

---

**NOTE 1 — Dict-style row access in `test_scheduler.py` may fail without `row_factory`**

The new `_dispatch_with_judge` helper returns `cursor.fetchone()` and all four callers access it with string keys:

```python
row, judge = await _dispatch_with_judge(...)
assert row["outcome_quality"] == "useful"
assert row["judge_score"] == 0.82
```

`aiosqlite.fetchone()` returns plain tuples unless `connection.row_factory = aiosqlite.Row` (or `sqlite3.Row`) is set on the connection. The pre-existing `_completed_verdict` helper uses integer-index access (`row[0]`), which doesn't need `row_factory` — suggesting the `db` fixture may not configure it. If so, each new test using `_dispatch_with_judge` will raise `TypeError: tuple indices must be integers or slices, not str` at the first dict-key access.

Worth confirming the `db` fixture (in `conftest.py` or top of `test_scheduler.py`) sets `row_factory` before the new tests land. If it doesn't, swapping to tuple-index access (`row[0]`, `row[1]`, `row[2]`) in `_dispatch_with_judge` is the safe fix.

---

**NOTE 2 — New params on `mark_completed` in `src/genesis/db/crud/surplus_tasks.py` lack docstring coverage**

`judge_score` and `judge_detail` were added to the signature but no docstring update is visible in the diff. If the function has an existing docstring (not shown in the diff), it should be extended to document the two new parameters and their semantics (continuous [0,1] score and JSON rationale, persisted for calibration/display, not read by the harvester).

---

No pattern violations, no PII or secrets in added lines, PR title/body/changelog wording is clean.

---

WingedGuardian:
Re: the harvest `reason_text` WARNING — checked, no test asserts that string. Harvest tests (`tests/feedback/test_harvest.py`) assert signal-type counts and polarity, not the reason text, and the `hollow` → verification-failed path is covered by count assertions. The full feedback + surplus + db suites pass, so no harvest test needed updating.

PR #830, opened 2026-07-01
fix(autonomy): fail-closed dispatch gate + drop web_agent cost auto-deny (WS5 Stage 1)

WingedGuardian:
## What & why

Stage 1 of the WS5 autonomy-gating hardening. Closes two fail-OPEN holes in the proposal dispatch gate and removes a cost-based auto-deny that violated Design Principle 3 (cost is observability, never automatic control).

### Fail-closed dispatch gate (risk-tiered)

The dispatch gate is the last safety check before an already-approved ego proposal runs autonomously in the background. Two ways it could silently fail open:

- **Init failure → gate unset.** If `ProposalDispatchGate` couldn't be built, the runtime attribute stayed unset and the ego wired it as `None` → every approved proposal dispatched **ungated**. Now a fail-closed `DenyHighRiskSentinel` is installed as the default *before* the real gate is built, so any failure path leaves a blocking gate. Init errors are surfaced (`autonomy.gate_init_failed`, ERROR). The post-execution auditor is observability (not a gate) → visibility upgrade only, no deny.
- **Gate-eval exception → allow.** The two ego dispatch except-blocks previously fell through to ALLOW on a gate error. Now **risk-tiered**: BLOCK external-write / represent-user / financial / self-modify domains (fail-closed), stay resilient (allow) for benign domains so a transient error can't wedge the ego cycle. Shared, dependency-free helpers live in `autonomy/proposal_gate.py`.

### web_agent (Design Principle 3)

`web_agent`'s daily-budget-EXCEEDED check auto-denied the tool. Cost is observability, never control — it now logs a warning and proceeds.

## Testing

- New unit tests: sentinel + risk-tier helpers + `init()` installs the sentinel when the gate can't be built.
- Full changed-area regression green (autonomy, ego, runtime, web tools).
- Runtime E2E: a real `init()` installs the blocking sentinel; sentinel/helper decisions verified with real objects.
- architect-reviewed; code-reviewed (a fail-open `prop_row`-unbound bug in the except handler was caught and fixed).

## Scope

Stage 1 of 3. No migration. Deferred to later stages: the CI external-I/O guard (a raw-POST grep proved a weak proxy), and the actual gating of the direct discord/poll/medium/reply publish paths (which carries a live-behavior decision).

### DIFF
--- a/src/genesis/autonomy/proposal_gate.py
+++ b/src/genesis/autonomy/proposal_gate.py
@@ -156,3 +156,85 @@ class ProposalDispatchGate:
                 highest = ProtectionLevel.SENSITIVE
 
         return highest if highest != ProtectionLevel.NORMAL else None
+
+
+# ---------------------------------------------------------------------------
+# Fail-closed fallback (Stage-1 hardening)
+# ---------------------------------------------------------------------------
+# The dispatch gate is the last safety check before an already-user-approved
+# proposal runs autonomously in the background.  Previously, if the real gate
+# could not be built (init failure) or raised while evaluating, the caller
+# SILENTLY dispatched — a fail-OPEN hole.  These primitives give a
+# deterministic, dependency-free risk tier so a degraded gate fails CLOSED for
+# high-consequence domains while staying resilient (allow) for benign ones —
+# the latter avoids wedging the whole ego proposal cycle on a transient error.
+
+#: Domains that act on outside parties or on Genesis's own code/identity.  This
+#: is exactly the set whose ``ACTION_DOMAIN_MIN_LEVEL`` is >= 2 (or ``None``),
+#: i.e. the domains the real gate would block below the earned autonomy level.
+#: KEEP IN SYNC when adding an ``ActionDomain``: ``classify_domain``'s fallback
+#: is ``EXTERNAL_READ`` (benign), so an UNMAPPED action is ALLOWED under a
+#: degraded gate — any new external/self-modify domain MUST be added here.
+HIGH_RISK_DOMAINS: frozenset[ActionDomain] = frozenset({
+    ActionDomain.EXTERNAL_WRITE,
+    ActionDomain.REPRESENT_USER,
+    ActionDomain.FINANCIAL,
+    ActionDomain.SELF_MODIFY,
+})
+
+
+def is_high_risk_domain(domain: ActionDomain) -> bool:
+    """True if *domain* touches external parties or Genesis's own code/identity."""
+    return domain in HIGH_RISK_DOMAINS
+
+
+def gate_failure_is_blocking(proposal: dict) -> bool:
+    """Decide fail-closed (True) vs resilient-allow (False) when the dispatch
+    gate is unavailable or raised while evaluating *proposal*.
+
+    Re-derives the action domain from the proposal via the PURE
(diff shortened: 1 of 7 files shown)


WingedGuardian (commented):
## First-pass review: structural + PII/wording scan

No API keys, private IPs, SSH keys, or credentials in added lines. PR title/body describe architectural hardening with no language implying a PII leak, credential exposure, or security incident. Nothing to flag.

---

### Structural review

**ERROR:** none

**WARNING — Missing test coverage for `web_tools.py` behavior change**

`src/genesis/mcp/health/web_tools.py` line ~594: the `EXCEEDED` branch was changed from a hard `return {"error": "Daily budget exceeded..."}` to a log-and-proceed. If there are existing tests that assert the error-dict return on budget-exceeded, they are now stale and will silently pass green only if mocked to never hit this branch. No `web_tools.py` test was added or updated in this PR to cover the new proceeding behavior. Worth verifying the existing `test_mcp_health` or `test_web_tools` suites either (a) exercise this path and have been updated, or (b) were already mocked past it and can't regress.

---

**NOTE — `gate_failure_is_blocking` type hint should be `dict | None`**

`src/genesis/autonomy/proposal_gate.py` line ~192: the signature is `proposal: dict` but `test_gate_failure_blocks_on_undeterminable_proposal` explicitly passes `None` and asserts `True`. The function handles `None` safely (the `AttributeError` from `None.get(...)` is caught by the blanket `except Exception` and returns `True`), but the type hint misrepresents the contract. The `prop_row is None or gate_failure_is_blocking(prop_row)` guard in `session.py` prevents `None` reaching it in production, but since the tested contract includes `None` → `True`, the hint should be `dict | None`.

---

**NOTE — Deferred imports inside `except` blocks in `session.py`**

`src/genesis/ego/session.py` lines ~1456 and ~1487 (and the `_sweep_approved_inner` path): `from genesis.autonomy.proposal_gate import gate_failure_is_blocking` and `from genesis.observability.types import Severity, Subsystem` are imported inside the `except` handlers rather than at the module top. If this is guarding against a circular import, a one-line comment explaining that would make it clear this is intentional. If there's no circular dependency, moving them to the top would be idiomatic. As-is, the intent is ambiguous to a future reader.

---

**Passing items (for the record)**
- No bare `except:` — all handlers use `except Exception:` ✓
- `HIGH_RISK_DOMAINS` comment ("KEEP IN SYNC when adding an `ActionDomain`") correctly flags the maintenance obligation ✓
- Sentinel pre-install before the real gate build is the right order; the comment in `runtime/init/autonomy.py` makes the invariant explicit ✓
- Both new test files in `tests/test_autonomy/` and `tests/test_runtime/` cover the sentinel + init path ✓
- Conventional commit prefix `fix(autonomy):` ✓

---

WingedGuardian:
Review coverage note: the automated Codex review was quota-limited (see above). This PR was covered by a genesis-architect design review + a code-reviewer pass that caught & fixed a fail-open `prop_row`-unbound bug (`a4ad64eb`). The first-pass structural review's 1 WARNING + 2 NOTES (0 ERROR) were addressed in `6fa91a4d` (web_agent budget test, `dict|None` hint, top-level import). Full CI green. Merging via --admin on CI + manual review.

PR #829, opened 2026-07-01
feat(attention): PR2 — dashboard Attention review/label tab (calibration cockpit)

WingedGuardian:
## What

PR2 of the passive-listening **attention engine** (Track 1 of the omnipresence layer). PR1 (#820) shipped the deterministic L1 shadow gate that persists "would have perked up" decisions to `attention_events` (refs + features + labels only — **never transcript text**). This PR adds the **calibration cockpit**: an auth-gated dashboard "Attention" tab to review each shadow decision, reveal the window's transcript on demand, and label it **should / shouldn't / skip** — closing the calibration loop that PR3's retune consumes.

## Why

The shadow table had no way for a human to say whether a fire was *right*, so the gate couldn't be tuned. PR1's live run already showed `multi_speaker` fires on 321/326 windows (over-dominant); this tab surfaces that per-trigger dominance and turns the shadow corpus into **labeled** ground truth.

## Changes

- **`db/crud/attention.py`** — label-preserving `bulk_upsert_events` (`ON CONFLICT DO UPDATE … WHERE acceptance_signal IS NULL`, so re-running the runner never clobbers a human label); `list_events` (activation / trigger / is_user / unlabeled filters via `json_each` **exact-name** match, not `LIKE`); `get_event`; `update_acceptance_signal` (returns the prior value); aggregates (activation / trigger / suppressor / label counts).
- **`attention/sources.py`** — `resolve_window_text`: reveals a window's transcript from the **transient snapshot** (read-only, NOT genesis.db); parametrized `IN()` with `int()`-cast ids; `None` → 410 when the snapshot is purged.
- **`attention/runner.py`** — canonicalize `snapshot_id` for `--snapshot` (strip the `ambient_` prefix) so the dashboard reveal can reconstruct the file.
- **`dashboard/routes/attention.py`** (+ registration) — `list` / `stats` / `reveal-text` / `label`, mirroring `references.py` (auth gate, `GenesisRuntime` guard). list/stats are value-free; `reveal-text` is the only text path and persists nothing.
- **`dashboard/templates/genesis_dashboard.html`** — the Attention tab: stats bar + per-trigger dominance bar + filters + per-event reveal/label.

## Firewall

Transcript text is never written to genesis.db. `list`/`stats` never open a snapshot. `reveal-text` reads the transient snapshot live for the authenticated human and stores nothing. A firewall test asserts no transcript text reaches `attention_events`.

## Verification

[...]

PR #831, opened 2026-07-01
fix(attention): enable label buttons — strict-boolean :disabled binding

WingedGuardian:
## What

Follow-up to #829 (Attention review tab). **Live browser E2E** revealed the should/shouldn't/skip label buttons rendered permanently **disabled**.

## Root cause

In this Alpine build, `:disabled="$store...attentionLabeling[ev.id]"` where the key is unset evaluates to `undefined`, which Alpine treats as truthy for the boolean `disabled` attribute (renders `disabled="disabled"`). Coercing to a strict boolean fixes it:

```
:disabled="$store.genesisDashboard.attentionLabeling[ev.id] === true"
```

The `attentionLabeling` map only ever stores `true` (while a label POST is in flight), so `=== true` is exact.

## Verification (live)

Swapped the fixed template onto the running dashboard and confirmed via the browser: all label buttons **enabled at rest**, disable during the in-flight request, and labels persist to `attention_events` (verified 2 rows written via real clicks). Reveal + list + stats + the per-trigger dominance bar all render correctly. Template-only change (no restart/migration).

### DIFF
--- a/src/genesis/dashboard/templates/genesis_dashboard.html
+++ b/src/genesis/dashboard/templates/genesis_dashboard.html
@@ -10701,11 +10701,11 @@
                   <span class="material-symbols-outlined" style="font-size:1rem; vertical-align:middle;"
                         x-text="$store.genesisDashboard.attentionRevealed[ev.id] !== undefined ? 'visibility_off' : 'visibility'"></span>
                 </button>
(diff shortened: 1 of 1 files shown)


WingedGuardian (commented):
Structural review: CLEAN — template-only change (HTML), no Python structural rules apply.

---

PR #833, opened 2026-07-01
fix(cc): break cc.contingency <-> awareness.loop circular import

WingedGuardian:
## What & why

A latent circular import between `genesis.cc.contingency` and `genesis.awareness.loop`. Importing `cc.contingency` first (e.g. running the `test_quota_exhausted_triggers_contingency` test in isolation) raised:

```
ImportError: cannot import name 'RATE_LIMIT_DEFERRAL_TTL_S' from partially initialized module 'genesis.cc.contingency'
```

It reproduces on clean `main` and is order-dependent — the full test suite masks it because an earlier test imports `awareness.loop` first. Found during unrelated merge verification.

### The cycle

- `cc/contingency.py` imports `Depth` from `awareness.types` (legitimate; used at module load).
- Touching the `awareness` package ran `awareness/__init__.py`, which **eagerly** imported `AwarenessLoop` from `awareness.loop`.
- `awareness/loop.py` imported `RATE_LIMIT_DEFERRAL_TTL_S` back from the still-initializing `cc.contingency` — a **layering inversion** (a low-level heartbeat reaching up into the CC dispatcher for a CC-domain constant).

## The fix (two independent break-points, both root-cause)

1. **Relocate the shared constant** to a new stdlib-only leaf `cc/constants.py`. `awareness.loop` and `cc.reflection_bridge._bridge` now import it from there instead of `cc.contingency`, removing the inverted `awareness -> cc` edge. `cc.contingency` re-imports the name, so its namespace (and any `from cc.contingency import RATE_LIMIT_DEFERRAL_TTL_S`) is unchanged.
2. **De-eager `awareness/__init__.py`** — `AwarenessLoop` / `JobRetryRegistry` are now exposed via a PEP 562 module-level `__getattr__` (lazy, cached). Importing the leaf types no longer drags in `loop.py`'s heavy async machinery. The documented public API (`from genesis.awareness import AwarenessLoop`) is preserved.

Constant value (14400) and all public behaviour are unchanged. No migration.

## Validation

- Fresh-process import probes: importing `cc.contingency` first now succeeds (was failing); `loop`-first OK; `from genesis.awareness import AwarenessLoop, Depth, JobRetryRegistry, SignalReading, TickResult` OK (lazy API preserved); `from genesis.awareness import loop, scorer` (submodule) OK under lazy init; unknown attribute still raises `AttributeError`; the constant is identical from both `cc.constants` and `cc.contingency`.
- `ruff` clean; `tests/test_cc/` + `tests/test_awareness/` = **598 passed, 2 skipped**. The previously-erroring `test_quota_exhausted_triggers_contingency` passes in isolation.
- Reviewed (automated code-reviewer): SHIP; the one nit raised (cache the lazily-resolved symbol per canonical PEP 562) is applied.

## Regression markers

If the fix is wrong, a fresh-process import of `cc.contingency` (or `awareness.loop`) would `ImportError` again, or `from genesis.awareness import AwarenessLoop` would stop resolving. Watch for either after deploy.

### DIFF
--- a/src/genesis/awareness/__init__.py
+++ b/src/genesis/awareness/__init__.py
@@ -1,15 +1,19 @@
-"""Genesis Awareness Loop — the system's heartbeat.
+"""Genesis Awareness package.
 
-Public API (import from genesis.awareness directly):
-- AwarenessLoop: Main loop class — periodic signal collection and depth-based ticking
-- JobRetryRegistry: Retry tracking for failed awareness jobs
-- SignalReading: Data class for individual signal measurements
-- Depth: Enum for tick depth levels (MICRO, LIGHT, DEEP, STRATEGIC)
-- TickResult: Result of a single awareness tick
+Public API:
+- AwarenessLoop: main heartbeat loop (periodic signal collection + depth ticks)
+- JobRetryRegistry: retry tracking for failed awareness jobs
+- Depth, SignalReading, TickResult: lightweight leaf types
+
+AwarenessLoop and JobRetryRegistry are exposed LAZILY (PEP 562 module __getattr__)
+so importing the leaf types — or any module that transitively touches this package
+(e.g. cc.contingency importing Depth) — does NOT eagerly load loop.py and its heavy
+async machinery. Eager-loading loop.py here previously created an import cycle
+(cc.contingency -> awareness.types -> this __init__ -> loop -> cc.contingency). For
+eager access, import from the submodules directly (genesis.awareness.loop /
+genesis.awareness.job_retry).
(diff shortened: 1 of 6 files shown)


WingedGuardian:

**NOTE** — `src/genesis/cc/constants.py` is a new `.py` file under `src/` with no corresponding test file in `tests/`. The module contains only a single constant with no logic, so a test would be `assert RATE_LIMIT_DEFERRAL_TTL_S == 14400` — low value, not blocking. Tracking per the missing-coverage convention.

---

PR #834, opened 2026-07-01
fix(install): align container Claude Code to the pinned version on install/bootstrap/update

WingedGuardian:
## Summary

The pinned Claude Code version (`CC_VERSION` in `scripts/lib/cc_version.sh`) could silently fail to reach an install. Two gaps:

1. **`install.sh` and `host-setup.sh` only installed Claude Code when it was *absent*** — a present-but-stale CC was never upgraded, so bumping the pin never reached an already-installed machine.
2. **`update.sh` synced only the *host VM's* Claude Code** (via the guardian `update-cc` op) — nothing aligned the container's own CC (update.sh runs inside the container).
3. `bootstrap.sh` didn't handle the CC version at all.

Net effect: a machine could run an old Claude Code indefinitely despite the repo pin.

## Fix

- Add a shared **`cc_ensure_local`** to `scripts/lib/cc_version.sh` that installs *or* aligns the local Claude Code to the pin:
  - Exact-match to the pin (so an intentional **downgrade** for incident rollback also applies; `npm install @X.Y.Z` pins exactly).
  - On **upgrade**, reinstalls to the **existing binary's own prefix** (`dirname(dirname($(command -v claude)))`), not `npm config get prefix` — the two can differ, which would install a second copy and leave `which claude` stale.
  - System prefixes use `sudo env "PATH=$PATH" …`; user prefixes don't. Command built as a bash array (quoting-safe).
  - **Post-install re-verify** (`hash -r` + re-read `claude --version`); returns non-zero + warning if still off-pin.
  - Idempotent, **non-fatal**.
- Call it from `install.sh` (Step 12), `bootstrap.sh` (after Node), and `update.sh` (a new **unconditional** container-sync block after the guardian block, `|| true` — `set -e` is active, so a CC hiccup must never abort an update). The host CC still syncs via the guardian op.

## Testing

- `scripts/spike_cc_ensure_local_test.sh` — stubbed `npm`/`claude` in a curated-PATH sandbox (never touches the real system), **7/7**: npm-absent skip, at-pin no-op, drift(user-prefix) install, fresh install, version-suffix parse, **sudo/system-prefix branch**, **post-install verify-fail → return 1**.
- `shellcheck` clean; live idempotency no-op verified; each script's exact block sequence verified end-to-end against the real environment (no-op at pin).

## Docs

- `docs/reference/cc-compatibility.md` governance model updated; user-facing CHANGELOG entry under Fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Genesis now keeps your Claude Code CLI at the version it's tested against — automatically.**
+  Previously the installer only put Claude Code in place when it was *missing*, so if you already
+  had an older Claude Code, bumping the pinned version never actually upgraded you — you'd silently
+  keep running the old one. Now `install.sh`, `bootstrap.sh`, and `update.sh` all install *or* align
+  Claude Code to the pinned version on every run (matching it exactly, so an intentional rollback
+  also applies), with no manual step. It's non-fatal: if the update can't run (e.g. no permissions),
+  your update still completes and Claude Code is left as-is.
 - **A casual message can no longer be mistaken for a permanent "hard rule" — and your
   steering-rules file keeps its structure.** Genesis auto-adds a steering rule only when you
   actually give it a terse directive ("stop doing X", "never do Y"); an ordinary status update
--- a/docs/reference/cc-compatibility.md
+++ b/docs/reference/cc-compatibility.md
@@ -16,8 +16,16 @@
 ## Current CC Version
 
 **Installed:** Claude Code 2.1.173 on the **container** (upgraded 2026-06-11 from 2.1.170 — Fable 5 `[1m]` model-ID normalization fix + 2.1.172 long-conversation render perf). The **host VM** runs **2.1.87** — deliberately held there after the 2.1.90 Linux scrollback regression (see the 2.1.90 row below). 2.1.173 (the fullscreen-renderer scrollback fix) is the version the host is now being moved to via the unified updater. **Both** container and host install Claude Code **via npm-global** (`npm install -g @anthropic-ai/claude-code@<version>` — the container auto-detects its npm prefix, the host uses `sudo npm install -g`). There is no native-installer path.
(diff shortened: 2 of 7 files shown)


WingedGuardian:

---

**Scope:** `scripts/lib/cc_version.sh`, `scripts/install.sh`, `scripts/bootstrap.sh`, `scripts/update.sh`, `scripts/spike_cc_ensure_local_test.sh`, `docs/reference/cc-compatibility.md`, `CHANGELOG.md` — shell + docs only, no Python files.

**Structural checks run:**
- `/tmp/` usage: test script correctly uses `$HOME/tmp` with an explicit comment explaining why (watchgod policy) ✓
- `nohup` / bare process starts: none ✓
- `git add .` / `git push --force`: not present ✓
- `task_states` direct inserts: not applicable ✓
- Conventional commit prefix: `fix(install):` ✓
- `IntervalTrigger` / `$CLAUDE_PROJECT_DIR` / hook-import patterns: not applicable ✓

**PII/Secrets scan (added lines only):**
- No API keys, tokens, or credentials ✓
- No private IPs (10.x, 192.168.x, Tailscale CGNAT) or IPv6 ULA addresses ✓
- No private hostnames or hardware identifiers ✓
- Hardcoded paths in install/bootstrap scripts: exempt per known-false-positives rule ✓

**PR wording scan:**
- Title, body, and CHANGELOG additions contain no language referencing PII leaks, credential exposure, or private infrastructure disclosure ✓

---

WingedGuardian:
## First-pass review — PR #834

### Structural review: CLEAN (1 NOTE)
No pattern violations (ERROR). Shell-script-only change — Python structural rules (import hygiene, bare `except:`, raw SQL, etc.) do not apply. `/tmp/` rule: compliant — test uses `mktemp -d -p "$HOME/tmp"` with an explicit rationale comment. Non-fatal callsites (`cc_ensure_local || true`) are correctly guarded in `bootstrap.sh` and `update.sh` where `set -e` is active.

---

**NOTE — `spike_` prefix on a merged test fixture** (`scripts/spike_cc_ensure_local_test.sh`)

The `spike_` prefix conventionally names a throwaway/exploratory script (one-shot verification, then delete). If `spike_cc_ensure_local_test.sh` is meant to be a **permanent** regression guard that lives in the repo and gets re-run on future `cc_version.sh` changes, consider renaming it (e.g. `test_cc_ensure_local.sh`) to signal intent. If it was a one-run validation that already passed, it would typically be deleted post-merge rather than committed. Either direction is fine — the current name just leaves the intent ambiguous for the next person who runs `ls scripts/`.

---

No secrets, private IPs, personal emails, absolute paths with usernames, or hardware identifiers found in added lines. PR title, body, and CHANGELOG entry describe the fix functionally with no language suggesting a secrets/PII cleanup or infrastructure exposure.

---

PR #836, opened 2026-07-01
feat(autonomy): WS5 Stage 2 — Discord capability shadow-gate (observe-only, all 3 doors)

WingedGuardian:
## WS5 Stage 2 — Discord capability SHADOW-gate (observe all 3 doors, hold nothing)

Autonomous Discord posting can currently reach the outside world through **three ungated
doors**. This PR installs an **observe-only** capability shadow-gate at every door: it
records what a capability gate *would* decide (hold vs allow) **without holding anything**
— gathering the volume/pattern data needed to choose the right enforcement posture, with
**zero behaviour change**.

> This **instruments** the doors; it does not **close** them. Enforcement (actual
> hold-for-approval) is a separate follow-on stage, informed by this shadow data.

### The three doors (exhaustively enumerated)
| Door | Path | Cell |
|---|---|---|
| A | `outreach/pipeline._deliver` → discord webhook (server process) | `discord:send:bulk` |
| B | `mcp/outreach_mcp.outreach_poll` → webhook (outreach subprocess) | `discord:poll:bulk` |
| C | `mcp/discord_bot_mcp.send_reply` → Discord API (discord-bot subprocess) | `discord:reply:standard` |

(`DiscordWebhookAdapter.send_poll` is dead — tests-only caller — so it is not instrumented; the CI guard backstops any revival.)

### Design
- **`autonomy/shadow_gate.observe_discord_send`** — **read-only** cell lookup (no
  `apply_event`/`ensure_cell`; a missing cell reads as `not_determined`), **best-effort**
  (swallows every failure at DEBUG), **never raises / never blocks / never mutates state**.
- Called **after the successful send** on all three doors, so the outbound post is never
  delayed — any DB contention only defers the internal record.
- **`db/crud/capability_shadow`** — best-effort `record()` with a cached table-existence
  guard. Migration `0044` is the *sole* DDL authority (no inline-DDL divergence); the
  subprocess writers (which don't run migrations) skip until the table exists, then
  self-heal.
- **discord-bot server** gains an *optional* best-effort `genesis.db` connection (opened
  once in its bootstrap lifespan, non-fatal) purely for shadow recording — it stays fully
  functional with no DB.
- **`content_preview`** is bounded (≤200 chars); **`content_hash`** is over the full
  content (a genuine fingerprint, distinct from the preview span). No full outbound
  content is stored.

### CI regression guard
`scripts/check_external_io.py` (new CI job) fails the build if any source file references
a Discord/webhook/public-social endpoint outside a reasoned allowlist — so a new ungated
door can't ship silently. Scoped precisely to external chat/webhook/social HTTP egress
(not a generic `.post(` scan); browser publish and owner-private channels are out of scope
by design.

### Verification
- **Architecture review** (pre-build): 2 blocker + 3 high findings, all resolved before coding.
- **Code review** (post-build): full-content hash fix, observe-moved-after-send, honest
  guard scope.
- **28 new + 52 regression tests green; lint clean; guard clean.**
- **End-to-end** against a copy of a migrated database (real migration runner + real code
  paths, incl. the discord-bot's own connection): all three doors record + still send;
  **0 capability cells created, 0 holds created, 0 approvals created**.

### Deploy notes
- Migration `0044` + door A go live on server restart; doors B/C (MCP subprocesses) apply
  on the next session — same pattern as the prior WS5 stage. No behaviour change
  (observe-only).
- Follow-on: the enforce stage (hold-for-approval + posture decision), informed by the
  `capability_shadow_events` volume.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -186,6 +186,19 @@ jobs:
           fi
           echo "Migration prefix check: CLEAN"
 
+  external-io-check:
+    # WS5 regression guard: no NEW ungated autonomous external-world egress
+    # (Discord REST/webhook, public social APIs) outside the reasoned allowlist.
+    # See scripts/check_external_io.py.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: External-I/O egress guard
+        run: python scripts/check_external_io.py
+
   test:
     runs-on: ubuntu-latest
     # CI is blocking — local full-suite runs are banned. See procedure
--- a/scripts/check_external_io.py
+++ b/scripts/check_external_io.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python3
+"""WS5 external-I/O regression guard — no NEW ungated autonomous external egress.
+
+The whole point of WS5 autonomy-gating is that Genesis must not post to the outside
+world (Discord community, public webhooks, public social APIs) through an ungated,
+unobserved path. This grep-based guard is the CI/pre-commit backstop: any source file
+that references a Discord/webhook/public-social endpoint must be on the ALLOWLIST
+(each entry pinned to a known, capability-gated-or-shadowed egress door). A new door
+added anywhere else fails the check — forcing it through the capability gate (or an
+explicit, reasoned allowlist entry) instead of silently shipping a bypass.
+
+SCOPE (deliberately precise, not a generic ``.post(`` scan — there are ~dozens of
+legitimate compute/read POSTs: embeddings, search, TTS, crawl, IPC):
+  * HTTP egress to Discord (REST API + webhooks) and public-social APIs (Twitter/X,
+    Slack). You cannot POST to these without referencing their endpoint/webhook-env
+    string somewhere — that reference is what this guard flags.
+OUT OF SCOPE (documented, handled elsewhere):
+  * Browser-based publishing (e.g. Medium via Playwright) — a different egress
+    modality; a browser-egress guard lands with that channel's gating stage.
+  * Owner-private notification channels (email-to-owner, Telegram-to-owner) — the
+    owner IS the recipient; these are not external-world posting.
+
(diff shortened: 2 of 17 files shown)


WingedGuardian:
## First-pass review

**Structural review:** 1 NOTE (no ERRORs, no WARNINGs)

---

### NOTE — `src/genesis/mcp/outreach_mcp.py` ~line 255: shadow call not try/except-guarded

The shadow gate import in `outreach_poll` is not wrapped in a `try/except`, unlike the analogous code in `discord_bot_mcp.py`:

**`discord_bot_mcp.py`** (defensive):
```python
try:
    from genesis.autonomy.shadow_gate import observe_discord_send
    await observe_discord_send(...)
except Exception:  # noqa: BLE001 — the reply path must never fail on shadow
    logger.debug("send_reply capability shadow observe failed", exc_info=True)
```

**`outreach_mcp.py`** (not guarded):
```python
from genesis.autonomy.shadow_gate import observe_discord_send
await observe_discord_send(...)
```

The comment says "Best-effort" but the import itself is unguarded. `observe_discord_send` catches all exceptions internally, so the `await` call is safe — but an import-time failure in `shadow_gate` (e.g. a missing `genesis.autonomy.types` in the outreach subprocess) would propagate and break `outreach_poll`. Recommend wrapping with the same pattern used in `discord_bot_mcp.py`.

---

Everything else is clean: all new `src/` files have tests, no bare `except:`, no mutable defaults, no raw SQL outside the CRUD module, `feat(autonomy):` prefix present, no PII or secrets in added lines, PR wording is neutral feature description.

---

WingedGuardian:
Merge basis: automated Codex review was quota-limited on this PR. An inline code-review pass was run before merge and surfaced one real issue, now fixed in `9168d1a0`: the shadow-observe import at the deliver/poll egress doors was not wrapped in its own guard (unlike the reply door), so an import/shadow failure could have propagated to an already-completed send and flipped it to an error return. All three doors now guard the observe call uniformly, keeping the observe-only invariant intact. Rebased cleanly over #838; all CI green (test/lint/migration-check/external-io/leak-detector/CodeQL).

PR #840, opened 2026-07-01
fix(autonomy): defer watchdog restarts during a deploy + bound the seed

WingedGuardian:
## What & why

`update.sh` intentionally stops `genesis-server` for its merge/bootstrap/migrate window. The `genesis-watchdog` timer (fires every 60s) could see the server "down" during that window and restart it — and the revived server's DB write lock deadlocked `bootstrap.sh`'s procedure seed (`seed_procedures.py`, an aiosqlite worker thread), hanging the whole update silently for ~30 minutes. `update.sh` already disarms systemd's `Restart=on-failure`, but was blind to the watchdog — a second, independent revival path.

## The fix

**1. New shared signal — `genesis.env.update_in_progress()`**
Reads both deploy signals under `~/.genesis/`: the CLI path's `update_state.json` (`{phase, pid, started_at, …}`, written per-phase by `update.sh::_write_state`) and the dashboard path's bare-int `update_in_progress.pid`. Returns `True` only while an owning `pid > 1` is alive (`os.kill(pid, 0)`) **and**, for the state file, `phase != "done"` **and** `started_at` is within a 4h cutoff. Everything else — dead / absent / corrupt / expired / `pid <= 1` — is `False`, and the whole body is wrapped so it **never raises into the watchdog** (fail-open to "no deploy"). The 4h cutoff bounds the rare PID-reuse window in which a leftover state file could otherwise suppress the watchdog.

**2. Watchdog guard** — `autonomy/watchdog.py::_restart_if_allowed()` returns `WatchdogAction.SKIP` when a deploy is in progress, *before* `_record_failure`, so a deploy window never trips backoff or the max-restart counter. This function is the sole producer of `WatchdogAction.RESTART`, so all three restart paths are covered by the one guard.

**3. Defense-in-depth** — `bootstrap.sh` wraps the procedure seed in `timeout 300` (keeps the existing `|| true`; a contended DB now fails fast at 5 min with the existing WARNING instead of hanging silently).

**4. Cleanup symmetry** — `update.sh::_do_rollback` and the "already up to date" early exit now clear the in-progress signal files, mirroring the success-path cleanup. The merge-conflict `exit 2` path deliberately keeps `update_state.json` (the `--post-merge` continuation reads it for the rollback tag); that path is already watchdog-safe because its PID dies on exit.

**5. DRY** — `learning/signals/genesis_version.py::_check_failure_file()` now reuses `update_in_progress()` instead of an inline orchestrator-PID read, so the update-failure alarm also defers during CLI updates (not just dashboard-orchestrated ones). Removed the now-unused `import os`.

## Validation

- `ruff` clean; `bash -n` clean on both scripts.
- New `tests/test_env.py` (15 cases: both signal paths, `phase=done`, dead/`<=1` PID, stale/naive `started_at`, corrupt JSON, never-raises). Watchdog deploy-guard tests (defers a stale-server restart; does not trip the failure counter; restarts normally with no deploy). A `genesis_version` defer test. **98 targeted tests pass**; import-cycle regression stays green.
- **E2E with the real `WatchdogChecker` + real `update_in_progress()`** (side-effect-free — `check()` only decides): a 20-min-stale server that would normally RESTART is **deferred (SKIP)** while a realistic `phase=bootstrap` deploy signal is present, and RESTARTs again on `phase=done` / a dead PID. `bootstrap`'s `timeout 300` was verified to fail fast and emit the WARNING on a simulated deadlock.

## Regression markers

- If the watchdog ever stops restarting a genuinely-down server: suspect a stale `update_state.json` / `update_in_progress.pid` whose PID was recycled to a live process — remove it (auto-cleared by the next bootstrap; the 4h cutoff bounds the window).
- A deploy must complete without the watchdog reviving the server mid-run — the "Restarting … via systemctl" line must NOT appear in the watchdog journal during the stop-window.

## Deploy note

This fix must be deployed with the watchdog timer paused (`systemctl --user stop genesis-watchdog.timer`) for its *own* update.sh run, then restarted — because the old (unguarded) watchdog is still live during the deploy that installs the guard. After it lands, future deploys are self-protected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **`update.sh` no longer hangs if the health watchdog restarts the server mid-update.**
+  During an update Genesis stops its server to swap in new code and migrate the database. The
+  background health watchdog could see it "down" and restart it right then — and the revived
+  server's database lock deadlocked the update's procedure-seeding step, leaving the whole update
+  stuck for as long as ~30 minutes with no error. The watchdog now defers restarts while an update
+  is in progress, and the seeding step is time-bounded so a contended database fails fast instead of
+  hanging silently.
 - **Genesis now keeps your Claude Code CLI at the version it's tested against — automatically.**
   Previously the installer only put Claude Code in place when it was *missing*, so if you already
   had an older Claude Code, bumping the pinned version never actually upgraded you — you'd silently
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -417,7 +417,11 @@ echo "--- Seeding baseline procedures ---"
 # Non-fatal, but surface a failure instead of swallowing it (a silent miss would
 # leave advisory surfacing untracked with no signal). Capture, then check for the
 # success line rather than relying on the exit code through the pipe.
-SEED_OUT="$("$VENV_DIR/bin/python" "$GENESIS_ROOT/scripts/seed_procedures.py" 2>&1)" || true
+# Defense-in-depth (incident IR-2): bound the seed with `timeout 300`. Seeding
+# normally finishes in seconds; if the DB write lock is ever contended (a mid-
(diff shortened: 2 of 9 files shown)


WingedGuardian:

---

PR #843, opened 2026-07-02
fix(surplus): correct gap_clustering and wing_audit judge-expected strings

WingedGuardian:
The surplus quality judge grades each completed background insight task's output against a per-task-type description of what a good output looks like. Two of those descriptions did not match what the tasks actually produce, so correct outputs were being scored as low-quality:

- **gap_clustering** clusters recent unresolved observations into recurring themes that suggest systemic issues, but was graded against "knowledge or capability gaps" — so correct systemic-issue theme clusterings were failed on relevance.
- **wing_audit** proposes reclassifying currently-uncategorized memories into specific wings/rooms, but was graded against "auditing the existing taxonomy for miscategorization / sparse-or-overloaded domains" — so correct reclassification proposals would be failed.

Both descriptions are rewritten to match each task's actual prompt. The judge stays measurement-only — it never gates anything or changes intake routing; only the grading target for these two task types is corrected. No logic or threshold changes. The other nine task-type descriptions were checked and already match their tasks.

**Verification:** re-grading a real gap_clustering sample with the corrected description recovers 5 false-negative "hollow" verdicts with zero regressions (genuine non-clustering outputs stay hollow); synthetic wing_audit controls discriminate correctly. `ruff` clean; targeted tests pass.

### DIFF
--- a/src/genesis/surplus/quality_judge.py
+++ b/src/genesis/surplus/quality_judge.py
@@ -81,10 +81,17 @@ _JUDGE_EXPECTED: dict[TaskType, str] = {
         "and references actual procedures or patterns."
     ),
     TaskType.GAP_CLUSTERING: (
-        "A clustered analysis of knowledge or capability gaps — coherent themes "
-        "grouped from underlying signals, each described specifically. Good output "
-        "is a structured set of real, distinguishable gap clusters, not a flat "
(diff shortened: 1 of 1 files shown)


WingedGuardian:

---

PR #845, opened 2026-07-02
feat(observability): surface silently-stale scheduled jobs

WingedGuardian:
## What & why

Scheduled jobs that run on schedule but silently never succeed were invisible. The weekly self-assessment and quality-calibration jobs failed on several recent Sundays — they returned empty Claude Code output (root cause: the shared Claude Code subscription hit its usage cap for ~2 days; investigated separately, not a per-provider 429) — but nothing surfaced because:

- `clear_stale_job_failures` resets `consecutive_failures` to 0 and clears `last_failure` on **every server restart** (no age check) — the ~10 restarts over two days erased each recorded failure.
- The awareness `JobHealthCollector` only alerts on `consecutive_failures >= 2`, which an infrequent (weekly) job cannot reach between restarts.

The failures were real (`job_health.total_failures > 0`) but the current-state fields kept getting wiped, so nothing surfaced.

## The fix

Surface staleness where job health is actually observed, keyed on the **`last_run − last_success` gap** — which `clear_stale_job_failures` never touches (so it survives restarts) and which is cadence-independent (a healthy job has gap 0, because a successful run writes both fields together):

- `_impl_health_alerts` emits a `job_stale:<name>` **WARNING** for any job whose gap exceeds ~1 weekly cadence. This flows automatically into the daily report, the `health_alerts` MCP tool, and the dashboard.
- `_impl_job_health` adds `days_since_success` + a `stale` flag per job (non-mutating — the runtime path returns `job_health` by reference).
- The shared threshold lives in a dependency-free `health/constants.py` leaf to avoid an import cycle (the package `__init__` imports both modules).

Deliberately does **not** modify `clear_stale_job_failures` (the gap signal is robust to it) or the awareness signal-collector path (which is dormant post-bootstrap — dropped by the learning-init collector swap and unweighted in scoring).

## Verification

- Architect + code-reviewer: all findings addressed (log level, shared-constant leaf, test/prod `row_factory` parity, mixed-tz normalization).
- 8 new tests (real `julianday` gap query against an in-memory DB + the annotate helper incl. mixed tz-awareness) + 424 existing tests green; ruff clean; no import cycle (fresh-process probes).
- E2E against the live DB: flags exactly the two stale weeklies (14d / 7d), healthy jobs `0.0 / False` — no false positives.

Note: the *underlying* cause (why the weekly cognitive jobs hit rate limits) is a separate, deeper follow-on — this PR makes the silent failure **visible**, it does not fix the rate-limiting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **Genesis now catches scheduled jobs that silently stop working — running on schedule but never
+  succeeding.** Some background jobs (like the weekly self-assessment and quality calibration) could
+  fail week after week without ever showing up as "failed," because the failure counter is reset
+  every time the server restarts. Genesis now watches the gap between a job's last run and its last
+  success: if a job has been running-but-not-succeeding for more than about a week, it raises a health
+  alert that reaches your daily report and the dashboard, and the job-health view now shows a
+  `days_since_success` figure and a `stale` flag for every job.
 - **You can now start an interactive Claude Code session on a different model with one command.**
   `gmodel <name>` launches `claude` on the model you pick: a Claude tier (`gmodel opus`) runs on
   your normal Max subscription, while a roster peer (`gmodel glm-5.2`) runs on that provider's
--- a/src/genesis/db/crud/job_health.py
+++ b/src/genesis/db/crud/job_health.py
@@ -1,0 +1,33 @@
+"""Read-side queries for the ``job_health`` table.
+
+Writes live in :mod:`genesis.runtime._job_health` (record_job_success /
+_failure / clear_stale_job_failures). This module holds the observability
+READ queries consumed by the health MCP + dashboard.
+"""
+
+from __future__ import annotations
+
+import aiosqlite
+
+
+async def get_stale_jobs(
+    db: aiosqlite.Connection, *, threshold_days: float
+) -> list[dict]:
+    """Jobs that have RUN more than ``threshold_days`` since they last SUCCEEDED.
+
+    Returns ``{job_name, last_success, gap_days}`` rows, widest gap first. The
+    ``last_run − last_success`` gap survives the per-restart ``consecutive_failures``
+    reset (``clear_stale_job_failures`` never touches ``last_run``/``last_success``),
+    so it is the honest "running but not succeeding" signal — one a healthy job
+    reads as 0 (a successful run writes both columns together).
+    """
+    cursor = await db.execute(
+        "SELECT job_name, last_success, "
+        "julianday(last_run) - julianday(last_success) AS gap_days "
+        "FROM job_health "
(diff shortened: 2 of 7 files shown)


WingedGuardian:
## First-pass review — PR #845

**Structural review: 1 ERROR · 1 WARNING**

> Note: `--request-changes` is unavailable on self-authored PRs; findings posted as a comment.

---

### ERROR — Raw SQL against genesis.db · `src/genesis/mcp/health/errors.py`

The new silently-stale-jobs block calls `_service._db.execute("SELECT … FROM job_health …")` directly. Convention requires DB access to go through `genesis.db.crud` modules (exception: migrations only). This pattern already exists in other blocks of `_impl_health_alerts`, but adding another instance compounds the anti-pattern rather than normalising it.

Options:
1. Add a `get_stale_jobs(db, threshold_days)` function in the appropriate `genesis.db.crud` module and call that instead.
2. If the health MCP is an accepted standing exception to the CRUD rule (read-only observability queries that cross multiple tables and don't belong in a single CRUD module), document it explicitly with a comment near the first raw-SQL block in `_impl_health_alerts` — something like `# health MCP: direct reads only; CRUD exception — see architecture note` — so future reviewers can distinguish intentional exceptions from accidental bypasses.

---

### WARNING — New `src/` file with no dedicated test file · `src/genesis/mcp/health/constants.py`

`constants.py` is a new source file. Convention flags new `.py` files in `src/` that have no corresponding file in `tests/`. `JOB_STALE_GAP_DAYS` is exercised indirectly by `test_job_staleness.py`, which is fine for coverage, but the mapping is implicit. A one-liner `tests/test_mcp/test_health_constants.py` that asserts `isinstance(JOB_STALE_GAP_DAYS, float) and JOB_STALE_GAP_DAYS > 0` satisfies the convention and guards against accidental type or value regressions.

---

### CLEAN areas

- No bare `except:` — the new `except Exception:` block is correctly typed. ✓
- No mutable default arguments, no `/tmp/` usage, no `nohup`, no `task_states` inserts, no force-push patterns, no star imports, no circular imports. ✓
- Conventional commit prefix (`feat(observability):`) present. ✓
- New `_parse_ts_utc` and `_annotate_staleness` have docstrings. ✓
- Test coverage is substantive: real in-memory SQLite DB for the `julianday` gap math, mixed-tz awareness, non-mutating behaviour, and sort order. ✓
- CHANGELOG wording: describes user-visible behaviour, no language referencing a security incident or infrastructure leak. ✓
- PII/Secrets: no keys, tokens, private IPs, private hostnames, or usernames in paths on any added line. ✓

---

WingedGuardian:
Structural-review ERROR (raw SQL against job_health) and WARNING (constants without a test) were **addressed in head commit `941aa59d`** — the query moved to `genesis.db.crud.job_health.get_stale_jobs`, plus `tests/test_mcp/test_health_constants.py`. The bot reviewed the pre-fix commit (10:35 EDT) and hasn't re-reviewed the fix (pushed 11:26 EDT); CI is green on head and Codex is quota-limited. Merging with a documented review-override since both findings are resolved.

PR #846, opened 2026-07-02
fix(knowledge): tombstone manifest entry on unit delete (unblock re-ingest)

WingedGuardian:
## Problem

The knowledge-ingestion **manifest** (`knowledge/manifest.py`, an on-disk JSON keyed by source, one source → many `unit_ids`) is a source-identity cache. The dashboard DELETE route removed a knowledge unit from SQLite + Qdrant but **never touched the manifest**. So once a source's *last* unit was deleted, the manifest entry survived → `has_source()` kept returning True → re-ingest of that file/URL was permanently skipped (serving now-dead cached unit IDs). Deleting then re-adding a knowledge source silently returned nothing.

## Fix

- New `ManifestManager.remove_unit(unit_id) -> bool` — **unit-granular** (matching the per-unit DELETE route): pops the id from its owning source entry and **tombstones the whole entry only when its last live unit is removed**, so re-ingest is unblocked. Idempotent; leaves legitimately-empty `no_units_extracted` entries untouched; leaves physical artifacts in place (re-ingest overwrites them).
- Wired into the DELETE route as a **best-effort** call after the physical delete (a manifest hiccup can never fail an already-successful delete); adds a `manifest_removed` field to the response.

No DB migration (JSON manifest).

## Testing

- 9 new `remove_unit` unit tests (tombstone-on-empty, multi-unit keep, idempotency, no-units-untouched, cross-reload disk persistence) + 1 orchestrator-level re-ingest-after-delete test. **21/21 green** in `test_knowledge/`; ruff clean.
- Reviewed inline (code-reviewer). The one flagged concern — a stale in-process manifest cache — was verified **not applicable**: `_get_orchestrator()` builds a fresh `ManifestManager` per ingest, so the tombstone is seen on the very next ingest without a restart.
- Mechanism E2E covered by tests; the live HTTP-DELETE → re-ingest E2E runs at deploy.

## Scope note

Content-hash idempotency (re-ingesting a *changed* source) is a separate, larger change and is **not** in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,11 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Deleting a knowledge item no longer blocks you from re-adding its source.** When you deleted
+  a knowledge unit from the dashboard, Genesis removed it from the search index but still remembered
+  having ingested its source — so re-adding that same file or URL was silently skipped and nothing
+  came back. Now deleting a source's last remaining unit clears that record, and re-ingesting the
+  source works normally again.
 - **`update.sh` no longer hangs if the health watchdog restarts the server mid-update.**
   During an update Genesis stops its server to swap in new code and migrate the database. The
   background health watchdog could see it "down" and restart it right then — and the revived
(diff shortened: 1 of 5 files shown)


WingedGuardian:

---

PR #809, opened 2026-06-30
feat(security): injection-scan and boundary-wrap external content at knowledge ingestion

WingedGuardian:
## What

Genesis ingests external documents (fetched pages, uploaded files, agent-supplied units) into the knowledge base, where they are later recalled into prompts. Until now that content reached the distillation LLM **raw** — a prompt-injection payload inside an ingested document (\"ignore previous instructions…\") was read by the model as if it were a trusted instruction.

This wires the existing `ContentSanitizer` into the knowledge-ingestion paths so external content is scanned and boundary-delimited on the way in. The defense is **detect-and-flag, never block**, and **fail-open** (a scanner error never aborts an ingest).

## Changes

- **Distillation path** (`knowledge/orchestrator.py`, `knowledge/distillation.py`): each ingested source is scanned for injection patterns and surfaced as an `injection_patterns_detected:N` quality flag; every chunk sent to the distillation LLM is wrapped in `<external-content source=… risk=…>` boundary markers, and the distillation prompt now instructs the model to treat wrapped content strictly as data (never as instructions) and never to echo the markers into its output. Already-wrapped upstream content is normalized so it is never double-wrapped.
- **Non-distillation paths** (`memory/knowledge_ingest.py`, `knowledge/ingest_upload.py`): the direct knowledge-unit ingest helper and the no-LLM \"store-as-is\" upload mode now scan and log on detection. Logs record matched pattern names and provenance only — never the content itself.
- **Shared helper** (`security/sanitizer.py`): a small `strip_boundary_markers` helper, with `knowledge/processors/web.py` refactored to reuse it.

Recall-time boundary-wrapping (delimiting external-world content where it is injected back into prompts) is a deliberate follow-up.

## Tests

- 158 targeted tests pass (16 new): boundary-wrapping per chunk, no double-wrap on already-wrapped content, no tag leakage into output, detection flags on injection content, no flag on benign content, and fail-open at every scan site.
- A live distillation A/B confirmed that wrapping does not degrade extraction, the boundary markers do not leak into stored knowledge, and an embedded injection payload is not obeyed by the model.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #848, opened 2026-07-02
feat(security): wrap external-world content at recall/inject time

WingedGuardian:
Closes the **recall half** of prompt-injection defense — the sibling to #809 (ingestion-side scan+wrap). External-world knowledge-base content that is recalled and injected into an LLM prompt now carries the `<external-content>` structural delimiter, not just a soft `KB·source` text label, so an injection payload inside an ingested document is treated as data rather than as Genesis's own instructions.

**Principle (unchanged from #809):** detect-and-delimit, fail-open, first-party memory untouched. No blocks, no dropped legitimate content.

## What changed
- **Shared helper** `memory/provenance.wrap_external_recall` — strip-then-wrap (idempotent, no double-wrap), provenance-derived risk tier (`crag_web`→web-fetch 0.6, `recon`→0.3, else ingested-KB→0.2), guards non-str/empty, fail-open with a warning.
- **Wrap at every external→LLM-prompt inject site** (gated on external-world provenance; first-party left alone): `memory_recall` (full path), `memory_expand`, `memory_proactive`, `knowledge_recall` (vector + FTS body), `resume_review` (both vector and FTS branches), the research-executor recall, a defensive guard in the CC context injector, and the voice handler's full-LLM system prompt.
- The proactive-hook **line format** strips leaked markers and keeps the `KB·` label (a multi-line wrapper would break its one-hint-per-line output); the 150-char compact `memory_recall` preview is intentionally not wrapped (always followed by a wrapped `memory_expand`). The voice **spoken** path keeps a soft label (XML markers can't be voiced) while its LLM path wraps.
- Removed the blunt `_is_garbage` `<external-content>` drop in the proactive hook — surface-clean instead.
- **Durable coverage guardrail** (`tests/test_security/test_recall_inject_coverage.py`): AST-enumerates every `.recall()` call site under `src/genesis` and fails if a new one is unclassified, so a future consumer can't silently reopen the gap.

No DB schema change, no migration. Ingestion body-wrapping is subsumed (those units land in the KB and are wrapped on recall).

## Verification
- Targeted unit/integration tests + coverage guardrail green; existing recall/voice/resume tests no-regression; ruff clean.
- Helper: idempotency, risk-tier correctness, non-str guard, and fail-open all covered.
- End-to-end against a live retriever (real embedding chain + populated vector store): `knowledge_recall` / `memory_recall` (full) / `memory_expand` wrap real external content; real first-party content is left unwrapped and first-party-labeled; a forced wrap exception still returns content (fail-open).

### DIFF
--- a/scripts/proactive_memory_hook.py
+++ b/scripts/proactive_memory_hook.py
@@ -272,8 +272,11 @@ def _update_and_format_trail(
 
 def _is_garbage(content: str) -> bool:
     """Filter out content that should never surface as proactive memory."""
-    if "<external-content" in content:
-        return True
+    # NOTE (PR2): we deliberately no longer drop content merely for containing
+    # `<external-content>` markers. That was a blunt defense against leaked
+    # boundary markers, but it also silently lost legitimate hits. `_format_results`
+    # now strips any leaked markers and re-labels external-world hits (`KB·…`),
+    # so surfacing-clean is safe. Do not re-add this drop as "protective".
     stripped = content.lstrip()
     if stripped.startswith("{") and any(
         k in stripped[:100] for k in ('"drift_detected"', '"tags"', '"type":', '"operation"')
@@ -845,6 +848,13 @@ def _format_results(results: list[dict]) -> str:
         is_rule = r.get("memory_class") == "rule"
         max_len = 300 if (rank == 0 or is_rule) else 200
         content = r.get("content", "")
+        # Injection defense (PR2): strip any leaked <external-content> boundary
+        # markers from BOTH first-party and KB hits, so raw markers never reach
+        # the model. External-world hits keep their soft `KB·source` label below
+        # (this line-oriented one-per-hint format can't carry a multi-line
+        # structural wrapper — the full-content MCP/expand paths do that).
+        from genesis.security.sanitizer import strip_boundary_markers
+        content = strip_boundary_markers(content)
         # Strip extraction-pipeline prefixes like [discovery], [feature], etc.
         # These are baked into stored content but waste display chars.
         if content.startswith("[") and "] " in content[:30]:
--- a/src/genesis/autonomy/executor/research.py
+++ b/src/genesis/autonomy/executor/research.py
@@ -21,6 +21,7 @@ from typing import Any
 
 from genesis.autonomy.executor.types import ResearchResult
 from genesis.cc.types import CCInvocation, CCModel, CCOutput, EffortLevel, background_session_dir
+from genesis.memory.provenance import is_external, wrap_external_recall
 
 logger = logging.getLogger(__name__)
 
@@ -221,6 +222,12 @@ class DeepResearcherImpl:
             lines = []
             for r in results[:5]:
                 content = r.content[:200] if hasattr(r, "content") else str(r)[:200]
+                # Injection defense (PR2): recall defaults to source="both", so a
+                # KB hit can reach the triage prompt — wrap external-world content.
+                if is_external(getattr(r, "collection", "")):
+                    content = wrap_external_recall(
+                        content, source_pipeline=getattr(r, "source_pipeline", None),
+                    )
                 lines.append(f"- {content}")
             return "\n".join(lines)
         except Exception as exc:
--- a/src/genesis/cc/context_injector.py
+++ b/src/genesis/cc/context_injector.py
@@ -5,6 +5,8 @@ from __future__ import annotations
 import logging
 from typing import TYPE_CHECKING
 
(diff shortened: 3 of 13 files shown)


WingedGuardian:
## First-pass review: PR #848

---

### WARNING — Import placed inside a loop body (`scripts/proactive_memory_hook.py` ~line 854)

```python
# Inside _format_results(), inside the for-loop over results:
content = r.get("content", "")
from genesis.security.sanitizer import strip_boundary_markers   # ← here
content = strip_boundary_markers(content)
```

The import is inside the body of `_format_results` (and inside its inner `for` loop), not at module level. Python caches modules after the first load, so this doesn't re-read the module on every iteration — but it still pays a `sys.modules` dict lookup + attribute access on every pass through the loop, and it surprises readers who expect all imports at the top of the file.

Additionally, `scripts/proactive_memory_hook.py` is a hook script; the Genesis convention is that hooks stay stdlib-only or use `genesis.env`. This hook already imports from `genesis.memory.provenance`, so adding `genesis.security.sanitizer` isn't a new category of violation, but it's another genesis dependency loading at hook cold-start. Placing it at module level alongside the existing genesis imports at least makes the dependency explicit and measurable. Move it there.

---

### Structural scan — everything else: CLEAN

Checked all added/modified `.py` files against the full Domain 1 checklist:

- No `nohup`, no direct inserts into `task_states`, no `/tmp/` usage
- No `git add .` / `git add -A`, no force-push patterns
- No bare `except:` (all uses are `except Exception:`)
- No mutable default arguments
- No `os.killpg`, no raw SQL against `genesis.db`
- Import order correct in all `src/` files (stdlib → local; no star imports, no visible circular imports)
- All new public functions have docstrings (`wrap_external_recall`, `_source_for`)
- New `tests/test_security/test_recall_inject_coverage.py` covers the new guardrail
- `feat(security):` commit prefix ✓
- No `IntervalTrigger` for intervals > 1h

---

### PII/Secrets/Wording scan (added lines only): CLEAN

- No API keys, tokens, passwords, or credentials
- No private IPs (RFC-1918, Tailscale CGNAT, or IPv6 ULA)
- No SSH keys or fingerprints, no database connection strings
- No personal email addresses, no hardcoded `/home/<username>/` paths
- No AWS/GCP account IDs, no hardware identifiers, no private hostnames

**PR title/description language:** clean — describes a proactive security hardening, not a breach or data-exposure fix. No language that would be damaging in permanent git history.

---

WingedGuardian (commented):
## First-pass review — PR #848

### Structural review

**WARNING — import inside function body (`scripts/proactive_memory_hook.py`)**

```python
# _format_results, ~line 851
from genesis.security.sanitizer import strip_boundary_markers
content = strip_boundary_markers(content)
```

Two issues with this placement:

1. **Import hygiene**: imports belong at module level (stdlib → third-party → local), not embedded inside a function. An import inside a hot path like `_format_results` re-executes the attribute-lookup on every call (CPython caches it after the first hit, but the pattern itself is non-standard and makes the dependency invisible to tooling).

2. **Hook import rule**: `scripts/proactive_memory_hook.py` is a hook script and must be stdlib-only or `genesis.env`. Placing the import inside a function sidesteps module-level detection but the dependency still loads at runtime. The fix is to hoist it to the top-level import block — and if the hook is subject to the stdlib-only constraint, use the `genesis.env` mechanism instead.

---

**WARNING — missing behavioral unit test: `research.py::_memory_recall`**

The wrapping added at `src/genesis/autonomy/executor/research.py` lines ~226–230 is only covered by the classification entry in `KNOWN_RECALL_SITES` — that guardrail verifies the site is *classified*, not that wrapping *fires*. The PR adds analogous behavioral tests for `memory_recall`, `memory_proactive`, and `knowledge_recall` in `test_mcp_integration.py`, but `_memory_recall` in the research executor has none. A test that mocks a KB result and asserts `<external-content` appears in the returned string would close this gap cleanly.

---

**WARNING — missing behavioral unit test: `resume_review.py::_query_knowledge_base` (FTS branch)**

The `wrap_external_recall(fts.get("body", ""), ...)` call in the FTS path of `_query_knowledge_base` is not directly exercised by any test in this diff. The PR notes "existing recall/voice/resume tests no-regression," but if those existing tests don't assert the wrapped output shape, the FTS branch's injection defense is unverified. A test that populates `fts_results` with a KB body containing a payload and asserts it exits `_query_knowledge_base` wrapped would be the same pattern as the vector-branch wrap.

---

No API keys, tokens, private IPs, credentials, hardware identifiers, or disclosure-type language found in the diff or PR metadata.

---

PR #849, opened 2026-07-02
fix(reflection): self-heal weekly jobs (daily fire, ≤1 success/week via idempotency)

WingedGuardian:
## What & why

The weekly self-assessment and quality-calibration jobs fired **once a week** (Sunday, `misfire_grace_time=None`), so a single failed run meant the next attempt was a full week later. Root cause of the recent failures: the shared Claude Code Max subscription hit its **usage cap** → `claude -p` returned **empty output** → the reflection parsed to nothing → recorded a failure. During a multi-day cap (late June) the jobs stayed broken for ~2–3 weeks.

## The fix — self-heal via daily fire + idempotency

Both jobs now fire **daily** (dropped `day_of_week`; kept the hour), and the **existing `_already_ran_this_week()` idempotency gate** holds each to **≤1 *successful* run per Monday-anchored week**. A day that fails leaves the idempotency window empty, so the next day retries — until one success lands that week. This recovers from *any* transient cause (a usage cap, a throttle, a one-off), not just the specific June outage.

**Behavior note:** the effective run-day shifts from **Sunday → the week's first fire (Monday)**; retries fill in later in the week after a failure. Monday-morning "review last week" is a natural fit, and it moves the job off the Sunday window where the cap landed. (Easy to pin back to Sunday via the idempotency anchor if preferred.)

## From review (architect + code-reviewer)
- **Guard the calibration regression check behind `result.success`** — a failed calibration has no new baseline, and daily firing would otherwise re-run it (and possibly re-emit the signal) every day during an outage.
- Replaced an obscure `__import__("datetime").timedelta` with a proper top-level `timedelta` import (the gate is load-bearing).
- Corrected the class docstring (said "Sunday"; now daily/Monday-anchored).
- Verified the idempotency gate is correct: `observations.query` sorts `created_at DESC`, so `limit=1` returns the newest row (a reviewer flagged a possible oldest-first bug — refuted).

## Verification
- 25 unit tests green (added: daily-fire assertion, failed-calibration-skips-regression). ruff clean.
- E2E: the trigger fires on consecutive days (07-01 → 07-02 → 07-03); a stuck 14-day-old state **runs** (self-heal); a this-week success **skips** (≤1/week).

## Deploy note
If deployed mid-week after a healthy prior run, expect **one** extra invocation on the first fire of the new regime (the prior Sunday run falls outside the new Monday-anchored window). One-time, ~$0.30, and moot for the currently-stuck jobs.

## Not in this PR (deliberately)
- **Live failover to a roster model (GLM-5.2) when capped** — needs cap *detection* first (the cap returns silent-empty, treated as success); a separate careful follow-on.
- **Reducing ego CC-subscription consumption** (~$88/wk Opus — the real driver of hitting the cap) — a separate analysis + proposal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,15 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   having ingested its source — so re-adding that same file or URL was silently skipped and nothing
   came back. Now deleting a source's last remaining unit clears that record, and re-ingesting the
   source works normally again.
+
+- **The weekly self-assessment and quality-calibration jobs now recover on their own instead of
+  going dark for weeks.** Previously each ran only once a week, so if that one run failed — for
+  example when the shared Claude Code subscription was capped and returned empty output — the next
+  attempt wasn't until the following week, and a multi-day outage could leave them stale for 2–3
+  weeks. They now run daily but still complete at most once per week (an idempotency check skips the
+  rest of the week once one run succeeds), so a failed day is simply retried the next day until it
+  succeeds. Side effect: the successful run now normally lands early in the week rather than on Sunday.
+
 - **`update.sh` no longer hangs if the health watchdog restarts the server mid-update.**
   During an update Genesis stops its server to swap in new code and migrate the database. The
   background health watchdog could see it "down" and restart it right then — and the revived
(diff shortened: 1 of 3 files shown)


WingedGuardian (commented):
## First-pass review — PR #849

PII / Secrets / Wording scan: **CLEAN**

---

PR #850, opened 2026-07-02
fix(contribution): block Tailscale CGNAT + IPv6 in the sanitizer floor

WingedGuardian:
## What

The public-repo contribution sanitizer's always-on portability floor missed two install-specific address classes — Tailscale CGNAT and the Tailscale IPv6 ULA prefix — so they could reach a public PR via the `genesis contribute` path. This closes the gap across every enforcement layer that already covers the CGNAT class:

- **Sanitizer floor** (`_check_portability`): added the CGNAT range and the Tailscale IPv6 prefix, and widened the private subnet match from two hard-coded addresses to the full range (deterministic, always-on).
- **gitleaks ruleset** (`.gitleaks.toml`): added the missing IPv6 rule (CGNAT was already present).
- **commit-message guard** and **CI PR-description scan**: added the IPv6 prefix (both already had CGNAT).
- Allowlisted the commit-msg pattern file so its own definitions don't self-match.

## Testing (TDD, red-first)

5 new tests in `test_sanitize.py`: CGNAT blocks, IPv6 blocks, the widened subnet range blocks, a public address outside the CGNAT range is NOT flagged (false-positive guard), and an IPv6-lookalike hex token without a colon is NOT flagged. Full file: **42 passing**, ruff clean. E2E-verified end to end: the floor blocks the real Tailscale address classes and passes clean diffs; the commit-message guard blocks and passes correctly.

## Scope notes

- The audit also suggested passing `--config` to the sanitizer's optional gitleaks layer. Intentionally **dropped**: reusing `.gitleaks.toml` inherits its global path-allowlist — correct for the repo's own at-rest scan, wrong trust boundary for untrusted contribution diffs. Filed as a follow-up (a rules-only diff-scan config).
- Separately filed follow-up: CI's `rg` portability scan doesn't scan committed source for private IPs at all (relies on the local gitleaks pre-commit hook) — a pre-existing defense-in-depth gap, out of scope here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,7 +155,7 @@ jobs:
         run: |
           set -euo pipefail
           body=$(gh pr view ${{ github.event.pull_request.number }} --json body -q .body 2>/dev/null || echo "")
-          PATTERNS='10\.176\.[0-9]|192\.168\.50\.[0-9]|100\.(6[4-9]|[7-9][0-9]|1[01][0-9]|12[0-7])\.|046056[0-9]{6}|moonandstar|zorror|assistant1'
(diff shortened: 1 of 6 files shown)


WingedGuardian:

---

PR #851, opened 2026-07-02
fix(dashboard): default attention filter to newest version + reframe labeling as attention

WingedGuardian:
## What
Two labeling-blocking fixes on the dashboard **Attention** review tab (template-only), so the next labeling pass is correct. Precursor to the larger "Genesis Voice" section (PR-2).

### 1. Duplication fix
The store now live-persists two config versions (`0.1.0-default` 326 + `0.2.0-taxonomy` 352); **226 windows fired under both**, and the version filter defaulted to "All versions" → every such window rendered twice, blocking labeling. Now the filter **defaults once to the newest version** (`config_versions` is ascending → last entry). Guards:
- `_attentionVersionDefaulted` — one-time, so a later manual **"All versions"** selection *sticks* (no snap-back).
- `_attentionEventsGen` — a request-generation guard on `fetchAttentionEvents` so a stale unscoped response can't clobber the scoped list during the init double-fetch (**race caught in code review; the reviewer's simpler "await" fix wouldn't have stopped the already-in-flight unscoped request**).

### 2. Labeling reframe (copy only)
Users were grading *"would I want it chiming in"* (a **speak** bar) instead of *"should it pay attention"* (the L1 gate's actual job) — because the tab never defined the criterion. Fixed:
- Header now states it explicitly: *"should Genesis have PAID ATTENTION here — was it worth being aware of / remembering? This is about attention, not interrupting you."*
- Buttons relabeled **"Worth noticing / Not worth it / Skip"**; the **stored `acceptance_signal` values (should/shouldnt/skip) are unchanged** (metrics/API untouched). The already-labeled badge maps them to noticed/ignored/skipped for display.

## Scope / risk
Template-only (`genesis_dashboard.html`); no Python, routes, schema, or metric changes. The `calibrate` precision tool and the `attention_events` firewall are untouched.

## Verification
- `node --check` on the module script (valid JS).
- Code-reviewer pass: confirmed termination/snap-back/stored-values/`config_versions`-availability clean; **found + fixed an init race** (unscoped fetch clobbering the scoped list).
- **Browser-E2E is post-deploy** (editable install serves the main-tree template) — will confirm: only the newest version shows (no duplicates), "All versions" selectable and sticky, reframed copy renders, buttons still POST should/shouldnt/skip.

## Follow-on
After merge + deploy: reset the 33 mis-graded labels (graded with the old question) for a clean slate, then re-label with the attention frame.

### DIFF
--- a/src/genesis/dashboard/templates/genesis_dashboard.html
+++ b/src/genesis/dashboard/templates/genesis_dashboard.html
@@ -240,6 +240,8 @@
         attentionEvents: [],
         attentionStats: null,
         attentionFilters: { activation: "", trigger: "", is_user: false, unlabeled: true, config_version: "" },
+        _attentionVersionDefaulted: false,   // one-time guard: default the filter to newest version once
+        _attentionEventsGen: 0,   // request generation — discard a stale events response superseded by a newer fetch
         attentionRevealed: {},   // {event_id: [window rows] | "unavailable"}
         attentionLabeling: {},   // {event_id: true} while a label POST is in flight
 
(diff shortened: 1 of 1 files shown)


WingedGuardian:
Structural review: CLEAN — no pattern violations found (template-only change; Python structural patterns N/A).

---

WingedGuardian:
Merge note: all 9 CI checks green (test 10m54s, CodeQL, leak-detector, Analyze×3). Automated review (Codex + structural bot) did not post — quota-limited/quiet. Self-reviewed via a genesis code-reviewer agent, which caught a real init race (stale unscoped fetch clobbering the scoped list) — fixed with a request-generation guard; `node --check` clean. Template-only, no Python/route/schema/metric change. Merging on CI + self-review.

PR #852, opened 2026-07-02
fix(dashboard): knowledge delete used nonexistent rt.qdrant_client (500 on every delete)

WingedGuardian:
## Problem

The dashboard knowledge DELETE route (`knowledge_delete`) referenced `rt.qdrant_client`, which `GenesisRuntime` has **never had** (bug since #83, `a0c0099f`). Every dashboard knowledge delete:
1. deleted the SQLite row (runs first), then
2. **crashed** at the Qdrant step with `AttributeError: 'GenesisRuntime' object has no attribute 'qdrant_client'` → **HTTP 500**,
3. orphaning the embedding, and
4. **never reaching the manifest tombstone** — so #846's re-ingest-after-delete fix could never actually run.

Confirmed live via traceback during the #846 deploy E2E. Enumerated: this was the *only* `rt.qdrant_client` reference left; the sibling `rt.qdrant→store.qdrant_client` bug in dream_cycle was already fixed in #385.

## Fix

Route the Qdrant deletion through the memory store's client: `store = getattr(rt, "memory_store", None)` then `store.qdrant_client.delete(...)`, guarded so a missing store degrades gracefully (skip Qdrant, no 500) instead of crashing. `rt.memory_store` is a property returning the `MemoryStore` whose `.qdrant_client` is the live client — mirrors `references.py:250` and the #385 pattern. The manifest tombstone (from #846) now actually runs.

## Tests

First tests of this route (`tests/test_dashboard/test_knowledge_api.py`). Key detail: they use a faithful `SimpleNamespace` fake runtime, **not** a `MagicMock` — a MagicMock would auto-vivify `rt.qdrant_client` and hide the bug. Against the old code these tests fail with the exact production `AttributeError`; against the fix they pass. Cases: happy path (200 + Qdrant `.delete` on the `knowledge_base` collection with the correct point id + manifest `remove_unit` called), `memory_store=None` (200, Qdrant skipped, tombstone still runs), 404 (missing unit), 503 (not bootstrapped). Full `test_dashboard` + `test_knowledge` suites green; ruff clean. No DB migration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,11 +78,12 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
-- **Deleting a knowledge item no longer blocks you from re-adding its source.** When you deleted
-  a knowledge unit from the dashboard, Genesis removed it from the search index but still remembered
-  having ingested its source — so re-adding that same file or URL was silently skipped and nothing
-  came back. Now deleting a source's last remaining unit clears that record, and re-ingesting the
-  source works normally again.
+- **Deleting a knowledge item from the dashboard now works end to end.** Previously the delete
+  button returned a server error, and behind the scenes the item was only half-removed — dropped
+  from search but with its stored embedding left behind and its source still marked as already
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

PR #854, opened 2026-07-02
fix(surplus): age-cap terminal task rows + reap two async leaks

WingedGuardian:
## WS-F — queue hygiene (audit 2026-07-02)

Three independent, low-risk fixes from the read-only audit. No DB migration.

### F7 — surplus terminal-row reaper (the main change)
`drain_expired` only ever deleted `pending` rows, so finished tasks accumulated forever. On the live DB that's **3,798 completed + 107 failed = 3,905 terminal rows** and climbing.

- New `surplus_tasks.delete_terminal_before(before)` — deletes `completed`/`failed`/`cancelled` rows aged by `COALESCE(completed_at, started_at, created_at)` (a `failed` row never gets a `completed_at`).
- `SurplusQueue.reap_terminal(older_than_days=30)` wrapper; wired into `scheduler.dispatch_once` (new `terminal_retention_days=30` ctor param) right after `drain_expired`.
- **30-day retention** sits far above every scheduler cooldown (max 24h) and the pending-drain window (72h), so `last_completed_at` cooldown checks are never disturbed.
- Rows still referenced by a **non-terminal follow-up** (`linked_task_id`) are preserved so `follow_ups.dispatcher` can still resolve them. The `linked_task_id IS NOT NULL` filter keeps NULLs out of the `NOT IN` set (avoids the classic footgun).

### F10.2 — outreach_mcp orphan-task on init
`init_outreach_mcp`'s standalone fallback used `asyncio.get_event_loop().create_task(ensure_table(db))`, which can spin up an orphan loop and swallows errors. Now `tracked_task(...)` (RuntimeError suppressed in the no-loop path), so a failed `ensure_table` surfaces via the error callback.

### F10.3 — code_audit leaked subprocess on timeout
`_run_cmd` returned on `TimeoutError` while leaving the child running. Now `proc.kill()` + `await proc.wait()` (guarded by `ProcessLookupError`), mirroring `guardian/_subprocess`.

### Dropped: F9 (not a bug)
The audit's F9 (`staleness_policy "drain"→"ttl"` in `outreach/pipeline.py`) was investigated and **dropped as a false positive**. `outreach_delivery` items are owned end-to-end by `OutreachRecoveryWorker`, which marks them terminal itself; `drain` deliberately keeps `expire_by_policy` from age-deleting them (WS-6 / AUTO-RECOV-01). Flipping to `ttl` would re-break that.

## Verification
- **TDD** — 8 new tests (4 crud reaper incl. the follow-up-link exclusion, 1 queue wrapper, 1 scheduler dispatch, 1 code_audit kill-on-timeout, 1 outreach init). 84 pass across the 5 touched test files; ruff clean.
- **Code review** — clean, no findings (NULL/`NOT IN` guard, ISO-timestamp comparison, cooldown headroom, `blocked` treated as open, `proc` binding all verified).
- **E2E dry-run on the live DB** (read-only): the exact reaper query is valid SQL and would reap **3,181** stale rows (→ ~724 remaining); **0** rows wrongly protected or reaped by the exclusion; newest `completed_at` per task type is all *today*, so no cooldown check is endangered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,12 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Completed background-task history no longer grows without bound.** Genesis's idle-time "surplus"
+  task queue kept every finished task row forever — only never-started tasks were ever cleaned up — so
+  the table crept upward over months of background work. Finished tasks (completed, failed, or
+  cancelled) are now aged out after 30 days, while any task still referenced by an open follow-up is
+  kept until that follow-up resolves.
+
 - **Deleting a knowledge item from the dashboard now works end to end.** Previously the delete
   button returned a server error, and behind the scenes the item was only half-removed — dropped
   from search but with its stored embedding left behind and its source still marked as already
--- a/src/genesis/db/crud/surplus_tasks.py
+++ b/src/genesis/db/crud/surplus_tasks.py
@@ -188,6 +188,35 @@ async def drain_expired(db: aiosqlite.Connection, *, before: str) -> int:
     return cursor.rowcount
 
 
+async def delete_terminal_before(db: aiosqlite.Connection, *, before: str) -> int:
+    """Delete terminal (completed/failed/cancelled) tasks older than *before*.
+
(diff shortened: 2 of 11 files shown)


WingedGuardian (commented):

---

PR #858, opened 2026-07-02
fix(memory): filter invalid + deprecated memories in the proactive recall hook

WingedGuardian:
## Problem

The proactive memory hook (`scripts/proactive_memory_hook.py`) injects relevant memories into **every** prompt's context. It searches three ways — FTS5 (`_search_fts5`) + two Qdrant episodic searches (`vector_results`, `wing_results`) — fused by RRF. Its FTS path diverged from the main retrieval path (`memory/retrieval.py` / `db/crud/memory.py::search_ranked`) on **two** validity filters:

| Filter | Meaning | Leaked via FTS (live DB) | Already filtered by |
|---|---|---|---|
| `invalid_at` | bitemporally expired (superseded fact) | 8 memories | `search_ranked` (in-SQL) + `_expired_candidate_ids` (post-query) |
| `deprecated` | superseded/merged by consolidation | **2,596 memories** | `search_ranked` default **and** the hook's own Qdrant `must_not` |

So memories the system had explicitly retired — expired facts, and dream-cycle-consolidated duplicates (old build snapshots, superseded notes) — could still surface on the hottest path. The `deprecated` gap in particular silently undercut memory consolidation at the injection layer.

## Fix

Restore parity, mirroring the main path exactly:

- **`_search_fts5`** gains a `now_iso` param + two in-SQL clauses on the existing `LEFT JOIN memory_metadata`:
  - `AND (invalid_at IS NULL OR invalid_at > ?)`
  - `AND (deprecated IS NULL OR deprecated = 0)`
- **New `_expired_memory_ids(db_path, ids, now_iso)`** helper (mirrors `_expired_candidate_ids`) drops expired ids from the two episodic Qdrant result sets before fusion. `knowledge_base` results are intentionally not filtered (not episodic `memory_metadata` rows). The Qdrant search already excludes `deprecated`, so no change needed there.
- `now_iso` is computed once in `_run` so FTS and Qdrant agree on the as-of instant.

**Safety property:** the filters are NULL-safe — `invalid_at IS NULL` ("valid forever") and `deprecated IS NULL OR = 0` ("not deprecated") are always kept, and the JOIN's orphan-row NULLs are preserved. A memory can only be dropped if it carries a concrete past `invalid_at` or `deprecated=1`, so the filter **cannot over-drop live context**.

## Verification

- **53 tests green** (12 new in `tests/test_hooks/test_proactive_invalid_at.py`, 41 regression incl. the cross-process subsystem-list-sync guard); ruff clean.
- **E2E against a copy of the live DB**, old-vs-new, `_MAX_RESULTS` raised so the LIMIT can't mask the effect:
  - `reboot`: 76 → 75 — dropped exactly **1** expired row, all **75** valid-live kept.
  - `incident`: 272 → 251 — dropped exactly **21** deprecated rows, all **251** valid-live kept.
  - Every dropped row provably expired-or-deprecated; **zero** valid rows dropped.
- Full-hook orchestration smoke ran clean and the expired "Network topology reference" was absent from injected context.
- Two independent code reviews (both clean).

## Deploy

The hook is a `scripts/` per-prompt Claude Code hook — **no DB migration, no server restart**. Deploy = merge + `git pull`.

## Out of scope (tracked as follow-ups)

- `invalid_at` timestamp-format normalization — the DB has 3 formats (`T…+00:00`, 55 legacy space-naive, 1 `Z`); the raw string compare has a same-day latent edge shared by **both** retrieval paths (0 overlap with any currently-leaking memory). Tracking the shared root cause rather than fixing only the hook (which would make the paths disagree).
- Remove the dead `memory_proactive` MCP tool (0 internal callers, redundant with this hook; exposed surface → own review).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Superseded memories no longer resurface in Genesis's automatic recall.** Before every prompt,
+  Genesis injects the most relevant memories into its working context. That fast-path recall wasn't
+  honoring a memory's expiry date — so a fact that had been explicitly replaced (an outdated network
+  note, stale career details, a deprecated index) could still surface, even though the main memory
+  system already filtered it out. The fast path now applies the same validity check across all of its
+  sources, matching the main retriever. Memories with no expiry (valid indefinitely) are never
+  affected, so nothing live is ever dropped.
+
 - **Completed background-task history no longer grows without bound.** Genesis's idle-time "surplus"
   task queue kept every finished task row forever — only never-started tasks were ever cleaned up — so
   the table crept upward over months of background work. Finished tasks (completed, failed, or
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,13 +78,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
-- **Superseded memories no longer resurface in Genesis's automatic recall.** Before every prompt,
-  Genesis injects the most relevant memories into its working context. That fast-path recall wasn't
-  honoring a memory's expiry date — so a fact that had been explicitly replaced (an outdated network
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

PR #859, opened 2026-07-02
fix(guardian): keep host Node + Claude Code aligned so recovery stays online

WingedGuardian:
## Problem

Guardian runs `claude -p` on the host VM for intelligent diagnosis and recovery. Without a working host Claude Code, `diagnose()` escalates and performs no autonomous recovery — but nothing surfaced when that happened:

- `host-setup.sh` installed Node 20, while the pinned Claude Code (2.1.198) requires Node >= 22.
- Every Node/CC install error on the host was redirected to `/dev/null`.
- `update.sh`'s host-CC sync only handled version *drift* — it treated "Claude Code not installed" as "gateway unreachable" and silently skipped.

Net effect: a host could end up with no working Claude Code and Guardian's LLM recovery offline, with no signal.

## Changes

- **Node version pin** in `scripts/lib/cc_version.sh` (`NODE_MAJOR`), derived from Claude Code's `engines.node`, as the single source of truth.
- **New `update-node <major>` gateway op** — installs a pinned Node major via NodeSource with a strict argument allowlist, a passwordless-sudo guard, verify-by-`node --version`, and distro→NodeSource conflict handling. Mirrors the existing `update-cc` op.
- **`update.sh`** now syncs the host Node before Claude Code, *installs* Claude Code when it's absent (not just on drift), distinguishes unreachable / absent / drift, and records any host-side gap as a degraded subsystem in `update_history` instead of skipping silently.
- **`host-setup.sh`** installs Node off the pin (was hardcoded to 20), stops silencing install failures, and adds explicit post-checks.

## Tests

New `update-node` coverage in `tests/test_scripts/test_guardian_gateway_update.py`: argument validation (including newline-injection rejection), the passwordless-sudo guard, idempotence, and verify-by-`node --version`.


WingedGuardian:

---

PR #860, opened 2026-07-02
feat(attention): auditable L1.5 verdict v2 (category + reason) + labeled-row backfill

WingedGuardian:
## Why

The L1.5 attention **judge** (offline shadow LLM that scores whether a moment is worth attention) returned two bare floats `{real, perk}` — as unauditable as the human binary labels we retired. This is the methodology reframe: the perk-up decision is an **LLM judgment**, so humans should **review the judge's reasoning** and iterate the **prompt**, not label fires to tune trigger weights. To review a verdict, you have to be able to *see it*.

## What

**Verdict v2** (`sampler.py`): the judge now returns `{real, perk, category, reason}`.
- `category` ∈ {question, task, decision, problem, chatter, garble}; anything else → `other`.
- `reason` is ONE short sentence the prompt instructs to be a **characterization, not a verbatim quote**.
- `real`/`perk` stay HARD-required + fail-closed; `category`/`reason` are best-effort and additive, so an old `{real, perk}` response still parses with no invented keys.
- `_first_json_object` is now **string-aware** — a stray `}` inside the free-text `reason` no longer prematurely closes the JSON scan (a silent verdict-loss the numeric-only scanner couldn't hit).
- `sample()` stamps `prompt_version` (comparability across prompt iterations) and the serving `model` when the router reports it.

**Firewall note:** the `reason` sentence is a judge *characterization* stored in the private DB — a deliberate, small, **explicitly-approved** relaxation of the prior floats-only rule, made so verdicts are auditable. No raw transcript text reaches the DB.

**Backfill fix** (`consumers.py` + new `crud.bulk_update_l15_verdicts`): `bulk_upsert_events`' `WHERE acceptance_signal IS NULL` guard **freezes a labeled row's derived columns**, so a `--l15` re-run would never attach a verdict to an already-labeled row — exactly the human∩judge rows a review needs. `flush()` now does a second, unconditional pass to backfill verdicts (None-verdict runs skipped, so a plain re-run never nulls a stored verdict).

## Tests

201 passed. New: v2 parse (happy / category coercion / reason cap / omit / **brace-in-reason**), old-format compat, prompt enum + no-quote instruction, `sample()` stamping, crud bulk-update on labeled rows, flush backfill onto a labeled row + no-null-on-plain-rerun.

This is a shadow-only path (no runtime consumer); the judge is still opt-in via the runner's `--l15` flag.

### DIFF
--- a/src/genesis/attention/consumers.py
+++ b/src/genesis/attention/consumers.py
@@ -82,16 +82,45 @@ class ShadowStoreConsumer:
         )
 
     async def flush(self) -> int:
-        """Write all buffered rows via the crud layer in one transaction; returns count."""
+        """Write all buffered rows via the crud layer; returns the upserted count.
+
+        Two passes on one connection: (1) the label-preserving upsert; (2) an unconditional
+        backfill of ``l15_verdict`` for buffered rows that carry one. Pass 2 exists because
+        pass 1's ``WHERE acceptance_signal IS NULL`` guard FREEZES a labeled row's derived
+        columns — so on a re-run WITH ``--l15``, a verdict would never reach an already-labeled
+        row without it (and those are exactly the human∩judge rows the review needs). Only
+        NON-None verdicts are backfilled, so a plain (no ``--l15``) re-run never nulls a stored
+        verdict.
+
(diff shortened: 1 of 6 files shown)


WingedGuardian:

---

PR #863, opened 2026-07-03
fix(scheduler): per-ego job_health keys + memory_extraction boot-run

WingedGuardian:
Two independent scheduler-correctness fixes, verified against live code + the live DB before building.

## 1. Dual-ego `job_health` key collision (observability)
Genesis runs two `EgoCadenceManager`s (user ego + Genesis ego). Each has its **own** `AsyncIOScheduler`, so the APScheduler job ids don't collide — but both called `record_job_success("ego_cycle")` / `record_job_failure("ego_cycle", …)` into the **one** global `job_health` table, so the two egos clobbered each other's health row. Live symptom on the shared row: `last_run` ahead of `last_success` with zero failures — impossible for a single well-behaved job.

**Fix:** use the ego's own `self._session._source_tag` (`user_ego_cycle` / `genesis_ego_cycle`, matching the existing ego_source convention) as the key. Reuses existing identity — no new param, no construction-site change. Blast radius is nil: every `job_health` reader is generic (`SELECT * …` / `WHERE consecutive_failures>0` / parameterized); nothing hardcodes the old literal key.

## 2. `memory_extraction` restart-starvation (reliability)
The memory-extraction job was wired `IntervalTrigger(2h)` with no `next_run_time` — first fire is boot+2h and the timer **resets on every restart**, so a server restarting more often than the interval can defer extraction indefinitely (the IntervalTrigger-resets-on-restart trap). Memory extraction turns recent conversations into long-term memories, so starving it is a real cognitive degradation.

**Fix:** extract a small `_wire_memory_extraction_job(...)` helper that pins `next_run_time` ~60s out (same boot-run pattern as the `schedule_code_audit` job), guaranteeing a run every boot while keeping the steady interval. The helper also gives the wiring a real unit-test seam.

## Tests
- 8 new tests (`tests/ego/test_cadence.py::TestJobHealthKeyPerEgo`, `tests/test_runtime/test_memory_extraction_wiring.py`), written red→green.
- Item 2's boot-run is verified end-to-end against a real started `AsyncIOScheduler` (`next_run_time` ~60s, not 2h).
- Regression: 771 passed / 1 skipped across `tests/ego/` + `tests/test_runtime/`.

## Deploy
Server-loaded code, so it needs a `systemctl restart`. One data step after restart: `DELETE FROM job_health WHERE job_name='ego_cycle'` to clear the now-orphaned shared row (reversible, one row). Definitive live E2E (dashboard shows two per-ego rows updating independently; extraction fires ~60s after restart) runs post-deploy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,18 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **The dashboard now reports each ego's cycle health separately.** Genesis runs two egos (a user-facing
+  one and its own), and both recorded their proactive-cycle health under a single shared key — so on the
+  health surface one ego's last run kept overwriting the other's, making it impossible to tell whether
+  either was actually cycling. Each ego now tracks its own health row, so a stalled or failing ego is
+  visible instead of masked.
+
+- **Memory extraction now runs reliably after a restart.** The job that turns recent conversations into
+  long-term memories was scheduled on a fixed 2-hour interval measured from server start — and that timer
+  reset on every restart, so a box that restarted more often than every two hours could keep deferring
+  extraction indefinitely. It now also runs shortly after each start, so extraction can't be starved by
+  frequent restarts.
+
 - **Superseded and expired memories no longer resurface in Genesis's automatic recall.** Before every
   prompt, Genesis injects the most relevant memories into its working context. That fast-path recall
   wasn't checking whether a memory had expired or been superseded (replaced by a newer, consolidated
--- a/src/genesis/ego/cadence.py
+++ b/src/genesis/ego/cadence.py
@@ -954,7 +954,10 @@ class EgoCadenceManager:
         try:
             from genesis.runtime import GenesisRuntime
 
-            GenesisRuntime.instance().record_job_success("ego_cycle")
+            # Per-ego job_health key (user_ego_cycle / genesis_ego_cycle) — the
+            # two egos write to the ONE global job_health table, so a shared
+            # "ego_cycle" key made them clobber each other's health row.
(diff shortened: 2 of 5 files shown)


WingedGuardian:
## Code Review — PR #863

**Structural:** 1 WARNING
**PII / Secrets / Wording:** CLEAN

---

### WARNING — Genesis convention: `IntervalTrigger` for a >1h interval

**File:** `src/genesis/runtime/init/surplus.py` — `_wire_memory_extraction_job`

The helper still registers the job with `IntervalTrigger(hours=extraction_hours)` for a 2 h cadence. CLAUDE.md Traps says: _"Use `CronTrigger` for intervals >1h — resets on restart."_

The `next_run_time=now+60s` pin is a genuine improvement: it guarantees a run shortly after every boot, so the restart-starvation symptom the PR is fixing is real and the mitigation works. But `IntervalTrigger` still measures its interval from the last run (or last start), whereas `CronTrigger` computes the next fire time from the wall clock — it never "resets" on restart in the same way. The convention exists specifically to avoid the trap this PR is correcting, so the remaining `IntervalTrigger` is still technically a convention violation.

Suggested direction for a follow-up (not a blocker here, since the boot-run mitigates the practical impact):

```python
from apscheduler.triggers.cron import CronTrigger

scheduler.add_job(
    coro,
    CronTrigger(hour=f"*/{extraction_hours}"),   # fires every N hours on the clock
    id="memory_extraction",
    max_instances=1,
    misfire_grace_time=300,
    next_run_time=now + timedelta(seconds=60),   # keep the boot-run pin
)
```

Everything else looks good — imports are correctly grouped, `_wire_memory_extraction_job` has a docstring, the new test file covers both the boot-run and the interval, and the `_record_success`/`_record_failure` fix is clean with no blast radius on existing `job_health` readers.

---

PR #868, opened 2026-07-03
feat(cc): real-time silent-cap detection + DLQ alert honesty

WingedGuardian:
## What & why

In late June a capped Anthropic **Max** subscription made `claude -p` return **silent empty output** (0 tokens, no error, no `rate_limit_event`) on Genesis's cognitive calls. Those blank runs recorded as *successes*, so ego cycles / reflections / weekly jobs produced nothing for ~2 days and **nothing alerted**. This PR makes that visible in real time, and fixes a related cry-wolf on the dead-letter alert.

Scope was set with the user to **detection + alert only** (real-time, invoker-level) + the DLQ threshold honesty fix. Failover of ego/reflection to a peer (and provisioning `ZHIPU_API_KEY` to re-arm the currently-inert CC roster) are **deferred** to the ego-quality spike — verified during planning that those sites have no failover loop and the roster has no authenticable peer today.

## Item 1 — real-time silent-cap detection (no raise, no failover, no control-flow change)

- **`CCInvocation.expect_output: bool = False`** (opt-in). The invoker fires a new injected `on_cc_empty_output` callback at the success returns when `expect_output and not output.text.strip() and not output.is_error` (+ `rate_limit_event` guard on the streaming no-result path). It never raises and never changes the returned output (mirrors `_fire_downgrade_callback`).
- **`cc_relay._on_cc_empty_output`** records each as a low-priority **internal** `cc_cap_empty_event` observation (mirrors `_on_model_downgrade`; 3-day TTL; no migration).
- **`awareness/loop._check_cc_cap_detection`** counts unresolved events in the last 60 min; **≥3 → one critical `infrastructure_alert`** (`cc_cap_monitor`) that rides the existing critical-observations job to Telegram. DB-backed dedup (`content_hash` + `skip_if_duplicate`) so a multi-hour cap yields **one** alert, and it **auto-resolves** on recovery (+ resets the cooldown). **Alert-only** — deliberately does *not* touch `CCStatus`/backoff (that would suppress the very signal + oscillate).
- **12 output-producing cognitive sites** opt in (ego ×2, reflection ×3, sentinel, decomposer, plan/adversarial review, step dispatcher, research, mail judge). Excludes foreground/streaming, direct sessions, the liveness probe, checkpoint resume, gauntlet, and the inbox monitor (already fails-on-empty).

**Calibration:** verified against real `cc_sessions` that `output_tokens=0` is *normal* for foreground/inbox (accounting gap) but the invoker sees the true empty output; and that a row-counting `cc_sessions` detector can't beat ~48h latency (normal cognitive-success gaps reach ~31h; the June cap was a 114h drought) — hence the invoker hook.

**Honest limits (documented in code):** fires only while cognitive work is *attempted* (a cap that also halts dispatch is a blind spot — a ~48h drought backstop is possible future work); targets the *silent* case only (`is_error=False`) — classified errors are already visible.

## Item 2 — DLQ alert counts genuinely-stuck items, not self-healing bursts

The critical DLQ-accumulation alert counted *all* pending equally, so a 7-min burst of ~63 `chain_exhausted:judge` grades (1h TTL, "worthless after minutes") fired a **critical** Telegram alert though it self-heals. New `DeadLetterQueue.get_stuck_count()` counts an item only if it's **not** a short-TTL self-healing type still inside its window (reusing `_OPERATION_TTL_HOURS`); types without a short TTL count immediately (no long-TTL regression). Raw total still shows on the dashboard/snapshot (`get_pending_count` unchanged).

## Verification

- **genesis-architect**: SHIP-WITH-CHANGES — all folded in (observations sink not cc_session metadata; alert-only; `not text.strip()` discriminant; the 12-site opt-in set).
- **code-reviewer**: SHIP-WITH-NITS — the one real finding (cap-alert write-spam under a persisting cap) fixed with dedup + resolve; one pre-existing DLQ boot-caveat tracked as a follow-up.
- **Tests**: `test_invoker_empty_output`, `test_cc_cap_detection`, `test_dead_letter_stuck_count`, updated `test_queues_dead_letter`. Regression: full `test_cc` (639p), `test_awareness`, `test_routing`, observation-TTL — all green. ruff clean on the full diff.
- **E2E** (real schema + live data): detector creates the critical alert → it surfaces via the exact `get_unsurfaced(critical)` Telegram query → raw events stay internal → dedup holds → recovery resolves; and `get_stuck_count` on live data correctly split 39 pending into 36 stuck (aged past 1h) / 3 fresh.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **Genesis now notices when its Claude Code subscription hits its usage cap — instead of quietly going dark.**
+  A capped Anthropic subscription makes `claude -p` return *empty* output with no error, which Genesis used to
+  record as a successful (but blank) run — so its background thinking (ego cycles, reflections, weekly reviews)
+  could silently produce nothing for days without anyone noticing. Genesis now watches for a run of empty
+  results on calls that should have produced output and, when it sees one, sends a single critical alert
+  ("CC subscription likely capped — degraded until the limit resets") so you know to check. It's detection only:
+  it never changes how a call runs, and it stays quiet during normal idle periods.
+
 - **The Genesis Voice add-on's attention surface now shows the judge's reasoning — and lets you review it.**
   For the optional passive-listening add-on, the buried "Attention" tab is now a top-level **Genesis Voice →
   Judgment** review. Each moment the attention gate noticed is scored by a lightweight LLM judge that says
@@ -111,6 +119,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   duplicate. Together these keep the procedure store both more complete and less cluttered.
 
 ### Fixed
+
+- **The dead-letter-queue alert no longer cries wolf on self-healing bursts.** A short burst of low-value
+  retry items (e.g. memory-relevance grades, which are discarded within an hour by design) could push the queue
+  past its alert threshold and fire a *critical* notification for something that clears itself minutes later.
+  The alert now counts only items that are genuinely stuck — pending past their designed self-heal window — so a
+  transient burst stays quiet while a real, un-draining backlog still alerts. The dashboard still shows the full
+  raw count.
 
 - **The dashboard now reports each ego's cycle health separately.** Genesis runs two egos (a user-facing
   one and its own), and both recorded their proactive-cycle health under a single shared key — so on the
--- a/src/genesis/autonomy/decomposer.py
+++ b/src/genesis/autonomy/decomposer.py
@@ -204,6 +204,7 @@ class TaskDecomposer:
 
         invocation = CCInvocation(
             prompt=prompt,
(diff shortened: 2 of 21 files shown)


WingedGuardian:
## Structural review

**ERROR — Raw SQL in non-crud module**

`src/genesis/awareness/loop.py` — `_check_cc_cap_detection` issues a raw `db.execute("SELECT COUNT(*) FROM observations WHERE type = 'cc_cap_empty_event' ...")` directly inside the awareness loop, bypassing the `genesis.db.crud.observations` module.

The DLQ half of this PR gets this right: the new query was encapsulated in `src/genesis/db/crud/dead_letter.py` as `list_pending_type_age`, and `routing/dead_letter.py` calls it via the CRUD layer. `_check_cc_cap_detection` should follow the same discipline — add a helper such as `count_recent_unresolved_by_type_and_source(db, type, source, since)` to `src/genesis/db/crud/observations.py` and call that instead of the inline `db.execute`.

The `resolve_by_source_and_type` call in the same function already uses the CRUD module correctly; just the COUNT query needs to move.

---

---

PR #871, opened 2026-07-03
feat(disk): label-aware attention-snapshot GC + ~/tmp age-prune

WingedGuardian:
## What

Adds two retention steps to the daily disk-hygiene groom, closing two unbounded-growth gaps (`~/tmp`, attention snapshots) found in a whole-data-layer retention audit. Preventive — disk is healthy (26%), but these stores had no periodic guard.

**`scripts/attention_snapshot_gc.py` (new, stdlib-only)** — label-aware GC for attention snapshots. Deletes `ambient_*.db` (>60d) / `omi_*.db` (>14d) **only when no LABELED `attention_event` references them**. A referenced-and-labeled snapshot is kept forever: purging it would make those labeled events permanently review-read-only (a 410 in `resolve_window_text`). Fail-safe by construction — a missing/unreadable DB or a failed label check keeps everything. `--dry-run` supported.

**`scripts/disk_hygiene.sh`** — adds `prune_tmp()` (age-prunes `~/tmp` direct children >7d, excluding `bg-cc-sessions` which is reaped separately at 24h) and a call to the snapshot GC. Refactored into functions + a guarded `main()` so a single step is unit-testable via `source` without running the full groom; the existing steps (worktree reap, `disk_reclaim`, bg-CC-sandbox reap) are preserved byte-for-byte inside `main()`.

## Safety (irreversible deletes)

- Crown-jewel join verified against the **live DB**: the labeled snapshot is kept, the unreferenced one would-delete only when forced past its window. The GC's snapshot-id derivation is grounded against `runner._snapshot_id_from_path` by a test, so it looks up exactly what the runner persists (incl. the `omi_` prefix).
- Fail-safe: no DB / unreadable DB / label-query error → delete nothing.
- Only `ambient_*.db` / `omi_*.db` are eligible; any other file is skipped. Windows are configurable (`--home-days` / `--omi-days`).

## Tests

`tests/test_scripts/test_attention_snapshot_gc.py` (17 cases incl. label-protection, fail-safe-on-missing-DB, unknown-file-untouched, dry-run-inert, runner-derivation grounding) and `test_disk_hygiene_tmp_prune.py` (exclusions + wall-clock-independent). Wall-clock-independent via injected `now` / `os.utime`.

Note: the crontab change that points the daily job at `disk_hygiene.sh` (so these steps run proactively rather than only reactively at ≥85% disk) is a deploy step, tracked separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,12 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **Daily disk hygiene now prunes stale scratch and old attention snapshots.** Housekeeping now
+  age-prunes leftover files in `~/tmp` (older than 7 days) and garbage-collects attention-engine
+  snapshots older than 60 days — but never one behind a moment you've labeled for review, so your
+  labeled history stays revealable. Keeps disk usage from creeping up between the reactive cleanups
+  that previously only fired when the disk was nearly full.
+
 - **Genesis now notices when its Claude Code subscription hits its usage cap — instead of quietly going dark.**
   A capped Anthropic subscription makes `claude -p` return *empty* output with no error, which Genesis used to
   record as a successful (but blank) run — so its background thinking (ego cycles, reflections, weekly reviews)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -54,8 +54,8 @@ systemctl --user list-units 'genesis-*' --all    # All units
 
 Other units: `genesis-bridge.service` (Telegram relay, on-demand),
 `genesis-tmp-watchgod.service` (/tmp protection), `genesis-watchdog.timer`
-(health check), `genesis-disk-hygiene.timer` (daily worktree reaping + cache
-reclaim; see `scripts/disk_hygiene.sh`). MCP servers are CC child processes
+(health check), `genesis-disk-hygiene.timer` (daily worktree reaping, cache reclaim, `~/tmp`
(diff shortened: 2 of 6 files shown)


WingedGuardian:
## Code Review — PR #871

### Structural Review

**WARNING — Raw SQL against genesis.db** (`scripts/attention_snapshot_gc.py:80–85`)

`_is_labeled_referenced` executes raw SQL directly against `genesis.db` rather than going through `genesis.db.crud` modules. Flagging per convention. Note that this is an intentional stdlib-only design (the docstring explicitly mirrors `worktree_lifecycle.py` / `disk_reclaim.py`), so if those scripts are an accepted precedent for the exceptions rule this is already blessed — just confirm that's the intent.

```python
cur = conn.execute(
    "SELECT 1 FROM attention_events "
    "WHERE snapshot_id IN (?, ?) AND acceptance_signal IS NOT NULL LIMIT 1",
    (sid, raw_stem),
)
```

---

**NOTE — Header comment inaccuracy** (`scripts/disk_hygiene.sh:4`)

Changed from `"Run by the genesis-disk-hygiene.timer systemd unit"` to `"Run by the daily cron"`. The script is still invoked by a systemd timer unit, not cron — the body of the same file even says "run under a hardened systemd sandbox". Minor, but could mislead future readers.

---

### PII / Secrets / Wording

**CLEAN** — No API keys, tokens, credentials, private IPs, or hardcoded usernames in added lines. PR title, body, and CHANGELOG wording contain no language about credential leaks, security incidents, or private infrastructure details.

---

PR #855, opened 2026-07-02
feat(observability): honest health surface — PSI/CPU status + per-CC-slot RSS

WingedGuardian:
## Thread 2 — honest health surface (PSI/CPU) + per-CC-slot RSS

Makes the dashboard container-health badge tell the truth, and adds per-session
memory visibility with a leak alert. Two clean commits on one branch (they touch
overlapping files).

### Background
Investigation found the container has **no memory problem** — memory PSI = 0,
anon ~40%, ~21 GB OOM headroom; the climbing `memory.high` counter is benign
page-cache reclaim. So this is a small, additive honesty/visibility improvement,
**not** a fix, and it adds **no** auto-degrade behavior (per the no-auto-throttle
design principle).

### Commit 1 — health-surface honesty (`6a262f75`)
- `_collect_cpu_usage()` derives status from the existing `used_pct` (healthy
  <80 / degraded 80–95 / error >95) instead of hardcoding `"healthy"`. Not
  loadavg (loadavg counts I/O wait and misrepresents CPU).
- New `_read_psi()` reads `/sys/fs/cgroup/{cpu,memory}.pressure`; PSI surfaced on
  both blocks. `memory_status()` = worse of anon% (85/95) and sustained memory
  PSI (`full_avg60` >10% → degraded / >30% → down) — benign reclaim (PSI 0) never
  alarms.
- Dashboard `containerSemantic()`: CPU added as a 3rd worst-of dimension; memory
  reason reports PSI when it's the driver. `resources.py`: PSI surfaced; memory
  status `critical` → `down` (canonical vocabulary; the value was unconsumed).

### Commit 2 — per-CC-slot RSS + leak alert (`0711dd68`)
- New stdlib leaf `observability/cc_slots.py`: `enumerate_cc_slots()` walks
  `/proc` for `claude` processes tagged `GENESIS_SLOT` and reads VmRSS. Same-uid
  `/proc` reads (work despite `ptrace_scope=1`, which gates ATTACH not READ).
- Infra snapshot + `/api/genesis/resources` + the dashboard Container card show
  per-slot RSS.
- `awareness/loop.py::_check_cc_slot_memory` — WARN 4 GB / CRIT 6 GB on the main
  `claude` RSS (normal ~0.7–1 GB). Mirrors `_check_wal_health`; creates a
  `type=infrastructure_alert` observation that rides the existing
  critical-observations job → Telegram (CRIT) / morning report (WARN). Runs
[...]

PR #873, opened 2026-07-03
fix(dashboard): honest memory GB label — stop pairing 47% with cache-inflated 32/36 GB

WingedGuardian:
## Problem
The dashboard health widget showed **Memory 47%** next to **32/36 GB** — a contradiction (32/36 ≈ 89%).

They measured different things:
- **47%** = `anon_pct` = non-reclaimable memory (anon + kernel) ÷ limit — the honest OOM signal.
- **32/36 GB** = cgroup `memory.current` ÷ `memory.max` — **includes ~16 GB of reclaimable page cache**.

`#855` (honest health surface) correctly switched the **percentage and bar** to `anon_pct`, but left the **GB label** at `genesis_dashboard.html:5156` still reading `current_gb` — so the honest % ended up paired with the cache-inflated absolute. (Before #855 they agreed at ~87%; the fix updated one half.)

It was never a real memory problem — the container had ~20 GB genuinely available; the extra "used" was reclaimable cache the kernel drops on demand.

## Fix
Show honest non-reclaimable used GB (`anon_gb + kernel_gb`, which matches `anon_pct`) instead of `current_gb`, with a tooltip noting GB available and that `memory.current` includes reclaimable cache. Falls back to `current_gb` if the breakdown is unavailable. One-line template change; the snapshot already exposes every field used.

## Verification
`anon_gb+kernel_gb` ≈ 16 GB → **"16/36 GB" next to 47%** (consistent). Node-checked both the normal and fallback expressions; confirmed all fields (`anon_gb`, `kernel_gb`, `current_gb`, `limit_gb`, `available_gb`) are set by the infrastructure snapshot.

### DIFF
--- a/src/genesis/dashboard/templates/genesis_dashboard.html
+++ b/src/genesis/dashboard/templates/genesis_dashboard.html
@@ -5153,7 +5153,8 @@
                          :style="(() => { const p = $store.genesisDashboard.health.infrastructure.container_memory.anon_pct ?? $store.genesisDashboard.health.infrastructure.container_memory.used_pct ?? 0; return 'width:' + p + '%; background:' + (p > 85 ? '#d9534f' : p > 75 ? '#f0ad4e' : '#4caf50'); })()"></div>
                   </div>
                   <span style="font-size:0.7rem; color:var(--color-text-secondary); white-space:nowrap;"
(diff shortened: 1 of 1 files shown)


WingedGuardian:
Structural review: CLEAN — no pattern violations found. (Single HTML template change; Python-specific rules not applicable.)

---

PR #874, opened 2026-07-03
fix(scheduler): restart-safe triggers + genesis.db drip-table retention

WingedGuardian:
Part of the whole-data-layer retention work (follows #871). Two restart-safety + retention fixes for the runtime scheduler.

## D2 — restart-safe maintenance + code-index triggers

`schedule_maintenance` (24h) and `schedule_code_index` (4h) used `IntervalTrigger(hours=N)`, which resets its countdown on every server restart — so under frequent restarts (daily / CC-OOM) the interval never elapses and the job **starves** (the CLAUDE.md >1h trap; the same class #863 fixed for `memory_extraction`). `schedule_maintenance` drives the surplus GC block (surplus TTL, `pending_embeddings`, heartbeat/weak-link pruning, transcript archival), so starving it directly worsens disk growth.

New `_restart_safe_hourly()` helper → wall-clock `CronTrigger` + a ~60s boot-run pin (mirroring the `code_audit`/`memory_extraction` pattern). Each job's `_recently_completed(..., N)` cooldown is unchanged and stays the true cadence gate, so the fixed fire only needs to be frequent enough. **`code_index` was enumerated as the same bug class**, not just the flagged maintenance job (`code_audit` is disabled + already boot-pinned; the minute-cadence jobs are harmless).

## D3 — retention for 3 genesis.db drip tables

`execution_traces` (per execution), `cost_events` (per LLM call), and the `file_modifications` audit trail (per file edit, written by `file_modification_audit_hook.py`) accrued a steady drip with no scheduled GC. Registers restart-safe `CronTrigger` retention jobs in the learning scheduler (staggered off-peak after `otel_span_prune`), each a 90d prune.

- Adds `prune_older_than(days=90)` to `execution_traces` + `cost_events` crud (single indexed DELETE on `created_at`, mirroring the existing `file_modifications.prune_older_than`, which had **no caller until now**).
- **cost_events floor:** 90d stays well clear of the monthly budget window (`cost_tracker._period_start('this_month')` looks back ≤ ~31d); a test locks it (a 30-day row survives).

## Retention windows (as approved)
`~/tmp` 7d (in #871) · execution_traces/cost_events/file_modifications **90d** · cost_events hard floor ≥35d.

## Tests
Trigger-class assertions + helper unit tests (`test_surplus/test_scheduler.py`); crud prune + floor tests, all **wall-clock-independent** (age via `created_at`/`timestamp`). `test_runtime` green (learning-init path). No expected-jobs registry to update (verified).

### DIFF
--- a/src/genesis/db/crud/cost_events.py
+++ b/src/genesis/db/crud/cost_events.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+from datetime import UTC, datetime, timedelta
 
 import aiosqlite
 
@@ -161,3 +162,19 @@ async def delete(db: aiosqlite.Connection, id: str) -> bool:
     cursor = await db.execute("DELETE FROM cost_events WHERE id = ?", (id,))
     await db.commit()
     return cursor.rowcount > 0
+
+
+async def prune_older_than(db: aiosqlite.Connection, days: int = 90) -> int:
+    """Delete cost_events older than N days (by ``created_at``). Returns count deleted.
+
+    Single indexed DELETE (``idx_cost_events_created``); runs in a daily maintenance job.
+    The 90d default stays well clear of the monthly budget window
+    (``cost_tracker._period_start('this_month')`` looks back at most ~31d), so retention
+    can never delete a row the budget accounting still needs.
+    """
+    cutoff = (datetime.now(UTC) - timedelta(days=days)).isoformat()
+    cursor = await db.execute(
+        "DELETE FROM cost_events WHERE created_at < ?", (cutoff,),
+    )
+    await db.commit()
+    return cursor.rowcount  # type: ignore[return-value]
--- a/src/genesis/db/crud/execution_traces.py
+++ b/src/genesis/db/crud/execution_traces.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+from datetime import UTC, datetime, timedelta
 
 import aiosqlite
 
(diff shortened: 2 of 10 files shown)


WingedGuardian (commented):
## First-pass review — PR #874

### Structural review

**ERRORs: none**

**WARNINGs**

**[WARNING] Missing registration test for 3 new learning-scheduler jobs** (`src/genesis/runtime/init/learning.py`)

`_prune_execution_traces`, `_prune_cost_events`, and `_prune_file_modifications` are registered via `rt._learning_scheduler.add_job(...)` but there is no test asserting those job IDs actually appear in the scheduler after init. Compare with `test_maintenance_and_code_index_triggers_are_restart_safe` in `tests/test_surplus/test_scheduler.py`, which does exactly this for the surplus scheduler. The crud prune functions themselves are well-covered, but if `add_job` calls were inadvertently removed or silently failed, no test would catch it. Suggest adding a `test_runtime` (or equivalent) assertion along the lines of:
```python
for job_id in ("execution_traces_prune", "cost_events_prune", "file_modifications_prune"):
    assert rt._learning_scheduler.get_job(job_id) is not None
```

**NOTEs**

**[NOTE] Redundant local import of `CronTrigger` inside `_restart_safe_hourly`** (`src/genesis/surplus/scheduler.py`)

`from apscheduler.triggers.cron import CronTrigger` is imported inside the function body, but the surrounding module already uses `CronTrigger` in other `add_job` calls (e.g., the Recon gather job), which means a module-level import already exists. The local re-import is harmless but can confuse readers scanning the module for its dependencies. Can be moved to the top of the file.

**[NOTE] Fragile `__str__`-based assertion in trigger test** (`tests/test_surplus/test_scheduler.py`, `test_restart_safe_hourly_subdaily_is_stepped_and_daily_differs`)

The test asserts `"minute='10'"` as a substring of `str(_restart_safe_hourly(...))`. This couples the test to APScheduler's `__str__` representation. If APScheduler changes its repr format in a future version, the assertion silently stops catching regressions. Consider asserting on the trigger's `.fields` dict directly (e.g., checking `trigger.fields` for the `minute` field value), which is more stable than string output.

---

**CLEAN** — no secrets, private IPs, credentials, personal emails, or private hostnames found in added lines. PR title and body contain no incident-disclosure language or private infrastructure references.

---

PR #877, opened 2026-07-03
fix(ego): restart-safe ego_cycle boot first-fire

WingedGuardian:
## Problem

The ego's proactive cadence (`EgoCadenceManager`) armed its `ego_cycle` job with `IntervalTrigger(minutes=self._current_interval)` and no `next_run_time`. APScheduler's `IntervalTrigger` counts from scheduler start, so the first proactive fire was `boot + interval` and reset on every restart. Because the interval adapts upward when things are idle (base → up to 72h), an install that restarts more often than the current interval could keep deferring the proactive cycle — the same IntervalTrigger-resets-on-restart trap #863 fixed for `memory_extraction`.

## Fix

Anchor the boot first-fire to this ego's own persisted `job_health.last_success` (per-ego since #863):

- `_compute_boot_first_fire()` returns `max(now + 60s, last_success + base_cadence)` — an overdue ego fires shortly after boot; an up-to-date one waits only the remaining base interval, so a restart doesn't re-run the expensive cycle prematurely.
- `start()` passes `next_run_time` only when non-`None`; a fresh install (no prior run) omits it and keeps APScheduler's default timing. It never passes `next_run_time=None`, which APScheduler treats as a *paused* job.

The ego interval is variable/adaptive, so unlike `memory_extraction` it keeps `IntervalTrigger` (a wall-clock `CronTrigger` can't express a changing interval) — only the boot arming changes. No new persistence: it reuses the per-ego `job_health` row. The after-cycle reschedule is unchanged, so this affects only the first fire after a restart. On a fresh install (or before the per-ego row exists) it degrades to the previous default timing.

## Testing

- 7 targeted tests in `tests/ego/test_cadence.py`: overdue → boot-pin, recent → remaining base, no-row / missing-table → safe fallback, per-ego isolation, `start()` pins `next_run_time` when overdue, fresh install not paused.
- `tests/ego/test_cadence.py` passes (81/81); `ruff` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -126,6 +126,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Genesis's background egos now keep their thinking rhythm across a restart instead of going quiet.**
+  The two egos run proactive cycles on an adaptive schedule that stretches out when things are idle. That
+  schedule was re-armed from scratch on every restart, so an install that restarts often (deploys, recovery)
+  could keep pushing the next cycle further out — in the worst case starving an ego for up to its full
+  backed-off interval. Each ego now anchors its first post-restart cycle to when it last actually ran: an
+  overdue ego runs shortly after startup, while an up-to-date one simply keeps its cadence.
+
 - **Re-ingesting a knowledge source with changed content now refreshes it instead of serving the stale
   version.** Previously, once a file or URL was ingested, re-ingesting the same source was skipped on source
   identity alone — so if the underlying content changed, the knowledge base kept serving the old distilled
--- a/src/genesis/db/crud/job_health.py
+++ b/src/genesis/db/crud/job_health.py
@@ -31,3 +31,20 @@ async def get_stale_jobs(
         (threshold_days,),
     )
     return [dict(r) for r in await cursor.fetchall()]
+
+
+async def get_job_last_success(
+    db: aiosqlite.Connection, job_name: str
+) -> str | None:
+    """Return the ISO ``last_success`` timestamp for ``job_name`` (or None).
+
+    Used by the ego cadence to anchor its restart-safe boot first-fire to the
+    last time this ego actually cycled (see
+    ``EgoCadenceManager._compute_boot_first_fire``).
+    """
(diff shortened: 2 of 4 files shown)


WingedGuardian:
## First-pass review: fix(ego) restart-safe ego_cycle boot first-fire

### Structural

**[ERROR] Raw SQL in production code — `src/genesis/ego/cadence.py`**

`_compute_boot_first_fire()` calls `self._db.execute()` with a hand-written SQL string:

```python
    "SELECT last_success FROM job_health WHERE job_name = ?",
    (self._session._source_tag,),
)
```

Genesis convention: raw SQL against genesis.db must go through `genesis.db.crud` modules; raw SQL is only permitted in migration files. Please route this query through the appropriate crud layer (or add a `get_job_health_last_success(job_name)` helper there if one doesn't exist).

---

**[WARNING] `IntervalTrigger` kept for an interval that can reach 72 h (> 1 h convention limit)**

The Genesis convention flags `IntervalTrigger` on intervals > 1 h and recommends `CronTrigger` to avoid the resets-on-restart trap. The PR body explains the rationale (the adaptive/variable interval can't be expressed as a wall-clock `CronTrigger`), and the fix itself mitigates the restart-starvation problem — but the deviation from convention is worth recording explicitly in a comment in `start()` or `_compute_boot_first_fire()` so future readers don't treat this as an oversight.

---

### PII / Secrets / Wording

**CLEAN** — no secrets, tokens, private IPs, or sensitive identifiers in added lines. PR title, body, and CHANGELOG additions contain no PII-leak disclosure or security-incident language.

---

WingedGuardian:
Addressed in 27ff9939:

- **[ERROR] raw SQL** — added `crud.job_health.get_job_last_success()` and routed `_compute_boot_first_fire` through it (raw SQL now stays in migrations/crud only).
- **[WARNING] IntervalTrigger >1h deviation** — added an explicit note in `start()`: the ego interval is variable/adaptive (base → up to 72h) so a wall-clock `CronTrigger` can't express it, and the boot first-fire anchoring is the reset-trap mitigation.

No behavior change; `tests/ego/test_cadence.py` 81/81 pass, ruff clean.

(Note: the automated Codex review was quota-limited this cycle.)

PR #878, opened 2026-07-03
fix(observability): health snapshot — offload blocking I/O, isolate section failures, coalesce callers (WS-D)

WingedGuardian:
## WS-D — health-snapshot correctness (F4 + F10 + single-flight)

`HealthDataService.snapshot()` ran three synchronous I/O calls **on the event loop** while building the return dict, and its `asyncio.gather` had no failure isolation. Under load this stalled the loop, and a single unguarded sub-snapshot raise 500-ed the whole health route.

Two files: `observability/health_data.py` + a small extract-refactor of `observability/snapshots/services.py`. No migration.

### What changed
- **F4 — no blocking I/O on the loop.** Fold the blocking work into the existing gather via `asyncio.to_thread`: `conversation_activity()` + `proactive_memory_metrics()` (pure file I/O) offload whole; `services()` is split so only `collect_service_status()` (the ~7 `systemctl` subprocesses) is threaded, while the **sentinel read stays on-loop** — `sentinel.state` is mutated in place field-by-field on the loop, so offloading it would open a torn-read window. Host-framework `detect()` is near-free (empty registry) so it stays on-loop too. All three run concurrently with the DB/network sub-calls → ≈0 added wall-clock vs. the old serial post-gather calls.
- **F10 — gather failure isolation.** `return_exceptions=True` + a uniform `_or_error()` map: a raising sub-snapshot (e.g. `call_sites`, which has no internal guard) degrades to a `{"status": "error"}` section instead of taking down the whole snapshot; `CancelledError`/`BaseException` still propagate (never swallowed).
- **Single-flight coalescer.** Concurrent callers that share the runtime singleton (both ego-context builders on an awareness tick, sentinel dispatch, morning report) coalesce onto one in-flight compute. The shared task is awaited through **`asyncio.shield()`** so one caller's cancellation (e.g. the dashboard route's 15s `_async_route` timeout) can't abort the computation for the others; `_inflight` is released via the task's done-callback. Callers are read-only, so the shared dict needs no copy.

### Review + verification
- **Code review caught a Critical** (unshielded shared task → cross-caller cancellation cascade) — fixed with `asyncio.shield()` + done-callback lifecycle, plus a regression test proving a cancelled caller can't abort its coalesced siblings.
- **Tests (TDD):** new `test_health_snapshot_offload.py` (offload-off-thread, gather isolation, coalescing, cancellation-isolation) + existing guards green (`test_snapshot_includes_services_key`, `test_all_none_produces_valid_snapshot`, `test_services_snapshot.py`, `test_health_cache.py`). Full `test_observability/` (290) + ego/sentinel/morning-report consumer suites (236) pass.
- **E2E (real code path):** a warm snapshot runs in **57ms with a 14ms max event-loop gap** (loop not blocked), returns the full section set with real systemd/sentinel data, and concurrent callers coalesce without error. Live `/health` is 8–11ms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -126,6 +126,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **The dashboard's system-health view stays responsive under load and no longer errors out when a single
+  check hiccups.** Building the health snapshot used to run its systemd service checks (several `systemctl`
+  calls) and a couple of file scans directly on the main event loop, so gathering health could briefly stall
+  other work; and if any one sub-check raised an unexpected error, the entire health request failed. Those
+  checks now run off the main loop, and a failure in one section degrades just that section to an error state
+  while the rest of the health view loads normally. Overlapping health requests now also share a single
+  computation instead of each recomputing from scratch.
+
 - **Uploading a large or malformed PDF to the Knowledge tab no longer freezes — or crashes — Genesis.**
   PDF text extraction used to run directly on the main event loop, so a big document could stall the whole
   server for seconds (health checks, background thinking, other requests all waited), and a corrupt or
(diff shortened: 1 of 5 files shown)


WingedGuardian (commented):

---

PR #880, opened 2026-07-03
fix(ego,routing): stop COO reactive-waste from provider-exhaustion storm

WingedGuardian:
## Problem
The genesis (COO) self-maintenance ego ran ~94 reactive Opus/high cycles in 14 days, **~92% producing zero proposals**. It kept waking on routing `all_exhausted` ERROR events it fundamentally cannot act on (provider config is outside its domain), investigating, and concluding "nothing I can do" — on the order of $250/mo of wasted cognition.

## Root cause
The `judge` and `38a_procedure_novelty_llm` chains are **all-deepseek with no non-deepseek fallback**. When the deepseek providers' circuit breakers trip (they flap intermittently), every call exhausts and storms the event bus with `all_exhausted` (judge alone: ~398 in 3 days). Those ERROR events wake the COO's reactive path. Sibling chains are unaffected because they use non-deepseek providers.

## Fix (the high-value core; 2 of 4 investigated layers)
**L1 — domain-gate the reactive path** (`runtime/init/ego.py`): drop routing/provider `all_exhausted` from the COO reactive path via a new module-level `_is_non_actionable_infra_event(subsystem, event_type)`. **Nothing is hidden** — the condition still reaches the ego via `ProviderEscalation` → `provider_failure` observation, read in *proactive* context (`genesis_context._observations_section`). Only `all_exhausted` from routing/providers is gated; `budget.exceeded` and every other event still wakes the ego.

**L2c — fail-fast in the novelty judge** (`routing/circuit_breaker.py` + `learning/procedural/extractor.py`): new `CircuitBreakerRegistry.chain_has_available(chain)`; the cross-type dedup skips its `route_call` when the 38a chain has no available provider, failing open to "novel" (same outcome as a failed judgment, minus the doomed call that emits another `all_exhausted`). Precision-safe: worst case is a paraphrase duplicate stored, never a false-merge.

## Deferred (intentionally not here)
- **L2a** DLQ redispatch breaker-gate — the DLQ is currently empty (1h TTL clears items before the 6h redispatch cron); hygiene follow-up.
- **L3** a non-deepseek fallback for `judge`/`38a` — precision-constrained (0-false-merge validated on deepseek-v4); needs a fresh validation run + a model decision.

## Tests & verification
- 65 targeted tests green (9 new): the L1 predicate, an end-to-end test through the subscriber closure (all_exhausted gated / budget.exceeded forwarded), `chain_has_available` (any-open / all-open / empty), and the extractor fail-fast + fall-through paths. Existing tests unchanged — the guard is a no-op when providers are available.
- Code review (feature-dev:code-reviewer): SHIP-WITH-NITS; both coverage nits closed.
- E2E against the **live** routing config + durable breaker state: the predicate gates correctly, and `chain_has_available` returned "proceed" while the deepseek breakers were recovered (flapping, not dead) — confirming the fail-fast fires only during down-windows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -139,6 +139,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Genesis's COO (self-maintenance ego) no longer burns cognitive cycles reacting to model-provider
+  outages it can't fix.** When a provider chain runs out of options (e.g. a temporary DeepSeek outage),
+  the "all providers exhausted" alerts no longer wake the ego into a full high-effort reasoning cycle that
+  can only conclude "nothing I can do" — these were the large majority of its zero-outcome reactive cycles.
+  The outage still reaches the ego through its normal system-health context, so nothing is hidden; it just
+  stops paying to react to it. Relatedly, the cross-type procedure-dedup check now skips its judgment call
+  outright when that chain is down, instead of firing a doomed request that would raise yet another alert.
+
 - **The dashboard's system-health view stays responsive under load and no longer errors out when a single
   check hiccups.** Building the health snapshot used to run its systemd service checks (several `systemctl`
   calls) and a couple of file scans directly on the main event loop, so gathering health could briefly stall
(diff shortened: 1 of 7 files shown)


WingedGuardian (commented):

---

PR #882, opened 2026-07-03
fix: autonomy context-ceiling gate + Guardian SSH key hardening

WingedGuardian:
## Summary

Two security-posture fixes surfaced by an architecture audit, plus a documented limitation marker.

### Autonomy gate honors the context ceiling (`proposal_gate`)
The background-dispatch gate compared an action's required autonomy level against the raw *earned* level for the background-cognitive category. An earned level can legitimately sit above a context's ceiling (the ceiling is a per-context clamp, not a cap on earning), so a category that had earned L4 could clear a check it shouldn't in the more restricted background context — most visibly letting a FINANCIAL action (which requires L4) pass the gate's level check even though background cognition is capped at L3. The gate now uses the ceiling-clamped effective level. Only autonomous FINANCIAL background dispatch is newly refused; every other domain is unaffected. Explicit user approval was, and remains, required for these actions, so this closes a defense-in-depth gap behind that approval.

### Guardian→host control SSH key hardening (`install_guardian.sh`)
The installer wrote the container→host control key without the `from=` source-IP restriction and `no-pty` that its own gateway header documents, so a stolen key was usable from any host on the LAN and could request a PTY. The key is now installed with `no-pty` and a `from=` bound to the container's actual host-facing source address (derived from its egress route, not from a listing that can return an unrelated address). A source address that doesn't match is caught by the installer's own connectivity test and rolled back to a working, `no-pty`-only entry, so the restriction can never lock Guardian out of the host. Re-running the installer upgrades an older un-hardened key in place and self-heals a stale restriction.

### Post-exec audit limitation marker (`audit.py`)
The post-execution auditor's protected-path scan reads Write/Edit tool calls only, so it cannot see file changes made via Bash. This is currently safe (no background session profile can write files via Bash), so rather than add a partial heuristic, a comment documents the bound and the condition under which it must be revisited.

## Testing
- New and existing autonomy gate tests pass; the full autonomy package suite is green.
- An end-to-end check with a real autonomy manager confirms the ceiling clamp blocks a background financial action while leaving lower-risk domains untouched.
- Installer changes validated with shellcheck and a logic harness covering source-address derivation, idempotent re-run, and the rollback path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,23 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   and several denial-of-service vectors). Genesis uses it only as a client for outbound calls, so
   real-world exposure was limited, but the newer version removes the advisories outright.
 
+- **The autonomy gate now honors each action category's context ceiling before auto-dispatching a
+  background action.** When deciding whether a background action may run, the gate compared the
+  required level against the raw *earned* level — so a category that had earned a high level could
+  clear a bar it shouldn't in a more restricted context, most notably letting a financial action pass
+  its level check. It now uses the ceiling-clamped effective level, so background auto-dispatch of
+  higher-risk categories (e.g. financial) is refused regardless of earned level. Your explicit
+  approval was, and still is, required for these actions — this closes a defense-in-depth gap behind
+  that approval.
+
+- **The Genesis→host control SSH key is now bound to the container's source address and denied an
+  interactive terminal.** The key that lets the container drive recovery on the host is installed with
+  a `from=` source-IP restriction and `no-pty`, so a copied key can't be used from another machine on
+  your LAN and can't request a shell. Re-running the Guardian installer upgrades an existing key in
+  place; if the source address can't be confirmed it keeps the terminal restriction and safely skips
+  the address lock (verified by the installer's own connectivity test) rather than risk locking
+  Guardian out.
+
 ### Added
 
 - **Off-site backups now self-prune on a grandfather-father-son schedule instead of growing forever.**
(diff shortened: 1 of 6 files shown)


WingedGuardian:
## First-pass code review

### Structural review

**WARNING — Genesis conventions: missing conventional commit prefix**

The PR title `Autonomy context-ceiling gate + Guardian SSH key hardening` lacks a conventional commit prefix. When this is squash-merged the commit message will be unprefixed, breaking the conventional-commit log. Suggest `fix: autonomy context-ceiling gate + Guardian SSH key hardening` (these are both correctness fixes rather than new features).

Everything else is clean:
- No `nohup`, no direct `task_states` inserts, no `/tmp/` misuse (installer temp file written to `$HOME/.ssh/` via `${_ak}.tmp.$$`, same FS — safe atomic `mv`)
- No bare `except:`, no mutable default args, no `os.killpg`, no raw SQL
- `audit.py` change is comment-only — no test file required
- `proposal_gate.py` functional change is fully covered by the new `test_financial_uses_effective_not_raw_level` test, including an explicit assertion that `effective_level` (not `get_state`) was consulted

CLEAN.

- All IP references in `install_guardian.sh` additions are runtime-derived (`ip -4 route get …`) — no hardcoded private IP literals committed
- No API keys, tokens, passwords, SSH private keys, or connection strings in any added lines
- No personal addresses, absolute paths with usernames, AWS/GCP IDs, or non-generic private hostnames
- CHANGELOG wording describes a security hardening appropriate for user-facing release notes — no incident language, leaked-credential mentions, or exploitable detail

---

WingedGuardian:
Review note: the Codex reviewer was quota-limited on this PR (no Codex pass). The structural review ran clean (one soft warning — PR-title conventional-commit prefix — now addressed) and the PII/secrets/wording scan was CLEAN. All CI checks pass (test, lint, migration-check, dependency-audit, leak-detector, external-io-check, CodeQL). Merge proceeds on CI + structural review per the review-gate policy.

PR #883, opened 2026-07-03
fix(guardian): age-prune stale snapshots + incus expiry to prevent thin-pool exhaustion

WingedGuardian:
## Problem

The Guardian's pre-recovery Incus snapshots could accumulate indefinitely and fill the host storage pool. On an LVM-thin backend this exhausts the thin pool, forcing the container's root filesystem read-only — a hard outage. Two gaps allowed it:

1. A `guardian-…-pre-recovery` snapshot sorts as the **"newest"** by name suffix, so the retention rule (keep newest + latest-healthy) protected it forever, accumulating copy-on-write divergence.
2. `prune()` only ran in the `HEALTHY` state, so a Guardian stuck in another state never cleaned up.

## Changes

- **Age-based prune** (`SnapshotManager.prune`): delete `guardian-*` snapshots older than `max_age_days` (default 14) — including a stale snapshot that currently sorts as newest — while **always** exempting the most-recent `-healthy` snapshot, which is the offline snapshot-rollback lifeline.
- **`enforce_expiry_policy()`**: idempotently set incus `snapshots.expiry` (default `2w`) so the daemon auto-deletes aged **scheduled** snapshots even if the Guardian process is dead. Deliberately does **not** set `snapshots.expiry.manual` (instance-wide; would expire snapshots an operator creates by hand).
- Snapshot maintenance (expiry + prune) now runs after **every** check cycle regardless of Guardian state, throttled to once per 24h via the existing marker.
- Robust `created_at` parsing for incus RFC3339 timestamps (handles `Z` suffix and nanosecond precision).
- Operator prompt (`guardian-claude.md`): timestamped pre-recovery snapshot names + delete-after-verify guidance.

## Testing

- `pytest tests/test_guardian/` — 410 passing.
- New tests cover: age-pruning a stale "newest" pre-recovery snapshot, preserving an aged healthy lifeline, expiry-policy sets `snapshots.expiry` (not `.manual`), maintenance runs in all states, and prune-failure still advances the marker (no retry storm).

### DIFF
--- a/config/guardian-claude.md
+++ b/config/guardian-claude.md
@@ -37,7 +37,7 @@ Container inspection (via incus exec):
 - `incus exec genesis -- su - ubuntu -c "cd ~/genesis && git log --oneline -5"` — Recent commits
 - `incus info genesis` — Container status
 - `incus restart genesis` — Restart container (last resort)
-- `incus snapshot create genesis guardian-pre-recovery` — Snapshot BEFORE recovery
+- `incus snapshot create genesis guardian-$(date +%Y%m%d-%H%M%S)-pre-recovery` — Snapshot BEFORE recovery (timestamped name so it sorts and prunes correctly; the guardian auto-prunes guardian-* snapshots older than ~14 days, but DELETE this one yourself once recovery is verified: `incus snapshot delete genesis <name>`)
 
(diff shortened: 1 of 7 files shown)


WingedGuardian:

---

PR #886, opened 2026-07-03
fix(surplus): intake fence-stripping + bare-array findings + WING_AUDIT

WingedGuardian:
## Summary

Batch 1 PR-1 of the 2026-07-03 context-layer audit remediation. The surplus → knowledge pipeline was severed at intake: model output arrives wrapped in ```json fences that the bare `json.loads()` never parsed, code_audit emits bare top-level arrays no branch handled, and WING_AUDIT was missing from the multi-finding set — so months of findings were stored as raw fenced blobs (388 polluted knowledge units on this install).

- `atomize()` unwraps a whole-message ```json fence (regex single-sourced in intake.py, imported by executor.py), parses bare top-level arrays via a shared per-item extractor (identity-key title fallbacks, one-level nesting flatten), and treats empty/unusable envelopes as a no-op instead of storing the raw envelope.
- Per-item extraction is hardened against untrusted shapes (null/string `sources`, null `relevance`) so a malformed field degrades that item only, never the whole payload.
- `wing_audit` joins the multi-finding set + task-type→source map.
- **Scheduler**: CODE_AUDIT output no longer routes through generic intake — FindingsBridge owns per-finding ingestion behind its 0.7-confidence gate and slop filter; the generic route would double-ingest every finding and bypass both. The quality judge still runs.
- **New one-off script** `scripts/cleanup_fenced_knowledge_units.py` (dry-run default) re-parses the polluted rows: recover → proper units (provenance/confidence preserved, original id + timestamp as tags); malformed whole-fence junk → delete (D2); prose with incidental inline fences → keep. Deletes run the memory-layer cascade first via `store.delete()` (Qdrant point + memory_fts/metadata/links — a knowledge_units-only delete would leave bodies recallable via FTS), keep the row on cascade failure (re-runnable) with nonzero exit, and dump every deleted row to a JSONL backup first.

## Verification

- NEW `tests/test_surplus/test_intake.py` (31 tests — no coverage existed); RED-proven against unfixed code. Scheduler test pins the CODE_AUDIT intake skip. 101 tests green across test_intake/test_executor/test_scheduler.
- Script dry-run against the live DB: 388 rows → 369 recovered (1275 proper units), 5 malformed junk, 14 legit prose kept, 0 data loss. Executes post-merge (backup + 6h cron backups in place).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -397,6 +397,16 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   report (the dedicated morning-report pipeline is the sole source — its daily-briefing cycle now
   focuses on what you need today), and the pending-proposal guidance matches the real threshold.
 
+- **Background research findings reach your knowledge base again — instead of piling up as unreadable
+  JSON blobs.** Genesis's idle-time research (brainstorms, gap clustering, code and wing audits) sends its
+  findings through an intake step that splits them into individual knowledge entries. That step couldn't
+  read the output format models actually produce (code-fenced JSON, and the bare-array shape code audits
+  emit), so months of findings were stored as single raw JSON dumps — useless to search and recall. Intake
+  now parses these formats, skips empty result envelopes outright, and wing-audit results are split into
+  individual findings like the other research types. A new `scripts/cleanup_fenced_knowledge_units.py`
+  (dry-run by default) re-parses the previously mangled entries into proper knowledge units and removes
+  the junk ones.
+
 - **The dashboard's per-session memory row is clearer: "Claude Code Sessions", green when healthy.** The
   cryptic "CC" row that listed sessions as gray "cc-1 …" chips is now labeled **Claude Code Sessions**,
   renders each healthy session in green (amber ≥ 4 GB, red ≥ 6 GB), and shows a hover tooltip explaining
--- a/scripts/cleanup_fenced_knowledge_units.py
+++ b/scripts/cleanup_fenced_knowledge_units.py
@@ -51,14 +51,6 @@ from pathlib import Path
 logging.basicConfig(level=logging.INFO, format="%(levelname)s %(message)s")
 logger = logging.getLogger("cleanup_fenced_ku")
 
-_SELECT_SQL = """
-SELECT id, project_type, domain, source_doc, source_pipeline, body,
-       confidence, ingested_at, qdrant_id
-FROM knowledge_units
-WHERE source_doc LIKE 'intake:%' AND body LIKE '%```json%'
-ORDER BY ingested_at
-"""
-
 
 def _derive_task_type(title: str, multi_types: frozenset[str]) -> str:
     """Reverse the single_item title back to a task type for re-atomization.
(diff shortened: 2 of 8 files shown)


WingedGuardian:
## Structural Review

### ERROR — must fix

**Raw SQL against genesis.db** — `scripts/cleanup_fenced_knowledge_units.py:100`

`_SELECT_SQL` is executed directly via `db.execute()` against the `knowledge_units` table (genesis.db). The convention requires going through `genesis.db.crud` modules; the only exemption is migration files. A one-off cleanup script does not qualify as a migration.

```python
# current — raw SQL, violates convention
cursor = await db.execute(_SELECT_SQL)
```

Suggested fix: add a targeted method to `src/genesis/db/crud/knowledge.py` — e.g. `async def select_fenced_intake_rows(db)` — that owns this query, then call it here instead of `db.execute(_SELECT_SQL)` directly. The `LIKE '%\`\`\`json%'` filter is unusual enough to deserve a named CRUD method with a docstring explaining the cleanup context.

---

### NOTE

**Private name imported across module boundary** — `src/genesis/surplus/executor.py:18`

```python
from genesis.surplus.intake import _FENCE_RE
```

The underscore prefix on `_FENCE_RE` signals module-private; importing it from a sibling module looks like leaking internals even though the PR description explains the single-sourcing intention. Rename to `FENCE_RE` (public constant) in `intake.py` — the intent becomes self-documenting and the cross-module import stops raising eyebrows. `_MULTI_FINDING_TASK_TYPES` imported in `test_intake.py` has the same character; worth making it public there too.

---

## PII / Secrets / Wording: CLEAN

No PII, hardcoded credentials, private IPs, SSH keys, or sensitive wording found in the diff or PR metadata.

---

WingedGuardian:
Structural review findings addressed in 6d584f21 (the bot has not re-reviewed since):

- **ERROR (raw SQL)**: the selection query moved into `genesis.db.crud.knowledge.select_fenced_intake_rows()` with a docstring covering the cleanup context; the script now calls the CRUD method — no raw SQL remains in the script.
- **NOTE (private-name imports)**: `_FENCE_RE` → `FENCE_RE` and `_MULTI_FINDING_TASK_TYPES` → `MULTI_FINDING_TASK_TYPES` — public constants, matching their deliberate cross-module use (executor, tests, cleanup script).

128 targeted tests green post-fix (test_intake / test_executor / test_scheduler / test_knowledge_crud). Merging with review-override on the stale ERROR comment; this comment documents the fix mapping.

WingedGuardian:
Merging after rebase onto current main (squash). The earlier structural-review **ERROR** (raw SQL against genesis.db in `scripts/cleanup_fenced_knowledge_units.py`) is **resolved on this branch**: the selection now goes through a crud-owned `knowledge.select_fenced_intake_rows()` method (commit "review: crud-owned selection + public FENCE_RE/MULTI_FINDING_TASK_TYPES") — the script carries zero raw `_SELECT_SQL`. The Codex reviewer remained quota-limited, so no fresh automated review re-fired on the rebased head; CI is green (11/11). Merging with `# review-override` because the flagged finding is fixed, not merely accepted.

PR #892, opened 2026-07-04
feat(memory): dream-cycle weekly-cluster / daily-drain split (shadow)

WingedGuardian:
## What

T2-D PR1: splits the dream cycle's two workloads onto their own cadences, SHADOW-first.

- **Weekly `run()` (Sunday 4am)** keeps the expensive full-collection scan + clustering + the additive link/centrality layer, and now persists a **value-ranked worklist** (size-DESC, top-500, oversize `>MAX_CLUSTER_SIZE` excluded and flagged for manual review) into the existing `deferred_work_queue` (`work_type="dream_synthesis_slice"`, `DRAIN` policy — the recovery expiry cadence never touches it). A fresh weekly run supersedes the prior list via a new `supersede`/`delete_by_work_type` queue primitive that preserves in-flight `processing` rows. No migration.
- **New daily drain (8am)** `run_dream_synthesis_drain` merges a bounded, value-ranked slice (`DAILY_SYNTHESIS_BUDGET=100` clusters/day) — spreading synthesis load across the week instead of a single ~1560-call Sunday spike, and deferring on bad days. **SHADOW hardwired in this PR**: full queue + rehydration lifecycle, zero LLM calls, zero memory/Qdrant mutations, reports what it *would* merge. The live flip (honoring `GENESIS_DREAM_CYCLE_LIVE`) is a separate, gated PR2.

## Why

Consolidation has been dry-run since 6-21; the backlog grows ~7-10%/week (1058 clusters at 6-28). The blocker to re-enabling was the weekly all-at-once spike (the 6-14 provider-outage grind: 838 clusters → 1 merge, 556 errors). Value-ranked + budget-bounded means the highest-value merges happen first and a slice can't exhaust providers; the low-value tail is dropped and re-ranked by the next weekly re-cluster.

## Hardening (from a pre-build failure-mode review)

- Oversize clusters excluded at enqueue — size-DESC ranking would otherwise put unmergeable clusters at the TOP of the worklist, wasting budget weekly and making shadow reports unfaithful to live behavior.
- Qdrant preflight + mid-drain infra reset — an outage can no longer mass-discard the week's top slices as "stale"; the drain aborts without consuming items.
- Rehydration is one batch `retrieve` per cluster, run via `asyncio.to_thread` (no event-loop blocking), and raises on infra errors instead of swallowing them.
- Stale slices (<2 live members) complete as no-ops instead of discarding — discarded rows surface as failures on the dashboard errors view.
- `mark_processing` return checked (supersede race guard); heavy_workload flag clears are owner-guarded (two jobs now share it); dead `aborted_capacity` branch removed from `run()` (capacity signal moved to the drain; equivalent entity-resolution protection is a PR2 item).
- Docstring drift fixed: the merge gate is the `GENESIS_DREAM_CYCLE_LIVE` env var — there is no config-file key. Weekly observation now reports `worklist_enqueued`/`oversize_flagged` instead of a permanently-zero merged count.

## Verification

- 86 targeted tests green (`tests/test_memory/test_dream_cycle.py`, `tests/test_resilience/test_deferred_work.py`, `tests/test_surplus/test_scheduler.py`) + routing-config guards.
- **E2E shadow against a consistent copy of the production DB + the real Qdrant instance**: 600 clusters → 500 enqueued (unfiltered pending stays ≤ the 1000 overflow threshold); supersede purged all 500; value-ordered drain; stale slice completed-not-discarded; before/after fingerprints of `memory_metadata` and Qdrant payloads identical; router/store stubs that raise on any call never fired; empty-worklist drain is a clean no-op.
- Code-reviewer pass: no blocking findings; both should-fixes applied.

## Expected shadow behavior (for the watch window)

Weekly obs: "N clusters found, K enqueued for daily drain". Daily obs: "N drained, M would merge". The worklist empties by ~day 5 (500 cap / 100 budget) — `drained=0` late-week is normal, not a failure. No `dream_synthesis_slice` rows should appear on the dashboard errors view.

No CHANGELOG entry: no user-visible effect until the PR2 live flip.

### DIFF
--- a/src/genesis/db/crud/deferred_work.py
+++ b/src/genesis/db/crud/deferred_work.py
@@ -32,6 +32,35 @@ async def create(
     return id
 
 
+async def delete_by_work_type(
+    db: aiosqlite.Connection,
+    *,
+    work_type: str,
+    exclude_status: str | None = "processing",
+) -> int:
+    """Delete all rows for a work_type, optionally preserving one status.
+
+    Used to supersede an ephemeral batch worklist (e.g. dream-cycle re-clustering
+    replaces last week's synthesis worklist and its completed/discarded residue).
+    ``exclude_status`` defaults to ``'processing'`` so an in-flight drain item is
+    never yanked out from under the worker. Returns the count deleted. This is an
+    explicit synchronous supersede — NOT a staleness policy — so it never races
+    the recovery ``expire_by_policy`` cadence (which only touches refresh/discard).
+    """
+    if exclude_status is not None:
+        cursor = await db.execute(
+            "DELETE FROM deferred_work_queue WHERE work_type = ? AND status != ?",
+            (work_type, exclude_status),
+        )
+    else:
+        cursor = await db.execute(
+            "DELETE FROM deferred_work_queue WHERE work_type = ?",
+            (work_type,),
+        )
+    await db.commit()
+    return cursor.rowcount if cursor.rowcount and cursor.rowcount > 0 else 0
+
+
 async def query_pending(
     db: aiosqlite.Connection,
     *,
--- a/src/genesis/memory/dream_cycle.py
+++ b/src/genesis/memory/dream_cycle.py
@@ -911,6 +911,11 @@ async def _synthesize_clusters(
     into a saturated chain. Genuine quality blocks reset the streak and never
     trip the breaker.
     """
+    # Deliberate re-sort: the worklist was value-ranked at enqueue, but
+    # rehydration can shrink clusters (members deprecated since Sunday) — the
+    # POST-rehydration size is the fresher value signal, so attempt order may
+    # diverge from queue order for shrunk clusters. PR2's per-item lifecycle
+    # supersedes this batch call.
     all_clusters.sort(key=len, reverse=True)
     merged = 0
     breaker = _CapacityBreaker(_CAPACITY_ABORT_THRESHOLD)
--- a/src/genesis/memory/dream_cycle.py
+++ b/src/genesis/memory/dream_cycle.py
@@ -183,19 +183,17 @@ async def run(
     report dict with cluster counts, worklist-enqueue count, and any errors.
     """
     run_id = str(uuid.uuid4())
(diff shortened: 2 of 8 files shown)


WingedGuardian:
## First-pass review: PR #892

**Structural review:** 1 WARNING, 0 ERRORs

---

### WARNING — Missing test coverage for scheduler wrapper

**`src/genesis/surplus/scheduler.py` — `run_dream_synthesis_drain` (new method, ~120 lines)**

`tests/test_surplus/test_scheduler.py` is not updated in this PR. The unit tests in `test_dream_cycle.py` cover the core drain logic (`run_synthesis_drain`) well, but the scheduler wrapper's distinct logic paths are untested:

- Pause check early return
- `heavy_workload` active → skip path
- Missing runtime dependency validation (db/store/router/qdrant `None` checks)
- `record_job_start` / `record_job_success` / `record_job_failure` wiring
- Observation write path (`obs_crud.create`)
- `finally` block: owner-guarded `heavy_workload` clear (`if rt._heavy_workload == "dream_synthesis_drain"`)

The `run_dream_cycle` wrapper has the same gap, but that's pre-existing. This is new code in the same mold. Not a hard block (shadow mode limits blast radius), but worth tracking before the live flip — especially the `heavy_workload` owner-guard, which is the mechanism preventing `run_dream_cycle`'s finally from clearing the drain's flag and vice versa.

---

No other structural findings. The core logic is well-tested. The `delete_by_work_type` / `supersede` pair, the FM-1 through FM-10 hardening, and the `CronTrigger` cadence (correct vs. `IntervalTrigger` for a daily job) all look right.

---

WingedGuardian:
Structural review WARNING addressed in 5d1336be: added 6 wrapper tests for `run_dream_synthesis_drain` covering the pause skip, the heavy_workload overlap skip (other job's flag untouched), dependency-checks-before-`record_job_start`, the happy-path job_health + observation write, failure recording, and the owner-guarded `finally` (a flag taken by the weekly job mid-run is not clobbered). The pre-existing `run_dream_cycle` wrapper gap is noted for the PR2 live-flip work.

PR #894, opened 2026-07-04
fix(autonomy): persist watchdog reclaim cooldown; mark dark verifiers as groundwork

WingedGuardian:
## Summary

Two small autonomy/watchdog correctness fixes.

### The watchdog's memory-reclaim cooldown now survives its own restarts

When container memory runs high, the watchdog reclaims a small, capped amount of page cache — throttled to at most once every 5 minutes to avoid an I/O spiral (the throttle exists because rapid repeated reclaims once caused an I/O death-spiral). That throttle lived only in memory, but the watchdog runs as a **systemd oneshot** — a fresh process on every timer fire — so the in-memory timestamp reset to zero each run and the cooldown never actually held. Under sustained memory pressure, reclaims could fire back-to-back.

The last-reclaim time is now persisted to a small sidecar file, using wall-clock time so it's comparable across the watchdog's separate processes. A future/backward-jumped timestamp (e.g. pre-NTP clock skew at boot) is treated as expired so the cooldown self-heals in one cycle rather than wedging reclaim off. Reads fail open (a missing or corrupt sidecar means "never reclaimed").

### Dark verifiers are marked as intentional groundwork, not dead code

Three verifier components have no runtime caller yet. Each is now tagged so it reads as deliberate future investment rather than dead code, with the reason it's parked:

- **TraceVerifier** — needs decision traces to be persisted first (a separate, parked pipeline).
- **TaskVerifier** — constructed at startup but its structural check isn't invoked on any path yet (the executor uses the LLM reviewer instead).
- **verify_chain** — cannot be wired as-is: the audit hash chain is a single chain written by two components concurrently, so expected concurrent-write forks would look like tampering to a linear walk. Real tamper detection needs a fork-tolerant check. The tag documents this so no one wires it naively.

Comment-only; no behavior change.

## Testing

- New tests cover the cross-process cooldown (persisted timestamp blocks a reclaim in a fresh process), timestamp-on-success, fail-open on a corrupt sidecar, and the future-timestamp self-heal.
- `tests/test_autonomy/test_watchdog.py` (51) and the verifier/integrity suites (61) pass; `ruff` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -163,6 +163,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **The watchdog's memory-reclaim cooldown now survives its own restarts, so it can't trigger an
+  I/O storm under sustained memory pressure.** When container memory runs high, the watchdog reclaims
+  a small, capped amount of page cache — throttled to at most once every 5 minutes to avoid an I/O
+  spiral. That throttle lived only in memory, but the watchdog runs as a fresh short-lived process on
+  each check, so it reset every run and never actually held. The cooldown is now persisted, so
+  repeated reclaims stay correctly spaced even across the watchdog's process boundary.
+
 - **Skill suggestions now actually fire — and nested skill packs show up in the catalog.** The
   prompt-time skill nudge scored matches against the length of your prompt, so on any wordy prompt a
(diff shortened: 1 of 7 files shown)


WingedGuardian:
Merging on green CI (all 10 checks pass). Automated Codex review was quota-limited; this branch went through a local multi-agent review before submission, which found and fixed one issue (a future-timestamp cooldown wedge).

PR #895, opened 2026-07-04
fix(guardian): wire the healthy-snapshot lifeline + recover stopped containers

WingedGuardian:
## What

Four structural gaps in the Guardian's offline recovery layer, found by tracing recent outages back through the code:

1. **The healthy-snapshot lifeline was never created.** `mark_healthy()` had zero callers, so `SNAPSHOT_ROLLBACK` — the only recovery action that works with the network down — could never find a target and always failed. The Guardian now takes a daily `-healthy` snapshot while all health checks pass, rotated create-then-delete so exactly one exists (CoW divergence bounded to one day; kill switch `snapshots.healthy_enabled`).

2. **The snapshot safety gate measured the wrong filesystem on LVM.** `safe_to_snapshot` ran `df` against a mountpath that lives on the host rootfs, blind to real thin-pool allocation. It now gates on `lvs` data%/metadata% at the storage-pool high tiers (reusing `pool.py`'s measurement); non-LVM backends keep the df byte-headroom logic, where df is correct.

3. **`take()`'s delete-before-create could evict the lifeline.** Eviction was purely count-based, so at default retention a pre-recovery snapshot would delete the healthy restore point right before the risky action that might need it. The latest healthy snapshot is now exempt (mirroring `prune()`'s exemption); retention default 1→2 so the lifeline and the newest pre-recovery coexist.

4. **A stopped container was unrecoverable.** `incus restart` errors on a stopped instance — exactly what an unclean host reboot leaves behind. `RESTART_CONTAINER` now falls back to `incus start`, and `host-setup.sh` sets `boot.autostart` on the container (idempotently, so re-running it upgrades existing installs).

Also: snapshot rollback deletes newer guardian-created snapshots and retries once when the storage driver refuses non-latest restores (documented ZFS behavior), and the snapshot-delete path is unified in one `delete()` helper with a 120s timeout (deleting a long-diverged thin snapshot does real kernel metadata work).

## Testing

- TDD: 15 new tests across `test_snapshots.py`, `test_check.py`, `test_recovery.py` (LVM-tier gating + df fallback, eviction exemption, rotation ordering incl. create-failure, healthy-only-when-HEALTHY wiring, daily-marker throttle, start-fallback, rollback retry)
- Full guardian suite: 449 passed; ruff clean

## Notes

- The healthy snapshot is only taken when the Guardian's state machine reports HEALTHY *this tick* — a broken container is never labeled healthy.
- A failed healthy-take logs loudly but does not block the daily maintenance marker (no retry storms against a failing incus).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,19 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **The Guardian now maintains a rolling "healthy" snapshot of the container — the restore point its
+  snapshot-rollback recovery always needed.** Once a day, while every health check passes, the Guardian
+  takes a `-healthy` container snapshot and deletes the previous one (exactly one exists, never more
+  than a day old, so copy-on-write growth can't quietly fill the storage pool the way stale snapshots
+  once did). Snapshot rollback — the one recovery action that works even with the network down — now
+  has a real target; before this, nothing ever created a healthy snapshot, so rollback could never
+  succeed. Kill switch: `snapshots.healthy_enabled: false` in `guardian.yaml`.
+
+- **The Genesis container now comes back on its own after a host reboot — including an unclean one.**
+  `host-setup.sh` sets incus `boot.autostart` on the container (applied to existing installs on
+  re-run, not just fresh ones). Previously an unclean host reboot could leave the container stopped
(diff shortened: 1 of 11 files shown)


WingedGuardian:
Merging on CI (10/10 green, full suite pass) — Codex review quota-limited; structural review clean; local high-effort review applied (2 findings fixed in 85e23cde-equivalent commit).

PR #897, opened 2026-07-04
chore(cc): bump pinned Claude Code 2.1.198 → 2.1.201

WingedGuardian:
## Summary

Moves the single-source-of-truth Claude Code pin (`scripts/lib/cc_version.sh`) from 2.1.198 to 2.1.201 (current npm `latest`, published 2026-07-03). One `scripts/update.sh` run after merge aligns the local install via `cc_ensure_local` and syncs a configured host VM via the guardian `update-cc` op.

## Changelog evaluation (2.1.199–2.1.201)

Bugfix-dominated delta; full evaluation recorded in `docs/reference/cc-compatibility.md` → Version History. Highlights relevant to Genesis installs:

- **Subagent reliability**: subagents cut off by rate limits or server errors now return their partial work or fail cleanly, instead of reporting empty output as success (@2.1.199/@2.1.200).
- **Background-agent daemon fixes**: Linux kill-loop after unclean shutdown (@2.1.199), stale `daemon.lock` with an OS-reused PID blocking all starts (@2.1.200), silent mid-turn stops after sleep/wake (@2.1.200).
- **`AskUserQuestion` no longer auto-continues** on an idle timeout by default (@2.1.200).
- `SessionStart`/`Setup`/`SubagentStart` hooks no longer swallow stderr on exit 2 (@2.1.199); project plugins now load in git worktrees (@2.1.200); tmux 3.4+ flicker fixed (@2.1.200).

## Compatibility checks

- Node floor unchanged (`>=22` across the range) — `NODE_MAJOR` stays 22, no lockstep bump.
- Grep-verified: no Genesis code depends on `AskUserQuestion` auto-continue; no `"default"` permission-mode pinning (renamed to "Manual" upstream, old value still accepted); `disabledMcpServers` config already a proper array.
- Behavioral note: @2.1.201 changes how Sonnet 5 sessions receive harness reminders (no mid-conversation system role) — flagged in the compatibility doc for post-deploy observation.

## Test plan

- [x] `source scripts/lib/cc_version.sh` yields `CC_VERSION=2.1.201`, `NODE_MAJOR=22`; inherited `CC_VERSION` override still honored
- [x] `bash -n` / shellcheck clean on touched scripts
- [ ] CI
- [ ] Post-merge: single `scripts/update.sh` run aligns container + host to 2.1.201 (first end-to-end exercise of the one-pin sync pipeline)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #898, opened 2026-07-05
fix(update): de-track settings.local.json, heal pin drift on no-op runs, codify host-deploy gate

WingedGuardian:
## Summary

Three pipeline gaps found during the first single-command `update.sh` end-to-end run (deploying the CC 2.1.201 pin, #897):

1. **`.claude/settings.local.json` re-untracked.** Install-local by definition and already `.gitignored`, it was accidentally re-tracked in #792 — making every live install permanently dirty and blocking `update.sh` at the clean-tree gate. This PR de-tracks it with a safe transition: back up the live copy pre-merge, let the upstream deletion merge cleanly, restore the copy (untracked) at the merge join point. Covers fresh-merge, `--post-merge`, and nothing-to-do paths; the live copy is never lost.

2. **Pin-drift healing no longer depends on pull delta.** The deploy-target sync stages (guardian redeploy, host Node/CC alignment, container CC alignment) only ran when the current invocation's merge brought commits — a manually pulled pin bump exited with "Nothing to do." while drift persisted. The sync region is extracted into `_sync_deploy_targets()` (verbatim move) and now also runs on the no-delta path.

3. **Host-Deploy Gate codified** in the genesis-development skill: merged ≠ deployed for guardian/host-touching paths — run `update.sh` + verify in the same session or create an explicit follow-up; host VMs are deploy targets, never edit-in-place dev environments.

## Test plan

- [x] New `tests/test_scripts/test_update_settings_local_transition.py` (5 tests) — extracts the actual marked script blocks and runs them against a scratch git repo: backup+clear on tracked-modified, clean merge of upstream deletion + restore untracked, nothing-to-do restore, untracked steady-state no-op, function-defined-before-call-sites invariant
- [x] `bash -n` + shellcheck (no new findings; 3 pre-existing warnings untouched)
- [ ] CI
- [ ] Post-merge live check: next `update.sh` run on an install with a locally modified settings.local.json completes without manual stashing


WingedGuardian:
No automated review posted (Codex quota-limited) — merging on green CI (all 10 checks pass) per the pre-merge gate. Post-merge: running the hardened update.sh, which regression-tests its own settings.local.json + drift-heal fixes.

PR #899, opened 2026-07-05
feat(guardian): self-heal the host SSH key — reharden-key verb + autonomous reconciler

WingedGuardian:
## What

Genesis now re-hardens its own host Guardian SSH key automatically — no human on the host required. This is Phase 1 of the Guardian self-maintenance initiative (escalation-inversion: Genesis acts via its own control plane; the human is the last resort).

The Guardian's control key should carry `no-pty` + a source-IP `from=` restriction, but before this those were applied **only at install time** — so an install set up before the hardening (or one whose container address changed) kept a weaker key until someone re-ran the installer by hand.

## Changes

- **`reharden-key` gateway verb** (`guardian-gateway.sh`) — zero-input; rewrites the guardian `authorized_keys` line to the canonical hardened options with a **self-proving `from=`** derived from the connection's own `$SSH_CONNECTION` source (the container always re-dials the same host, so the next connection presents the same source). Key material is read from the file; options are hardcoded. **Un-brickable:** snapshot + a 120s systemd dead-man's-switch restores the previous file unless a fresh connection confirms the rewrite works; `ssh-keygen` validates the rebuilt line before an atomic `mv`; exactly-one-guardian-line and line-count invariants guard the write.
- **`version` verb** now reports `authkey_no_pty`, `authkey_has_from`, `authkey_from_matches`, `authkey_opts_hash`, and `authkey_observed_src_hash` — **booleans + sha256 hashes only**, never the raw key blob, the `from=` literal, or the source address.
- **`_check_authkey_hardening` reconciler** (`watchdog.py`) — runs each watchdog tick beside the gateway-staleness check. A hardening **regression** (no-pty stripped, or from= missing with an observable source) heals on the first tick; a **from= mismatch** heals only after a confirmed **stable-source streak** (same observed-source hash for `DRIFT_ALERT_THRESHOLD` consecutive ticks); a **flapping** source is never chased — escalated once — and the guard is non-sticky so a later-stable source self-recovers.
- **`GuardianRemote.reharden_key()`** — when the line changed, confirms with a second fresh connection (proves the rewrite still authenticates **and** cancels the host dead-man's-switch); a failed confirm reports `restore_pending` rather than false success. Plus an `AddressFamily=inet` pin for v4/hostname targets so a dual-stack resolution can't flip the source the host's `from=` matches (skipped for v6 literals).
- **`install_guardian.sh --reharden-key-only`** — Step 11 factored into `setup_bidirectional_ssh()` + `derive_ts_ip()` (behavior-identical normal path); the flag re-hardens the key without a full re-install, as the manual out-of-band fallback.
- **Divergence guardrail test** — asserts the canonical `GUARD_BASE_OPTS` is byte-identical between `guardian-gateway.sh` and `install_guardian.sh` (the duplication is deliberate — the gateway must stay hermetic — so it's mechanically enforced).

## Security posture (SSH trust boundary)

The verb is **theft-neutral** (to reach it a caller's source must already satisfy the current `from=`, or `from=` is absent = a pre-existing weakness this closes), **strictly less powerful** than the existing `sudo`-running `update`/`update-cc`/`update-node` verbs, takes **zero caller input**, and is **fail-safe on every path** — no armed switch → no write; every error leaves the file untouched or the switch restores it. It can only make the key *more* restrictive.

## Tests

TDD RED→GREEN. 34 new tests (483 guardian tests total, all passing):
- `test_gateway_reharden.py` runs the **real** gateway script as subprocess E2E with real ed25519 keys + real file rewrites (self-proving from=, other-lines-untouched, idempotency, dead-man's-switch arm/cancel, refuses on 0/>1 guardian lines, refuses when the switch can't arm, no-from fallback) + version authkey fields + the divergence guardrail.
- `test_watchdog.py::TestAuthkeyHardening` — the full reconciler state machine (regression heal-once-then-escalate, stable-move heal at the streak, flap never-heals-escalates-once, flap-then-stable recovery, resolved reset + re-arm).
- `test_remote.py` — the confirm flow + AddressFamily pin.

ruff + shellcheck clean; private-data scan clean (fixtures use synthetic `10.0.0.42`).

## Deploy (Genesis-driven, no manual host access)

On merge, the container's `update` gateway verb self-installs the new gateway on its next call, and the reconciler heals the live (pre-hardening) key automatically — the initiative's thesis proof.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,18 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **Genesis now re-hardens its own host SSH key automatically — no human on the host required.** The
+  Guardian's control key (the one the container uses to manage the host) is supposed to carry `no-pty`
+  and a source-IP `from=` restriction, but before this those were applied only at install time, so an
+  install set up before the hardening — or one whose container address changed — kept a weaker key
+  until someone re-ran the installer by hand. The watchdog now checks the key's hardening every cycle
+  and, when it has drifted, calls a new `reharden-key` gateway operation that rewrites the key to the
+  hardened form and proves the rewrite still works before committing to it (a built-in 120-second
+  dead-man's-switch restores the previous key otherwise, so it can never lock Genesis out of its own
+  host). A moved-but-stable source address heals on its own; a flapping address is left alone and
+  reported instead of chased. For the rare manual case, `install_guardian.sh --reharden-key-only`
+  re-hardens the key without a full re-install.
+
 - **The Guardian now maintains a rolling "healthy" snapshot of the container — the restore point its
   snapshot-rollback recovery always needed.** Once a day, while every health check passes, the Guardian
   takes a `-healthy` container snapshot and deletes the previous one (exactly one exists, never more
--- a/scripts/guardian-gateway.sh
+++ b/scripts/guardian-gateway.sh
@@ -625,7 +625,12 @@ PYEOF
         OLD_N=$(grep -c '' "$AK") || OLD_N=0
         NEW_N=$(grep -c '' "$AK_TMP") || NEW_N=0
         if [ "$OLD_N" -ne "$NEW_N" ]; then
+            # Aborting WITHOUT writing → the file is unchanged, so disarm the
+            # switch we armed above (keep "armed iff we wrote" symmetry). The
+            # restore would be a harmless self-copy either way, but leaving a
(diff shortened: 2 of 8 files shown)


WingedGuardian:
Merging on CI (all 10 jobs green, 0 open CodeQL alerts). Codex code review was quota-limited this window, so no bot review findings — merging on CI per the pre-merge gate. Self-reviewed inline (SSH trust-boundary analysis + the two high-risk surfaces); the one CodeQL high was a false positive (name-heuristic on an integer tick-counter), resolved in 7f1be055.

PR #900, opened 2026-07-05
feat(ci): enforce CC↔Node pin lockstep at PR time

WingedGuardian:
## Summary

Makes it mechanically impossible to bump the Claude Code pin past its Node floor without also bumping `NODE_MAJOR`. A host VM recently ran Node 20 while its pinned CC required Node >=22, so `claude -p` — the Guardian's recovery brain — could not start. `update.sh` already *heals* deployed drift; this closes the remaining **authoring-time** gap, where the "bump NODE_MAJOR in lockstep" instruction was only a comment in `scripts/lib/cc_version.sh`.

## What it does

`scripts/check_cc_node_lockstep.py` reads `CC_VERSION` + `NODE_MAJOR` from the pin file, fetches the pinned CC version's `engines.node` from the npm registry (stdlib `urllib`, no dependencies), and asserts `NODE_MAJOR` satisfies the required Node major. Wired as a new `cc-node-lockstep` CI job mirroring `external-io-check`.

**Error policy (deliberate):**
- **Fail closed** (block the PR): `NODE_MAJOR` below the floor; an unparseable pin file; or an **HTTP 404** for the pinned CC version (a version that doesn't exist is a typo'd pin, not weather).
- **Fail open** (warn + pass): transient network errors (one retry first) or an `engines.node` comparator shape the parser doesn't recognize. A flaky npm registry must never wall off every PR, and a parser gap must never false-fail a valid spec.

## Test plan

- [x] 21 offline tests (`tests/test_scripts/test_cc_node_lockstep.py`) via an injected registry opener: lockstep pass, below-floor fail, 404 fail-closed, 5xx/URLError fail-open, missing/exotic `engines.node` fail-open, comparator-shape parser units, real-file parse smoke
- [x] Live: passes at the current `22 >= 22` lockstep; fails closed (exit 1) under a sabotaged `NODE_MAJOR=20`
- [x] `ruff` clean; `ci.yml` YAML validated
- [ ] CI (the new job runs against this PR — proves the wiring E2E)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -237,6 +237,22 @@ jobs:
       - name: External-I/O egress guard
         run: python scripts/check_external_io.py
 
+  cc-node-lockstep:
+    # Enforce the CC↔Node pin lockstep: NODE_MAJOR in scripts/lib/cc_version.sh
+    # must satisfy the pinned Claude Code's engines.node floor. Prevents the
+    # "bumped CC past a Node floor, forgot NODE_MAJOR" mistake that left a host
+    # unable to run `claude -p`. Fails closed on a real mismatch / typo'd pin;
+    # fails open (warn) on a transient npm-registry error. See
+    # scripts/check_cc_node_lockstep.py.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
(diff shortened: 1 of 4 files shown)


WingedGuardian:
No automated review posted (Codex quota-limited) — merging on green CI (all 11 checks pass, including the new cc-node-lockstep job proving its own wiring) per the pre-merge gate.

PR #881, opened 2026-07-03
fix(dashboard): clearer per-session memory row — "Claude Code Sessions", green when healthy

WingedGuardian:
The infrastructure health card's per-session RSS row (added in #855) was labeled **"CC"** with **gray** `cc-N` chips. Gray reads as inactive/unknown, and "CC" is opaque — a user asked what it even was and why it looked gray.

**Change (display-only; `cc_slots` data + thresholds unchanged):**
- Rename the row **"CC" → "Claude Code Sessions"** (stacked label so the full name fits the compact card).
- Color **healthy sessions green** (`#4caf50`, matching the CPU/Disk rows) — amber ≥ 4 GB, red ≥ 6 GB unchanged.
- Chips **`cc-N` → `#N`**, plus hover tooltips (row purpose on the label; per-session pid + GB on each chip).

Verified the template still parses (jinja2). One file, +8/−6.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #903, opened 2026-07-05
fix(dashboard): infra-card failed-first ordering, drop raw cc_slots row, emit probe transitions

WingedGuardian:
## What & why

Three user-facing Infrastructure-card bugs on the dashboard overview:

- **B1 — raw `cc_slots` row leak.** `cc_slots` is an array with its own dedicated "Claude Code Sessions" section (#881), but the probe grid's exclusion filter omitted it, so it also rendered as a raw unlabeled `cc_slots` row. Added to the grid `x-show` exclusion + a defensive `infraLabel` entry.

- **B2 — failed-probe-first ordering.** Probes rendered in dict-insertion order, burying a down probe below healthy ones. New `sortedInfrastructure` getter orders worst-first (down/error → degraded → unknown → healthy), deriving severity from `probeStatusColor` so the order always matches the status dots the operator sees. Stable within-rank (backend order preserved); rebuilds an object (infra keys are non-integer strings → insertion order preserved) so the existing template body is untouched.

- **B3 — probe-transition events.** New `observability/probe_transitions.py` `ProbeTransitionTracker` detects healthy↔unhealthy boundary crossings and `HealthDataService` emits a `probe_transition` event to the activity feed per crossing. Isolated by construction from `resilience.state` (the routing-driving state machine) — probe transitions are pure observability and never influence routing. Includes a 90s startup grace (swallow the restart transient) and a 15min/3-crossing flap window (dampen a bouncing probe). Single emitter: `HealthDataService` is a singleton in the server process; the standalone health MCP calls `infrastructure()` directly and never emits.

- **Drive-by:** bounded the previously-unbounded append-only `ResilienceState.transitions` log to `deque(maxlen=200)`. It has zero routing readers; flap detection uses a separate per-axis window, so the bound is behaviorally inert.

## Verification

- `ruff` clean; `node --check` on dashboard.js clean.
- Targeted tests green (81 passed): new `test_probe_transitions.py` (9) + `test_health_data_probe_emit.py` (8, incl. storm-guard/emit-failure isolation/severity), plus existing `test_types`, `test_health_psi_cpu`, `test_service_status`, `test_resilience/test_state`.
- B2 sort + object-order assumption verified in Node.
- Wiring confirmed: `runtime/init/health_data.py` passes the live `event_bus`, so emits fire in production.
- Frontend browser E2E and B3 activity-feed E2E run post-deploy (dashboard is served from the installed package, not a worktree).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -200,6 +200,15 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **The dashboard's Infrastructure card is clearer at a glance: failed probes sort to the top, and a
+  stray raw row is gone.** Probes now order worst-first (down/error, then degraded, then healthy) so
+  problems surface at the top of the card instead of wherever they happened to fall in insertion
+  order. The "Claude Code Sessions" data no longer also leaks out as a raw, unlabeled `cc_slots` row
+  — it appears only in its dedicated section. And when an infrastructure probe crosses the
+  healthy↔unhealthy line (e.g. Qdrant goes down, then recovers), that transition is now recorded in
+  the Activity feed — with startup-grace and flap dampening so a restart blip or a rapidly-bouncing
+  probe won't spam it.
+
 - **The watchdog's memory-reclaim cooldown now survives its own restarts, so it can't trigger an
   I/O storm under sustained memory pressure.** When container memory runs high, the watchdog reclaims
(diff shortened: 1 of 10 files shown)


WingedGuardian:
Codex review is quota-limited (no automated review this round). Lean single-agent review completed clean (verdict: SHIP) — verified storm-guard, emit dict-unpack, flap/warmup accounting, the deque-bound safety (zero routing readers; flap uses a separate window), and the B2 sort/object-order assumption. All 11 CI checks green. Merging on CI + lean review per the pre-merge gate.

PR #905, opened 2026-07-06
feat(sentinel): shadow reversibility classifier for proposed actions

WingedGuardian:
## What

The Sentinel now classifies every action it proposes as **would-run-autonomously** (`would_auto_run`) or **would-still-ask** (`would_gate`) and logs the verdict to `sentinel_log.jsonl` — **observe-only**. Nothing about execution changes: every proposed action still requires approval through the existing gates, exactly as before.

This is groundwork for a future opt-in autonomous tier for reversible, programmatic fixes (service restarts, journal vacuums, cache drops). The allowlist cannot be calibrated without real incident data, so this ships shadow-first: accumulate real would-decisions, then decide.

## How

- **New `sentinel/auto_eligibility.py`** — `classify_action()` is deliberately pessimistic:
  1. **Self-fatal deny first** (matches anywhere): any `systemctl` invocation naming `genesis-server` (the Sentinel runs inside that process), kill/pkill/killall in any form, recursive `rm` (short or `--recursive`).
  2. The diagnosis's own `safe` + `reversible` tags must both be true.
  3. The command must **full-match an anchored allowlist shape** (known-safe units only, tight unit charset) — commands execute via `create_subprocess_shell`, so an allowlisted prefix with a chained payload can never match.
  Anything else → `gated`. `find … -delete` is deliberately excluded: deletion is irreversible.
- **New `config/sentinel.yaml`** — `autonomy.mode: shadow` (default). `live` is reserved and **not implemented**: the dispatcher coerces it to shadow with a warning, so no config value changes behavior. Per-install overrides go in the gitignored `.local.yaml` overlay.
- **Dispatcher** — one exception-proofed hook at the phase-3 block logs the classifications; insertions-only diff, both approval paths untouched. A `GROUNDWORK(sentinel-live-autonomy)` marker records that a future live mode must reclassify at the approval-resume path.
- **SENTINEL.md** — tagging guidance: `safe`/`reversible` feed calibration; when unsure, tag false.

## Testing

- 170 sentinel tests pass (49 new): allowlist hits, self-fatal exclusions incl. flag-interposition bypasses, chained-payload rejections, malformed/non-dict actions, config-loader fallbacks, and dispatcher wiring (shadow logs while the gate is still consulted; pending approval executes nothing; a crashing classifier cannot break dispatch; zero-action dispatches emit no event).
- Unpatched end-to-end run: a real dispatch wrote the `shadow_classification` line to disk with correct verdicts while the dispatch parked awaiting approval.
- Design reviewed (architect pass, 4 corrections applied) + two independent code-review passes (6 findings fixed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,17 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **The Sentinel now learns which infrastructure fixes would be safe to run itself — observe-only.**
+  Every fix the Sentinel proposes still requires your approval, exactly as before. What's new: each
+  proposed action is additionally classified as "would run autonomously" (reversible, programmatic,
+  and matching a conservative allowlist of known-safe command shapes — service restarts, journal
+  vacuums, cache drops) or "would still ask you" (anything permanent, data-mutating, unlisted, or
+  that could take down the Sentinel's own host process), and the verdict is logged to
+  `sentinel_log.jsonl`. This shadow data calibrates a future opt-in autonomous tier against real
+  incidents before it is ever allowed to act. The mode lives in `config/sentinel.yaml`
+  (`autonomy.mode: shadow`); `live` is reserved and not yet implemented — nothing executes without
+  approval regardless of the setting.
+
 - **Genesis now re-hardens its own host SSH key automatically — no human on the host required.** The
   Guardian's control key (the one the container uses to manage the host) is supposed to carry `no-pty`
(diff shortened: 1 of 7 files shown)


WingedGuardian:
Codex review quota-limited (no automated review posted — see comment above). Substituting: architect design review pre-build (92%, 4 corrections applied) + two independent code-review finder passes post-build (6 findings fixed: flag-interposition deny bypass on genesis-server, `--recursive` rm, over-broad systemd-unit trust, ambiguous journal vacuum-time unit, missing zero-action regression test, deny-reason clarity). All 11 CI checks green.

PR #907, opened 2026-07-06
Backup scheduling via systemd timer + honest health readouts

WingedGuardian:
## Why

The backup script and docs claimed backups ran "every 6h via cron," but no installer ever created that schedule — an install could silently have **no scheduled backups at all**. This wires real scheduling, and fixes three health-readout papercuts found alongside it.

## What changed

**Backup scheduling**
- New `genesis-backup.{service,timer}` systemd user units (6h cadence, `Persistent=true`). Bootstrap installs the unit *files* via its existing template-sync loop but leaves the timer **disabled** — enabling stays a deliberate, verified opt-in (a schedule that quietly leaves your DB local-only is worse than none). Enable with `systemctl --user enable --now genesis-backup.timer`.
- The backup unit intentionally does **not** use `ProtectSystem=strict`/`ReadWritePaths` (unlike the watchdog/disk-hygiene units, which only touch `$HOME`): `backup.sh` drives `gpg-agent` over its socket in `/run/user` and writes a Tier-2 marker via `mktemp` in `/tmp`, both read-only under `strict`.
- Docs (`SETUP.md`, `README.md`, `CLAUDE.md`, `backup.sh` header, onboarding skill) updated from the stale "cron" framing to the timer, with the explicit enable step. The onboarding skill previously installed a 6h **crontab** entry — removed, so it can't double every backup alongside the timer.

**Health readouts**
- `bootstrap_manifest` MCP tool returned `{bootstrapped:false, manifest:{}}` whenever called out-of-process (i.e. always, from the CC-child MCP server). It now reads the manifest genesis-server persists verbatim to `~/.genesis/bootstrap_manifest.json` at bootstrap — exact fidelity, and reported as `source:persisted_manifest` with `bootstrapped:false` so a persisted file can never masquerade as current liveness (use `subsystem_heartbeats` for that).
- Ego-calibration `no_data` message now states its real precondition (ego-sourced ground-truth rows are required) instead of implying the Outcome Bus lacks tier-1 volume. Fixed in both places the message lived.

## Tests
New unit tests for the persisted-manifest writer/reader (round-trip fidelity, readonly-skip, missing/corrupt/malformed-file handling). `78 passed` across affected files, ruff clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -21,7 +21,8 @@ reflection) → Services (routing, memory, outreach, autonomy, surplus) → Data
 - **Qdrant**: `localhost:6333` (systemd service)
 - **GitHub**: configured in `~/.genesis/config/genesis.yaml` (`github.user` / `github.public_repo`)
 - **Database**: `~/genesis/data/genesis.db` (NOT `~/genesis/genesis.db`)
-- **Backups**: encrypted 6h cron via `scripts/backup.sh` → your private
+- **Backups**: encrypted, every 6h via `genesis-backup.timer` (systemd user
+  unit; enable deliberately after configuring) running `scripts/backup.sh` → your private
   `genesis-backups` repo (SQLite, Qdrant, memory, transcripts, config,
   secrets). Restore via `scripts/restore.sh` or `python -m genesis restore`.
 - **Env scrub**: `CLAUDE_CODE_SUBPROCESS_ENV_SCRUB=1` is NOT used — Genesis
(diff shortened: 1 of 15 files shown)


WingedGuardian:
Automated review was quota-limited (Codex usage limit reached — no findings posted). Coverage substituted by a local multi-angle review (correctness / cross-file / cleanup-altitude / conventions), whose findings are folded into commit `f0555a75`: de-sandboxed the backup unit (ProtectSystem=strict breaks gpg-agent + mktemp), replaced the lossy manifest reverse-mapping with a persisted raw manifest, removed the duplicate onboarding crontab, and swept the second stale calibration message. All 11 CI checks green.

PR #912, opened 2026-07-06
feat(attention): text-only clarity (has_audio) + device provenance (source) on attention events

WingedGuardian:
## Summary

PR-4 of the omnipresence track — the two pure-core attention-engine changes the OMI wearable connector (PR-5) builds on. Both are additive with defaults that keep every existing code path byte-identical.

### 1. Text-only clarity path (`has_audio`)

A text-only source (OMI transcripts) has no `rms` — today that pins `capture_clarity`'s loudness sub-score to 0 (a permanent −25% penalty) and makes `is_blip` drop short utterances as near-silence junk.

- `AmbientUtterance.has_audio: bool = True` (appended last), decided in ONE place: `row_to_utterance` sets `has_audio=bool(meta.audio)`. Verified against both live snapshots (5,423 rows each): zero rows lack `meta.audio`, so every existing home row keeps the audio path.
- `capture_clarity(has_audio=False)` averages the three text sub-scores (confidence/length/richness); `is_blip(has_audio=False)` treats only a token-less row as junk. Keyword-only args, default `True`; the audio path is regression-pinned to the exact 4-term math.
- Threaded at the 3 call sites (`engine.py` clarity + blip, `runner.py` pre-filter). The edge-vendored `garble_features.py` copy is deliberately untouched (backward-compatible default).

### 2. Device provenance (`source` on events)

`AmbientUtterance.source` was dropped at the event boundary — the store couldn't tell which device a perk came from.

- `AttentionEvent.source: str = ""` (appended last), populated from the trigger utterance in `evaluate()`.
- Persisted via `crud.attention.COLUMNS` + `ShadowStoreConsumer._to_row` (order-match preserved; machine-derived column, rides the label-preserving upsert — labeled rows stay frozen).
- **Migration 0046** (`attention_events.source`, PRAGMA-guarded idempotent ADD COLUMN, mirrors 0045) + the canonical DDL in `_tables.py`.
- Dashboard: `list_events` gains an exact-match `source` filter; a new `sources()` aggregate (mirrors `config_versions()`) feeds a device dropdown on the Judgment tab; each event row shows a source pill.

## Tests

TDD throughout (RED verified before each implementation). New coverage: 3-term clarity math + rms-invariance + 4-term regression pin; text-only blip rule; `row_to_utterance` discriminator (incl. corrupt/absent meta → text-only, fail-safe); engine/runner pass-through of short text-only utterances; `evaluate` sets `source`; consumer persists it; crud filter + `sources()`; migration 0046 idempotency/no-op; dashboard list threading + stats `sources`. All three COLUMNS-order test tuple builders updated. 274 passed across the affected surface (attention, crud, migrations runner, stripped-db schema, dashboard API).

## Verification still to run (post-review)

E2E over a live snapshot with `--persist`: confirm `source` lands on `attention_events`, the device filter scopes correctly, existing events reproduce identical clarity, and no transcript text reaches genesis.db.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,6 +103,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **Attention events now record which device they came from, and the Judgment tab can filter by it.**
+  Every "perk-up" event the passive-listening engine emits now carries the source device (the home
+  edge connection, or `omi` once the wearable connector lands), and the dashboard's Judgment review
+  gains a device dropdown plus a per-event source pill — so you can review one device's moments in
+  isolation. The engine also stops penalizing text-only sources: an utterance with no audio capture
+  (no loudness to measure) is scored on its text signals instead of being docked 25% clarity or
+  dropped as near-silence junk, which is what OMI wearable transcripts will need.
+
 - **The Sentinel now learns which infrastructure fixes would be safe to run itself — observe-only.**
   Every fix the Sentinel proposes still requires your approval, exactly as before. What's new: each
   proposed action is additionally classified as "would run autonomously" (reversible, programmatic,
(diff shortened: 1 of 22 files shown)


WingedGuardian:

---

PR #914, opened 2026-07-06
fix(inbox): fold parked files into the refresh batch (30-min approval nag)

WingedGuardian:
## Problem

Inbox approvals nagged via Telegram every 30 minutes (38 system-cancelled approval requests on 2026-07-06 alone; same storm pattern Jun 28-29). Mechanism, verified live: two files parked on one pending approval; the refresh path in `_phase_detect_changes` cancels the approval and invalidates ALL rows parked on it, but re-dispatches only the newly-detected file. The other file's invalidated rows are excluded from `get_all_known` by design, so it re-surfaces as phantom-new on the next scan, cancels the fresh approval in turn, and the two files leapfrog forever — a new approval request (= new Telegram message) every scan with no disk change.

No notification-policy change caused this: fire-once + 24h re-ask is intact; the churn creates genuinely-new approvals. Latent since the #323 refresh path; the crud docstrings' design intent ("fresh rows with fresh approvals" via re-detection) is what the leapfrog exploits — re-detection lands one scan too late.

## Fix

The refresh's invalidation loop now collects the file paths of the rows it invalidates and folds them (deduped against detected files, still-existing files only) into the returned `modified_files`, so ONE merged batch covers parked + new in the same cycle. Folded files flow through the identical modified-file machinery (including #810 URL-delta batching) they would have taken via re-detection one scan later — the fix changes when, not how.

## Tests

- Oscillator repro: two files parked on one approval, one edited → ONE cancel, merged re-park of BOTH files, and quiet scans (no cancel, no new request) on unchanged disk. Fails on unfixed code at the merged re-park assertion.
- Folded file with a prior completed baseline re-evaluates ONLY the unprocessed URL delta (never re-evals baselined URLs).
- Deleted parked file: end-to-end non-resurrection, plus an isolated test of the fold's `exists()` guard (mutation-verified: removing the guard fails the test).
- Stateful gate mock models the real stable-approval-key contract (one pending per site; cancel clears; next route mints a fresh id).

Review: inline code review clean, no blocking findings. One reviewer observation on record: the pre-existing `modified_files` loop in `_phase_create_records` lacks the `count_url_failures` retry-storm guard that the `new_files`/`retry_files` loops have; no realistic exploit path found (parked rows are pre-dispatch; URL-failure history is post-dispatch), so intentionally not changed here.

## Verification plan (post-deploy)

With files parked on a pending approval: first scan after new content → one refresh message covering the merged batch; subsequent unchanged-disk scans → journal shows "Inbox detection skipped — approval pending, no new files", no cancel, no new `approval_requests` rows. Regression markers: no "cancelling stale approval to refresh" on unchanged-disk scans; inbox-eval approval creations ≤1/day absent real drops; superseded rows stop accumulating.

### DIFF
--- a/src/genesis/inbox/monitor.py
+++ b/src/genesis/inbox/monitor.py
@@ -682,7 +682,15 @@ class InboxMonitor:
                     exc_info=True,
                 )
 
-            # Invalidate inbox_items parked on the cancelled approval.
+            # Invalidate inbox_items parked on the cancelled approval and
+            # fold their files into THIS refresh. Invalidated rows are
+            # excluded from get_all_known and from retry by design ("fresh
+            # rows with fresh approvals"), so a parked file that is not
+            # re-dispatched here re-surfaces as phantom-new next scan and
+            # cancels the fresh approval in turn — two parked files then
+            # leapfrog forever: a cancel + a new request (a Telegram
+            # message) every scan with no disk change.
+            parked_paths: list[str] = []
             try:
                 awaiting = await inbox_items.get_awaiting_approval(self._db)
                 for row in awaiting:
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

PR #913, opened 2026-07-06
feat(dashboard): voice Bridge tab cockpit — full live edge health

WingedGuardian:
## What

The voice dashboard's **Bridge tab** becomes a full cockpit. Previously it showed only status/latency/verdict — the rich edge health payload never reached the frontend (`probe_ambient_health` forwards only the verdict into the shared infrastructure snapshot).

- **`bridge_snapshot()`** (`observability/ambient_health.py`): composes the module's existing config-load / SSH-read / verdict pieces into one never-raising call; full `ambient_health.json` passthrough under a `health` sub-key so future edge keys surface without a Genesis-side change. TDD, 5 tests.
- **`GET /api/genesis/voice/bridge`** (`dashboard/routes/voice.py`): thin on-demand `_async_route` wrapper — deliberately NOT widening the shared infrastructure probe (avoids leaking ~25 keys into the generic health cards, and skips its two cache layers, wrong for a live cockpit). 20s timeout is strictly a backstop above the SSH read's ~15s internal bound. 2 wiring tests.
- **Bridge panel** (`genesis_voice.html`): verdict as a shared status chip + ages via `fmtAge` (reusing #893's `ui-utils.js` — its docstring names this page as an intended consumer), grouped sections: memory leak-watch (RSS parent/child/total, ORT-arena state, pool recycles), capture, diarization, connection stability, speaker ID, plus a forward-compat "Other" k/v for unknown health keys.
- CHANGELOG: this + the previously-missed #876 Device-tab entry.

## Verification

- TDD RED→GREEN; blast-radius sweep 58/58 (`test_ambient_health` + `test_voice_api` + `test_health`), ruff clean.
- **Full-stack browser E2E pre-PR** via a throwaway local server running this branch: panel renders live edge data end-to-end (chip styled, ages humanized, all groups populated, "Other" catches un-rendered keys); "not configured" degradation verified visually.
- Code review: 1 finding (`conn_dark_since` listed as shown but rendered nowhere) — fixed in `7f1ed78a`.
- Private-data scan of the full diff: clean (test fixtures use RFC 5737 TEST-NET only; edge host/user stay in install-local config).

## Notes

- The voice page's horizontal overflow at narrow viewports is **pre-existing** (reproduces on the untouched Device tab) — left for the Dashboard P1 restyle workstream.
- Deploy note: needs a genesis-server restart (new Python route); the template half is served from disk per-request, so FF + restart should happen together.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #915, opened 2026-07-06
feat(observability): RSS-threshold leak-regression alert for the ambient bridge

WingedGuardian:
## What

Closes the loop on the ambient-bridge memory-leak saga: now that the arena-off soak has **passed** (4.6 days: child RSS flat ~170 MB, parent slope ≈ 0, total plateau ~470 MB vs the old 1.67 GB pathology), `evaluate_ambient_health` gains a leak-class regression watch:

- **degraded** when `rss_total_mb` > 1000 MB or `rss_diar_child_mb` > 450 MB — ceilings ~2× the worst burst observed in the soak, so workload breathing can never fire them.
- Null/absent `rss_*` keys (lazy diar-pool spawn, older edge builds) are never a breach.
- A dead bridge stays **down**; the RSS reason is still appended for the operator.
- Flows to the existing transition-gated Telegram alert (one alert on entering the bad state, no nagging), the infrastructure card, and the voice Bridge tab's verdict chip (#913).

## Verification

- TDD RED→GREEN: 6 new tests (both ceilings, plateau-ok, null/absent no-op, non-numeric ignored, down-not-masked). Blast radius 57/57 (`test_ambient_health` + `test_health`), ruff clean.
- Downstream trace: `OutreachScheduler._ambient_health_job` alert is transition-gated and includes verdict reasons verbatim.
- Threshold calibration data: 4.6-day soak via the edge `rss_*` health keys (GENesis-Voice #27/#29).

Closes follow-up `6c11696b`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **Ambient bridge memory-leak regression alert.** If the edge bridge's RSS ever climbs past the
+  healthy plateau again (total > 1000 MB or diarization child > 450 MB), ambient health flips to
+  degraded and you get a one-time Telegram alert naming the breach — no nagging, and normal
+  workload bursts never trigger it. Requires an ambient edge reporting the `rss_*` health keys.
+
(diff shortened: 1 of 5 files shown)


WingedGuardian (commented):

---

PR #918, opened 2026-07-07
fix(memory): stop internal subsystem noise from leaking into recall

WingedGuardian:
## What this fixes

Genesis's memory recall was surfacing its **own internal machine-generated
noise** — reflection observations, task-executor retrospectives, and
ego-dispatch bookkeeping — alongside genuine user-relevant memories. This
created a subtle feedback-loop risk: the system reading back its own
decisional output as if it were external signal.

The `source_subsystem` infrastructure — a memory tag that routes
internal-subsystem writes to keyword-only storage and excludes them from
default semantic recall — shipped earlier, but **coverage and backfill were
never completed**: the machine write paths didn't set the tag, and legacy
rows were embedded in the vector store without it, so the recall-exclusion
filter never matched them.

## Changes

**Forward fix — tag the machine writers** so their output is keyword-only
and excluded from default recall (verified no recall consumer depends on
these tags):
- Perception awareness loop (the live producer) → `reflection`
- Autonomy task-executor traces + notepad promotion → `autonomy`
- Ego-dispatch outcome records → `ego`

**Recall robustness** — `only_subsystem` / `include_subsystem` now raise a
clear `ValueError` on an unknown subsystem name instead of silently
returning nothing (a typo like `only_subsystem="reflction"` was previously a
silent empty result).

**Backfill + cleanup for existing installs** (dry-run by default, `--apply`
to commit):
- `scripts/backfill_source_subsystem.py` — tags legacy rows by their
  recorded pipeline.
- `scripts/cleanup_subsystem_qdrant.py` — fixes a WHERE-clause bug that
  skipped some tagged subsystems, and adds a payload-based orphan sweep so
  vector-store points with no metadata row are also removed. Content
  survives in keyword storage; only the redundant vector copy is deleted.

**Module-vs-subsystem hardening (future-proof template).** A capability
*module* (an external pluggable capability under `src/genesis/modules/`) is
never a Genesis *subsystem*. A new AST coverage-guardrail test enforces this
mechanically: any `.store()` call under `modules/` that sets
`source_subsystem` is a hard CI failure, and every new memory-writer must
either tag itself or be explicitly classified as user-context. The taxonomy
is now documented at the definition sites (`modules/base.py`,
`memory/retrieval.py`, `memory/store.py`).

## Testing

- New AST coverage guardrail (`test_store_subsystem_coverage.py`) — proven
  RED both ways (an untagged writer and a tagged module each fail CI).
- Subsystem-name validation tests, backfill script tests, and cleanup
  orphan-sweep tests (including a regression test for the pure-orphan
  install case).
- Targeted suites pass; ruff clean.

## For existing installs

After updating, run the two scripts (dry-run first to review the counts,
then `--apply`) to tag and clean up legacy memory rows. The cleanup deletes
redundant vector-store points only — the content remains fully retrievable
via keyword recall.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,15 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Memory recall no longer surfaces Genesis's own internal noise.** Machine-generated
+  decisional output — background reflections, autonomy task retrospectives, and ego-dispatch
+  records — was leaking into normal recall because several writers didn't mark themselves as
+  internal-subsystem writes. Those writers are now tagged, so the content stays available to the
+  subsystem that produced it but no longer pollutes user-facing recall (which measurably improves
+  results on reflection-adjacent queries). Existing installs can purge already-embedded legacy
+  noise with the new `scripts/backfill_source_subsystem.py`, then
+  `scripts/cleanup_subsystem_qdrant.py` — both dry-run by default; add `--apply` to commit.
+
 - **A false "deferred work backlog" health warning no longer fires every few minutes.** The
   weekly memory dream-cycle parks a large synthesis worklist that drains a little each day by
   design; the health check was counting that scheduled batch against the "postponed due to
--- a/scripts/backfill_source_subsystem.py
+++ b/scripts/backfill_source_subsystem.py
@@ -1,0 +1,168 @@
+#!/usr/bin/env python3
+"""Backfill ``memory_metadata.source_subsystem`` for legacy machine writes.
+
+Background
+----------
+PRs #315/#317 added the ``source_subsystem`` column + default-exclude recall
+filter, but legacy machine-generated memories (written before the tagging
+code landed) still have ``source_subsystem = NULL``. They remain embedded in
+Qdrant and leak into default semantic recall.
+
+This script attributes each legacy row from its Qdrant ``source_pipeline``
+payload and backfills the SQLite tag so default recall (and the FTS5 path)
+excludes it. It pairs with ``cleanup_subsystem_qdrant.py``, which deletes the
+now-tagged points from Qdrant (run the backfill FIRST, then the cleanup).
+
+Attribution map (source_pipeline -> source_subsystem):
+    reflection | deep_reflection | quality_calibration |
+    weekly_assessment | surplus_promotion            -> "reflection"
+    module:automaton_supervisor                       -> "autonomy"
+
+Everything else is left untouched (NULL) — user-sourced content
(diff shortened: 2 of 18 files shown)


WingedGuardian:
## First-pass review — PR #918

### Structural review

**No ERRORs.** Three warnings on test coverage for the changed write paths.

---

**WARNING — Changed write paths in `autonomy/executor/trace.py` with no visible test update**

Five methods now pass `source_subsystem="autonomy"` to `.store()`:
- `finalize` (line ~120)
- `_record_skill_observation` (line ~539)
- `_record_calibration` (line ~567)
- `_record_optimization` (line ~591)
- `_record_drift` (line ~619)

This changes routing from full-embed to FTS5-only. If `tests/test_autonomy/test_trace.py` (or equivalent) uses a real or partial MemoryStore, the store call assertions should cover the new kwarg — otherwise the tag will be silently dropped by any test that stubs `.store()` without it. No test update visible in the diff.

---

**WARNING — Changed write path in `autonomy/executor/engine.py::_promote_notepad` with no visible test update**

`source_subsystem="autonomy"` added at line ~1139. Same concern: existing tests for notepad promotion that assert on `.store()` kwargs should include the new parameter, or the routing change goes unverified.

---

**NOTE — `cc/direct_session.py::_record_proposal_outcome` changes are inside broad `except Exception` swallowers**

The two `source_subsystem="ego"` additions (lines ~881, ~911) live inside exception-swallowing blocks that silently log and continue on any error. The new behaviour (FTS5-only routing) is therefore untestable via the failure path. Not blocking, but worth knowing if you later want to add a regression test — you'd need to wire up a non-swallowing variant.

---

**CLEAN** — no secrets, private IPs, credentials, or private hostnames in added lines. PR title, body, and CHANGELOG entry use "leaking into recall" as memory-subsystem terminology; this does not describe a security or PII incident. No wording issues.

---

WingedGuardian:
Thanks — addressed the review in \`055feeb9\`:

- **WARNING (trace.py, 5 methods)** and **WARNING (engine.py \`_promote_notepad\`)** — the routing change is already enforced at source by the new AST coverage guardrail (`test_store_subsystem_coverage.py` hard-fails if any of these `.store()` calls drops `source_subsystem`), but the behavioral tests only asserted "store was called". Strengthened the two existing tests (`test_finalize_stores_memory`, `test_promote_detects_added_content`) to pin `source_subsystem="autonomy"` so the FTS5-only routing intent is directly verified at runtime.
- **NOTE (direct_session.py, exception swallowers)** — accepted as-is. The two `source_subsystem="ego"` additions are inside failure-path `except Exception` bookkeeping; they carry no behavior a regression test could exercise without rewiring the swallower, and the guardrail already guarantees the tag is present. Not worth a non-swallowing test variant for dispatch-failure records.

PR #916, opened 2026-07-06
feat(audit): AI-code failure-mode taxonomy for auditor + dev guidance

WingedGuardian:
## Summary

Folds the AI-generated-code audit research (ingested to the KB under `engineering.ai-code-audit`) into the two surfaces that act on it:

- **`CODE_AUDITOR.md`** (idle-time surplus auditor briefing): a new hunting section for the defect classes iterative LLM development is documented to produce — swallowed async errors (catch-log-return-nothing), orphan state without teardown, race surfaces, phantom guards, near-duplicate helpers from context loss, cosmetic single-implementation abstractions, pattern abandonment at context-decay seams, and constraints quietly removed across refinement cycles. The JSON output contract is unchanged (and pairs with the bare-array intake handling shipped in #886).
- **genesis-development skill**: new `references/ai-code-audit.md` — the distilled taxonomy plus ordered deep-audit passes (inventory → structure → async/state) — added to the reference router, and a short inline **iterative-refinement discipline** rule: iterate with scoped explicit prompts (never "improve this"), be security-explicit when touching validation/auth, and diff each refinement for what it *removed*.

Companion change made outside the repo: the user-level code-voice skill gained the same tells for reviewer-facing code.

## Testing
- Prompt/docs only — no runtime code paths. Auditor output contract unchanged (pinned by existing `tests/test_surplus/test_code_audit.py` parsing tests, untouched).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #919, opened 2026-07-07
chore: release v3.0b17

WingedGuardian:
175 commits since v3.0b16 (PRs #742–#916). Full list in [CHANGELOG.md](https://github.com/WingedGuardian/GENesis-AGI/blob/main/CHANGELOG.md) (89 entries).

### Added
- Voice dashboard gets a full cockpit: Bridge-tab live edge health, Device-tab HA-polled vitals, and a Judgment review that shows the attention judge's reasoning — now with per-device provenance and a regression alert if the bridge's memory climbs again.
- Honest system health: the container badge factors real CPU/PSI instead of raw memory alone, per-session Claude Code memory is tracked with leak alerts, and scheduled jobs that quietly stop succeeding are now caught.
- A new **Campaigns tab** to see and steer autonomous campaigns — including one where Genesis now stewards its own upstream open-source pull requests.
- Claude Fable 5 and the full effort range everywhere a model is chosen; `gmodel` launches an interactive session on any roster model; the **gauntlet** proves a model can actually drive Claude Code before you rely on it.
- A self-improvement loop: weekly cognitive-quality grades in the morning report, regression alerts with investigation proposals, and an Evo loop that proposes — never auto-applies — reflection-prompt improvements against a golden set.
- Off-site backup retention pruning, daily disk hygiene, a Sentinel shadow-mode reversibility classifier, a self-hardening host SSH key, and a Guardian rolling healthy snapshot — the restore point its rollback recovery always needed.

### Changed
- The idle-time code auditor now hunts AI-generated-code failure modes specifically — swallowed async errors, orphan state, phantom guards, iteration scars. The Sentinel only wakes for alarms it can actually remediate.
- Procedure learning now extracts concrete, replayable playbooks instead of vague summaries, with cross-type dedup and relevance-ranked surfacing — including flagged unproven drafts.

### Fixed
- Health-alarm honesty: false backlog/overdue/dark alarms, DLQ crying wolf, and per-ego cycle health are all fixed. Guardian/Sentinel recovery targets the live service, prunes stale snapshots, and gained a real rollback restore point.
- `update.sh`'s host-VM sync (silently dead) is restored, Guardian redeploy reconciles against the actual host commit, and Claude Code stays aligned to its pin instead of drifting.
- Code intelligence: all index spawns route through one locked, capped entrypoint — fixing a real container freeze — and the codebase-memory server is hard-capped at 2 GB.
- Surplus/knowledge intake: fenced and bare-array findings now parse into the knowledge base instead of raw JSON dumps; deleting a knowledge item finally works end-to-end.
- Inbox: the phantom re-detection storm is stopped, dispatch is at-most-once, partial failures auto-retry, and two parked files can no longer leapfrog cancel-and-recreate every 30 minutes.

### Security
- The contribution sanitizer blocks the full Tailscale range, `aiohttp` is raised past 11 CVEs, the autonomy context-ceiling gate uses the clamped effective level, and the Guardian SSH key is bound to source IP with no interactive terminal.

## Process notes
- CHANGELOG `[Unreleased]` had drifted into 9 split Added/Changed/Fixed/Security headings across concurrent merge sessions; consolidated into one of each here, zero duplicates found, entry count verified independently (89).
- Two gaps found and filled during curation: #915 (ambient-bridge RSS regression alert) had no changelog entry at all; #914 (inbox parked-file leapfrog fix) had a highlights citation but no full entry.


WingedGuardian:
Merging on CI alone — Codex is quota-exhausted and the structural-review bot hasn't posted (docs/version-only diff: CHANGELOG.md + VERSION + pyproject.toml version string, no source code). Content was independently privacy-scanned before this PR was opened (see the earlier grep pass over the full curated section).

PR #923, opened 2026-07-07
fix(guardian): report a failed Proxmox resize as a failure (poll the PVE task)

WingedGuardian:
## What

A Proxmox VE disk/RAM resize is **asynchronous**: `PUT .../qemu/<vmid>/resize` returns HTTP 200 with a `UPID:` task string in `data`, and the real work runs as a background worker that can **fail after the 200** (a missing `Datastore.AllocateSpace` on the storage surfaces exactly this way). The adapter ignored the UPID and only re-read the VM config, so a **failed** resize task and a merely **slow** one both looked identical — an "unverified" grow with no diagnostic.

This adds `ProxmoxAdapter._await_task(upid)`: poll `GET /nodes/<node>/tasks/<upid>/status` (audit token — it has `Sys.Audit`) until `status == "stopped"`, mapping `exitstatus == "OK"` → success. Wired into `grow_vm_disk` (after the resize PUT 200) and defensively into `grow_vm_memory`. On a non-OK task it returns `ok=False` / `verified=False` with the **real exitstatus** in the error; on OK it keeps the existing verify-by-re-read.

## Safety contract (preserved)

- **Never-raise** — `_await_task` only issues reads and is null-safe on every field.
- **Exactly one mutating PUT** — the poll never re-issues the resize; every failure/timeout path returns a typed failure.
- **Bounded** — a wall-clock deadline (honors the ~120s bound even when a connected-but-hung status endpoint makes each read block up to the per-request timeout) plus a poll-count cap (deterministic when reads return fast).
- Reads use the audit token; the mutation stays provision-token only.

## Docs

Corrects `docs/reference/proxmox-provisioning.md` pveum setup to match what the live failure proved:
- `GenesisProvision` needs `Datastore.AllocateSpace` (a disk grow allocates storage), granted at `/storage/<STORAGE>` for the provision token.
- `GenesisAudit` must be **re-granted** at `/storage/<STORAGE>` — a specific-path ACL *replaces* the propagated one, silently breaking the storage read the moment you touch that path.
- Documents the async-resize task behavior and that `require_recent_backup: true` currently refuses every grow (the Proxmox backup-age query is a stub returning `None`; follow-up filed).

## Design note (conscious choice, not a gap)

`exitstatus == "OK"` treats a rare `"WARNINGS: N"` task completion as non-OK. This is the *safe* direction (reports unverified, never auto-retries; a resize doesn't emit warnings), and #920's anti-stack guard self-corrects an unverified-but-landed grow on the next cycle. Kept strict deliberately rather than masking a potential partial failure.

## Tests

`tests/test_guardian/test_provisioning_proxmox.py` — fake task-status transport: running→stopped-OK (verified), stopped-non-OK 403 (exitstatus surfaced, one PUT, no re-read), UPID url-encoding (`%3A`), poll timeout, persistent non-200, and memory-PUT-no-UPID falls through. 21 file / 106 provision-slice tests green; ruff clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,10 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   Genesis itself is down in the outage, the host-side guardian can do it as part
   of recovery. It is **off by default**, every change is approval-gated and
   rate-capped, grows are one-attempt/never-auto-retried, and only two
-  read/write-split Proxmox tokens are ever stored. Setup and the full safety
-  model: `docs/reference/proxmox-provisioning.md`.
+  read/write-split Proxmox tokens are ever stored. A grow that Proxmox accepts
+  but then fails to carry out in the background (e.g. a storage-permission gap)
+  is now reported as a clear failure rather than a vague "couldn't confirm."
+  Setup and the full safety model: `docs/reference/proxmox-provisioning.md`.
 
 - **Ambient bridge memory-leak regression alert.** If the edge bridge's RSS ever climbs past the
   healthy plateau again (total > 1000 MB or diarization child > 450 MB), ambient health flips to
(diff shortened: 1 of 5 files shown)


WingedGuardian:

---

PR #924, opened 2026-07-07
fix(guardian): storage-expand targets the real thin-pool LV (autoextend now actually arms)

WingedGuardian:
## What

Found during the first real disk grow that reached the storage-expand autoextend steps. `expand.py` assumed the thin-pool **LV** name equals the incus **pool** name (`thinpool = pool_name`). In the **default incus LVM layout** the pool is named `default` but its backing LV is `IncusThinPool` — so the two `lvchange` steps (attach autoextend profile, enable dmeventd monitoring) ran against `vg0/default`, a nonexistent LV, and failed:

```
"apply metadataprofile", ok=false, "Failed to find logical volume \"vg0/default\""
"enable monitoring",     ok=false, "Failed to find logical volume \"vg0/default\""
```

The grow still corrected `vg_free` (pvresize worked), but **autoextend was left unarmed** — the profile was never attached to the pool, so the pool would never auto-grow into the new free space. That silently defeats the point of the grow.

## Fix

`_resolve_thinpool_lv(pool_name, vg, run)` resolves the LV name honestly, most-authoritative first:
1. incus's own `lvm.thinpool_name` config (exactly which LV incus uses),
2. the single `lvs -S segtype=thin-pool <vg>` LV in the VG,
3. last resort: the pool name (correct only where the names coincide — never worse than the old behavior).

Also fixes the misleading `pool.py` `_lvm_source` comment that seeded the wrong assumption.

## Impact

This affects **every** incus-LVM install using the default layout (pool `default` / LV `IncusThinPool`), not just one machine — anyone who runs an approval-gated grow would get a corrected `vg_free` but a still-unarmed pool.

## Verified live

On the affected host: `lvchange --metadataprofile genesis-thinpool vg0/IncusThinPool` attaches correctly → `LProfile=genesis-thinpool`, `monitored`, `vg_free=32G`. This PR makes the code produce that outcome automatically on the next grow.

## Tests

`tests/test_guardian/test_provisioning_expand.py` — all three resolution branches, the ambiguous-multiple-pools fallback, and a regression asserting the profile/monitoring target `vg0/IncusThinPool`, never `vg0/default`. 11 file / 137 provision+pool-slice tests green; ruff clean.

### DIFF
--- a/src/genesis/guardian/pool.py
+++ b/src/genesis/guardian/pool.py
@@ -166,7 +166,9 @@ async def _lvm_source(pool_name: str) -> str | None:
             source = s.split(":", 1)[1].strip()
     if driver != "lvm" or not source:
         return None
-    # source is the VG name; the thin pool LV is conventionally the pool name.
+    # ``source`` is the VG name. NOTE: the backing thin-pool LV name is NOT
+    # necessarily the incus pool name (default layout: pool "default" / LV
+    # "IncusThinPool") — resolve it via lvm.thinpool_name, not by assumption.
     return source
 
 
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

PR #925, opened 2026-07-07
fix(dashboard): merge duplicate resolveApproval store keys

WingedGuardian:
## Problem

`dashboard.js` defines `resolveApproval` twice inside the same Alpine store object literal. In a JS object literal the second key silently wins, so the first implementation — the one carrying the `approvalActions` saving/error state machine, the optimistic `approvals` filter, and `readApiError` handling — was dead code. The chat-tab approval buttons therefore resolved approvals with no saving state and no error reporting.

## Fix

One merged implementation at the original (first) location, signature `(requestId, decision, outreachMsgId = null)`:

- Saving/error state machine, optimistic filter, and `readApiError` path kept from the first variant.
- The second variant's optional outreach engage POST (3-arg call site) and outreach/approvals/comms refresh folded in after a successful resolve.
- Post-resolve refresh failures are logged only — they no longer paint an error onto an approval that already resolved.
- Second definition deleted. No template changes: both existing 2-arg and 3-arg call sites work with the merged signature.

## Prevention

New `tests/test_dashboard/test_webui_js_integrity.py` parses the store literal and asserts member-name uniqueness (plus a member-count floor so the parser failing to match fails loudly instead of passing on an empty list). Verified RED on the pre-fix file (detects exactly this duplicate) and GREEN after.

## Verification

- Guard test: RED pre-fix / GREEN post-fix
- `node --check` parses the fixed file
- `tests/test_dashboard/test_api.py`: 29/29 pass
- Inline code review: no findings (call sites, fetchApi semantics, and test heuristic independently verified)

### DIFF
--- a/src/genesis/dashboard/webui/js/dashboard.js
+++ b/src/genesis/dashboard/webui/js/dashboard.js
@@ -2692,7 +2692,7 @@
           }
         },
 
-        async resolveApproval(requestId, decision) {
+        async resolveApproval(requestId, decision, outreachMsgId = null) {
           const current = this.approvalActions[requestId] || {};
           this.approvalActions = {
             ...this.approvalActions,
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

PR #928, opened 2026-07-07
feat(skills): add aws-fde-delivery skill (AWS FDE build-first deliverable contract)

WingedGuardian:
## Summary

**First-ever autonomous task-executor build (V0 canary).** Source: the "New Genesis Capabilities" notepad → inbox evaluation (2026-07-02, ADAPT verdict) → /task intake → autonomous execution (task `t-21400a`) → branch push by the engine. PR opened by the foreground session per V0 protocol; merge is human.

- New `src/genesis/skills/aws-fde-delivery/SKILL.md` — encodes the AWS FDE delivery contract: build-first over advisory; mandatory artifact set (runnable POC, CDK/CloudFormation IaC, grounded cost estimate + scaling projection, Well-Architected review, evolution roadmap); cost figures grounded in AWS Pricing API/documented pricing or explicitly marked UNVERIFIED; AWS MCP Server preferred over raw CLI.
- Registered in `SKILLS_INVENTORY` (phase 7, `cc_background_task`).

Per the eval's own recommendation, this extracts the 1-star third-party skill's *methodology* rather than installing the package.

## Testing (run by the executor, verified)

- `pytest tests/test_learning/test_skills.py -v` — passed
- `ruff check src/genesis/learning/skills/inventory.py` — clean
- Discoverability + phase-7 registration asserted via `PYTHONPATH=src` import checks
- Two-file diff confirmed (`git show --stat`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #931, opened 2026-07-07
fix(executor): downgrade shell-syntax deterministic commands at decompose time

WingedGuardian:
## Summary

First-contact fix from the V0 executor canary (task `t-21400a`, PR #928). The decomposer emitted `source …/activate && pytest …` and `git add … && git commit …` as deterministic steps, but the deterministic runner is **exec-only** (`create_subprocess_exec` + `shlex.split`, no shell — by design, as an injection guard). Both steps failed instantly (`Errno 2: 'source'`; git exit 129 on a literal `&&` argument) and the recovery layer rerouted them to paid CC sessions — correct outcome, wasteful path.

- `decomposer._validate_steps`: commands containing shell operators (`&& || ; | > < $( ` `), shell builtins (`source`, `cd`, `export`, …), `VAR=value` prefixes, or interpreter invocations now **downgrade to `code` steps at decompose time** (same pattern as the existing missing-command downgrade; interpreter set shared with the runner's blocklist via lazy import — top-level would cycle through `executor/__init__` → `engine` → `decomposer`).
- `TASK_DECOMPOSE.md`: documents the exec-only contract — one program per step, venv already on PATH (no `source`), cwd already set (no `cd`), chain via `dependencies` not `&&`.
- `TASK_STEP.md`: code-step constraint now requires committing changed files in-worktree (the executor never commits and delivery only pushes — uncommitted work is invisible; second V0 finding).

Known acceptable false positive: `&&` inside a quoted commit message downgrades to `code` (substring check) — safe direction, pinned by a test.

## Testing

- `pytest tests/test_autonomy/test_decomposer.py` — 35 passed (new: all six V0-shaped failure commands downgrade; clean single-program commands untouched)
- `ruff check` clean

### DIFF
--- a/src/genesis/autonomy/decomposer.py
+++ b/src/genesis/autonomy/decomposer.py
@@ -13,6 +13,7 @@ import contextlib
 import json
 import logging
 import re
+import shlex
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Protocol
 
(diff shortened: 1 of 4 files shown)


WingedGuardian (commented):

---

PR #927, opened 2026-07-07
feat(db): build_candidates ledger + task_states.source (build-lane groundwork)

WingedGuardian:
## Summary

Groundwork PR-2 of the autonomous capability-build lane. Pure-additive DB changes; nothing consumes them yet (BuildLane lands separately).

- **Migration 0047 — `build_candidates`**: one row per (notepad item, verdict episode) — verdict (`build`/`dont_build`/`needs_discussion`) + reasoning + pinned `build_spec`, greenlight approval ref, the user's actual decision (`approved`/`rejected`/`discussed` — calibration ground truth), and build outcome (task/branch/PR/scope-gate result). **Partial unique index** `ON build_candidates(item_key) WHERE user_decision IS NULL` = rescan guard: re-evaluating an unchanged notepad can never spawn a second pending card.
- **Migration 0048 — `task_states.source`** (`TEXT NOT NULL DEFAULT 'user'`): the dispatcher always accepted `source` but never persisted it; delivery-side branching needs provenance. Existing rows backfill as `'user'` — the only path that ever existed.
- Canonical DDL mirrored in `db/schema/_tables.py` (fresh-DB path); new `crud/build_candidates.py`; `crud/task_states.create()` gains `source`; dispatcher passes it through.

## Testing

- 131 passed across `test_schema.py`, `test_migration_0047/0048`, `test_build_candidates.py`, `test_autonomy/test_crud.py`, `test_executor/test_dispatcher.py`, `test_migrations_runner.py`
- New tests: CHECK-constraint rejections, partial-unique-index semantics (open vs decided), migration idempotency + backfill, dispatcher `source` persistence
- `ruff check` clean on the full diff
- Inline code review: clean — migration/_tables DDL equivalence, placeholder counts, no-commit-in-migration, and all `task_states.create` call sites verified

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #932, opened 2026-07-07
feat(executor): build-lane diff allowlist gate at delivery (fail-closed)

WingedGuardian:
## Summary

PR-3 of the autonomous capability-build lane: the **containment boundary**. Ships safe-dark — with no `task_states.source` column yet (PR #927) and no build-lane submitter (PR-6), `_task_source` always resolves `'user'` and the gate is inert until those land. Nothing changes for user-submitted tasks, ever (pinned by test: their first git call is still the push).

- **New `executor/scope_gate.py`** (pure functions): builds may only touch capability trees — `src/genesis/modules/**`, `src/genesis/skills/**`, `src/genesis/mcp/**` (minus `mcp/health/**`, the admin/cognitive-control tool surface), `tests/**`, `docs/**` (minus narrative subtrees). **Deny overrides win everywhere**: env/dotenv files, `*secret*`, service/timer units, `settings*.json`, anything with `approval` in the path, and the task-submission MCP surface. Everything else (autonomy/, ego/, inbox/, db/, runtime/, scripts/, .claude/, .github/, config/) is denied by absence — the allowlist fails **closed**. The gate lives in `src/genesis/autonomy/` — a tree it does not allow — so a build can never PR the gate that judges it.
- **`engine._deliver` graft**: `source='build_lane'` → dirty tree, any git failure, empty diff, or any out-of-scope path **parks the build** (no push, `scope_gate`/`scope_blocked` recorded on the task, user notified). Widening scope is a deliberate human PR on Stage-1 evidence.
- **Adversarial review hardening (second commit)**: case-insensitive deny matching (fnmatch doesn't fold case on POSIX — `SECRETS.ENV` bypassed the globs); `.env.local`/`.env.*` family covered; `mcp/health/**` denied wholesale; delivery diffs with `--raw` and blocks any **symlink** (mode 120000) — path-only gating can't see that a link inside an allowed tree points at `/etc/passwd`.

## Residual risks (documented, accepted for Stage 1)
- Quoted non-ASCII paths from git fail the prefix match and therefore **over-block** (fail-safe, never a bypass).
- Content-level review remains the draft-PR human review; this gate bounds *where*, not *what*.

## Testing
- 67 scope-gate tests (allowlist matrix, deny overrides incl. case variants + dotenv family, malformed paths, raw-diff/symlink parsing, engine fail-closed branches, byte-identical user path) + 49 engine regressions — all green; `ruff` clean.

### DIFF
--- a/src/genesis/autonomy/executor/engine.py
+++ b/src/genesis/autonomy/executor/engine.py
@@ -1234,6 +1234,14 @@ class CCSessionExecutor:
         wt_path = self._worktree_paths.get(task_id)
 
         if has_code and wt_path:
+            # Build-lane tasks pass the diff allowlist gate before anything
+            # leaves the worktree. Blocked (or any git failure computing the
+            # diff) => parked: no push, no PR. User tasks are untouched.
+            if await self._task_source(task_id) == "build_lane":
+                gate_result = await self._run_scope_gate(task_id, wt_path)
+                if not gate_result.allowed:
+                    return
+
             branch = f"task/{task_id[:8]}"
             try:
                 proc = await asyncio.create_subprocess_exec(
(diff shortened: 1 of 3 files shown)


WingedGuardian (commented):
## First-pass review — PR #932

### Structural review

**WARNING — missing timeout on `_git` inner function in `_run_scope_gate`**

`src/genesis/autonomy/executor/engine.py`, `_run_scope_gate` → `_git`:

```python
async def _git(*args: str) -> tuple[int, str]:
    proc = await asyncio.create_subprocess_exec("git", *args, ...)
    out, err = await proc.communicate()   # ← no timeout
    ...
```

A stuck git lock file in the worktree (e.g. left by an aborted prior git operation, an index write interrupted by SIGKILL) causes `git status --porcelain` to block indefinitely — which wedges the executor delivery semaphore with no recovery path.

This is the exact failure mode that motivated the 300s timeout in `pr_open._run` (PR #933). Local git calls almost never hang, but when they do (lock contention, corrupted index, slow network FS), there is no external watchdog and the delivery path is stuck permanently for that task.

Suggested fix — wrap `communicate()` and treat `TimeoutError` as a gate failure (same pattern as `pr_open._run`):

```python
try:
    out, err = await asyncio.wait_for(proc.communicate(), timeout=30.0)
except TimeoutError:
    proc.kill()
    await proc.wait()
    return 128, f"git {args[0]} timed out"
```

30s is generous for a local diff; adjust to taste. The important property is that a stuck git call degrades to a scope-gate failure (fail-closed) rather than a permanent wedge.

---

**NOTE — missing return type annotation on `_run_scope_gate`**

```python
async def _run_scope_gate(self, task_id: str, wt_path: Path):  # no -> ScopeGateResult
```

The method always returns `ScopeGateResult`. If `engine.py` has `from __future__ import annotations`, the lazy-import inside the method body doesn't prevent annotating the return type.

---

**CLEAN** — no credentials, private IPs, private hostnames, personal email addresses, absolute paths with usernames, or hardware identifiers in added lines. Test fixture paths (`/etc/passwd`, `../outside/file.py`) are threat-model data for the malformed-path tests, not real paths. PR title and description describe a new security control — no language referencing leaks, incidents, or infrastructure exposure.

---

PR #934, opened 2026-07-07
test(sentinel): isolate save_state in the autouse fixture

WingedGuardian:
## Problem

The autouse `_isolate_state` fixture in `tests/test_sentinel/test_dispatcher.py` patched only `load_state`. `save_state` — which atomically overwrites the live sentinel state file on the machine running the suite — was instead guarded by **35 hand-rolled per-test `with patch(...)` wrappers** covering the dispatcher's ~21 call sites. One forgotten wrapper = a test silently writing live operational state.

## Fix

- The fixture now also monkeypatches `genesis.sentinel.dispatcher.save_state` to a no-op.
- All 35 per-test wrappers removed (mechanical unindent / clause removal only — no test-body logic touched; none had an `as` binding, so nothing asserted on the removed mocks).
- The one intentional round-trip test (`test_rejection_persists_across_save_load`) imports from `genesis.sentinel.state` with an explicit `path=` and is unaffected — the fixture patches the dispatcher module's rebound name, not the state module.

## Verification

- `ruff` clean; 68/68 tests pass.
- sha256 of the live state file identical before/after the full test run — with zero per-test wrappers remaining, that is the end-to-end proof the fixture covers what the wrappers used to.
- Independent code review confirmed the no-op signature matches all dispatcher call sites and no wrapper removal changed more than structure.

### DIFF
--- a/tests/test_sentinel/test_dispatcher.py
+++ b/tests/test_sentinel/test_dispatcher.py
@@ -23,10 +23,20 @@ from genesis.sentinel.state import SentinelState, SentinelStateData
 
 @pytest.fixture(autouse=True)
 def _isolate_state(monkeypatch):
-    """Ensure load_state returns fresh state, not the live system's state file."""
+    """Isolate every test from the live system's state file.
+
+    load_state returns fresh state and save_state is a no-op — without the
+    latter, any dispatcher call site not individually wrapped in a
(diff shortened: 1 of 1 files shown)


WingedGuardian:

---

PR #935, opened 2026-07-07
fix(scripts): worktree guard for bootstrap.sh editable install

WingedGuardian:
## Problem

`bootstrap.sh` ran `pip install -e` with **no worktree guard**, while `install.sh` carried its own inline guard for the identical operation. An editable install is system-wide state — pointing it at a git worktree redirects every Genesis process (server, bridge, watchdog) to the worktree's code, which previously caused an I/O death spiral and repeated crashes. Direct `bootstrap.sh` invocation from a worktree was unprotected (update.sh refuses worktrees by path only, and only on its own entry path).

## Fix

Root cause is duplication, so the guard + install + import verification move to a single implementation:

- New `scripts/lib/venv_setup.sh` → `editable_install_guarded <repo_dir> <venv_path>` (rc 0 = installed+verified, 1 = blocked worktree, 2 = not importable).
- The guard now checks the **target repo dir** explicitly (`git -C`) instead of the caller's cwd — the old install.sh guard silently checked whatever directory the installer was run from.
- Both entry points source it; caller semantics preserved: install.sh warns + `SETUP_WARNINGS=1` on import failure (continues on a blocked worktree, as before); bootstrap.sh exits 1 on either.

## Verification

- `bash -n` on all three scripts.
- Functional test in a scratch git repo with a fake pip: worktree run → rc 1, pip **never invoked**; main-checkout run → pip invoked with correct args, rc 0; failed import verification → rc 2.
- Independent code review confirmed the `set -euo pipefail` interactions (`|| rc=$?` capture, pipefail inside the lib), sourcing order in both scripts, and that no other `pip install -e` call site was missed.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -394,8 +394,17 @@ if [[ ! -d "$VENV_DIR" ]] || [[ ! -x "$VENV_DIR/bin/python" ]] || [[ ! -x "$VENV
     "$PYTHON_BIN" -m venv "$VENV_DIR"
 fi
 echo "  Syncing dependencies..."
-"$VENV_DIR/bin/pip" install -e "$GENESIS_ROOT" --quiet 2>&1 | tail -1 || true
-if ! "$VENV_DIR/bin/python" -c "from genesis.runtime import GenesisRuntime" 2>/dev/null; then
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

WingedGuardian:

---

PR #936, opened 2026-07-07
fix(scripts): single source for the CLAUDE.md network-identity block

WingedGuardian:
## Problem

`update.sh` and `host-setup.sh` each carried their own writer for the network-identity sentinel block in the user-level CLAUDE.md, and they drifted:

- `update.sh` (which rewrites the block on **every** update) dropped the Tailscale line entirely.
- Its container-IP detection (`hostname -I | awk '{print $1}'`) reports the Tailscale CGNAT address as the "Container IP" on installs where the tailscale interface sorts first — observed live.
- Its IPv6 detection has the same flaw (Tailscale's ULA is scope global).

Net effect: one run of host-setup writes a correct block; the next update.sh run silently degrades it.

## Fix

New `scripts/lib/claude_md_blocks.sh` — single source of truth:

- `build_network_identity_block` — one canonical builder (container v4/v6, host v4/v6, optional Tailscale line, dashboard URL). Tailscale absent → line omitted, everything else intact (it is default-but-optional across installs).
- `detect_container_lan_ip` / `detect_container_lan_ipv6` — first global address **not** on a tailscale interface, with fallback to the old behavior.
- `detect_tailscale_ip` — CLI first, interface fallback, empty + rc 0 when absent.
- `write_sentinel_block` — the shared sed-delete + append.

`update.sh` sources it in-container. `host-setup.sh` sources it from its host-side checkout and keeps piping the built block into the container via incus — same mechanism, shared content.

## Verification

- `bash -n` on all three scripts.
- Builder run standalone in-container: container IP is now the bridge address (not CGNAT), Tailscale line present, no-tailscale variant degrades cleanly.
- `write_sentinel_block` idempotent on a scratch file (two runs → one marker pair), user-editable sections preserved.
- Full e2e via the post-merge `scripts/update.sh` run (host-deploy gate applies to this PR).

### DIFF
--- a/scripts/host-setup.sh
+++ b/scripts/host-setup.sh
@@ -991,20 +991,13 @@ fi
 # instructions, not tracked in the repo). Idempotent via sentinel blocks.
 echo ""
 echo "  Writing network identity into ~/.claude/CLAUDE.md..."
-# Build the network block on the host side where all variables are available
-_NET_LINES="## Network Identity"
-_NET_LINES="${_NET_LINES}
-"
-_NET_LINES="${_NET_LINES}
-- **Container IP**: ${CONTAINER_IPV4}"
-[ -n "$CONTAINER_IPV6" ] && _NET_LINES="${_NET_LINES} (v6: ${CONTAINER_IPV6})"
-_NET_LINES="${_NET_LINES}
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

PR #937, opened 2026-07-07
fix(channels): legacy Telegram approval fixes — shared vocabulary, scoped standalone resolution, human-scale waiter TTL

WingedGuardian:
## Summary

PR-9 of the build-lane campaign (user-approved scope): the three long-standing Telegram approval bugs share one root theme — three subsystems parsed decision words with three drifted vocabularies, and a machine-scale 300s timeout guarded a human-scale wait.

- **One canonical vocabulary** (`genesis/util/approval_words.py`, tokens + standalone phrases + emoji). The CLI approval gate, the Approvals-topic bare-text path, and the ego proposal parser all import from it — "go" and "sounds good" now work everywhere a quote-reply "ok" does. Strictly widened, never narrowed (pinned by test).
- **"Ok did nothing" fixed**: Approvals-topic bare text gains leading-token semantics ("Ok sounds good" resolves), matching the gate's own quote-reply behavior. Still hard-scoped to the Approvals topic; non-decision openers never match. Review hardening: **quote-replies never take the bare-text path** — without that guard, a quote-reply onto an OLDER approval would have resolved the most recent one instead.
- **Scoped standalone resolution**: `resolve_any_pending` stays retired. Its replacement `resolve_scoped_pending` resolves a waiter only when the message arrives in the exact chat+topic the prompt was **delivered** to (`OutreachResult` now carries `chat_id`/`thread_id`; the pipeline records the context on the waiter). The DM-resolves-alert-topic conflation that forced the original disable is structurally impossible; contextless waiters are never eligible; aliases count as one waiter; two eligible waiters = ambiguous = no resolution.
- **Button TTL stall fixed**: ReplyWaiter/pipeline default timeout 300s → 7200s (policy floor — waiting on a human), `outreach_send_and_wait` MCP default 300s → 1800s. Blast radius verified: every existing call site passes an explicit timeout, so the default change is inert today and only stops future stalls.
- **Honest expired-tap feedback**: "Approved *(expired)*" implied the decision was recorded; it now says plainly nothing was recorded and Genesis will re-ask.

## Testing
- 166 tests green across approval_words (new), reply_waiter (7 new scoped/alias/lifecycle tests), callback handler (incl. the quote-reply-shadowing guard), proposal parser, gate reply-decision, pipeline, MCP, and handler flow tests; `ruff` clean.
- Adversarial review: no blocking findings; its one sub-bar catch (quote-reply shadowing) fixed before push; timeout blast radius enumerated (all call sites explicit).

### DIFF
--- a/src/genesis/autonomy/approval_gate.py
+++ b/src/genesis/autonomy/approval_gate.py
@@ -17,14 +17,16 @@ from typing import Any
 from genesis.autonomy.cli_policy import load_autonomous_cli_policy
 from genesis.cc.types import CCInvocation
 
-logger = logging.getLogger(__name__)
+# Canonical vocabulary — shared with the Telegram bare-text path and the
+# ego proposal parser so the three surfaces can never drift again.
+from genesis.util.approval_words import (
+    APPROVE_TOKENS as _APPROVE_WORDS,
+)
+from genesis.util.approval_words import (
+    REJECT_TOKENS as _REJECT_WORDS,
+)
 
-_APPROVE_WORDS = frozenset({
-    "approve", "approved", "ok", "yes", "go", "lgtm",
-})
-_REJECT_WORDS = frozenset({
-    "reject", "rejected", "deny", "denied", "no", "nope",
-})
+logger = logging.getLogger(__name__)
 
 
 def _json_loads(raw: str | None) -> dict[str, Any]:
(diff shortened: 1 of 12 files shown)


WingedGuardian (commented):

Style note (pre-existing, not introduced by this PR): `_bare_decision` in `src/genesis/channels/telegram/_handler_messages.py` imports `scoped_decision` inside the function body rather than at module level. The old code already had `import re` there; the new code inherits that pattern. No circular-import risk — `approval_words.py` only depends on stdlib (`re`). Worth hoisting to module level in a follow-up cleanup.

---

PR #940, opened 2026-07-07
fix(guardian+inbox): gateway-staleness reference + modified-path storm guard

WingedGuardian:
Two small, independent container-side fixes (take effect on next `genesis-server` restart; no host deploy required — `watchdog.py` is container-side per `guardian/DEPLOYMENT.md`).

## 1. Guardian gateway-staleness inversion (`guardian/watchdog.py`)

`_check_gateway_staleness` computed its expected gateway sha from `code_version` (the install-dir's `git rev-parse HEAD`). The guardian `redeploy` action tar-extracts new code and records a new `deployed_commit` **without ever running git** in the install dir, so that HEAD systematically lags what was deployed. Result: after every healthy redeploy the check fired a spurious "gateway stale" alert, and its `sync-gateway` self-heal (which copies the install-dir gateway) could roll the deployed gateway **back** to the older one.

Fix: key the expected sha off `deployed_commit` (the authoritative record of what the last redeploy shipped) — matching the sibling `_check_code_drift_inner`, which already measures against `deployed_commit`. A fully-successful redeploy now matches and stays quiet; a genuinely frozen gateway (deployed_commit advanced but the atomic gateway `cp` failed) still mismatches, fires, and self-heals from the tar-updated install-dir file. Verified against `scripts/guardian-gateway.sh` and by running the check unstubbed against the real container git.

## 2. Inbox modified-path storm guard (`inbox/monitor.py`)

`_phase_create_records` has three `_queue_drop` sites — new, modified, retry. The new-file and retry paths both gate on `count_url_failures(f) >= max_retries`; the modified path did not, so a file repeatedly edited with genuinely-new content whose evaluations keep failing on dead URLs would re-queue a drop every scan. Added the same guard to the modified path, placed after the shared supersede and below the empty-delta branch, writing a completing row (advancing the known hash, stopping re-detection) — mirroring the new-file/retry shape. Closes a 2-of-3 consistency gap (defense-in-depth).

## Tests
- `TestGatewayStaleness`: existing cases switched to `deployed_commit` + `test_uses_deployed_commit_not_code_version`, `test_unknown_deployed_commit_skips`.
- `test_modified_path_respects_url_failure_storm_guard`, `test_modified_under_cap_still_queues`.
- 899 guardian+inbox tests pass; ruff clean. Empirical RED confirmed for both fixes (each fix's test fails without the fix). Code-reviewed clean (no P1/P2).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,15 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **The Guardian no longer false-alarms that the host gateway is "stale" after a
+  healthy update — and can no longer roll it back.** A guardian redeploy advances
+  the host's deployed code without moving its install-dir git checkout, so the
+  staleness check (which compared against that lagging checkout) fired a spurious
+  "gateway stale" alert after every clean update, and its self-heal could copy the
+  *older* checked-out gateway over the current one. The check now measures against
+  the commit that was actually deployed, so a healthy update stays quiet while a
+  genuinely frozen gateway is still detected and repaired.
+
 - **Memory recall no longer surfaces Genesis's own internal noise.** Machine-generated
   decisional output — background reflections, autonomy task retrospectives, and ego-dispatch
   records — was leaking into normal recall because several writers didn't mark themselves as
--- a/src/genesis/guardian/watchdog.py
+++ b/src/genesis/guardian/watchdog.py
@@ -476,11 +476,12 @@ class GuardianWatchdog:
         except Exception:
             return None
 
-    async def _expected_gateway_sha(self, code_version: str) -> str | None:
-        """sha256 of the gateway script at the host's install-dir commit.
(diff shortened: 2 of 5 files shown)


WingedGuardian (commented):

---

WingedGuardian:
**Structural review: 1 NOTE**

**NOTE** · `tests/test_inbox/test_drop_dispatch.py` — `test_modified_path_respects_url_failure_storm_guard` and `test_modified_under_cap_still_queues`

Both new tests use raw SQL for post-condition assertions:

```python
await db.execute("SELECT id FROM inbox_items WHERE file_path=? AND status='completed' AND content_hash=?", ...)
await db.execute("SELECT id FROM inbox_items WHERE file_path=? AND status IN ('pending','processing')", ...)
```

The "Raw SQL against genesis.db" rule excepts migrations only — test files are not listed. This pattern appears established in the existing file and is standard for integration-test assertions, so no change requested. Noting for consistency tracking.

---

**PII/Secrets/Wording scan: CLEAN** — no secrets, private IPs, credentials, private hostnames, or sensitive wording in diff, PR title, body, or CHANGELOG additions.

---

PR #949, opened 2026-07-07
fix(outreach): task notifications go verbatim + concise voice TL;DR (stop fabrication)

WingedGuardian:
## Problem

A routine task-executor progress notification ("Proceeding with task…") was **LLM-rewritten into a fabricated alert** — it invented a problem that didn't exist and a fake shell command — and then **spoken aloud** over the voice device. Root cause: every task notification was run through the `ContentDrafter` (LLM) under `OUTREACH_ALERT.md`, whose prompt demands "the issue + the action needed" with **no fact-preservation rule**. An issue-less progress ping was forced into an alert shape, so the model manufactured one. (A sibling "Blocked" ping the same day was accurate — it genuinely *was* an issue, so the alert-shaped prompt fit. The bug only bites benign inputs.)

## Fix — remove the LLM from the task-notification path entirely

Telegram and voice now render **separately**, and neither can fabricate:

- **`OutreachRequest`** gains two optional fields: `verbatim` (`submit()` skips the drafter and delivers `context` exactly — governance still runs first) and `voice_text` (the voice fan-out speaks this short factual TL;DR instead of the full text, so paths/branches/commands are never read aloud). Both default off; every existing caller is byte-for-byte unchanged.
- **`engine._notify`** is now kind-aware (`progress` / `success` / `alert` / `blocked`): all delivered verbatim, each with its own voice `signal_type` and a hand-authored `voice_text`. Category (blocker vs alert) is preserved, so governance/routing is unchanged. All 9 call sites classified.
- **Voice eligibility split:** `task_progress` is silent (Telegram only); `task_complete` / `task_alert` are on the voice allowlist (`config.py` **and** the authoritative `config/outreach.yaml`). Routine progress no longer interrupts by voice.
- **`OUTREACH_ALERT.md`** hardened with a fact-preservation contract — defense-in-depth for the health/sentinel alerts that still draft.
- **`drafter.py`** router-failure fallback now degrades to the real `context`, not the opaque `topic`.

Result: **no LLM anywhere in the task path.** Telegram is verbatim (tokens exact); voice speaks a fixed factual line. Fabrication is structurally impossible, not merely discouraged.

## Bonus latent bug fixed (second commit)

Due-diligence surfaced that task notifications are ALERT/BLOCKER, which bypass salience/quiet-hours but **not** dedup, and `task_notification` fell through to the **24h default** dedup window keyed on `(signal_type, topic, category)`. Since `topic` is `"Task <id>"` for every ping, a "Task completed" (ALERT) collided with the earlier "Proceeding" (ALERT) and was **suppressed for 24h** — a delivered ping silently swallowed the task's terminal notification. The kind→signal_type split de-collides cross-kind pings; `task_progress/complete/alert` now get a dedup window of `0` (never suppress, like `cli_approval`) to close the rest.

## Testing

- `submit(verbatim=True)` never invokes the drafter; delivers `context` exactly; non-verbatim still drafts (regression guard).
- Voice speaks `voice_text` (not the tokened text) when set, falls back to full text without it (unchanged for non-task callers).
- `_notify` kind→(category/signal_type/salience/voice_text) mapping; voice allowlist per kind (`task_progress` silent).
- Task lifecycle notifications are never dedup-suppressed.
- Drafter falls back to context, not the opaque topic.
- **E2E:** `engine._notify` → a real `OutreachPipeline` → Telegram receives the exact detailed text (paths intact) while voice speaks only the token-free TL;DR; drafter asserted never called.

167 targeted tests green; full `test_outreach/` green (162); ruff clean.

Affects **all** task-executor runs, not just the (still-disabled) build lane. Deploys on the next server restart.

### DIFF
--- a/config/outreach.yaml
+++ b/config/outreach.yaml
@@ -81,8 +81,8 @@ health_alerts:
 # Everything still goes to Telegram regardless — this only controls what
 # interrupts the user out loud during `hours`. This block is authoritative;
 # editing it changes behavior on next server restart. Keep entries specific
-# (prefix match) and keep `task_notification` in sync with the signal_type
-# set in src/genesis/autonomy/executor/engine.py _notify.
+# (prefix match) and keep the `task_*` entries in sync with the signal_type
+# values set in src/genesis/autonomy/executor/engine.py _notify.
 voice:
   alert_ids:
     # Memory-system + resource emergencies (matched via source_id)
@@ -95,6 +95,9 @@ voice:
     - "sentinel_escalation"
     - "sentinel_approval"
     - "sentinel_action_approval"
-    # Autonomous task hit a blocker/alert
-    - "task_notification"
+    # Autonomous task reached an attention-worthy state (complete / alert).
+    # `task_progress` is intentionally omitted — routine progress pings go to
+    # Telegram only and never interrupt by voice.
+    - "task_complete"
+    - "task_alert"
   hours: [9, 2]   # 9am-2am local (wraps midnight); end hour exclusive
--- a/src/genesis/autonomy/executor/engine.py
+++ b/src/genesis/autonomy/executor/engine.py
@@ -385,7 +385,7 @@ class CCSessionExecutor:
                     task_id,
                     f"Proceeding with task: {description} "
                     f"({len(steps)} steps)",
-                    "alert",
+                    "progress",
                 )
 
                 step_results = []
@@ -676,7 +676,8 @@ class CCSessionExecutor:
                 f"Task completed. {len(step_results)} steps executed.",
             )
             await self._notify(
-                task_id, f"Task completed: {description}", "alert",
+                task_id, f"Task completed: {description}", "success",
+                voice_text="Task completed successfully.",
             )
             return True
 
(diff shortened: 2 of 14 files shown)


WingedGuardian (commented):
## First-pass review — structural + PII/secrets/wording

### Structural: 1 WARNING

**`src/genesis/outreach/pipeline.py` — `verbatim=True` with `context=None` is a silent footgun**

In the verbatim path added here:

```python
if request.verbatim:
    formatted = self._formatter.format(request.context, format_target)
```

`OutreachRequest.context` is typed `str | None = None`. Every current caller in `engine._notify` always supplies `context`, so this is safe today. But the interface creates a footgun: any future caller (or an accidental refactor) that sets `verbatim=True` without providing `context` will pass `None` to the formatter — which may produce garbage output or raise.

Recommend narrowing the contract, e.g. an assertion at the guard:

```python
if request.verbatim:
    assert request.context is not None, "verbatim=True requires context to be set"
    formatted = self._formatter.format(request.context, format_target)
```

Or document the invariant in the `verbatim` field comment in `types.py` ("`context` must be set when `verbatim=True`"). Either is sufficient.

---

Scanned all added lines (+) across all 13 changed files: no API keys, tokens, private IPs, SSH material, database credentials, personal email addresses, absolute user paths, private hostnames, or hardware identifiers. PR title and body contain no language referencing credential leaks, security incidents, or infrastructure exposure.

---

### Everything else: CLEAN

No pattern violations (no nohup, no `/tmp/`, no `git add .`, no direct `task_states` inserts, no raw SQL, no bare `except:`, no mutable defaults, no `os.killpg`, no star imports, no `IntervalTrigger`). All changed `src/` modules have corresponding test files with targeted new tests. Conventional commit prefix `fix:` present.

---

PR #948, opened 2026-07-07
fix(guardian): read non-LVM pool used% from df, not a bogus incus flag

WingedGuardian:
## Summary

Fixes a latent bug in the Guardian's storage-pool measurement, surfaced while validating #944 (btrfs-on-LVM support) on real hardware.

`measure_storage_pool` derived the backend-agnostic `pool_used_pct` from `incus storage info <pool> --format json` — but **`--format` is not a valid flag for `incus storage info`**. The call therefore always errored (`Error: unknown flag: --format`), and `pool_used_pct` was **perpetually `None`**.

This went unnoticed because LVM-thin pools tier on `data%`/`metadata%` (from `lvs`); `pool_used_pct` is only the *sole* tiering signal for **non-LVM (btrfs/dir) pools** — which never ran this path before #944. On a btrfs pool the effect was that `worst_tier` stayed permanently `unknown`: **no early warning as the pool fills toward a read-only-rootfs outage** (the exact failure the storage tiers exist to pre-empt).

## Fix

- Derive `pool_used_pct` from `df -B1 --output=used,size <mount>` on the pool mount — the same source `snapshots.py` already uses for this pool — via a new defensive `_pool_used_pct_via_df` helper. Every failure path (rc≠0, short/garbage output, zero size) returns `None`, never raises.
- Scope it to the **non-LVM branch only**. The LVM-thin path is unchanged: `pool_used_pct` stays `None` there (unused — `data%`/`metadata%` remain authoritative), byte-for-byte identical to prior behavior.
- Remove the now-unused broken incus call and the `import json` it needed.

## Validation

- **Live on the btrfs-on-LVM host:** the old call returns `Error: unknown flag: --format`; `df` on the pool mount reports ~16% used. After deploy, `disk-status` should report `detected: true`, `pool_used_pct ≈ 16`, `tier: ok` (previously `detected: false`, `tier: unknown`).
- New tests (`tests/test_guardian/test_pool.py`): `_pool_used_pct_via_df` success + every failure path (rc≠0, single-line, non-numeric, zero-size), and the `measure_storage_pool` non-LVM branch end-to-end (populates used% → tiers) plus df-failure → `detected: false`.
- Full `tests/test_guardian/` green (615 passed); ruff clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #944, opened 2026-07-07
feat(guardian): teach storage-expand the btrfs-on-LVM substrate + durable DB nodatacow

WingedGuardian:
## Summary

Teaches the Guardian's approval-gated storage-expand (shipped in #920 for **LVM-thin** substrates) a second substrate: **btrfs-on-LVM**. On installs where the incus storage pool is a btrfs filesystem sitting on an LVM logical volume, #920's absorb was a no-op — an L3 (hypervisor) disk grow would land but never reach the guest filesystem. This adds the missing absorb path plus a durable `nodatacow` fix for the SQLite DB on btrfs.

Both changes are additive and gated: the LVM-thin path is unchanged, and nothing here can mutate storage without an existing live human APPROVE.

## WS-A — btrfs-on-LVM absorb (`guardian/`)

`expand_storage` now forks on the incus pool driver:

- **LVM-thin path: unchanged.**
- **btrfs path:** resolve the backing `(vg, lv)` from the pool's *live mount* (`findmnt -no SOURCE <mount>` → `lvs -o vg_name,lv_name`), then `pvresize` the PV(s) → `lvextend -L +<bytes>b <vg>/<lv>` → `btrfs filesystem resize max <mount>`, then re-read the filesystem size and require strict growth before reporting success. Each step early-returns on failure, so a failed step never cascades into the next.
  - The extend is **clamped to the actual `vg_free`** rather than issued as a literal `+<N>G` — a +N GiB hypervisor grow yields slightly fewer usable extents (PV metadata rounding), so a literal request would fail exactly when used as intended. It is never `+100%FREE` (that guard stays scoped to the thin path).
  - Resolution is anchored to the specific pool's mount, so it can only ever target that pool's LV — never a sibling LV.
- `worst_tier` gains a pool-used% fallback that engages **only** when both LVM data% and metadata% are unavailable (i.e. non-LVM backends). On any LVM install both percents are present, so behavior there is unchanged. This lets a full btrfs pool actually raise warn/high/crit tiers (alerting + autonomous-propose parity with LVM).
- New `pool_used_warn/high/crit_pct` config fields (default 75/85/92) back that fallback.
- The absorb alert is now driver-aware (`fs_size=` for btrfs vs `vg_free=` for thin).

## WS-B — nodatacow the SQLite DB on btrfs

A CoW SQLite DB on btrfs suffers WAL write-amplification and chronic fragmentation. This install sets `chattr +C` on the data directory so DB files inherit `nodatacow` at creation:

[...]

PR #951, opened 2026-07-07
feat(scripts): offline nodatacow conversion for an existing DB

WingedGuardian:
## Summary

Adds `scripts/convert_db_nodatacow.sh` — a one-time **offline** maintenance script to convert an existing install's `genesis.db` to btrfs `nodatacow` (`chattr +C`). Complements the durable pieces already shipped (bootstrap `+C` for fresh installs in #944, df-based pool tiering in #948, the drift monitor).

## Why a dedicated offline script (not an in-session step)

`chattr +C` silently no-ops on a non-empty file, so applying nodatacow to an existing DB means rewriting its data into a `+C` directory — a **new inode**. That inode swap is only safe with **exclusive** access, and `genesis.db` is held open not just by `genesis-server` but by the MCP servers of every running Claude Code session (~30 persistent connections). So the conversion cannot be done from inside a CC session — it needs a quiet window with all sessions closed.

## What the script does

1. **`--check`** (read-only): reports btrfs/`+C` state and current DB holders, changes nothing.
2. Pauses the host guardian by writing `paused.json` — read as a fallback while the server API is down — so it won't try to "recover" the deliberately-stopped server. Prior state restored on exit.
3. Stops `genesis-server`, then **refuses to proceed unless it is the sole DB holder** (aborts with the list of holders if any CC session's MCP servers are still attached).
4. `PRAGMA wal_checkpoint(TRUNCATE)` → backup → `chattr +C` dir → `VACUUM INTO` a fresh `+C` file (compacts + defrags) → integrity-check → **re-check exclusivity** → atomic swap → integrity-check again → restart.
5. EXIT trap: rolls back from the backup on a post-swap failure, always restarts the server and restores the guardian pause state. Idempotent (exits 0 if already `+C` or not on btrfs).

## Safety review

Reviewed adversarially for data loss. Two guards added from that review:
- **WAL completeness**: checkpoint before backup, so the backup is self-contained even if the server was SIGKILLed with dirty WAL (a plain `cp` + `integrity_check` would otherwise pass on an incomplete copy and a rollback could discard committed data).
- **Snapshot-not-a-lock**: a second holder re-check immediately before the swap, since a new same-user MCP connection could attach mid-run. (True lockout of same-user opens isn't possible; the operator gate — close all sessions — plus this re-check is the proportional fix, and the header states it explicitly.)

Validated: `bash -n` clean, `shellcheck -S warning` clean, `--check` exercised live (correctly enumerates the current ~31 non-server holders → would abort, as intended).

## How it will be run

Not run by CI or automatically. Tracked as a follow-up to run in a quiet window (all Claude sessions closed), verifying with `--check` first. The existing DB stays CoW until then — WAL already banks the bulk of the perf win, so this is a fragmentation/perf optimization, not urgent.

### DIFF
--- a/scripts/convert_db_nodatacow.sh
+++ b/scripts/convert_db_nodatacow.sh
@@ -1,0 +1,241 @@
+#!/usr/bin/env bash
+#
+# One-time OFFLINE conversion of the Genesis SQLite DB to btrfs nodatacow (+C).
+#
+# WHY
+#   On btrfs a copy-on-write SQLite DB suffers WAL write-amplification and
+#   chronic fragmentation. Fresh installs get nodatacow via bootstrap.sh
+#   (`chattr +C` on data/, inherited at file creation). An EXISTING install's
+#   live genesis.db predates that flag, and `chattr +C` silently no-ops on a
+#   non-empty file — the only way to apply nodatacow to existing data is to
+#   rewrite it into a +C directory, which produces a NEW inode.
+#
+# WHY IT MUST BE OFFLINE (read this before running)
+#   genesis.db is held open not only by genesis-server but by the MCP servers
+#   of every running Claude Code session. Swapping the inode while any of them
+#   hold it strands those connections on the old (unlinked) inode — split-brain
+#   writes and data loss. This script therefore REFUSES to run unless it is the
+#   SOLE holder of the DB after genesis-server is stopped. Close ALL Claude Code
+#   sessions first, then run this from a PLAIN terminal — never from inside a CC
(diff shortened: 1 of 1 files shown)


WingedGuardian (commented):

---

WingedGuardian:
Review note: Codex was quota-limited (usage-limit notice, no findings). Coverage: inline Claude code-review (2 data-loss findings raised and fixed in this branch — WAL checkpoint before backup + pre-swap holder re-check) + structural-review bot (CLEAN — PII/Secrets/Wording clean), all 12 CI checks green.

PR #952, opened 2026-07-07
feat(guardian): configure-provisioning gateway verb (state-dir override)

WingedGuardian:
## Summary

Adds a `configure-provisioning` guardian gateway verb so activating the Guardian's approval-gated Proxmox provisioning (#920) on an install no longer requires hand-editing the host `guardian.yaml`.

The host `guardian.yaml` is git-tracked (not skip-worktree) and installer-generated, so a raw edit risks being clobbered by a future checkout and leaves an un-landed host divergence. Instead:

- **Gateway verb** `configure-provisioning key=value ...` → `python -m genesis.guardian --configure-provisioning ...` writes a `provisioning.local.yaml` **override into the guardian state dir** — outside the git checkout, so `update.sh` redeploys never touch it.
- **Config loader** merges the override onto the `provisioning` sub-config in `load_config` (via `_finalize`) *before* env overrides, so `GUARDIAN_PROVISIONING_ENABLED` still wins as the kill switch.
- Only `ProvisioningConfig` fields are accepted (unknown key → error); values are type-coerced from the `key=value` strings. **No secrets** here — the two Proxmox tokens cross the credential bridge, never this verb.

## Security

`SSH_ORIGINAL_COMMAND` is untrusted. The verb requires **every token to be strictly `key=value`-shaped** (`^[a-z_][a-z0-9_]*=[-A-Za-z0-9_.]*( ...)*$`). Without that, a bare flag-shaped token (e.g. `--storage-expand`, `--provision-grow-disk scsi1 5`) would word-split into `argv` and hijack `main()`'s `if "--x" in sys.argv` dispatch to a *different* guardian verb (an argument-injection footgun; not a privilege escalation, since a key-holder can call those verbs directly). Verified: valid `key=value` passes; `--test`, `--storage-expand`, `--provision-grow-disk scsi1 5`, `; rm -rf /`, `$(whoami)` all rejected. The override write is atomic (temp + `os.replace`).

## Testing

- `tests/test_guardian/test_config_provisioning_override.py` (7): write+merge roundtrip, partial merge onto defaults, unknown-field rejection, env-kill-switch precedence, unreadable-override resilience (never raises in the health-check hot path), native-YAML-type acceptance.
- Full `tests/test_guardian/` green (629); ruff + shellcheck + `bash -n` clean; CLI + gateway-regex smoke-tested live.

Reviewed adversarially (found + fixed the injection vector above); config-loader ordering, exception-safety, coercion, and unknown-key handling all verified clean.

### DIFF
--- a/docs/reference/proxmox-provisioning.md
+++ b/docs/reference/proxmox-provisioning.md
@@ -172,9 +172,18 @@ curl -sk -X PUT -H "Authorization: $AUD" "$H/nodes/<NODE>/qemu/<VMID>/config" -d
 2. **Credential bridge** — the awareness tick propagates just those two keys to
    `<state_dir>/shared/guardian/proxmox_creds.env` (0600) for the host guardian
    to read. Host/node/vmid are non-secret config, not bridged.
-3. **guardian.yaml** — fill the `provisioning:` block (`enabled: true`,
-   `api_host`, `node`, `vmid`, `verify_tls`). See the commented template in
-   `config/guardian.yaml`.
+3. **provisioning config** — land the `provisioning` fields (`enabled: true`,
+   `api_host`, `node`, `vmid`, `target_disk`, `storage`, `verify_tls`,
+   `require_recent_backup`). Preferred: the audited, repeatable gateway verb
+   ```
+   configure-provisioning enabled=true api_host=<PVE> node=<node> vmid=<id> \
+       target_disk=scsi1 storage=local-lvm verify_tls=false require_recent_backup=false
+   ```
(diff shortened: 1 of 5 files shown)


WingedGuardian:

---

WingedGuardian:
Review note: Codex quota-limited (usage-limit notice, no findings). Coverage: inline Claude code-review (found + fixed an argument-injection vector — bare flag-shaped tokens could hijack the guardian flag dispatch; now strict key=value only) + structural-review bot (CLEAN — PII/Secrets/Wording clean), all 12 CI checks green.

PR #955, opened 2026-07-08
fix(observability): edit-failure sensor blind to failures — register PostToolUseFailure

WingedGuardian:
## Problem

`tool_call_outcomes` has 12,737 rows and **zero recorded failures** since the sensor shipped (May 17). The sensor was built specifically to measure Edit/Write failure rates — it has been blind since birth.

## Root cause

The sensor was registered only for **PostToolUse**, which per the Claude Code hooks contract fires **only on successful tool calls**. A failed Edit ("old_string not found" etc.) never reaches the hook at all; the failure-marker scan inside the script was unreachable in practice. Failures arrive via the separate **PostToolUseFailure** event (supported by the installed CC 2.1.201 — verified against current hooks docs and the installed binary), which was never registered.

## Fix

- Register `PostToolUseFailure` (matcher `Edit|Write`) → same sensor script.
- Script branches on `hook_event_name`: failure events write `success=0` with the error from `tool_error` (defensive key fallbacks, 200-char snippet cap). Success path unchanged, marker scan kept as a soft-error guard.
- `GENESIS_DB_PATH` env override added for tests only (script stays stdlib-only).

## Tests

9 new tests in `tests/test_scripts/test_edit_failure_sensor.py`: failure/success paths, missing-error-text fallback, snippet cap, non-Edit/Write filtering, malformed stdin, plus a subprocess E2E piping JSON through the real script into a temp DB.

## Verification & regression marker

Hooks load at session start, so the new registration cannot be live-fired from the authoring session. Regression marker: **the first failed Edit in a post-merge CC session must produce a success=0 row**; if the table is still 100% success after a week, the registration is not firing.

Found by the WS-0 observed-harm evidence pass (instrumentation meta-finding M2).

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -372,6 +372,18 @@
                     }
                 ]
             }
+        ],
+        "PostToolUseFailure": [
+            {
+                "matcher": "Edit|Write",
+                "hooks": [
+                    {
+                        "type": "command",
+                        "command": "${CLAUDE_PROJECT_DIR}/.claude/hooks/genesis-hook edit_failure_sensor.py",
+                        "timeout": 500
+                    }
+                ]
+            }
         ]
     },
     "worktree": {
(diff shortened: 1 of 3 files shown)


WingedGuardian (commented):

---

WingedGuardian:
Merging on CI + inline Claude review + clean structural scan — Codex review quota-limited at review time.

PR #957, opened 2026-07-08
fix(mcp): bridge send_and_wait/provision_grow to server via in-process RPC

WingedGuardian:
## Problem

`outreach_send_and_wait` and `provision_grow` block on a live owner Telegram reply, so they need the real `OutreachPipeline` + the single-owner reply-waiter — which exist only in the **genesis-server** process. The standalone MCP subprocess (what a Claude Code session talks to) bootstraps with `pipeline=None`, so both tools dead-ended at **"pipeline not initialized"** from any MCP context. (This forced a recent `provision_grow` E2E to fall back to the guardian gateway verb.)

The guardian's *autonomous* provisioning path is unaffected — it delivers via its own channel, not this pipeline.

## Fix — in-process HTTP RPC (chosen over a DB request/reply bridge)

The server already hosts a Flask API in the **same process/event loop as the pipeline** (`@_async_route` dispatches handlers onto the runtime loop and supports unbounded waits). Two new POST routes run the op in-process; the MCP tools POST to them when their own `_pipeline` is `None`.

A DB request/reply bridge was rejected — it would need a reply column + a fast, atomic, non-blocking drain (a whole class of concurrency bugs the HTTP path avoids).

- **`outreach/rpc.py`** (new) — one shared impl (`send_and_wait_via_pipeline`, `grow_via_pipeline`) called by **both** the `@mcp.tool` wrappers (in-server, where `_pipeline` is set) and the routes → no duplicated provision body to drift.
- **`routes/outreach.py`** — `POST /api/genesis/outreach/send_and_wait`.
- **`routes/provision.py`** (new, registered in `routes/__init__.py`) — `POST /api/genesis/provision/grow`.
- **`outreach_mcp.py`** — `_server_rpc` httpx helper (read timeout = wait + buffer, short connect, clean error dict on a down server — never raises); each tool's `pipeline=None` branch bridges instead of erroring.

## Access posture (deliberate)

No auth token: these inherit the dashboard's existing local-network reachability via the dashboard proxy (which makes every request present as loopback, so an IP guard would be non-functional). `provision/grow` stays **owner-APPROVE-gated** before any mutation; `send_and_wait` is a trusted-network send primitive, documented as such. Consistent with the existing `/api/*` posture.

## Testing

- Shared-impl unit tests (`tests/test_outreach/test_rpc.py`), MCP-tool bridge + connect-error tests (`tests/test_mcp/test_outreach_mcp.py`), Flask route tests (503 when pipeline absent, dispatch on success). 25 targeted tests green; ruff clean.
- Reviewed (architect pre-impl + code-reviewer on the diff): confirmed faithful extraction, correct wiring, no drift from the original tool bodies.

## Deploy

Dashboard route → picked up on `genesis-server` restart. MCP-side change → next CC session. Not a guardian host-deploy.


WingedGuardian:

---

PR #958, opened 2026-07-08
fix(memory): recover procedure extractions lost to provider exhaustion + storm alert (N1)

WingedGuardian:
## Context

An observed-harm evidence pass over production found that provider-exhaustion storms silently dropped internal-cognition work via the dead-letter queue. Two distinct gaps, both addressed here.

### 1. Procedure extractions permanently lost (the learning leak)

The whole-session procedure builder (`38_procedure_extraction`) runs **after** the extraction watermark advances, and `_call_and_parse_list` swallowed provider-exhaustion as an empty list. So when the builder hit "All providers exhausted": the failure was invisible, the watermark had already moved, and the next cycle skipped the session (`if not messages: continue`). The procedure — a reusable playbook — was lost for good (~174 dead-lettered over two months).

**Fix:** the builder now raises `ProcedureBuilderUnavailable` on call failure (distinct from a genuine no-procedures `[]`) and passes `suppress_dead_letter=True` (it owns its own retry; the DLQ replay only re-issued the transport and discarded the parsed procedures). On that failure the session is re-queued on the existing `deferred_work_queue` (`work_type=procedure_rebuild`); a drain pass at the top of each extraction cycle rebuilds from the full transcript. Idempotent via the `store_procedure_checked` novelty gate. Capped at 8 attempts; exhaustion or a rotated transcript discards and surfaces a `procedure_extraction_lost` observation, so the loss is visible, never silent. Reuses the `outreach_delivery` recovery pattern — no new schema.

### 2. Short-intense storms produced no alert (safety-net)

The existing DLQ alert is depth-based (`get_stuck_count`) and only counts an item once it ages past its TTL, so a short storm that self-heals or expires inside the TTL window fired no signal. Add a **rate-based** early warning: `_alert_dead_letter_storm` counts non-judge dead-letters enqueued in a rolling 15m window and pages once per storm.

- `count_recent` / `recent_optype_counts` CRUD + `idx_dead_letter_created` (migration 0049)
- distinct `source=dead_letter_storm` + hash prefix so it never collides with the accumulation alert's dedup/resolve; same band + cooldown + skip_if_duplicate throttle
- `chain_exhausted:judge` excluded (1h self-heal TTL, worthless-late)
- threshold 40/15m calibrated on the real storm timeline (normal <40 non-judge/15m, every observed storm >45)
- rides the existing critical-observations → notification channel

Safety-net for provider outages (e.g. a future model EOL). The depth alert is kept for sustained multi-hour backlogs.

## Tests

- storm alert: fires on rate, throttled same-band, resolves on drain, **judge burst excluded**, aged rows ignored, no-dup-rapid-fire, distinct-from-accumulation
- CRUD `count_recent` / `recent_optype_counts` (window + exclude-prefix); migration 0049 (upgrade + idempotent + no-op)
- builder raises on exhaustion / transport error and suppresses dead-letter
- rebuild: enqueue → drain-completes → counts; still-exhausted stays pending; transcript-gone discards; attempt-cap exhausts + observes; reference-only skips drain

## Follow-up

The single-extraction path (`extractor.py` via the retrospective-triage pipeline) shares the `38_procedure_extraction` call site and may have a sibling skip-on-failure. Tracked separately — it is a different pipeline with its own trigger semantics and was not traced end-to-end here.

### DIFF
--- a/src/genesis/db/crud/dead_letter.py
+++ b/src/genesis/db/crud/dead_letter.py
@@ -174,3 +174,47 @@ async def query_recent(
     db.row_factory = aiosqlite.Row
     cursor = await db.execute(sql, params)
     return [dict(r) for r in await cursor.fetchall()]
+
+
+async def count_recent(
+    db: aiosqlite.Connection,
+    *,
+    since: str,
+    exclude_prefix: str | None = None,
+) -> int:
+    """Count dead-letter rows enqueued at/after ``since`` (any status).
+
+    ``exclude_prefix`` drops operation_types starting with it — the rate-based
+    provider-exhaustion storm detector passes ``chain_exhausted:judge`` so a
+    self-healing 1h-TTL judge burst (worthless-late, drains itself) never counts
+    toward the storm trigger. Backed by ``idx_dead_letter_created``.
+    """
+    sql = "SELECT COUNT(*) FROM dead_letter WHERE created_at >= ?"
+    params: list = [since]
+    if exclude_prefix:
+        sql += " AND operation_type NOT LIKE ?"
+        params.append(exclude_prefix + "%")
+    cursor = await db.execute(sql, params)
+    row = await cursor.fetchone()
+    return int(row[0]) if row else 0
+
+
+async def recent_optype_counts(
+    db: aiosqlite.Connection,
+    *,
+    since: str,
+    exclude_prefix: str | None = None,
+) -> list[tuple[str, int]]:
+    """Per-operation_type counts of dead-letters enqueued at/after ``since``.
+
+    Feeds the storm alert's breakdown line (``light-reflection ×N``). Ordered by
+    count descending. ``exclude_prefix`` behaves as in :func:`count_recent`.
+    """
+    sql = "SELECT operation_type, COUNT(*) AS c FROM dead_letter WHERE created_at >= ?"
+    params: list = [since]
+    if exclude_prefix:
+        sql += " AND operation_type NOT LIKE ?"
+        params.append(exclude_prefix + "%")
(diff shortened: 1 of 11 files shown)


WingedGuardian:

PII/Secrets/Wording scan: **CLEAN** — no sensitive content, private identifiers, or problematic wording in diff or PR metadata.

---

WingedGuardian (commented):
## First-pass review

### Structural

**WARNING — import grouped incorrectly** in `tests/test_observability/test_queues_dead_letter.py`:

```python
# ── Rate-based storm alert ────────────────────────────
from datetime import UTC, datetime, timedelta  # noqa: E402
```

This import is appended mid-file after existing test code. The `# noqa: E402` suppresses ruff, but the standard is to keep all imports at the top of the file. Low friction to move — just add `timedelta` to the existing `datetime` import at the top of the file and drop the mid-section line.

No other structural violations found. All new `src/` files (`0049_dead_letter_created_index.py`) have corresponding test files. No bare `except:`, no mutable defaults, no `/tmp/`, no `task_states` direct inserts, no raw SQL outside CRUD/migration files, conventional commit prefix present.

### PII/Secrets/Wording

CLEAN — no secrets, PII, private IPs, or security-incident language in the diff or PR metadata.

---

WingedGuardian:
Addressed the code-review finding (shared-connection rollback): the builder/drain exception handlers no longer call `db.rollback()` on the shared `SerializedConnection`. It protected nothing of ours (procedure stores self-commit; the deferred `memory_events` write is committed by the watermark update before the builder runs) and a rollback there would discard concurrent jobs' pending writes. Removed all four calls (two pre-existing main-loop, two new in the drain) + a regression test. Fixed in f32ec4b5.

PR #960, opened 2026-07-08
fix(reflection): handle null scores in weekly_assessment parser

WingedGuardian:
## Problem

The `weekly_assessment` scheduled job has been failing for 5 days with:

```
TypeError: float() argument must be a string or a real number, not 'NoneType'
```

The reflection LLM legitimately emits an explicit `null` score for a dimension when it has `data_available: false` (no data that window). `float(None)` raises **`TypeError`**, not `ValueError`, so the parser's `except (ValueError, KeyError)` guard didn't catch it and the entire assessment parse aborted — taking down the cognitive-state maintenance pipeline.

## Fix

`src/genesis/reflection/output_router.py`:
- Add a module-level `_safe_float(value, default)` — returns the default for `None`/non-numeric input, catching both `TypeError` and `ValueError`.
- In `parse_weekly_assessment_output`, **skip** dimensions whose score is `null`/absent rather than coercing to `0.0`. A stored `0.0` would false-fire the learning-regression alarm in `stability.py`; a skipped dimension is simply not found by that name-keyed check, which correctly reads as "not enough data" instead of a regression.
- Route `overall_score` and the surplus-task `priority` through `_safe_float`; widen the dimension-loop `except` to include `TypeError`.

Null handling is now covered at the parser; the dashboard display side was already guarded separately.

## Verification

- `tests/test_reflection/test_output_router.py`: 9 new tests — null-score skip, all-null → empty dims (still a successful parse), missing-score-key skip, null `overall_score` → 0.0, non-numeric score coerced-and-kept, a production-shaped mixed payload, and null-priority handling on the deep-reflection path. Full file: **58 passed**, ruff clean.
- Replayed the **actual preserved failing payload** through the fixed parser: the old loop raises the exact production `TypeError`; the fixed parser succeeds, keeping the data-bearing dimensions and dropping the null ones.

## Post-merge

The daily scheduler fire self-heals within 24h (the weekly gate only advances on success). Watch `job_health` for the next `weekly_assessment` success and a fresh `self_assessment` observation + reflections markdown.

### DIFF
--- a/src/genesis/reflection/output_router.py
+++ b/src/genesis/reflection/output_router.py
@@ -31,6 +31,23 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+def _safe_float(value: object, default: float) -> float:
+    """Coerce ``value`` to float, returning ``default`` for null/non-numeric.
+
+    The reflection LLM legitimately emits an explicit ``null`` (e.g. a score or
+    priority it cannot assign). ``float(None)`` raises ``TypeError`` — not
+    ``ValueError`` — so a plain ``float(...)`` inside an ``except ValueError``
+    guard still crashes the whole parse. Route every model-supplied numeric
+    through here.
+    """
+    if value is None:
+        return default
+    try:
+        return float(value)  # type: ignore[arg-type]
+    except (TypeError, ValueError):
+        return default
+
+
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

PR #962, opened 2026-07-09
fix(metrics): un-blind outreach engagement grading (WS-1 A1)

WingedGuardian:
## What

Un-blinds outreach **engagement** measurement (WS-1 "A1" — the Bench foundation). Two bug classes, one canonical fix, no schema migration, outreach throttle deliberately untouched.

### Bug 1 — always-false grading (dead/incomplete positive set)
Consumers compared `engagement_outcome == "engaged"` or `IN ('useful','engaged')`, but the real positive vocabulary is `{useful, engaged, acted_on, acknowledged}` (a reply writes `useful`; the dashboard `/engage` endpoint writes `engaged`; behavioural signals write `acted_on`/`acknowledged`). Result: real replies were graded incorrect and engagement ratios undercounted.

The vocabulary had drifted across **six** independent consumers, each with its own incomplete copy. Rather than patch them individually (which is what let the drift happen), this defines one canonical set — `genesis.outreach.types.POSITIVE_ENGAGEMENT_OUTCOMES` — and routes every consumer through it:

- `calibration/reconciler.py` (prediction grading)
- `learning/signals/outreach_engagement.py` (awareness signal ratio)
- `feedback/harvest.py` (`_OUTREACH_MAP` omitted `engaged`)
- `observability/snapshots/outreach.py` (dashboard tile: missed `acted_on`/`acknowledged`)
- `db/crud/outreach.py` `get_engagement_stats` (morning report: `useful` only)
- `outreach/api.py` `/engagement` (rate: literal `engaged` only)

### Bug 2 — no-reply conflated with negative (WS-0 finding)
A 24h timeout writes `outcome='ignored', signal='timeout'` — the sole source of `ignored`. It was harvested as a **negative** learning signal and graded as a wrong prediction. The learning pipeline (harvest + reconciler) now treats no-reply (`ignored` + `timeout`) as **unknown** — skipped, not scored negative. An explicit dismissal via the `outreach_engagement` tool carries a non-`timeout` signal and is still a real negative.

The outreach spam-throttle (`governance.py`, counts `ignored`) is **not** changed — it keeps backing off on silence as outreach hygiene; only the *learning* interpretation changes.

## Why it matters (verified E2E on a copy of the live DB)
- **663 false-negative outreach signals** were sitting in the learning bus (no-reply treated as "user rejected"). Re-harvesting with the fix: **0 negative** (was 663), 196 neutral, 18 positive.
- `get_engagement_stats` engaged count: **18** (full positive set) vs the old `useful`-only **10**.

## Scope / deferrals
- **M12 (user-denied approvals)** → A2, where the intervention-rate metric consumes it (`rejected` already exists; `resolved_by` is free text — no schema change needed).
- **No-reply predictions accumulate unmatched** (reviewer finding #3): not a correctness regression — the calibration reader filters `outcome IS NOT NULL`, so they're correctly *excluded*, not mis-graded. A "resolved-as-unknown" state belongs with A2's calibration-reader rework.
- **The `engagement_outcome` CHECK is a no-op** (a `NULL` inside its `IN`-list defeats enforcement — verified). Tracked as a separate follow-up (needs a vocabulary decision + drifted-row reconciliation + table rebuild).

## Tests
Rewrote the tests that encoded the bug (they asserted the never-a-reply `engaged` vocabulary), added dashboard-`engaged` grading, no-reply-skip, explicit-dismissal-stays-negative, and full-positive-set coverage. Fixtures set `PRAGMA ignore_check_constraints=ON` for parity with the existing `test_harvest.py`. Targeted suites green locally.

Note: `test_health_mcp.py::TestBackupsEnabledHelper::{test_unset_everywhere,test_secrets_file_empty_value}` fail **identically on untouched `main`** in this environment (they detect a real local backup clone) — pre-existing and unrelated.

### DIFF
--- a/src/genesis/calibration/reconciler.py
+++ b/src/genesis/calibration/reconciler.py
@@ -7,6 +7,7 @@ import logging
 import aiosqlite
 
 from genesis.db.crud import predictions as pred_crud
+from genesis.outreach.types import POSITIVE_ENGAGEMENT_OUTCOMES
 
 logger = logging.getLogger(__name__)
 
@@ -40,10 +41,11 @@ class PredictionReconciler:
                 continue
             if outcome == "ambivalent":
                 continue  # neutral implicit-activity signal — not gradeable
-            # Positive engagement set (mirrors feedback.harvest._OUTREACH_MAP).
-            # Was `outcome == "engaged"`, a value never written to the column,
-            # so every outreach prediction was graded incorrect unconditionally.
-            correct = outcome in ("useful", "acted_on", "acknowledged")
+            # Was `outcome == "engaged"` alone. 'engaged' IS written (dashboard
+            # /engage endpoint) but a real reply writes 'useful' — so every
+            # reply-engaged prediction was graded incorrect. Use the canonical
+            # positive set so no real engagement value is missed.
+            correct = outcome in POSITIVE_ENGAGEMENT_OUTCOMES
             await pred_crud.record_outcome(
                 self._db, pred["id"], outcome=outcome, correct=correct,
             )
--- a/src/genesis/calibration/reconciler.py
+++ b/src/genesis/calibration/reconciler.py
@@ -23,7 +23,7 @@ class PredictionReconciler:
         for pred in unmatched:
             action_id = pred["action_id"]
             cursor = await self._db.execute(
-                "SELECT engagement_outcome FROM outreach_history "
+                "SELECT engagement_outcome, engagement_signal FROM outreach_history "
                 "WHERE id = ? AND engagement_outcome IS NOT NULL",
                 (action_id,),
             )
@@ -31,7 +31,19 @@ class PredictionReconciler:
             if not row:
                 continue
             outcome = row[0] if isinstance(row, tuple) else row["engagement_outcome"]
-            correct = outcome == "engaged"
+            signal = row[1] if isinstance(row, tuple) else row["engagement_signal"]
+            # No-reply (the 24h timeout) is not a graded outcome: silence is
+            # unresolved, not a wrong prediction (WS-0: "ignored" != no value).
+            # An explicit dismissal via the outreach_engagement tool carries a
+            # non-'timeout' signal and is still graded below.
+            if outcome == "ignored" and signal == "timeout":
+                continue
+            if outcome == "ambivalent":
+                continue  # neutral implicit-activity signal — not gradeable
+            # Positive engagement set (mirrors feedback.harvest._OUTREACH_MAP).
+            # Was `outcome == "engaged"`, a value never written to the column,
+            # so every outreach prediction was graded incorrect unconditionally.
+            correct = outcome in ("useful", "acted_on", "acknowledged")
             await pred_crud.record_outcome(
                 self._db, pred["id"], outcome=outcome, correct=correct,
             )
--- a/src/genesis/db/crud/outreach.py
+++ b/src/genesis/db/crud/outreach.py
@@ -4,6 +4,11 @@ from __future__ import annotations
 
 import aiosqlite
 
+from genesis.outreach.types import POSITIVE_ENGAGEMENT_OUTCOMES
+
+# SQL IN-list of positive engagement values (trusted constants, safe to inline).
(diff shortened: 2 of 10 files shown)


WingedGuardian:
## First-pass review: structural + PII/wording

### Structural: CLEAN (3 notes)

**NOTE 1 — Missing test for `ambivalent` skip (`src/genesis/calibration/reconciler.py:38`)**

The new `if outcome == "ambivalent": continue` branch has no test in `tests/test_calibration/test_reconciler.py`. The no-reply-timeout skip and the explicit-dismissal case are both tested; the ambivalent skip is not. Low risk (neutral outcomes aren't harmful to leave ungraded), but it's a new code path with no coverage. Suggest adding a `test_reconcile_outreach_ambivalent_is_skipped` analogous to `test_reconcile_outreach_noreply_is_skipped`.

---

**NOTE 2 — Manual sync between `_OUTREACH_MAP` and `POSITIVE_ENGAGEMENT_OUTCOMES` (`src/genesis/feedback/harvest.py:77`)**

The comment correctly calls out that positive keys must stay in sync, and they are currently (all four values appear in both sets). But there's no enforcement: if a value is added to `POSITIVE_ENGAGEMENT_OUTCOMES` without a matching `_OUTREACH_MAP` entry, harvest will silently drop it (`mapping is None → continue`). A one-line guard in the test suite would enforce this:

```python
from genesis.feedback.harvest import _OUTREACH_MAP
from genesis.outreach.types import POSITIVE_ENGAGEMENT_OUTCOMES

def test_outreach_map_covers_all_positive_outcomes():
    assert POSITIVE_ENGAGEMENT_OUTCOMES.issubset(_OUTREACH_MAP), (
        "All POSITIVE_ENGAGEMENT_OUTCOMES must have an _OUTREACH_MAP entry "
        "or they will be silently dropped from harvest."
    )
```

---

**NOTE 3 — Duplicate `_POSITIVE_IN` computation (`src/genesis/db/crud/outreach.py:8` and `src/genesis/observability/snapshots/outreach.py:26`)**

The identical `", ".join(f"'{o}'" for o in sorted(POSITIVE_ENGAGEMENT_OUTCOMES))` expression is evaluated independently in two modules. Both derive from the same frozenset so there's no functional divergence risk today, but it's minor duplication. Not worth a refactor on its own; just noting for awareness if a third consumer appears.

---

No secrets, tokens, private IPs, SSH material, personal email addresses, absolute paths with usernames, cloud account IDs, or private hostnames in added lines, PR title, or PR body.

---

*Structural review complete — no ERRORs or WARNINGs. Three notes above.*

---

WingedGuardian:
## First-pass review

### Structural

**WARNING — `_OUTREACH_MAP` / `POSITIVE_ENGAGEMENT_OUTCOMES` sync gap (`src/genesis/feedback/harvest.py`)**

The PR keeps both in sync and the comment calls out the coupling — but there's no CI enforcement. `_OUTREACH_MAP` in `harvest.py` and `POSITIVE_ENGAGEMENT_OUTCOMES` in `outreach/types.py` independently enumerate the same four positive vocabulary values. If a new positive outcome is added to the canonical set in the future, `harvest.py` will silently skip it (its key is absent from `_OUTREACH_MAP`, so `mapping is None` → `continue`). Every other consumer would count it as positive; `harvest.py` would not — a quiet divergence.

Consider adding an assertion to the test suite to catch drift at CI time, e.g.:
```python
from genesis.outreach.types import POSITIVE_ENGAGEMENT_OUTCOMES
from genesis.feedback.harvest import _OUTREACH_MAP

def test_harvest_map_covers_positive_outcomes():
    map_positives = {k for k, (_, pol, _) in _OUTREACH_MAP.items() if pol == "positive"}
    assert POSITIVE_ENGAGEMENT_OUTCOMES == map_positives
```

---

**NOTE — f-string SQL with inlined constants (`src/genesis/db/crud/outreach.py`, `src/genesis/observability/snapshots/outreach.py`)**

`_POSITIVE_IN` is spliced into f-string SQL rather than using `?` binding. Both sites carry a defensive comment ("trusted constants, safe to inline") and the values are module-level frozenset literals — no injection risk today. Noting it so the pattern isn't normalized for future cases where values might not be constants.

---

No credentials, private IPs, personal identifiers, or incident-disclosure language in added lines or the PR body. The "verified E2E on a copy of the live DB" phrasing describes a dev-testing practice; no data is exposed.

---

PR #964, opened 2026-07-09
feat(dashboard): P3 nav/layout — wrap tabs, clear footer, Comms, icon a11y

WingedGuardian:
## What

Dashboard roadmap **P3 (Navigation & layout)** — pure template/CSS, no JS/route/hash changes. Four fixes, all grounded against a live 1280×720 measurement.

- **Tab bar overflow (core):** `.genesis-tabs` gets `flex-wrap: wrap` + a small row-gap so the 15 tabs wrap to a second row instead of clipping. Previously the flat `nowrap` bar overflowed its container and caused a **page-wide horizontal scroll**, with "Backup & Updates" clipped off the right edge at ≤1280px.
- **Pause-footer overlap:** `.genesis-dashboard` bottom padding `24px → 56px` to clear the fixed **36px** kill-switch footer that was overlapping the last card (Container card on Overview).
- **"Chat" → "Comms" label:** that tab is the comms hub (terminal + approvals + proposals + outreach). **Label text only** — the hash key `chat` is unchanged (it's referenced in 8+ call sites), so deep links and routing are untouched.
- **Icon a11y leak:** `aria-hidden="true"` on the decorative `material-symbols` icons in the chrome header, so ligature text (`neurology`/`terminal`/`warning`) no longer leaks into accessible names.

## Verification (live, at 1280×720)

Rules injected into the running dashboard and re-measured (nothing booted from the worktree):

| | before → after |
|---|---|
| page horizontal scroll | bodyScrollW 1318 → **1265** (= clientW) — gone |
| tab-bar clip | scrollW 1294 → **1217** (= clientW); wraps, "Backup & Updates" visible |
| footer clearance | padding-bottom 24px → **56px** (footer top y=684) |
| heading accessible name | "neurology Genesis Dashboard…" → **"Genesis Dashboard Restart Server Terminal"** |
| nav button | "Chat" → **"Comms"** |

No test asserts on the changed strings and nothing parses `header.html`/`dashboard.css`, so there's no unit coverage for these static assets — verified live instead. Definitive re-check at 1280 **and** 1024px will follow post-deploy.

## Scope

- The 07-06 audit's "dead left rail (~230px)" was **verified not-a-bug** and dropped — it's the `max-width:1400px; margin:0 auto` container's symmetric centering at >1400px viewports, not a defect.
- No guardian/host paths touched → no Host-Deploy Gate. Deploy is code-only (`systemctl --user restart genesis-server`).

### DIFF
--- a/src/genesis/dashboard/templates/partials/chrome/header.html
+++ b/src/genesis/dashboard/templates/partials/chrome/header.html
@@ -11,7 +11,7 @@
 
     <!-- Error banner -->
     <div class="error-banner" x-show="$store.genesisDashboard.error" x-transition>
-      <span class="material-symbols-outlined">warning</span>
+      <span class="material-symbols-outlined" aria-hidden="true">warning</span>
       <span x-text="$store.genesisDashboard.error"></span>
     </div>
 
@@ -34,7 +34,7 @@
 
     <!-- Header -->
     <h1>
-      <span class="material-symbols-outlined">neurology</span>
+      <span class="material-symbols-outlined" aria-hidden="true">neurology</span>
       Genesis Dashboard
       <span style="margin-left: auto; display: flex; gap: 0.5rem; align-items: center;">
         <button @click="$store.genesisDashboard.restartBridge()"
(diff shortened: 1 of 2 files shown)


WingedGuardian:
Structural review: CLEAN — template/CSS-only change; no Python structural patterns apply.

---

PR #965, opened 2026-07-09
fix(guardian): fall back to a writable work_dir so a misconfigured cc.work_dir can't blind the recovery brain

WingedGuardian:
## What

`diagnosis.py` created the CC diagnosis work dir with a bare
`work_dir.mkdir(parents=True, exist_ok=True)` on the configured `cc.work_dir`.
On a vintage or misconfigured install that path can be **uncreatable** — e.g. a
root-owned `/var/lib/guardian-snapshots` on an install predating the setup
script's mkdir/chown. The mkdir then raised mid-tick and dropped the Guardian to
**alert-only** instead of running its intelligent diagnosis — the exact failure
observed live during a freeze (Guardian could detect + alert but not self-heal).

## Fix

Extract `_ensure_work_dir(configured, fallback)`:
1. try to create the configured dir → use it;
2. on `OSError`, log and fall back to `~/.local/state/genesis-guardian/cc-sessions`
   (always writable by the guardian user) → use that;
3. if the fallback is *also* uncreatable, the `OSError` propagates — with no
   writable work dir anywhere, alert-only is the honest degradation (we never
   silently return an unusable path).

## Scope note

This is the last genuinely-unbuilt sliver of the planned "F-detect" workstream.
The larger F-detect pieces (deploy/build **drift detection**, guardian-down
heartbeat recovery) already shipped via container-side `guardian/watchdog.py`
(+ #940), so this PR deliberately builds only the remaining gap rather than
duplicating that.

## Tests

`tests/test_guardian/test_diagnosis.py` — configured-dir-used, fallback-on-
uncreatable (verifies the configured path is never created and the fallback is),
and propagate-when-fallback-also-uncreatable. 25 file tests green; ruff clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **The Guardian's recovery brain no longer goes dark when its work directory is
+  misconfigured.** On some installs the Guardian's configured CC work dir points
+  at a path it can't create (e.g. a root-owned dir on an install that predates
+  the setup script) — which used to make its intelligent diagnosis fail mid-
+  incident and silently drop it to alert-only. It now falls back to a writable
+  state dir so the self-healing path keeps working.
+
 - **Asking Genesis to grow a disk or wait on your reply now works from a Claude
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

PR #966, opened 2026-07-09
feat(eval): WS-1 A2 — first-class metric series (precision@3, approvals, goals, noise)

WingedGuardian:
## What

Adds honesty-first metric series to the J-9 weekly eval pipeline (WS-1 A2 of the gap-closure campaign). **Zero schema migrations** — all snapshot-only.

### Memory dimension (extended)
- **`precision_at_3`**: top-3 of the rank-sorted judged memories per recall; only recalls where every judged memory carries a real `rank` qualify (unranked pre-fix events would make the slice arbitrary); denominator `min(3, judged)`; coverage visible via `precision_at_3_recalls`.
- **Defensive top-5 truncation**: j9_batch judges `memory_ids[:5]`, so this is a no-op on well-formed data, but a malformed emitter can no longer inflate the `@5` denominator. `precision_at_5` semantics unchanged (historical series preserved).
- **`judge_prompt_versions`**: the relevance judge is a raw prompt (not a registry rubric) — the prompt is now versioned, stamped into every event, and the versions seen in-window are reported so a judge change reads as a series break instead of silent drift.

### New snapshot dimensions
- **`approvals`** — gate-throughput buckets over `approval_requests`. Deliberately NOT labeled "intervention rate": this table measures how designed approval gates get resolved, not corrective intervention. Resolver classes come from ONE canonical `classify_resolver` (new, in `db/crud/approval_requests.py`); unknown `resolved_by` values surface as `unknown_resolver_*` keys, never silently bucketed. `datetime()` windowing handles the mixed space-format/ISO-T timestamps (both UTC). Includes the M12 scaffold `user_denied_count` (zero live instances to date — documented as unvalidated).
- **`goals`** — `user_goals` status ratios. `completion_rate` is **null on 0/0** (all 8 live goals are active — never a fake 0%), achieved/abandoned always reported separately. Ex-ante `success_criteria` deliberately deferred (needs writer/consumer plumbing — tracked as a follow-up).
- **`noise`** — loop-closure leak gauges (documented as all-time stocks, not flows) + windowed empty-ego-cycle % (opportunity denominator = cycles run) + rejected/withdrawn/tabled proposal buckets + realist-verdict counts. Raw buckets only, no composite score.

### Dashboard honesty fix
Trend arrows previously colored "up" green for every dimension — a rising noise metric would render as improvement. The arrow glyph stays factual (direction of the value); its **color** is now driven by a per-dimension valence map (`noise` lower-is-better; `approvals` direction-ambiguous → neutral gray).

## Guards
- Registration test: `run_weekly_aggregation` swallows per-dimension failures silently, so a test asserts every registered dimension produces a snapshot on a fresh DB.
- Resolver drift test: pins every in-tree `resolved_by` writer literal + the live value inventory; new writers must extend it.
- Headline/valence map coverage tests keep the dashboard in lockstep with the aggregator.

## Verification
- 285 targeted tests pass (`tests/test_eval/`, resolver classes, dashboard route maps, loop-closure, schema); ruff clean on the full diff.
- **E2E against a copy of the live production DB**: all dimensions produced with no silent failures; `precision_at_3` 0.7193 vs `precision_at_5` 0.7149 on 57 fully-ranked recalls; goals correctly null; approvals buckets internally consistent.
- Subsystem-map guard clean; CHANGELOG + `docs/architecture/CURRENT.md` updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,18 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **Three new honesty-first metric series on the dashboard's compounding
+  panel, plus retrieval precision@3.** The weekly eval now tracks how your
+  approval gates actually get resolved (by you vs. auto-expired vs.
+  fail-closed cancels — unknown resolver values are surfaced, never silently
+  bucketed), goal completion (reports "no data yet" instead of a fake 0%
+  until goals actually close), and a noise/passivity view (stale follow-ups,
+  empty ego cycles, rejected-proposal clusters). Memory retrieval gains a
+  precision@3 series alongside precision@5, with judge changes marked as
+  series breaks. Trend arrows are now honest too: the arrow shows which way a
+  metric moved, while its color shows whether that's an improvement — a
+  rising noise metric no longer glows green.
+
 - **Genesis can now grow this VM's disk or RAM from the hypervisor — with your
   approval — to fix the one storage failure nothing else could.** When the
   container's storage pool has no room left to auto-expand into (the structural
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -360,7 +360,7 @@ Self-improvement loops and the instrumentation that keeps them honest.
 ```yaml subsystem-map
 entry: learning-evaluation
 modules: [learning, eval, experimentation, feedback, calibration]
-verified: 9037d45b 2026-07-07
+verified: b2fde2bb 2026-07-09
 ```
 
 - **learning/** is the de-facto cron host: `rt._learning_scheduler` registers
@@ -375,7 +375,15 @@ verified: 9037d45b 2026-07-07
   "cannot break production" contract — hooks must never raise) + weekly Sunday
   aggregation (hard 7-day window) + an on-demand batch judge as a surplus
   task. The model gauntlet is weekly but OFF by default (paid inference) and
-  NEVER auto-mutates the roster.
+  NEVER auto-mutates the roster. Nine snapshot dimensions: the original five
+  (memory/system/ego/cognitive/procedure) + cognitive_drift + three
+  snapshot-only WS-1 A2 series (approvals gate-throughput, goals scaffold,
+  noise/passivity). `run_weekly_aggregation` swallows per-dimension failures
+  silently — the registration test in `tests/test_eval/test_j9_eval.py` is
+  the guard; keep it in sync when adding a dimension. Resolver-origin
+  classification for the approvals series is canonical in
+  `db/crud/approval_requests.py classify_resolver` (free-text convention —
+  a new `resolved_by` writer must extend the prefix tuples + drift test).
 - **experimentation/**: Crucible A/B + Evo fan-out — on-demand via MCP tools
   only; **recommend-only is the safety invariant** (no autonomous promotion,
   no live-cognition writes; Bonferroni + held-out re-validation).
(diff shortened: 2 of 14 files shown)


WingedGuardian:

---

WingedGuardian:
**Structural review: 2 WARNINGs** — no ERRORs, no hard blocks

---

### WARNING — Raw SQL in new compute functions (`src/genesis/eval/j9_aggregator.py`)

`_compute_approvals`, `_compute_goal_completion`, and `_compute_noise_passivity` all issue raw `db.execute()` queries against `approval_requests`, `user_goals`, `ego_cycle_outcomes`, and `ego_proposals` instead of going through CRUD modules. The business-logic delegation is correctly done (`classify_resolver`, `loop_closure.followup_funnel/observation_funnel/proposal_funnel`), but the analytical SELECT queries bypass the CRUD layer.

This pattern is pre-existing throughout all six other dimension functions in the file, so this PR isn't introducing a new anti-pattern — it's extending one. Flagging per rule; resolution is an architectural decision about whether the analytics aggregator warrants a documented exemption (similar to migrations).

---

### WARNING — No test update for modified MCP health endpoint (`src/genesis/mcp/health/j9_eval.py`)

`_impl_j9_eval_status` was extended to include `"approvals"`, `"goals"`, and `"noise"` in the snapshot retrieval loop. The aggregator-level `test_run_weekly_aggregation_writes_new_dimensions` confirms the snapshots are written, but the MCP health endpoint path itself has no test asserting the three new dimensions appear in the endpoint response. Low risk (the change is a mechanical list extension), but worth a targeted test for the endpoint's output shape.</body>
</invoke>

---

WingedGuardian:
Structural review: 1 WARNING, 1 NOTE

**WARNING — Raw SQL** (`src/genesis/eval/j9_aggregator.py`): The three new aggregator functions (`_compute_approvals`, `_compute_goal_completion`, `_compute_noise_passivity`) query `approval_requests`, `user_goals`, `follow_ups`, `observations`, `ego_proposals`, and `ego_cycle_outcomes` via raw `db.execute()` rather than `db.crud.*` modules. Per project conventions this is a pattern violation (ERROR severity), but it is consistent with every pre-existing compute function in this module — this file has used raw SQL throughout. Not requesting a block on this PR, but flagging for team awareness: if the project wants to bring `j9_aggregator.py` into compliance, it needs to happen across the whole module, not just the new functions.

**NOTE** (`src/genesis/eval/j9_aggregator.py`): `_NOISE_STALE_DAYS = 14` duplicates `_STALE_DAYS` from `mcp/health/loop_closure_status.py` (called out by an inline comment in the diff). Non-blocking — just noting so it's on the radar for a future consolidation pass.

---

WingedGuardian:
Review findings — dispositions (commit e8268630):

- **Codex P2 (decision windowing)**: FIXED — rejected/withdrawn/tabled buckets now window on `resolved_at` (verified: the reject/table/withdraw CRUD transitions set it; live coverage 174/176 on terminal proposals). Live-DB E2E confirms the fix catches a boundary-straddling rejection the old query missed. Regression test added.
- **Codex P2 (updated_at as abandonment time)**: FIXED by removal — `abandoned_in_period` is dropped rather than mislabeled; `user_goals` has no abandonment timestamp, so the honest signal is the `abandoned_count` stock (weekly deltas in the series). Noted in the fn docstring; an `abandoned_at` column can ride the deferred `success_criteria` migration if/when that lands.
- **Structural WARNING (raw SQL)**: ACCEPTED — consistent with every pre-existing compute fn in this module, as the review itself notes. Whole-module CRUD-compliance is an architectural decision for its own pass; business logic (resolver classification, funnels) is already delegated to crud modules.
- **Structural WARNING (MCP endpoint test)**: FIXED — `test_j9_eval_status_reports_new_snapshot_dimensions` asserts the three dims appear in the endpoint response.
- **Structural NOTE (duplicated stale-days constant)**: FIXED — `STALE_DAYS` consolidated in `db/crud/loop_closure.py`; both consumers import it.

PR #969, opened 2026-07-09
fix(autonomy): protect systemd unit templates, drop stale watchdog config duplicates

WingedGuardian:
## Problem

Two related defects around the watchdog systemd unit:

1. **Stale duplicates.** `config/genesis-watchdog.{service,timer}` are dead copies — live units are rendered from `scripts/systemd/*.template` (install.sh Step 7 / bootstrap.sh), and nothing reads the `config/` pair. They also carry wrong content (60s cadence vs the real 300s, `After=genesis-bridge`, a `TMPDIR` override, no hardening directives), so they actively mislead.
2. **Misdirected protection (the real bug).** `config/protected_paths.yaml` marked the dead `config/genesis-watchdog.*` path CRITICAL while the *actual* unit sources — `scripts/systemd/*.template` — were unprotected: the generic `*.service`/`*.timer` patterns don't match the `.template` suffix. A relay-channel edit to a template would become the live unit on the next render, bypassing the protection the generic patterns exist to provide. Fixed for **all** unit templates, not just watchdog — the suffix escape applies to every one.

## Changes

- Delete `config/genesis-watchdog.service` + `config/genesis-watchdog.timer` (verified orphans: no installer/script reads them; a live install's unit is provably template-derived)
- Repoint the CRITICAL pattern to `scripts/systemd/*.template`
- Update coupled tests: `test_protection.py` (fixture + real-yaml classify assertions), `test_pretool_check.py` (blocked-path, matcher, and config-completeness cases)
- Fix stale docs: the migration reference implied the `config/` units are live; the self-healing design doc + `watchdog_runner.py` docstring carried the dead 60s cadence

## Verification

- 118 tests pass across both touched test files; ruff clean
- Live hook subprocess E2E: template path → exit 2 BLOCKED, normal path → exit 0
- `ProtectedPathRegistry.from_yaml` truth-table against the edited yaml: all templates CRITICAL, normal paths NORMAL
- `config/genesis-tmp-watchgod.service` deliberately untouched — it IS a live source (install.sh copies it directly)
- No protection lost by removing the old pattern: fnmatch `*` crosses `/`, so the generic `*.service`/`*.timer` rules already covered the deleted paths


WingedGuardian:

---

WingedGuardian:
Codex P2 (fallback list missing the template rule) — **fixed in `db9aca80`**: `scripts/systemd/*.template` added to `_FALLBACK_CRITICAL` and pinned in `test_fallback_critical_contents`. All checks green on the updated head.

WingedGuardian:
@codex review

PR #970, opened 2026-07-09
fix(mcp): resolve short id: handles in memory_expand instead of silent []

WingedGuardian:
## Problem

The proactive memory hook surfaces memories as \`id:<8-char>\` handles, and the documented contract (CLAUDE.md L2 layer) tells sessions to pass those handles to \`memory_expand\`. But the tool passed IDs raw to Qdrant \`retrieve\` — which requires full-UUID point IDs — so every handle-based expand silently returned \`[]\`, with no diagnostic. Found live while verifying an unrelated plan's target memory.

Second defect in the same path: when *nothing* resolved, the early-return skipped the \`not_found\` trailer entirely — the silent-empty case gave callers no signal anything went wrong.

## Fix

- \`_resolve_id_prefixes\`: strips an optional \`id:\` prefix (case-insensitive), and resolves hex-ish handles (4–35 hex/dash chars) against \`memory_metadata\` via a parameterized \`LIKE\` (LIMIT 2). Unique match → full UUID. Ambiguous → reported, never guessed. Unknown → passes through into \`not_found\`. DB errors fail open (identical to pre-fix behavior).
- Non-hex short IDs (e.g. literal test IDs) bypass resolution entirely — no DB call.
- Empty results now return \`[{"not_found": [...], "ambiguous": [...]}]\` instead of a bare \`[]\`.
- Zero writes introduced; the resolver is a single SELECT.

## Verification

- 6 new tests (prefix resolve, \`ID:\` case variant, ambiguous, silent-empty, non-hex bypass, DB-error fail-open); 27/27 in the file, plus the two other memory_expand-touching test files green (12/12).
- E2E against live stores: 8-char handle → full UUID → real Qdrant point retrieved; a real colliding 4-char prefix correctly reported ambiguous.
- No non-test Python callers pattern-match the old \`[]\` shape (searched; the consumers are LLM sessions via MCP).

## Notes

MCP servers load code at CC session start — the fix takes effect for new sessions after merge; verified post-merge by expanding a handle in a fresh session.

### DIFF
--- a/src/genesis/db/crud/memory.py
+++ b/src/genesis/db/crud/memory.py
@@ -339,6 +339,26 @@ async def get_metadata(
     }
 
 
+async def match_id_prefix(
+    db: aiosqlite.Connection,
+    prefix: str,
+    *,
+    limit: int = 2,
+) -> list[str]:
+    """Memory IDs starting with *prefix* (e.g. an 8-char ``id:`` handle).
+
+    ``limit=2`` lets callers distinguish unique from ambiguous without
+    counting every match. Parameterized LIKE; callers are expected to
+    pre-validate the prefix shape (hex/dash).
+    """
+    rows = await db.execute_fetchall(
+        "SELECT memory_id FROM memory_metadata"
+        " WHERE memory_id LIKE ? || '%' LIMIT ?",
+        (prefix, limit),
+    )
+    return [str(r[0]) for r in rows]
+
+
 async def delete_metadata(db: aiosqlite.Connection, *, memory_id: str) -> bool:
     """Delete a memory_metadata row. Returns True if deleted."""
     cursor = await db.execute(
--- a/src/genesis/mcp/memory/core.py
+++ b/src/genesis/mcp/memory/core.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+import re
 from dataclasses import asdict
 from datetime import UTC, datetime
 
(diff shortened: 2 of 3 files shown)


WingedGuardian:
## First-pass review — PR #970

**Structural review:** 1 WARNING (no ERRORs)

---

### WARNING — Raw SQL in `_resolve_id_prefixes` outside the crud layer

**File:** `src/genesis/mcp/memory/core.py` (new function `_resolve_id_prefixes`)

```python
    "SELECT memory_id FROM memory_metadata"
    " WHERE memory_id LIKE ? || '%' LIMIT 2",
    (mid,),
)
```

The convention is to route reads against genesis.db through `genesis.db.crud` modules rather than inline SQL (exceptions: migrations). The query itself is fine — read-only, parameterized, no injection risk — but it belongs in a crud helper (e.g. `memory_crud.resolve_id_prefix(db, prefix)`) and called from here. If this table's crud layer doesn't have a prefix-lookup yet, adding one closes the pattern gap.

---

WingedGuardian:
Addressed the structural WARNING: the prefix lookup now lives in the crud layer as `genesis.db.crud.memory.match_id_prefix(db, prefix, limit=2)`; `_resolve_id_prefixes` calls it and keeps only the ambiguity/fail-open semantics. Tests unchanged and green (27/27 — they exercise the same `db.execute_fetchall` seam through the helper).

PR #973, opened 2026-07-09
fix(cc): truthful terminal states for crashed/cancelled sessions (T2-B reconcile)

WingedGuardian:
## Summary

Crashed or cancelled background sessions read as SUCCESSES: they lingered `active` until the 6h `session_reaper` cron bulk-relabeled them `completed` — and J-9 counts only `status='completed'` as success (session_success_pct and the cognitive-loop rate). Full reconcile of the three overlapping defects:

- **CancelledError handler** in `DirectSessionRunner._run_session` — `CancelledError` is a `BaseException`, so the existing `except Exception` failure path never saw it. Cancels now record `failed` (reason=cancelled), feed the outcome back to ego proposals (matching the generic failure path), and re-raise.
- **Reaper reconcile** — the `session_reaper` job now routes through `SessionManager.cleanup_stale` (same 6h cutoff, verified bit-for-bit equivalent): policy-aware sweep of stale non-foreground `active` rows → `expired` (outcome UNKNOWN — the process is gone; `failed` would over-claim exactly like `completed` over-claimed), foreground never swept, session end-hooks fired. This wires the previously caller-less `cleanup_stale` and deletes the misleading crud `reap_stale`.
- **Boot-time sweep** — orphans from a crashed process no longer wait up to ~6h: an immediate `tracked_task` kick runs the reaper at startup (init order verified: cc_relay sets `rt._session_manager` before learning init).
- **Review P2 (fixed in-PR)** — nothing ever cancelled in-flight session tasks during runtime shutdown; asyncio's teardown delivered CancelledError only AFTER the DB closed, so the handler's writes no-oped on the most common trigger (`systemctl restart`). `GenesisRuntime.shutdown()` now cancel-and-awaits the runner's tasks (`DirectSessionRunner.shutdown()`, 10s grace — bounded because shutdown runs under systemd's TimeoutStopSec hard deadline) BEFORE closing the DB, with an ordering test proving the DB is still open when it runs.

## Verification

- TDD with empirical RED for every behavior change (cancel→failed; end-hooks on sweep; cutoff respected; shutdown persistence ×3).
- 53 tests green (direct_session, session_manager, cc_sessions crud, runtime shutdown); ruff clean; subsystem-map guard clean.
- Independent code review verified the risky assumptions with evidence: the boot sweep cannot catch a legitimately-running session (systemd cgroup kill covers CC children; max session timeout 1h vs 6h cutoff; background runs never update `last_activity_at`), and no `cc_sessions.status` consumer breaks on `completed`→`expired` (J-9, capability aggregator, dashboard — which already has an "Expired" filter — and the extraction job's source-tag allowlist all enumerated).

## Notes

- Regression markers: journal shows `initial_session_reap` at boot and `Cancelled N in-flight direct sessions` on restarts with sessions in flight; no `active` rows older than 6h; success metrics may dip slightly as crashes stop counting as wins (that's the fix working).
- CHANGELOG (user-visible entry) + `docs/architecture/CURRENT.md` execution-cc entry refreshed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   included knowledge sources could only surface knowledge via vector
   similarity — exact-phrase and keyword matches on ingested docs were
   silently dropped. It now searches every requested collection.
+
+- **Background sessions that crash or get cancelled no longer show up as
+  "completed."** A session interrupted mid-run used to sit "active" until a
+  cleanup job quietly relabeled it completed — so the dashboard and
+  Genesis's own success metrics counted crashes as wins. Cancelled sessions
+  are now recorded as failed on the spot, orphaned ones are marked expired
+  (outcome unknown) at boot and every 6 hours, and only genuinely finished
+  sessions read as completed.
 - **Asking Genesis to grow a disk or wait on your reply now works from a Claude
   Code session, not only from inside the running server.** The two tools that
   block on your Telegram reply (`provision_grow`, `outreach_send_and_wait`) need
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -102,9 +102,11 @@ verified: e01e0c49 2026-07-09
   + a boot-time kick) routes through `SessionManager.cleanup_stale` —
   stale non-foreground 'active' rows → `expired` (outcome unknown),
   end-hooks fired. Known interruptions record `failed`: `_run_session` has
-  an explicit `CancelledError` handler (T2-B 2026-07-09; the old crud
-  `reap_stale`, which relabeled orphans 'completed', is deleted). J-9
-  counts only `completed` as success.
+  an explicit `CancelledError` handler, and `GenesisRuntime.shutdown()`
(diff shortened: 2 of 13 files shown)


WingedGuardian:

---

WingedGuardian:

---

WingedGuardian:

---

**NOTE** — Task-reference prefixes in inline comments

The new code embeds `T2-B:` task-ID tags in several inline comments and one docstring:

- `src/genesis/db/crud/cc_sessions.py` — tombstone comment replacing `reap_stale`
- `src/genesis/cc/direct_session.py` — `CancelledError` handler comment + `shutdown()` docstring names its caller (`GenesisRuntime.shutdown()`)
- `src/genesis/runtime/_core.py` — shutdown block comment
- `src/genesis/runtime/init/learning.py` — `_reap_stale_sessions` rewrite comment
- `tests/test_db/test_cc_sessions.py` — tombstone comment replacing the removed tests

Per project conventions: "Don't reference the current task, fix, or callers… since those belong in the PR description and rot as the codebase evolves." The *substance* of each comment (BaseException-not-caught-by-Exception invariant, DB-close ordering constraint, reap_stale removal rationale) is the kind of non-obvious WHY that should stay. Consider stripping the `T2-B:` prefix and the caller call-out in the `shutdown()` docstring, leaving only the invariant explanation.

---

PR #975, opened 2026-07-10
fix(qdrant): payload indexes for filtered search recall

WingedGuardian:
## Problem (found 2026-07-09, during WS-C ambient-layer E2E)

\`episodic_memory\` (48.7K points) has **zero payload indexes**, and every production vector search is filtered — at minimum the unconditional \`deprecated\` must_not. Filtered HNSW without payload indexes measurably **drops valid results**: exact-vs-approximate comparisons of the same query+filter returned different result sets (an internally inconsistent store — a point scoring 0.734 was absent from a filtered top-100 whose floor was far below that). Production recall quality has been silently degraded across the board.

## Fix

- \`ensure_payload_indexes()\`: idempotent bool/keyword index creation for every filter-used payload field (\`deprecated\`, \`tags\`, \`wing\`, \`room\`, \`life_domain\`, \`project_type\`, \`source_subsystem\`, \`source_type\`, \`source_pipeline\`, \`memory_class\`) on both collections, wired into \`ensure_collections()\` so server init keeps them in place. Per-field try/except: one failure can't block init.
- \`scripts/apply_qdrant_payload_indexes.py\`: one-off online apply (non-destructive; Qdrant builds payload indexes without blocking reads/writes) + resulting schema report — for applying without waiting for a server restart.

## Verification

- 3 new tests: full field×collection coverage, one-failure-doesn't-block, ensure_collections wiring. 13/13 in the file.
- The apply step runs only on explicit approval; before/after recall comparison on the same probe queries will be posted after applying.

### DIFF
--- a/scripts/apply_qdrant_payload_indexes.py
+++ b/scripts/apply_qdrant_payload_indexes.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python3
+"""One-off: create payload indexes on the live Qdrant collections.
+
+The server creates them at next init via ensure_collections(); this
+script applies them immediately (online, non-destructive — Qdrant
+builds payload indexes without blocking reads/writes) and prints the
+resulting payload schema per collection.
+
+Usage:
+    python scripts/apply_qdrant_payload_indexes.py
+"""
+
+from __future__ import annotations
+
+import json
+import sys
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

PR #977, opened 2026-07-10
feat(memory): headless-Haiku arbiter for ambient candidates (WS-C PR3)

WingedGuardian:
## What

The judgment stage of the ambient session-awareness layer. On each drift-trigger fire the worker now runs one pinned \`claude-haiku-4-5-20251001\` headless call that judges which ranked candidates genuinely deserve surfacing — still SHADOW: picks are recorded in the verdict/shadow log, nothing is injected.

Locked design properties, all test-enforced:
- **No fallback chain**: failed / timed-out / unparseable → recorded verdict with empty picks, never a guess.
- **Fail-closed parser** (mirrors \`attention/sampler.py\`): CLI JSON envelope unwrap, fence strip, first brace-balanced object, ints only (bools explicitly rejected), range-checked, deduped, hard cap 2. Adversarially traced: string-embedded braces cannot produce wrong picks — only \`None\`.
- **Process-group kill** on the single 90s timeout (\`preexec_fn=os.setpgrp\`, pgid>1 guard, \`killpg\` SIGKILL — claude spawns MCP children; mirrors \`cc/invoker.py\`). Timeout test uses real processes with explicit pids and proves the grandchild dies.
- **Injection resistance**: candidate previews are sanitized DATA (\`strip_boundary_markers\`), the model can only echo candidate numbers, and \`picked_memory_ids\` re-bounds-checks before resolving.
- \`GENESIS_CC_SESSION=1\` on the subprocess (no Genesis-hook re-entry), \`--strict-mcp-config\` + repo \`no_mcp.json\`, no \`--effort\` (Haiku).

Hook spawn drops \`--no-arbiter\`: the full retrieve → rank → judge pipeline now runs in shadow on every fire.

## Verification

- 25 tests: parser matrix (fences/garbage/bools/floats/out-of-range/dedupe-cap/injection), argv pinning, fake-claude ok/exit/unparseable paths, real-process group-kill, worker pick-resolution.
- Live E2E in-suite (skipif-gated on the claude binary): real Haiku call picks the decision-class candidate deterministically (~11-18s).

PR #976, opened 2026-07-10
feat(memory): ambient worker + 3-lane ranking + spawn + slots (WS-C PR2)

WingedGuardian:
## What

Second stage of the ambient session-awareness layer: on a drift-trigger fire, the hook spawns a **detached worker** (\`start_new_session=True\`, the genesis_session_end idiom) that retrieves and ranks candidate memories — SHADOW only: verdicts recorded, nothing injected.

**Three candidate lanes, all exact-search** (filtered HNSW without payload indexes measurably drops valid results — found hunting the OMI kill-criterion memory; production-wide fix in #975):
- **Vector**: theme EMA → Qdrant (deprecated filtered).
- **Decisions**: EMA against \`tags~decision\` only — the OMI-incident class. Measured: the repo-split fact sat past rank 200 unfiltered behind a wall of near-duplicate operational memories, rank 12 in this lane; E2E it entered the final candidate set at rank 8.
- **Drift**: entity-ledger top keywords through \`drift_recall\` (read-only path).

Rank = \`(1 + 0.05·log1p(inbound)) × confidence × CLASS_WEIGHT × cosine\` (mirrors the retrieval graph-boost shape) + bitemporal parity via \`_expired_candidate_ids\`. Stratified floor keeps ≥2 decision-lane candidates in the final 8 — otherwise the operational wall crowds out exactly what this layer exists to surface.

**Zero writes**: no retrieved_count bump, ever — MEM-005/H-1 baselines stay clean (asserted by test and verified E2E with before/after payload snapshots).

Infra: 2-slot flock semaphore (auto-release on death, 15s fail-closed acquire → recorded \`slots_busy\`), atomic \`ambient_verdict.json\`, size-capped \`shadow_log.jsonl\` (cap counted, never silent). \`qdrant_ops.search\` gains additive \`tags_any\`/\`exact\` params (default behavior byte-identical).

## Verification

- 16 new tests (ranking formula/lanes/stratified-floor/zero-write, slots contention/timeout, worker verdict paths incl. error + cap, subprocess entry E2E).
- Live E2E vs real stores: kill-criterion memory in candidates at rank 8; target \`retrieved_count\` unchanged before/after; two held slots → third worker records \`slots_busy\`.
- Worker cold start measured ~3.3s — detached, session never waits.

PR3 (arbiter) and PR4 (replay harness + verdict) follow on this branch chain.

PR #980, opened 2026-07-10
feat(memory): WS-C PR4 — replay harness, replay-derived tuning, soften-not-reset staleness

WingedGuardian:
## Summary
Final PR of the WS-C ambient session-awareness lane (PR0-PR3 merged: #970, #972, #976, #977; indexes #975).

### Replay harness (`scripts/ambient_replay.py`)
Offline acceptance instrument: streams a recorded CC transcript through the live fold→trigger→rank gates (hook-module import for gate parity), read-only against prod stores, with an embed cache, `--as-of` temporal cutoff (no post-session memory leakage into candidates), per-turn file context from transcript tool_use entries, and `--json` output.

### Replay-derived tuning (4 documented iterations)
- `FIRED_DIST_MIN` 0.35→0.15, `MAX_FIRES` 3→8 (0.35 starved all later fires)
- Fold gates: `FOLD_MIN_KEYWORDS=3`, non-thematic prefix skip (continuation/interrupt/bash-passthrough turns no longer fold)
- `PIVOT_CONFIRM_COS=0.5` — ring resets only on semantically corroborated pivots (Jaccard pivots fired on ~40% of turns and kept resetting stability)
- `OUTLIER_ESCAPE_RUN=3` unchanged from #972

### Kill-criterion verdict (documented; gate honored)
**FAIL, STRUCTURAL** on the incident session: the theme lived in the action stream (1 of 238 genuine user turns mentions the target topic) — a user-prompt EMA cannot see it; no constants fix this. Validated elsewhere: narrated-theme control session fires sanely (3 fires/34 turns) and a synthetic on-theme EMA ranks the target memory 8th. Per the locked gate, PR5 (delivery) is NOT proceeding on this design alone; the redesign (entity layer) was brought back for a design decision and is planned separately.

### Soften-not-reset staleness
A >24h-idle session now keeps its EMA + ema_turns (theme identity survives idle/compaction gaps), halves entity-ledger weights (prune <0.05), and clears only trigger state (ring, fired regions, claims) so the trigger re-arms. Hard reset remains new-session-id only.

### Docs
GENesis-Voice scope wording in CLAUDE.md: the invested edge platform is the HAOS Voice PE device.

## Testing
- `tests/test_session_awareness/` 79 passed in worktree (incl. new soften test, replay-adjacent ranking `created_before` tests)
- ruff clean on the full diff
- Replay runs documented against two real sessions + synthetic control

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -142,7 +142,7 @@ Genesis to research, summarize, write content, or do non-Genesis tasks).
 
 Your Genesis install is one operational system: the public `GENesis-AGI`
 codebase (the cognitive core), the public `GENesis-Voice` repo for
-voice/edge-device software (OMI + Voice PE firmware, esphome configs, S2S/
+voice/edge-device software (HAOS Voice PE device firmware, esphome configs, S2S/
 ambient audio bridges, edge deploy — `GENesis-AGI` keeps only its internal
 channel code), your private fork for customizations, and your private
 `genesis-backups` repo for encrypted data. Full model:
--- a/scripts/ambient_replay.py
+++ b/scripts/ambient_replay.py
@@ -1,0 +1,328 @@
+#!/usr/bin/env python3
+"""WS-C replay harness: run the ambient pipeline over a recorded session.
+
+Acceptance instrument for the session-awareness layer (NOT a CI test).
+Streams a CC session transcript, feeds each GENUINE user turn through
+the exact live gates (the hook's envelope + keyword functions, imported
+from the hook module itself), folds the theme EMA, traces every trigger
+decision, and on each fire runs the real retrieve+rank lanes (read-only)
+to capture the candidate set.
+
+KILL CRITERION (OMI incident, session 246fe52b…):
+    PASS = memory 9d36f039… appears in a fire's candidate set within
+    the first 10 genuine turns. FAIL → tune constants (authorized) or
+    STOP and redesign if it's structural.
+
+Usage:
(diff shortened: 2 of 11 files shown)


WingedGuardian:
**Structural review:** 1 note
**PII/Secrets/Wording scan:** 1 soft warning

No hard blocks. No request-changes findings.

---

### [NOTE] Docstring not updated — `src/genesis/session_awareness/hook_api.py`

`hook_fold` gained a new `prompt_text: str = ""` parameter but the function's existing docstring doesn't mention it or the new early-return behaviour it enables. Low priority (defaults to `""`, existing callers are unaffected), but the public API changed.

---

### [SOFT WARNING — PII] Hardcoded path encodes username — `scripts/ambient_replay.py:47`

```python
_TRANSCRIPT_DIR = Path.home() / ".claude" / "projects" / "-home-ubuntu-genesis"
```

The segment `-home-ubuntu-genesis` is the Claude Code project key for `/home/ubuntu/genesis`, encoding the username `ubuntu`. This has two consequences:

1. **Portability** — the default transcript lookup silently fails on any machine with a different username or install path. The `--transcript` flag works around it, but the default is always wrong outside this specific host.
2. **Username disclosure** — the literal string `ubuntu` ships in the public repo.

Consider sourcing this from `genesis_env` (e.g. derive from `genesis_env.genesis_db_path()` or a config key), or removing the default entirely and requiring `--transcript` to be explicit when the session-id is known.


---

PR #981, opened 2026-07-10
fix(dashboard): P4 monitor/correctness — badge, log defaults, staleness, time ranges, source filter, empty states

WingedGuardian:
Dashboard P4 — the monitor/correctness phase of the dashboard roadmap (follows #893 de-monolith, #903 infra card, #964 nav/layout). Six fixes for places the dashboard misled the operator, all verified against the live UI and current routes.

## Fixes

- **Observations badge no longer reads 0 on cold load.** The unsurfaced-count summary only fetched on tab activation; it now loads with the page and refreshes on the standard 15s poll. The summary endpoint also switches from fetching up to 1000 full rows (to `len()` them) to a proper `COUNT` query (`count_unsurfaced()`), so the more frequent poll is O(1) rows and the badge no longer silently caps at 1000.
- **Event Log defaults to INFO+.** The DEBUG+ default drowned the feed in the once-a-minute web-UI heartbeat (~90% of rows). DEBUG stays one click away; the Reset button now restores the same defaults the page loads with.
- **Neural Monitor rolls ages to days past 48h.** Long-stale runs read "67d" instead of "1613.5h"; the L1 EK row uses the same formatter instead of its own raw-hours string.
- **Error Log time ranges are same-day aware.** A group spanning days now shows the date on both ends instead of `Jul 1 01:55 → 18:04:19`; start/end precision matches.
- **Observation source filter folds `session:<uuid>` values into one "session (any)" option.** The list endpoint expands the sentinel to a prefix match via a new `source_prefix` kwarg on `observations.query()`. The sources feed also excludes internal observation types (same exclusion the list applies), so the dropdown never offers a source that filters to zero rows. The endpoint's inline SQL moved into crud helpers.
- **High-traffic tabs stop flashing a false "No X found".** Observations, work (tasks + sessions), follow-ups, chat (outreach/proposals/approvals), and autonomy now gate their empty states on the shared fetch-state machinery: "Loading…" during the first fetch, an explicit "Couldn't load — will keep retrying" on failure, and the true empty state only after a successful load. The autonomy tab's calm "No standing autonomy" line can no longer render on a failed fetch.

## Consciously out of scope

- Missing-key → failure-count linkage on the API Keys card (audit bug #13): an earlier attempt in this branch was built on `dead_letter.target_provider` and reverted after review proved that column only ever holds the `'all'` chain-sentinel in practice; the correct source is the structured `events.details` JSON and needs its own per-event-type design. Tracked separately.
- The ego-card focus-summary junk/truncation: the dashboard renders the stored value verbatim; the bug is in the writer, tracked separately.
- Low-traffic empty-state offenders (campaigns/knowledge/overview cards) — same gate pattern, deferred.

## Tests

New/extended coverage: `count_unsurfaced`, `source_prefix` + mutual exclusion, `distinct_unresolved_types/sources` (incl. type exclusion), filters-endpoint sentinel collapse, list-endpoint sentinel expansion. Full targeted sweep (142 tests across db/observability/dashboard surfaces) passing; JS integrity + node syntax checks clean.


WingedGuardian:
**Structural review: CLEAN** — no pattern violations found (no nohup, no raw SQL outside crud, no bare `except:`, no mutable defaults, no /tmp/, commit prefix correct, new functions docstring-covered, new test file present).

Two NOTEs for awareness:

**NOTE 1 — `panelState('approvals')` may not be tracked** (`src/genesis/dashboard/templates/partials/tabs/chat.html`)

Three new template guards use `panelState('approvals')`, but `approvals` was not added to the `fetchState` panel entries in `dashboard.js`. If `approvals` already exists in the pre-existing `fetchState` map (before this diff), no problem. If it doesn't, the loading/error templates for the Approvals sub-tab are dead code — `panelState('approvals')` will never equal `'loading'` or `'error'`, so empty-state flash protection isn't achieved there. Worth a quick grep for `approvals:` in the existing `fetchState` block to confirm.

**NOTE 2 — `timeRange` out of sync with comment** (`src/genesis/dashboard/templates/genesis_logs.html`)

The comment added to the `filters` block says _"Keep in sync with resetFilters()"_, but `timeRange` is not synced: the initial `filters` object keeps `timeRange: "1h"` (unchanged in this diff) while `resetFilters()` now uses `"24h"`. Either the initial `timeRange` should also be updated to `"24h"`, or the comment scope should clarify that only `severities` is being kept in sync.

---

WingedGuardian:
Both NOTEs verified against the branch — no changes needed:

**NOTE 1**: `approvals` is a pre-existing `fetchState` entry (dashboard.js:274), and `fetchApprovals()` already calls `startFetch("approvals")`/`failFetch(...)` — the new chat-tab guards are live, not dead code.

**NOTE 2**: the initial `filters` block has `timeRange: "24h"` (genesis_logs.html:42) and `resetFilters()` now also uses `"24h"` (line 116) — the two literals are fully in sync; the pre-diff inconsistency (reset=1h) is what this PR fixed.

PR #984, opened 2026-07-10
fix(autonomy): atomic dispatch claim + refuse-taxonomy (F2 duplicate execution)

WingedGuardian:
## Summary

Fixes **F2 (duplicate task execution)** from the 2026-07-02 audit. The autonomy dispatcher could execute a task more than once. Two root causes plus the real harm:

- **Bug A — no atomic claim.** Path 1 spawned `execute()` guarded only by an in-memory `_dispatched` set; `execute()` read `current_phase` once and `_transition` wrote it back unconditionally, so two overlapping dispatches could both run.
- **Bug B — Path 1b re-fired every 120s.** A blocked task with an approved, unconsumed approval was re-dispatched on every poll with no effective in-flight guard.
- **The harm — completed-refresh hole.** `execute()`'s phase branch treated any non-resumable phase (completed / synthesizing / delivering / retrospective) as a fresh run, so a stacked re-dispatch that landed after the first finished re-ran the whole task — re-pushing branches, re-notifying.

## The fix

- **`TaskPhase.DISPATCHING`** transient phase + transitions (`PENDING→DISPATCHING→OBSERVING`; `DISPATCHING→PENDING` for reaper reset).
- **`task_states.claim_for_dispatch()`** — one guarded `UPDATE … WHERE current_phase='pending'`, stamps `updated_at`, returns True only for the winner.
- **`execute()` phase taxonomy** (decided up front, claim after the plan read): `pending`→atomic claim; RESUMABLE {executing, verifying, blocked, **paused**}→resume from persisted steps; RESTARTABLE {observing, reviewing, planning}→safe fresh re-run; everything else (completed / tail phases / dispatching / terminal)→refuse with `task.dispatch_claim_lost`.
- **`_transition()` now stamps `updated_at`** (it never did) — the reaper relies on it and it also fixes `list_active`/dashboard recency ordering.
- **`recover_stale_dispatching()` reaper** resets claims stranded by a crash; wired at startup (before `recover_incomplete`) and on the dispatch poll with its own `task_dispatch_reaper` job-health key.
- **Dispatcher `_dispatch_inflight` guard** — populated before the semaphore acquire, so Path 1b also covers a dispatch still *queued* on the single-worker semaphore (the window where `_active_tasks` isn't set yet). `recover_incomplete` leaves crashed tail-phase tasks for review instead of a guaranteed-refusal dispatch.
- **Dashboard** color for the transient `dispatching` phase.

No DB migration (`current_phase` is `TEXT`, no CHECK).

## Verification (repro-test-first)

- 18 targeted WS-C tests (claim atomicity, one-execution under overlap, completed/tail/dispatching refusal, restartable-phase fresh re-run, **paused resume without redoing completed work + re-pause on resume**, reaper reset/spare, Path 1b queued-window guard, transitions).
- **E2E integration test**: real `TaskDispatcher._guarded_execute` + real executor + real `Semaphore` + real claim — two overlapping dispatches run the task exactly once.
- Full `test_autonomy/` suite green (**1020 passed**); executor dir 418; ruff clean.

## Review history

Two Claude inline review rounds each caught real defects, both fixed with regression tests: (1) the initial refuse-taxonomy wrongly orphaned `paused`/`observing`/`reviewing`/`planning` recovery — narrowed to a restartable set; (2) the Path 1b guard's semaphore blind spot — closed with `_dispatch_inflight`; (3) paused tasks were re-run fresh (silently redoing completed steps + re-spending cost) — now resumed. Commits are kept separate for that history.

Live-core change → single execution and the reaper reset are verified pre-merge; a live-server E2E (submit a queued task, observe single execution + dashboard `dispatching`) will run post-deploy.

### DIFF
--- a/src/genesis/autonomy/dispatcher.py
+++ b/src/genesis/autonomy/dispatcher.py
@@ -44,6 +44,15 @@ _TERMINAL_PHASES = frozenset({
     TaskPhase.CANCELLED.value,
 })
 
+# Non-resumable tail phases: a task that crashed here cannot be safely re-run
+# from scratch (execute() refuses them), so crash recovery leaves them for
+# review instead of spawning a guaranteed-refusal dispatch.
+_NON_RESUMABLE_TAIL = frozenset({
+    TaskPhase.SYNTHESIZING.value,
+    TaskPhase.DELIVERING.value,
+    TaskPhase.RETROSPECTIVE.value,
+})
+
 
 def _validate_plan_path(path_str: str) -> Path:
     """Validate that a plan path is under allowed directories.
@@ -276,6 +285,12 @@ class TaskDispatcher:
             if phase != TaskPhase.BLOCKED.value:
                 continue
 
+            # Bug B guard: if this task is already executing in-process, don't
+            # stack another dispatch. Without this a blocked+approved task
+            # re-fires every cycle until execute() consumes the approval.
+            if task_id in self._executor._active_tasks:
+                continue
+
             # Check if there's an approved-but-unconsumed approval for this task
             try:
                 cursor = await self._db.execute(
@@ -439,6 +454,18 @@ class TaskDispatcher:
                     logger.info("Found blocked task %s, re-notifying", task_id)
                     self._dispatched.add(task_id)
                     recovered += 1
+                continue
+
+            # Non-resumable tail phase (crashed mid synth/deliver/retro):
+            # execute() would refuse a fresh re-run, so leave it for review
+            # instead of spawning a guaranteed-refusal dispatch.
+            if phase in _NON_RESUMABLE_TAIL:
+                logger.info(
+                    "Found task %s stuck in tail phase %s, leaving for review",
+                    task_id, phase,
+                )
+                self._dispatched.add(task_id)
+                recovered += 1
                 continue
 
             # Re-dispatch executing/reviewing/planning/etc tasks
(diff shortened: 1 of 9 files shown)


WingedGuardian:
## First-pass code review — PR #984

### Structural Review

**WARNING** — `src/genesis/runtime/init/tasks.py`: duplicate import alias for the same module.

`task_states` is imported twice within `init()` under two different aliases:

```python
# Inside the try: block (crash-recovery section):
from genesis.db.crud import task_states as _task_states

# Outside the try:, just before _dispatch_poll_loop is defined:
from genesis.db.crud import task_states as _task_states_poll
```

Python closures capture variables by name from the enclosing scope; `_dispatch_poll_loop` can access `_task_states` directly without a second import. The second import is redundant. Suggested fix: drop `_task_states_poll` and rename the two references inside `_dispatch_poll_loop` to `_task_states`.

---

No other pattern violations found (no `nohup`, no direct `task_states` table inserts, no `/tmp/` usage, no bare `except:`, no mutable defaults, no raw SQL outside the crud module).

**CLEAN** — No API keys, private IPs, SSH keys, credentials, or hardware identifiers found in added lines. PR title and description contain no PII-leak language, security-incident references, or private infrastructure identifiers.

---

WingedGuardian:
Addressed the P2 (paused tasks delivering without an explicit resume after recovery) in `609dc5d2`: extracted the pause-wait logic into a shared `_honor_pause()` helper and added a pre-VERIFYING pause gate, so a task paused with no remaining steps stays paused across a restart until resumed. Tasks with remaining steps are unaffected (the last per-step checkpoint already cleared the pause). Regression test `test_paused_at_verify_boundary_stays_paused` added; full `test_autonomy/` suite green.

WingedGuardian:
Structural WARNING (duplicate `task_states` import alias in `init/tasks.py`) fixed in `bb892b3c`: hoisted a single `as _task_states` import before the crash-recovery try (unconditionally bound for the poll closure) and dropped the `_task_states_poll` alias. Wiring test + ruff green.

PR #985, opened 2026-07-10
fix(dashboard): event-log severity select shows the active filter on load

WingedGuardian:
Follow-up to #981. The Event Log's severity `<select>` displayed **DEBUG+** while the page was actually filtering **INFO+**: Alpine applies `x-model`'s initial value before the `x-for`-rendered options exist, so the select fell back to showing its first option. The old `debug` default coincidentally matched the first option, which is why this was invisible until #981 changed the default.

The option list is a hardcoded five-entry set, so render the `<option>` tags statically (letting `x-model` sync correctly on init) and drop the now-unused `severityOptions` getter. Caught by post-deploy live verification: store said `info` and the feed was heartbeat-free, but the combobox showed DEBUG+ selected.

One file, template-only. `tests/test_dashboard/test_logs_api.py` passing.

### DIFF
--- a/src/genesis/dashboard/templates/genesis_logs.html
+++ b/src/genesis/dashboard/templates/genesis_logs.html
@@ -230,15 +230,6 @@
                   "perception", "learning", "inbox", "reflection", "providers", "web", "outreach", "dashboard"];
         },
 
-        get severityOptions() {
-          return [
-            { value: "debug", label: "DEBUG+" },
-            { value: "info", label: "INFO+" },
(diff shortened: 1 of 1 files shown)


WingedGuardian:
**Structural review:** SKIPPED — template-only change (single HTML file, no Python).

**PII/Secrets/Wording scan:** CLEAN — no credentials, private IPs, personal emails, private hostnames, or sensitive wording in the diff or PR metadata.

---

PR #986, opened 2026-07-10
fix(memory): faceting parity on embedding recovery + real created_at for FTS-only rows

WingedGuardian:
## What

Two related memory-retrieval correctness fixes (one concern: `memory_metadata` parity).

### D3 — recovered memories keep their faceting filters
`EmbeddingRecoveryWorker.drain_pending` rebuilt the Qdrant payload of a memory that was stored FTS5-only during an embedding outage, but **dropped its `wing`/`room`/`life_domain`** — so the re-embedded point was silently excluded from every faceted (`wing=`/`room=`/`life_domain=`) recall, because Qdrant `must` filters exclude points that lack the key.

Recovery now restores them:
- **wing/room** ← `memory_metadata` (written by `create_metadata` in the same `store()` that enqueued the pending row, so it's present at drain time)
- **life_domain** ← the `life_domain:` tag `store.py` appends (not a metadata column)
- **project_type** stays absent — it is never persisted on this path (metadata/pending/tags); documented + follow-up filed, not a regression this introduces

Also drops a stray `memory_id` payload key the normal write path never sets.

### D4 — keyword-only results are no longer ranked as artificially fresh
`_compute_activations` gave FTS-only rows (no Qdrant hit) `created_at = now_str`, yielding `recency = exp(0) = 1.0` (max freshness) and a phantom age of 0 in the MEM-005 entrenchment metric. It now batch-fetches their real `created_at` from `memory_metadata` (NOT NULL there); legacy FTS ghosts with no metadata row keep the `now_str` fallback. Guarded to skip the async call when there are no FTS-only rows.

## How
- New crud helpers `memory.get_taxonomy` + `memory.batch_created_at` (mirror `memory_links.batch_link_counts`).

## Tests / verification
- 63/63 targeted tests green; RED-proven (both fixes fail without the change).
- Independent code review: clean, no correctness findings.
- **Live E2E vs real Qdrant** (scratch collection): the recovered point carried wing/room/life_domain and a `wing=` filter surfaced it, while a keyless control point was excluded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,15 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Memories recovered after an embedding outage keep their wing/room/life_domain
+  filters, and keyword-only results are no longer ranked as artificially fresh.**
+  When the embedding provider was down, memories were stored keyword-only and
+  re-embedded on recovery — but the rebuilt vector dropped its wing/room/life_domain
+  tags, so it silently vanished from any domain-scoped ("wing=…") recall. Recovery
+  now restores those fields. Separately, keyword-only memories were being scored as
+  if created just now (maximum freshness), letting old notes outrank genuinely recent
+  ones; they now use their real creation time.
+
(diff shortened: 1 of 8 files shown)


WingedGuardian (commented):

---

PR #983, opened 2026-07-10
feat(memory): entity-layer substrate — entities, mentions, bi-temporal links (E2)

WingedGuardian:
## Summary
Second PR of the entity-layer lane (WS-H Pillar 2, Graphiti blueprint on the existing SQLite+Qdrant substrate — no new server). Adds typed entity nodes with identity, memory↔entity mentions, and bi-temporal entity↔entity relations. Wired by E3 (extraction + backfill) and read by E4 (session-awareness entity lane, shadow-first).

### Data model (migration 0051 + DDL twins)
- **`entities`** — typed nodes (`code_file/code_symbol/pr/commit` mechanical anchors + `product/device/repo/subsystem/person/org/concept` named), `UNIQUE(norm_name, entity_type)`, merge lifecycle (`active/merged/gone`).
- **`entity_mentions`** — memory↔entity with provenance tags (`EXTRACTED/INFERRED/AMBIGUOUS`); no validity columns (a mention is an event-time fact; memory validity stays in `memory_metadata`).
- **`entity_links`** — entity↔entity, bi-temporal `valid_at/invalid_at` (canonicalized via #982's write gate), provenance + confidence, `evidence_memory_id`. **Open `link_type` vocabulary** (slugified) — deliberately NOT the 5-place `memory_links` CHECK registry; a maintenance report will surface vocab sprawl.

### Code
- `db/crud/entities.py` — upserts keep the stronger claim; **undirected ≤2-hop traversal** with validity filtering and provenance-weighted path confidence (plain queries; no NetworkX cache at 10²–10⁴ rows); adjudicated merge (mentions/links rewritten to survivor, norm lookups follow the merge); fuzzy matches enqueue `entity_adjudication` on `deferred_work_queue`.
- `memory/entity_registry.py` — write-time resolution tiering: mechanical anchors exact by construction → named exact (concept-cluster cross-type reuse) → fuzzy ≥0.85 creates AMBIGUOUS + queues adjudication → new. **Never an inline LLM call.**
- `memory/entity_seed.py` + `scripts/apply_entity_seed.py` — ~20 curated entities + relations incl. the repo-split spine; idempotent; the apply script prints a reachability spot-check.
- `entity_resolution.py` docstring disambiguation (that module = memory-PAIR dedup, not entity nodes); subsystem map updated.

### Design invariant
[...]

PR #988, opened 2026-07-10
feat(memory): entity-layer write wiring — extraction identity, anchors, backfill (E3)

WingedGuardian:
## Summary
Third PR of the entity-layer lane. **Zero new LLM calls** — the extraction pipeline already emits entity and relationship names; this PR gives them identity in the #983 substrate.

### Write wiring
- **`entity_registry.record_extraction()`** — `extraction.entities` → `entity_mentions`; `extraction.relationships` → bi-temporal `entity_links` (`valid_at` from the extraction temporal, canonicalized; `evidence_memory_id` set; per-relationship `confidence`/`ambiguous` honored). Wired into `run_extraction_cycle` immediately after `create_typed_links` — the `memory_links` path is untouched and parallel; fail-open per memory.
- **`entity_anchors.py`** — conservative regex extraction of repo paths, `genesis.*` symbols, PR#s, and SHAs (digit-required hex so hex-alphabet words never match; capped 16/memory). Wired at the `MemoryStore.store()` seam so EVERY write path records anchors; failure-isolated. Mechanical anchors skip the fuzzy tier entirely — no O(entities) scan on the hot path (reviewer-verified).
- **Extraction prompt** — preferred relation vocabulary (`is_a`, `part_of`, `constrained_by`, `depends_on`, …) + optional per-relationship `confidence`/`ambiguous`. Additive; unknown types are still skipped by the `memory_links` validator.
- **`scripts/entity_backfill.py`** (approved hybrid strategy) — mechanical anchor pass over all ~49.7K `memory_fts` rows + seed-entity FTS mention pass (`INFERRED`, 0.6) that makes the repo-split decision memory reachable organically. Dry-run by default, `--apply` gated. Forward-only for LLM named entities (cost + honest engine-re-eval load data).

### Code-review findings (inline reviewer over the combined E2+E3 diff) — both fixed here
1. `parse_extraction_response` stripped the optional per-relationship `confidence`/`ambiguous` fields, silently forcing every link to `EXTRACTED`; now preserved, with a regression test through the REAL parser path.
2. `merge_entity` minted a self-loop edge when the merged pair already linked to each other; rewrite now filters `dst == survivor`.

Reviewer verified clean: fail-open at both seams, commit discipline, regex boundedness, migration idempotence, backfill gating.

## Testing
- 24 tests across `test_entity_write_wiring.py` (anchor regex matrix incl. hex/heading false-positive guards, record_extraction incl. end-to-end parser path, store fail-open, merge self-loop guard) + `test_entity_layer.py`
- Neighbors green: `test_store.py`, `test_extraction_job.py`, `test_extraction.py`; ruff clean

## Post-merge (this session)
Apply seed + run backfill with `--apply` (user-gated), then verify the OMI spine end-to-end against the live DB.

### DIFF
--- a/scripts/entity_backfill.py
+++ b/scripts/entity_backfill.py
@@ -1,0 +1,149 @@
+#!/usr/bin/env python3
+"""Entity-layer backfill (approved HYBRID strategy, 2026-07-09).
+
+Two passes, both mechanical — NO LLM calls:
+
+1. **Anchor pass** — regex code-anchor extraction over every
+   ``memory_fts`` row (paths, genesis.* symbols, PR#s, SHAs) →
+   ``entity_mentions`` with ``source='backfill_mechanical'``.
+2. **Seed-FTS pass** — for each seed entity, FTS5 MATCH on its name →
+   ``entity_mentions`` with provenance ``INFERRED``, confidence 0.6,
+   ``source='backfill_fts'``. This is what makes the repo-split memory
+   reachable ORGANICALLY (its content names GENesis-Voice / voice edge),
+   not only via the hand-seeded spine.
+
+Forward-only for LLM-extracted named entities (no 49.7K re-extraction):
+cost, near-duplicate enrichment waste, and it would distort the
+engine-re-evaluation load data (follow-up 3ae26455).
+
+Default is DRY-RUN (counts only). Pass --apply to write.
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import sys
+from pathlib import Path
+
+SRC_DIR = Path(__file__).resolve().parent.parent / "src"
+if str(SRC_DIR) not in sys.path:
+    sys.path.insert(0, str(SRC_DIR))
+
+BATCH = 2000
+
+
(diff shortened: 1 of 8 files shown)


WingedGuardian:
Structural review: CLEAN — no pattern violations, no WARNINGs. 2 NOTEs below.

---

### NOTE — Direct `entity_mentions` read in `main()` outside crud layer

`scripts/entity_backfill.py` ~line 129:
```python
    "SELECT entity_id, provenance, source FROM entity_mentions WHERE memory_id = ?",
    (target,),
) if args.apply else []
```
The migrations/backfill exception clearly applies here (and the `memory_fts MATCH` queries have no CRUD counterpart anyway). Flagging only for awareness — no change required.

---

### NOTE — Hardcoded memory UUID in `main()`

`scripts/entity_backfill.py` line ~128:
```python
target = "9d36f039-3126-4721-8c71-027df1a94e2a"
```
Gated on `--apply` and harmless if stale (query just returns zero rows), but a brief inline comment naming the specific memory this ID targets would help future readers understand why this particular UUID is here rather than treating it as an arbitrary test value.

---

PR #989, opened 2026-07-10
refactor(surplus): remove disabled code-audit job + boot-time orphan sweeps (T2-C, D1/D2)

WingedGuardian:
## Summary

Three scheduler-polish items from the Tier-2 backlog:

- **T2-C — remove the long-disabled `schedule_code_audit` job end-to-end.** Gate fn, registration, boot-run, facade, ctor params (`enable_code_audits`, `code_audit_hours`), Protocol fields, init kwargs, config key, dashboard setting labels, and the awareness liveness-list entry. **Kept deliberately:** `CodeAuditExecutor` (wired unconditionally), the `CODE_AUDIT` task type and its dispatch/judge consumers, and the executor-path test — the capability remains manually enqueueable. Migration 0052 deletes the stale persisted `job_health` row so it can't linger as an ever-staler entry.
- **D1 — boot-immediate surplus orphan reclaim.** `recover_stuck_with_retries` gains `bump_attempt: bool = True`; the scheduler's boot block calls it with `older_than_hours=0, bump_attempt=False`: dispatch is single-worker, so any `running` row at boot is orphaned by the previous process — reclaim it now (was up to 2h + one cycle) and don't burn the task's retry budget on a restart. The per-cycle 2h-gated recovery is unchanged. Coupling recorded in-code: re-gate on worker ownership if parallel dispatch ever ships.
- **D2 — boot kick for the recovery orchestrator** (dead-letter replay / stuck-processing expiry / pending embeddings), mirroring the adjacent `_validate_keys` kick; previously nothing ran it until the first 30-min interval fire.

Also drops the caller-less crud `recover_stuck` wrapper (review finding).

## Review

Claude structured review: clean (verified SQL branch, caller enumeration, APScheduler first-fire timing — no race with the boot sweep). Codex adversarial: two findings adopted (boot-arm SQL now matches its every-running-row contract incl. `started_at IS NULL`; migration 0052 for the stale job_health row); others verified false or by-design (boot requeue of maxed rows is the intended restart-≠-failure semantics; per-cycle recovery still fails genuinely stuck tasks).

## Verification

- Targeted suites green: `tests/test_surplus/` (incl. new boot-sweep test), `tests/test_db/test_surplus_tasks.py` (bump/no-bump contracts), migration 0052 tests, `tests/test_runtime/` learning-init tests.
- The D2 kick follows the proven `_validate_keys` pattern (same scope, defined-before-use verified); deploy verification will confirm the `initial_recovery` task in the boot journal.


WingedGuardian:
## First-pass review — PR #989

**Structural:** 1 NOTE

---

### NOTE — `SurplusQueue.recover_stuck` docstring doesn't mention `bump_attempt`

**`src/genesis/surplus/queue.py`**

The wrapper retains its one-liner `"""Recover tasks stuck in 'running' state."""` but now exposes `bump_attempt: bool = True` whose `False` path has a meaningfully different contract (no age gate, never permanently fails, single-worker assumption). The full contract is well-documented in `surplus_tasks.recover_stuck_with_retries`; a caller looking only at the queue API won't see the single-worker caveat or the boot-sweep semantics. A brief addition like `"Pass bump_attempt=False for the boot sweep (restart ≠ task failure); see crud module for contract."` would close the gap.

---

**PII/Secrets/Wording scan:** No secrets, private IPs, SSH keys, credentials, private paths, or hardware identifiers in added lines. PR title, description, and CURRENT.md additions contain no incident-disclosure or vulnerability-detail language. CLEAN.

---

PR #990, opened 2026-07-10
fix(telegram): single-poller ownership — bridge yields to server; stall + dead-remediation fixes

WingedGuardian:
## Summary

Root-cause fixes for the dual-poller incident (two full runtimes polled getUpdates on one bot token for ~2 days: thousands of `telegram.error.Conflict` errors, split/dropped updates, flaky approval buttons, and two processes fighting over status.json):

- **Bridge yields to the server (primary fix).** `bridge.py` is a legacy fallback that boots a full second runtime; it now probes the `genesis-server` process lock at startup and exits 200 (`RestartPreventExitStatus` — no crash-loop) when the server is alive, regardless of who started it. A second probe after bootstrap and before `adapter.start()` closes the co-start race (a deploy restarting both units simultaneously — the actual incident trigger). Legacy fallback (server down → bridge runs) is unchanged. New `ProcessLock.is_locked()` probe helper (flock, non-destructive, errno-aware).
- **Dead self-heal removed, honest alert added.** The `awareness_restart` remediation could never fire (its probe was never registered — `collect_probe_results`'s only caller passes no scheduler) and its command restarted **genesis-bridge**, a stale mapping from when the awareness loop lived there — i.e. the self-heal would have *caused* this incident class. Deleted; the restart response stays with the external watchdog's zombie-scheduler path (heavy-workload/stabilization/backoff guards). New `probe_scheduler_heartbeats` (reads the same `rt.job_health` entries the status writer publishes) feeds an L4 alert-only action. Regression pins: `awareness_restart` can't return; no remediation may target genesis-bridge.
- **Stall-detector false alarms fixed.** The polling watchdog only counted *arriving* updates as liveness, so any quiet 15+ minutes looked like a hung poller and restarted the updater all day (`consecutive=21` observed). `LivenessHTTPXRequest` on the dedicated getUpdates pool records every successful HTTP-200 round trip, empty or not; errors/exceptions still starve the watchdog so genuine outages trip it. Verified against PTB 22.7 internals (do_request seam, builder `get_updates_request` defaults reproduced exactly).
- **Server-side belt:** the server logs CRITICAL if a rogue bridge holds its lock when Telegram starts (log-only by design — the adapter being started is the very channel an alert would need).

## Review

Claude structured review: clean (PTB 22.7 compatibility verified against installed source; slots/builder checks; both sub-threshold observations fixed — TOCTOU guard in `is_locked`, `peek()` over `instance()` in the probe). Codex adversarial: co-start race + errno conflation adopted and fixed; remaining findings verified false (L4 empty-command path returns before execution; exception→HEALTHY is the documented can't-evaluate-≠-dead policy; timestamp semantics deliberately mirror the external watchdog's).

## Verification

- Targeted suites green: transport/adapter, bridge yield guard (real flock subprocess tests incl. SIGKILL release), remediation defaults + regression pins, heartbeat probe cases, hosting belt.
- Live E2E after merge: start genesis-bridge with the server up → immediate exit 200 with the yield log, zero new Conflict lines; single status.json writer (uptime_s already re-verified matching server uptime after the ops stop).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,23 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   now restores those fields. Separately, keyword-only memories were being scored as
   if created just now (maximum freshness), letting old notes outrank genuinely recent
   ones; they now use their real creation time.
+- **Telegram no longer drops messages or breaks approval buttons when the
+  legacy bridge gets started alongside the server.** Two Genesis processes
+  polling the same bot token split incoming updates between them (observed:
+  thousands of `Conflict` errors, unresponsive approval buttons, and a
+  corrupted health-status file monitored by the watchdog). The bridge now
+  yields cleanly at startup whenever the server is running — however it was
+  started (deploy script habit, manual, self-heal) — and the server calls out
+  a rogue bridge loudly instead of failing silently. The broken self-heal
+  rule that restarted the bridge on a stale awareness heartbeat (a leftover
+  from when the bridge owned the awareness loop) is gone; a health alert
+  takes its place, and server restarts for genuinely dead schedulers remain
+  with the external watchdog.
+- **Quiet hours no longer restart the Telegram poller.** The stall detector
+  only counted arriving messages as signs of life, so any 15+ minutes of
+  silence looked like a hung poller and triggered an updater restart, all day
+  long. Successful empty polls now count as liveness; real network hangs
+  still trigger the restart.
 
 - **Protected-path guarding now covers the real systemd unit sources.** The
   critical-path rules protected a stale copy of the watchdog unit under
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -52,7 +52,10 @@ journalctl --user -u genesis-server -n 50        # Logs
 systemctl --user list-units 'genesis-*' --all    # All units
 ```
 
-Other units: `genesis-bridge.service` (Telegram relay, on-demand),
+Other units: `genesis-bridge.service` (LEGACY fallback — full stack incl.
+Telegram, only when genesis-server is DOWN; it yields/exits 200 if the server
+lock is held, and must never run alongside the server — dual getUpdates
+pollers split updates and break approval buttons),
 `genesis-tmp-watchgod.service` (/tmp protection), `genesis-watchdog.timer`
 (health check), `genesis-backup.timer` (6h encrypted backup via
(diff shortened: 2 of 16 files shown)


WingedGuardian:

---

WingedGuardian:
## First-pass review — PR #990

**Structural:** 1 WARNING, 1 NOTE

---

### WARNING — Async tests missing `@pytest.mark.asyncio` in `TestLateYieldCheck`

**`tests/test_channels/test_bridge.py`** — new `TestLateYieldCheck` class

Both async methods in the class have no `@pytest.mark.asyncio` marker, and `test_bridge.py` has no module-level `pytestmark = pytest.mark.asyncio`:

```python
class TestLateYieldCheck:
    async def test_shuts_down_and_exits_200_when_server_appeared(self, tmp_path):
        ...
    async def test_noop_when_server_absent(self, tmp_path):
        ...
```

Every other async test added in this same PR is correctly marked:
- `TestLivenessHTTPXRequest` in `test_transport.py` — per-method `@pytest.mark.asyncio` ✓
- `test_standalone_telegram_guard.py` — module-level `pytestmark = pytest.mark.asyncio` ✓
- `TestProbeSchedulerHeartbeats` in `test_health.py` — per-method `@pytest.mark.asyncio` ✓

If the project's `asyncio_mode` is not `"auto"`, these two tests will be silently uncollected (pytest treats the coroutine as a plain object, does not await it, and reports a false pass). That would mean the late-yield co-start race guard — the fix for the actual incident trigger — has zero live test coverage.

Fix: add `@pytest.mark.asyncio` to each method in `TestLateYieldCheck`, or add `pytestmark = pytest.mark.asyncio` at the top of `test_bridge.py`, whichever matches the rest of that file's conventions.

---

### NOTE — `LivenessHTTPXRequest.do_request` override has no docstring

**`src/genesis/channels/telegram/transport/polling.py`** — `LivenessHTTPXRequest.do_request`

The class-level docstring explains the full intent clearly. A one-liner on the override itself (e.g. `"""Fire on_success on HTTP-200; silent on errors/exceptions so genuine hangs still starve the watchdog."""`) would make the contract readable without scrolling to the class header, especially since this is the method future readers will step into when tracing stall-watchdog behaviour.

---

**PII/Secrets/Wording scan:** No API keys, tokens, private IPs, SSH keys, credentials, private hostnames, or hardware identifiers found in added lines. PR title, description, and CHANGELOG additions contain no incident-disclosure or vulnerability-detail language. CLEAN.

---

PR #967, opened 2026-07-09
feat(guardian): credential resilience — passphrase escrow + backup coverage (G.2+G.3)

WingedGuardian:
## Why

Two latent disaster-recovery gaps in the backup system:

1. **Circular passphrase trap.** `GENESIS_BACKUP_PASSPHRASE` lived *only* inside the `secrets.env` that the backup encrypts *with* it. A secrets.env loss (the exact zeroing/corruption class that has bitten this environment) → all auth gone **and** the encrypted backup undecryptable. The safety net had no anchor.
2. **Backup coverage stopped at secrets.env.** SSH keys (incl. the guardian control-plane key), gh/Claude credentials, and host/network wiring config weren't backed up — the difference between "restore" and "reprovision from scratch."

## What

**G.2 — break the circular trap (escrow):** `credential_bridge.py` gains `propagate_backup_passphrase`, which escrows the passphrase to the host-side shared mount (`0600`, outside the container's blast radius, alongside the existing telegram/proxmox token escrow the awareness tick already writes). `restore.sh` falls back to the escrowed passphrase when the env/secrets copy is gone. **The escrow lives outside every backup source, so it never re-enters the encrypted set** — the circle stays broken.

**G.3 — expand coverage:** `backup.sh` now also GPG-encrypts (Tier-1 + Tier-2) `~/.ssh/*`, gh `hosts.yml`, Claude `.credentials.json`/`.claude.json`, and `guardian_remote.yaml`/`genesis.yaml`. Missing files are skipped; plaintext is refused without a passphrase. `restore.sh` decrypts them to a **staging dir** (never auto-clobbering live keys), private-by-creation (`umask 077` + `0700`).

## Safety review (this is credential code)

A focused security review verified the critical invariant: **the plaintext escrow is never included in any backup source** (traced every backup.sh path vs `~/.genesis/shared/`), all `creds/*.gpg` are committed encrypted only, and the passphrase is only ever passed via `--passphrase-fd 0` (never argv). Three findings raised and fixed: the restore-staging perms window (`umask 077` before decrypt), a passphrase parser-divergence note (keep hand-set passphrases shell-safe), and Tier-2 creds coverage (so the "COMPLETE snapshot" contract stays true for a git-less restore).

## Tests / verification

- `tests/test_guardian/test_credential_bridge.py`: escrow propagate/load, absent-passphrase skip, combined-bridge 3-leg fanout (27 file tests green; ruff clean; `bash -n` on both scripts).
[...]

PR #995, opened 2026-07-10
feat(guardian): host-side credential mirror (G.4)

WingedGuardian:
## What & why

Every copy of Genesis's credentials lived **inside the container**: the live files *and* the Tier-1 backup clone (`~/backups/genesis-backups/`). A destroyed container loses both, and the off-site Tier-2 copy needs the very git/GitHub credentials it would restore (chicken-and-egg). G.4 mirrors the **already-encrypted** credential bundle onto the host — outside the container's blast radius — so a rebuilt container can recover credentials (including the guardian control-plane SSH key) with zero network.

This is the last piece of the G credential-resilience track, building on the G.2 passphrase escrow (#967) which already proved the shared-mount escape hatch.

## How

- **Container side** (`credential_bridge.py`): `mirror_credential_backup()` copies `creds/*.gpg` + `secrets/*.gpg` from the Tier-1 clone to `<shared>/guardian/creds-mirror/` on each awareness tick (copy-if-changed, containment-guarded prune, `MIRROR_STAMP` written last). Wired into the existing `propagate_guardian_credentials` composite — zero new plumbing.
- **Guardian side** (`cred_watch.py`): warns if the mirror goes stale (newest credential older than `mirror_stale_hours`, default 48h — an early "backups aren't landing" signal), and keeps a second, **host-only** archive (`creds-archive/`) the container can't reach. The archive is **grow-only**: it never deletes a credential based on one possibly-partial view of the mirror, because it is the last line of defence.
- **Restore** (`restore.sh`): falls back to the mirror (or, host-side, the archive) and its escrowed passphrase when the Tier-1 clone payload is absent — the rebuild path with no network.
- Config field + `guardian.yaml`, container-loss runbook in `docs/reference/recovery-and-portability-workflow.md`, user-voiced CHANGELOG.

## Security posture

Only GPG-encrypted `*.gpg` files cross; the passphrase is escrowed separately. The mirror sits next to the escrow on the host — but the host already has root over the container (`incus exec`) and can read every plaintext original today, so this is not new exposure. Mirror/archive dirs `0700`, files `0600`.

## Verification

- `tests/test_guardian/` — 718 pass (incl. 24 new G.4 tests); ruff + `bash -n` clean.
- **Code review** (2 finders, single round): 3 findings fixed — grow-only archive (don't let a partial mirror destroy the last copy), stale-state clear, payload-keyed restore fallback.
- **Live pre-merge E2E**: real `restore.sh` rebuilt creds from a mirror-only sandbox via the escrowed passphrase (incl. empty-placeholder-`creds/` edge); mirror one-shot against the real backup clone produced a byte-identical 0600 tree of all 18 production credentials.
- **Post-merge** (host-deploy gate — `cred_watch.py` is guardian-deployed): redeploy the host guardian via `update.sh`/`git archive | ssh redeploy`, then verify the mirror appears host-side after one awareness tick, the archive populates after a guardian tick, and a backdated-stamp drill fires the stale WARNING.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,17 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   row now shows "in N fallbacks (24h) · last …" beside its key status, so an
   ongoing fallback storm is visible at a glance instead of hiding in the logs.
 
+- **Your credentials now survive losing the whole container.** Genesis mirrors
+  its encrypted credential bundle (secrets, SSH keys including the guardian
+  control-plane key, and Claude/GitHub credentials — all GPG-encrypted) onto the
+  host, outside the container's blast radius, and the host guardian keeps a
+  second copy the container can't touch. If the container is ever destroyed, a
+  fresh one can be rebuilt with credentials intact from the host — no network and
+  no chicken-and-egg (previously the only backup copies lived *inside* the
+  container). The guardian also warns you if that mirror goes stale, so you find
+  out backups have stopped landing *before* you need them. See the
+  container-loss runbook in `docs/reference/recovery-and-portability-workflow.md`.
+
 - **Genesis now detects and repairs corrupted credential files on its own.**
   If a critical credential or wiring file (`secrets.env`, your Claude Code and
   GitHub credentials, SSH keys, `guardian_remote.yaml`, `genesis.yaml`) gets
--- a/config/guardian.yaml
+++ b/config/guardian.yaml
@@ -103,6 +103,7 @@ cred_integrity:
   max_restores_per_day: 3  # per-target guardian step-in cap
   check_timeout_s: 30
   restore_timeout_s: 60
+  mirror_stale_hours: 48   # WARN if the host-side credential mirror's newest cred is older
   # backup_dir: ""         # container-side backup path; empty = module default
   # container_home: ""     # empty = container's real home; set only for E2E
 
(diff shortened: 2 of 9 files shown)


WingedGuardian:
## Code Review — PR #995 `feat(guardian): host-side credential mirror (G.4)`

### Structural Review

**WARNING — Import grouping** · `tests/test_guardian/test_cred_watch.py`

`import os  # noqa: E402` is appended after all existing test code rather than moved to the top of the file with the other stdlib imports. The `noqa` silences ruff, but the import should live in the import block at the top of the file. Quick fix: move it up alongside the existing `import json`, `import os` imports at the head of the file and drop the `noqa` comment.

---

**NOTE — Private symbol imported across module boundary** · `src/genesis/guardian/cred_watch.py:33`

```python
from genesis.guardian.credential_bridge import _atomic_copy
```

`_atomic_copy` is a private function (leading underscore). Importing it by name in a sibling module creates a silent breakage risk if `credential_bridge` ever renames or inlines it. The comment acknowledges this is intentional and side-effect-free, but consider either:
- Rename it to `atomic_copy` and add it to `__all__` in `credential_bridge.py` (makes the public contract explicit), or
- Move it to a shared `_fs_utils.py` under `genesis.guardian` since both modules need it.

Not a blocker — the current cross-import is safe today and the intent is well-documented.

---

**CLEAN** — No secrets, tokens, private IPs, SSH keys, or hardcoded usernames found in added lines. PR title, body, and CHANGELOG additions describe a new feature; no language suggesting an existing credential exposure or incident fix.

---

WingedGuardian:
Fixed in 6c206df5. The single `_MIRROR_ROOT` probe is replaced with independent per-payload resolution: the secrets source and the creds source each scan all candidate roots (shared mirror preferred, then the host-only archive) for their own payload, so an incomplete mirror can no longer mask a complete archive. Verified via an E2E sandbox where the mirror had creds but no secrets and the archive had both — secrets correctly sourced from the archive, creds from the mirror.

PR #1000, opened 2026-07-10
feat(guardian): host CC auth-health signal + setup-token fallback sync

WingedGuardian:
## What

Makes the host Guardian's `claude -p` recovery brain survive a dead `claude login` without manual host access — insurance against the silent-brain-death class (the login has no refresh; today, if it dies the brain silently goes dark, discovered only mid-incident).

Two mechanisms:

1. **Auth-health signal.** The gateway `version` verb additively reports `cc_logged_in` (a 1h-cached `claude auth status` probe), `cc_token_present`, and `cc_token_age_days`. A container-side reconciler (`GuardianWatchdog._check_cc_auth`) alerts over Telegram when the login is dead with no usable fallback (3-tick threshold) and warns ~30 days before a synced fallback token expires. Only booleans + an age ever cross the wire — never the token or account identity.

2. **Fallback setup-token (optional, lazy).** Mint a 1-year token with `claude setup-token` from anywhere and pipe it to `scripts/store_cc_token.sh` (stdin only). The credential-bridge awareness tick syncs it to the host shared mount, and `diagnosis.py` injects it via `CLAUDE_CODE_OAUTH_TOKEN` **only** when a pre-flight `claude auth status` confirms the host's own login is dead — a working login is never overridden. It is a subscription-OAuth token, **not** an `ANTHROPIC_API_KEY`.

## Design notes

- **Token lives in a dedicated file, never `secrets.env`** — `runtime/init/secrets.py` does `load_dotenv(override=True)`, which would hijack the *container's* own CC auth.
- **Fallback-only injection** — the pre-flight probe runs with the same inherited env the brain gets, so `loggedIn:true` faithfully predicts the real invocation; injection can only degrade a working login under a false-negative coinciding with a stale token, and even then it lands as the existing `cc_unavailable` fail-safe, never silent death.
- **1h probe cache** — `version()` runs every 5-min tick; `claude auth status` is a ~290MB transient spawn, so the `loggedIn` probe is cached host-side (token presence/age stay per-tick, cheap stats). The diagnosis path always probes fresh.
- **`_VERSION_TIMEOUT=30s`** — the auth probe can exceed the 10s default, which would make `version()` return `unreachable` and skip every reconciler.

## Verification

- 733 guardian tests pass (~33 new): gateway tri-state + token age + cache reuse/expiry/corruption + **no-leak invariant**; reconciler 9-scenario state machine (expiry-warns-while-healthy, re-arm lifecycles, dead-brain threshold, ambiguity-never-alarms); `_resolve_cc_env` injection matrix (working login never overridden, env preserved); credential-bridge roundtrip; version timeout.
- ruff + shellcheck clean.
- genesis-architect design review + 2-finder code review (Python + bash/security): no behavior-breaking bug; 2 findings fixed with regression tests (base-10 age arithmetic, non-object-JSON guard).

## Not covered (by design)

The exact logged-out `auth status` output shape and a real setup-token in the guardian's exact invocation are unexercised until a token is minted (lazy) — the tri-state design makes any shape safe and the diagnosis fail-safe backstops a bad token.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,20 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **The host Guardian now tells you before its AI recovery brain loses its
+  login — and can survive it without you touching the host.** The Guardian's
+  autonomous diagnosis runs on Claude Code authenticated by a one-time login at
+  install, which never refreshes; if it died, the brain silently went dark and
+  you'd only find out mid-incident. Genesis now watches that login's health and
+  alerts you (over Telegram) if it goes dead, and warns ~30 days before a
+  fallback token would expire. Optionally, mint a one-year token with `claude
+  setup-token` from any machine and pipe it to `scripts/store_cc_token.sh` —
+  Genesis syncs it to the host and uses it **only** as a fallback when the
+  host's own login is dead, never overriding a working login. The token is a
+  subscription token (not an API key), is stored 0600 and never logged, and the
+  health signal sends only status booleans — never the token or your account
+  details. You can set this up at install or just wait for the first alert.
+
 - **`genesis eval bench` — a Genesis-vs-bare-Claude A/B benchmark you can run
   in one command.** Each task in a private task set runs through two arms: a
   cognition-enabled Genesis session (identity + read-only recall from your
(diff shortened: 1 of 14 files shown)


WingedGuardian:
Automated review (Codex/structural bot) did not post — quota-limited. Merging on green CI per the pre-merge gate. Review substance was covered pre-PR: a genesis-architect design pass + a 2-finder code review (Python correctness + bash/security); both real findings (base-10 age arithmetic, non-object-JSON guard) fixed with regression tests. Rebased onto #995 (G.4 credential mirror) — additive coexistence, 754 guardian tests green.

PR #1002, opened 2026-07-10
feat(observability): put the ambient arbiter on the neural monitor (call-site registration + telemetry)

WingedGuardian:
## What

The WS-C ambient arbiter — a headless pinned-Haiku CC subprocess spawned by the detached session-awareness worker on every drift-trigger fire — was firing **completely unobserved**: absent from the routing registry, `call_site_last_run`, and every dashboard surface. Its only trace was a JSONL file on disk. This registers it as a `dispatch: cli` call site (the `14_weekly_self_assessment` species) so each run renders on the neural monitor.

**Scope decision (2026-07-10): register + telemetry, model stays PINNED.** The YAML `cc_model: Haiku` reflects reality but the code never reads it — no governance path, zero risk to the determinism guarantee. Governance (reading `cc_model` live) is explicitly deferred.

## Changes

- **`config/model_routing.yaml`**: `ambient_arbiter`, `dispatch: cli`, `cc_model: Haiku`, `chain: []` **by design** — no API fallback exists (a failed arbiter call records an empty verdict, never a guess) and the router never dispatches it. First empty-chain cli site; loader-sanctioned (`config.py` keeps empty-chain cli sites).
- **`_call_site_meta.py`**: manual `cost_policy` (empty chain would auto-derive "Not configured"); `dispatch`/`cc_model` mirror YAML for the metadata drift guard.
- **`neural_monitor.html`**: memory-group grid cell + `SITE_META` label; tooltip skips the API fraction for CC-only sites (`0/0 API` read as a broken chain).
- **`session_awareness/worker.py`**: records one `call_site_last_run` row per arbiter **attempt** (judged verdicts with candidates; empty sets short-circuit without spawning CC). Own short-lived RW connection — the retrieval connection stays `mode=ro`, so the zero-memory-row-write invariant holds (test-asserted). `arbiter`/`picks`/`candidates`/`lat_ms` fold into `response_text` (no schema change).
- **`call_site_recorder.py`**: `record_last_run` now returns a landed-bool (additive; all callers ignored the return); new `record_last_run_detached` shared helper used by both detached emitters (this worker + `version_gate._record_to_monitor`, which previously hand-rolled the pattern).
- **`config.py`**: the dashboard save-path validator (`update_call_site_in_yaml`) now honors the same empty-chain cli exemption the loader has — without it the new site was permanently un-editable via the dashboard (loader accepts it, save path 400'd). The exemption does **not** leak to api/dual sites (test-locked).

## Verification

- **Targeted tests green**: `tests/test_session_awareness/` (incl. new `test_worker_telemetry.py`), `tests/test_routing/test_config.py` (54→55 + shape lock), `tests/test_routing/test_config_save.py` (3 new empty-chain cli tests), `tests/test_observability/test_call_site_metadata.py` + `test_call_site_recorder.py`, `tests/test_contribution/test_version_gate.py`. Ruff clean; subsystem-map guard clean.
- **Full E2E against real infra** (real Qdrant + real DB + **real Haiku arbiter subprocess**): worker pass judged 9 candidates, arbiter ran (`arbiter=ok`, 22s, picked 1), telemetry row landed in the real `genesis.db` (`success=1`, correct model + `response_text`), worktree snapshot rendered it as a proper registered site (cost policy, dispatch, full last-run detail), and the **live deployed `/health`** surfaces the row (as `routing=false` orphan pre-merge — upgrades to a proper cell after deploy).
- **High-effort multi-agent code review** run before merge: 2 confirmed bugs (dashboard-uneditable site; `telemetry_recorded` false-positive on swallowed INSERT failures) + 4 cleanups, all fixed in the second commit.

### DIFF
--- a/config/model_routing.yaml
+++ b/config/model_routing.yaml
@@ -699,6 +699,19 @@ call_sites:
       deepseek-v4-pro (100% clean, different vendor) fallback; both free, so never_pays
       keeps a valid chain. NON-Groq deliberately \u2014 Groq's 8B is EOL 2026-08-16 and its
       gpt-oss-20b replacement breaks strict JSON.
+  ambient_arbiter:
+    chain: []
+    dispatch: cli
+    cc_model: Haiku
+    never_pays: true
+    description: >-
+      Ambient session-awareness arbiter: judges which drift-fire candidate
+      memories deserve surfacing (headless CC, model PINNED in
+      session_awareness/arbiter.py). The chain is empty BY DESIGN: no API
+      fallback exists (a failed call records an empty verdict, never a guess)
+      and the router never dispatches this site. The detached ambient worker
+      spawns CC directly and reports each run via record_last_run; registered
+      here for neural-monitor visibility and central declaration.
 retry:
   # max_total_s: aggregate wall-clock ceiling (seconds) across a single
   # route_call's retries x chain length. A safety net against pathological
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -364,7 +364,7 @@ The loops that make Genesis think between conversations.
 ```yaml subsystem-map
 entry: ambient-cognition
 modules: [awareness, perception, reflection, attention, session_awareness]
-verified: 8dac642c 2026-07-09
+verified: 780cc8de 2026-07-10
 ```
 
 - **awareness/**: the 5-min heartbeat. ~23 signal collectors (the richer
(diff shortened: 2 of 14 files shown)


WingedGuardian:
**Structural review: 1 WARNING — no ERRORs or hard blocks**

---

**WARNING — Existing test file not updated for changed return type**

`src/genesis/observability/call_site_recorder.py:27` — `record_last_run` changed its return type from `-> None` to `-> bool`, with a semantically meaningful distinction: `False` signals a swallowed INSERT failure rather than silently disappearing. The corresponding `tests/test_observability/test_call_site_recorder.py` is not in this diff.

The PR states those tests are green — that's consistent with them not asserting on a return value that didn't previously exist. `test_swallowed_insert_failure_reports_false` in the new `test_worker_telemetry.py` covers this end-to-end through the worker, but a unit-level assertion in `test_call_site_recorder.py` (e.g. that `record_last_run` returns `False` when the underlying `db.execute` raises) would make the new contract explicit and protect against a future refactor silently reverting the behavior.

Low-urgency — the end-to-end coverage is real and the PR description shows this false-positive was a confirmed bug that was found and fixed. Worth a follow-up assertion in the recorder's own test file.

---

**PII/Secrets/Wording scan: CLEAN** — no private IPs, secrets, credential references, private hostnames, or security-incident language found in added lines or PR metadata.

---

PR #1004, opened 2026-07-10
feat(backup): complete #907 crontab→systemd-timer migration + Backup tab overhaul

WingedGuardian:
## Why

The dashboard **Backup tab misreported backup state** — showing "Not scheduled" while backups actually ran and succeeded every 6h on both tiers. Root cause: `genesis-server.service` runs `NoNewPrivileges=yes`, which neutralizes the setgid `crontab` binary, so `crontab -l` returns *Permission denied* from inside the service. The crontab **read and write** paths were both silently dead.

PR #907 already established the policy (systemd `genesis-backup.timer`, installed-but-disabled) and removed crontab from onboarding — but never migrated the dashboard. **This completes #907**: the tab now drives the timer over `systemctl --user` (D-Bus, unaffected by `NoNewPrivileges` — the same path `services.py` uses), and finally answers the four questions the old tab couldn't: *is it running, where, on what schedule, what can I change.*

## What changed

**Route (`backup.py`)** — crontab helpers → `_timer_state` / `_set_timer_schedule` (OnCalendar drop-in with an additive-reset) / `_set_timer_enabled`:
- `GET /status`: drops `cron_schedule`; adds `schedule` (timer state) + `destinations` (Tier-1 GitHub + Tier-2 off-site health). `last_backup` is **allowlist-projected** so a future status field can't auto-leak through this unauthenticated route; the off-site target is auth-gated.
- `GET/POST /config`: preset `schedule_interval` (3h/6h/12h/daily) + on/off. Disable path skips interval validation.
- `POST /trigger`: `systemctl --user start --no-block genesis-backup.service` — same unit the timer fires (fixes a latent bug where an in-process `bash backup.sh` under the hardened server namespace would break gpg/tmp), and `--no-block` avoids the `Type=oneshot` blocking-start hang.

**Script (`backup.sh`)** — `_write_status` enriched: `tier2_backend`, `snapshot_id`, `snapshot_count`, `pruned_count` (JSON `null` until the GFS prune runs), `tier1_pushed`.

**Frontend** — health headline, at-a-glance schedule line, a **destinations block that makes the off-site NAS visible** (backend, retained-snapshot count, latest id; target only when authenticated), preset-interval dropdown + toggle. Redundant read-only card removed.

**Docs** — SETUP.md: dashboard-managed timer + legacy-crontab removal note.

## Testing

[...]

PR #1005, opened 2026-07-10
fix(backup): resolve Tier-2 backend via backward-compat (NAS-only → smb)

WingedGuardian:
## Why

Follow-up to #1004, found during live E2E. A NAS-only install — `GENESIS_BACKUP_NAS` set with **no** explicit `GENESIS_BACKUP_TIER2_BACKEND` selector, the common legacy setup — resolves to `smb` in `backup_backends.sh` (`_backend_resolve`), but the dashboard read the selector env directly and reported backend **`none`**.

Two problems:
1. **Misreport** — the off-site destination showed "Not configured" even though NAS backups succeed (exactly the "where do backups actually go" confusion this work set out to fix).
2. **Footgun** — the config form loaded `none`, and since `saveBackupConfig` always sent `tier2_backend`, a save could write `GENESIS_BACKUP_TIER2_BACKEND=none` and **silently disable off-site backups**.

## Fix

- `_resolved_backend()` mirrors `backup_backends.sh:_backend_resolve` (explicit selector wins; NAS-only → `smb`; else `none`). Used by `/config` GET and `_destinations`' fallback. The status-file `tier2_backend` still wins once the enriched script has run — this covers the config dropdown and the pre-first-run window.
- `saveBackupConfig` only sends `tier2_backend` when the user actually changed it, so a plain save no longer pins an explicit selector over the backward-compat resolution.

## Tests

`test_resolved_backend_backward_compat`, `test_get_config_backend_backward_compat`, `test_status_destinations_backend_backward_compat` (31 pass total). Verified live: after the first enriched backup run, `/status` shows `tier2.backend: "smb"`.

### DIFF
--- a/src/genesis/dashboard/routes/backup.py
+++ b/src/genesis/dashboard/routes/backup.py
@@ -42,10 +42,7 @@ _BACKUP_DIR = _HOME / "backups" / "genesis-backups"
 # Systemd user units (source of truth for the backup schedule).
 _TIMER_UNIT = "genesis-backup.timer"
 _SERVICE_UNIT = "genesis-backup.service"
-_TIMER_DROPIN = (
-    _HOME / ".config" / "systemd" / "user"
-    / "genesis-backup.timer.d" / "schedule.conf"
-)
+_TIMER_DROPIN = _HOME / ".config" / "systemd" / "user" / "genesis-backup.timer.d" / "schedule.conf"
 
 # Preset schedules exposed by the dashboard. The value is the SHORT OnCalendar
 # spec we write into the drop-in; systemd normalises it on load to the long form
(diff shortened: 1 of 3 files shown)


WingedGuardian (commented):
**First-pass review — PR #1005**

---

### Structural review

**Pattern violations (ERROR):** none found.

**Import hygiene (NOTE):** `_resolved_backend()` opens with a deferred local import:

```python
# src/genesis/dashboard/routes/backup.py — new _resolved_backend()
from genesis.dashboard.routes.secrets import _key_value
```

However, `_key_value` is used directly (no local import) in the existing body of `_destinations()` visible in the same diff, which means it is already a module-level name in this file. The local import is therefore either:
- **redundant** — in which case it should be removed for consistency, or
- **intentional** (e.g. guarding against a circular-import edge case or making the function self-contained) — in which case a one-line comment inside the function explaining why would prevent future readers from deleting it as "obvious cleanup."

No other structural, convention, or coverage findings.

---

**Code diff:** CLEAN — no API keys, tokens, private IPs, SSH keys, or credentials found in added lines. UNC paths in test fixtures (`//host/share`, `//secret-host/share`) are clearly synthetic. `ghp_tok` in the existing URL-stripping test is an obviously truncated fixture, not a real token.

**PR title / body / CHANGELOG:** CLEAN — no language suggesting a secret/credential cleanup or security incident.

---

PR #1006, opened 2026-07-11
feat(observability): surface Model Fusion (deliberate) on the neural monitor

WingedGuardian:
## What

The `deliberate` MCP tool — **Model Fusion** (OpenRouter Fusion: a custom panel-of-models + judge for high-stakes/contested decisions) — POSTs raw httpx **outside** the router and `call_site_recorder`, so it fired completely unobserved: absent from `call_site_last_run` and every dashboard surface. This registers it as a **non-routing call site** so each real run renders on the neural monitor, and seeds it **idle** so the node is always present even before its first run.

Follows the `ambient_arbiter` (#1002) precedent for putting an off-router call site on the monitor, adapted for a non-CC, non-routed paid consult.

## Changes

- **`deliberation_tools.py`** — `_impl_deliberate` records every deliberation via `record_last_run_detached` (the sanctioned detached seam — the health MCP server holds no server DB handle). Best-effort: a failed record is logged and swallowed so it can never break the tool. `success` mirrors `result.error`.
- **`snapshots/call_sites.py`** — `setdefault`-seeds `model_fusion` as **idle** so it renders before its first run (non-routing sites otherwise never appear); no-ops once a `last_run` row has created it **active**. Backend meta layered per-key onto both states for a self-describing payload.
- **`_call_site_meta.py`** — `model_fusion` entry with a manual `cost_policy` (never touches a routing chain, so cost is not auto-derived; no `dispatch`, no `wired:False`).
- **`neural_monitor.html`** — new single-node **"Model Fusion"** subsystem group at free grid cell `[7,7]` + border color + `SITE_META` label (`Fusion`, category `reasoning`).

## Not in `model_routing.yaml` — by design

The Fusion backend never routes through Genesis, so `model_fusion` is **not** a routing call site. The routing site count is unchanged (no `test_config` churn).

## Tests

- `test_deliberation_tools.py` (new): `_impl_deliberate` records `model_fusion` on success/failure, recording is best-effort (recorder exception can't break the tool), and the empty-question short-circuit records nothing.
- `test_call_site_metadata.py`: `call_sites()` seeds `model_fusion` **idle** with no run, and flips **active** with a recorded row.
- Targeted suites green locally (metadata + MCP: 30 passed; health-assembly + config + recorder: 132 passed).

## Verification plan (post-merge)

1. Deploy (code-only: pull main + restart genesis-server) → idle node appears on the monitor with **zero** paid calls.
2. Data-flow: `curl /api/genesis/health` shows `call_sites.model_fusion` idle; synthetic row → flips active; row removed.
3. True E2E: one paid `budget` `deliberate` from a **fresh** session/direct call (the MCP-child recording code only loads on a new CC session) → live node active with the real verdict.

### DIFF
--- a/src/genesis/dashboard/templates/neural_monitor.html
+++ b/src/genesis/dashboard/templates/neural_monitor.html
@@ -466,6 +466,10 @@ const SITE_META = {
         name: 'G. Background', desc: 'Autonomous background sessions for reflections, inbox, and surplus tasks.',
         category: 'core', frequency: 'On demand', costPolicy: 'Subscription',
     },
+    'model_fusion': {
+        name: 'Fusion', desc: 'On-demand panel-of-models consult (OpenRouter Fusion) — synthesized verdict plus explicit dissent for high-stakes decisions. Invoked via the deliberate MCP tool.',
+        category: 'reasoning', frequency: 'On demand (opt-in)', costPolicy: 'Paid (OpenRouter Fusion)',
+    },
     // ── Routing call sites — name + category only (desc/frequency/costPolicy from backend) ──
     '3_micro_reflection': { name: 'Micro Reflect', category: 'reflection' },
     '4_light_reflection': { name: 'Light Reflect', category: 'reflection' },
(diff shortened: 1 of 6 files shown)


WingedGuardian:

---

WingedGuardian:
Addressed the Codex **P2** (wired-idle mislabeled as groundwork) in 041d78e2: the monitor now special-cases a `status_reason === 'WIRED'` idle node — detail panel shows *"Idle — on-demand (not yet run)"* with a truthful "wired and available" banner, and the grid tooltip reads *"On-demand · not yet run"* instead of *"Groundwork"*. Guard is specific (true groundwork sites are all `wired=False` with non-WIRED reasons), and the seed's `status_reason=="WIRED"` is now locked by the snapshot test.

PR #1007, opened 2026-07-11
feat(guardian): F.0 — verify redeploy tree integrity before recording deploy

WingedGuardian:
## What & why

The guardian `redeploy` verb recorded `deploy_state.deployed_commit` verbatim from the SSH arg, with **tar's exit code as the only gate**. A truncated/corrupt tar stream — or an archive built with the wrong pathspec — could overwrite a healthy install and leave a good-looking `deploy_state.json` that every downstream drift check (`watchdog` code-drift, gateway-staleness, `update.sh` reconcile) then trusts. This is the credential/git-metadata resiliency track's deploy-integrity gap.

## Gateway (`guardian-gateway.sh`)

- `redeploy` accepts an optional 2nd arg = **sha256 of the tar stream**. The stream is spooled and its sha verified **before** the running guardian is disturbed (timer up, no backup churn), so a bad transfer can't touch a healthy install.
- **Required-file gate checks the archive contents** (`tar -tf` membership), not the extracted tree. The install is a **merge extract** so host-specific `config/guardian.yaml` + `secrets.env` (which live in the install dir but are *not* in the archive) survive — a post-extract check would be masked by stale files. Membership catches wrong-pathspec / partial archives, also pre-extraction.
- `deploy_state` gains `tree_sha256`; `version` advertises `redeploy_verify` + the deployed tree sha (folded into the existing `deployed_commit` read — one python spawn, since `version` runs every tick).
- Legacy 1-arg `redeploy <hash>` still works (sha skipped) for mid-rollout old gateways.
- An EXIT trap guarantees the guardian timer is restarted even if a post-stop step aborts under `set -e` (a failed self-update/cp/write must never leave the guardian DOWN); best-effort cp steps are `|| true`-guarded.
- Both arg validators hardened to bash `[[ =~ ]]` (not line-oriented `echo | grep`).

## Sender (`update.sh`)

- Materializes the archive to a `~/tmp` file (big-temp discipline), hashes it, sends the sha-checked form; falls back to the bare 1-arg form (old gateway), then `update`+retry (gateway with no redeploy verb). `mktemp` guarded so an unwritable `~/tmp` skips redeploy non-fatally rather than aborting the run.

## Tests

`tests/test_guardian/test_gateway_redeploy.py` (real script, sandboxed `$HOME`): verified + legacy deploy, sha mismatch leaves install untouched + timer never stopped, missing-member + corrupt-tar rejected pre-extract, **timer restarted when a post-stop step aborts**, broken `.local/bin` doesn't abort the deploy, `version` fields. Structural guards + `bash -n` in `test_update_host_sync.py`. shellcheck + ruff clean; 46 guardian/script tests green locally.


WingedGuardian:
Structural review: no pattern violations (ERROR). No import hygiene, missing-coverage, or Genesis-convention issues (WARNING).

**NOTE** `tests/test_guardian/test_gateway_redeploy.py` line 138 — misleading inline comment

The comment reads:
```
# New tree actually landed; old sentinel gone (clean extract, backup pruned).
```
The gateway performs a **merge** extract (`tar -xf "$SPOOL" -C "$INSTALL_DIR"`) — the PR's own inline docs explain this is intentional so host-specific files like `config/guardian.yaml` survive. That means `OLD_SENTINEL` is *not* removed by the deploy; it stays in the install dir. The assertion below the comment is correct (it only checks that the new `__init__.py` arrived, and that the spool file was cleaned up), but the "old sentinel gone" wording misrepresents merge-extract semantics and could mislead a future reader into expecting wipe-first behaviour. Incidentally, the `sandbox` fixture's own docstring correctly says "a sentinel that must survive any rollback." Suggested fix: drop "old sentinel gone" or replace with "merge extract adds new files; host-only files survive; deploy-backup pruned on success."

---

PII/Secrets/Wording scan: CLEAN — no hard blocks, no soft warnings, no wording issues in PR title or body.

---

WingedGuardian:
Fixed in bdb60530 — the unguarded 1-arg fallback is now gated on the gateway's `redeploy_verify` capability (parsed from the `version` output already fetched). A verify-capable gateway that rejects the verified form (e.g. sha mismatch) is reported non-fatal and retried next run; the bare form is only used for gateways that don't advertise the capability. Good catch — this was the crux of F.0.

WingedGuardian:
Re: the re-posted P1 at update.sh:278 — this is already fixed in this same commit. The bare-form `elif _redeploy_ssh "redeploy $DEPLOY_HASH"` is now guarded by the preceding `elif [ "$_gv_verify_capable" = 1 ]`: a verify-capable gateway that fails the verified form takes the non-fatal branch and **never reaches** the unguarded retry. The bare form is only evaluated when the gateway does NOT advertise `redeploy_verify` (genuinely old). Merging with a documented override since the finding is fixed, not accepted.

PR #1008, opened 2026-07-11
fix(guardian): F.0 hotfix — redeploy membership gate fails on real archives (pipefail+SIGPIPE)

WingedGuardian:
## Hotfix for #1007 (F.0)

Found during live E2E of #1007. The redeploy required-file gate used:

\`\`\`bash
printf '%s\n' "$TAR_LIST" | grep -qxF "$_req"
\`\`\`

under \`set -o pipefail\`. On a **real** git archive the \`tar -tf\` listing is ~86 KB — over the 64 KB pipe buffer — so when \`grep -q\` short-circuits on an early match and exits, \`printf\` gets **SIGPIPE** and \`pipefail\` makes the pipeline exit non-zero. The gate then reports **present** required files as missing and rejects **every** real redeploy with \`archive missing required files\`.

The gate runs on both the verified and legacy redeploy forms, so the deployed F.0 gateway rejected all redeploys. The unit fixtures used ~4-entry tars (<64 KB listing → no SIGPIPE), so CI stayed green — the bug only surfaced live.

## Fix

Whole-line membership via a **pure-bash** substring match against the newline-wrapped listing (\`case "$_tar_wrapped" in *$'\n'"$_req"$'\n'*\`) — no subprocess, no pipe, no SIGPIPE, correct at any archive size.

## Regression test

\`test_verified_deploy_with_large_archive\` builds a ~3000-entry archive (listing >> 64 KB) with the required files present. Passes with the fix; **verified to FAIL** against the pre-fix piped-grep gate by swapping the old block back in. 42 gateway/redeploy tests green; shellcheck clean.

## Deploy note

The currently-deployed gateway can't accept redeploys, so this ships to the host via full-shell recovery (the redeploy verb is the thing that's broken), then a verified redeploy is confirmed end-to-end.

### DIFF
--- a/scripts/guardian-gateway.sh
+++ b/scripts/guardian-gateway.sh
@@ -465,11 +465,20 @@ PYEOF
         # like the sha check — runs BEFORE the running guardian is disturbed.
         TAR_LIST="$(tar -tf "$SPOOL" 2>/dev/null || true)"
         REDEPLOY_TREE_OK=true
+        # Whole-line membership via a PURE-BASH substring match — deliberately
+        # NOT `printf '%s\n' "$TAR_LIST" | grep -qxF`. Under `set -o pipefail`,
+        # grep -q short-circuits on a match and the writer (printf) then gets
+        # SIGPIPE, so the pipeline exits non-zero and a PRESENT file reads as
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

PR #1011, opened 2026-07-12
fix(memory): heal embed-status orphans, restore project_type on recovery, unblock the store hot path

WingedGuardian:
Cross-store integrity + hot-path latency fixes for the memory pipeline, which spans three stores that must stay reconciled: SQLite `memory_metadata` (`embedding_status`), the `pending_embeddings` retry queue, and Qdrant vectors.

## Integrity (`fix`)
- **Failed embeds no longer leave a memory in permanent limbo.** When the embedding provider gave up on a queued memory, its metadata status was left saying "still queued" forever — a stale marker with no vector behind it that could drive a doomed write to the vector store. The recovery worker now records the failure in `memory_metadata` too (metadata-first, self-healing), and a maintenance sweep (`reconcile_orphaned_metadata`) heals any rows that were already stranded (aged, with no queue row).
- **Re-queued failures stop lying.** `reset_failed_to_pending` now flips the `memory_metadata` mirror back to `pending` in the same transaction, so a retried item isn't reported as failed.
- **Supersession only touches Qdrant when a vector exists.** `_mark_superseded`'s guard tightened from `!= 'fts5_only'` to `== 'embedded'`, so it never fires an `update_payload` against a `pending`/`failed` row that has no point.
- **Recovered memories keep their `project_type` recall filter.** `project_type` was written to the Qdrant payload only, so it was lost on the FTS5-only fallback path; it's now mirrored to a tag breadcrumb (like `wing`/`life_domain`) and re-hydrated on re-embed.
- **Deleting a memory now clears its entity mentions too** (previously left dangling rows keyed by the gone `memory_id`).
- **A maintenance tripwire counts FTS↔metadata drift both directions** (via an O(n log n) set-difference) so a store-path/migration regression surfaces instead of silently accumulating.

## Hot path (`perf`)
- `store()` / `_mark_superseded()` / `delete()` made blocking synchronous vector-store HTTP calls directly on the event loop; they now run off-thread (`asyncio.to_thread`, matching the existing background paths), so a slow round-trip during a memory write no longer stalls concurrent work.

## Notes
- No schema change; introduces a fourth `embedding_status` value `'failed'` (the column has no CHECK constraint).
- No rollbacks on the shared serialized DB connection (a rollback there would discard unrelated coroutines' pending writes); divergence is handled by ordering + a self-healing re-drain.
- Tests: new coverage for the recovery failure/success paths, the reconcile sweep + age-gate, the mirror reset, the supersede guard (RED-proofed), `project_type` recovery, `delete()` cascade, the drift counter, and the hot-path offload.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -181,6 +181,20 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   now restores those fields. Separately, keyword-only memories were being scored as
   if created just now (maximum freshness), letting old notes outrank genuinely recent
   ones; they now use their real creation time.
+- **A failed re-embed no longer leaves a memory in a permanent limbo state, and
+  recovered memories keep their project_type recall filter.** When the embedding
+  provider gave up on a memory, its status was left saying "still queued" forever —
+  a stale marker that could trigger doomed writes to the vector store for a memory
+  that has no vector. Failed embeds are now recorded truthfully, a maintenance sweep
+  heals any that were already stranded, and deleting a memory now also clears its
+  entity mentions. Memories re-embedded after an outage also keep their `project_type`
+  so they stay visible in project-scoped recall.
+- **Storing a memory no longer briefly stalls other work while it talks to the
+  vector store.** The store, supersede, and delete paths made blocking vector-store
+  HTTP calls directly on the event loop, so a slow round-trip could momentarily
+  freeze concurrent Telegram, dashboard, and reflection activity. Those calls now
+  run off-thread (matching the background paths), keeping the system responsive
+  under load.
 - **Telegram no longer drops messages or breaks approval buttons when the
   legacy bridge gets started alongside the server.** Two Genesis processes
   polling the same bot token split incoming updates between them (observed:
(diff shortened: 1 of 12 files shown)


WingedGuardian:
## Structural review

**1 ERROR — must fix**

**`src/genesis/resilience/embedding_recovery.py` — raw SQL against `memory_metadata` outside a CRUD module**

The new failure path adds a direct `db.execute` against `memory_metadata` in a non-CRUD module:

```python
await self._db.execute(
    "UPDATE memory_metadata SET embedding_status = 'failed' "
    "WHERE memory_id = ?",
    (item["memory_id"],),
)
```

The success-path equivalent (`SET embedding_status = 'embedded'`) was pre-existing; the failure-path update is new and hits the same violation. Both belong in `genesis.db.crud.memory` — a small helper like `set_embedding_status(db, memory_id, status)` covers both callsites and keeps the raw SQL in one place. The metadata-first ordering and no-rollback constraints documented in the comments are orthogonal to which module owns the SQL.

---

---

WingedGuardian:
Addressed both review findings in `a3f910a5`:

- **Structural ERROR (raw SQL in a non-CRUD module):** added `crud.memory.set_embedding_status(db, memory_id, status)`; the recovery worker's success and failure paths now route through it.
- **Codex P2 (confirm no Qdrant point before marking orphans failed):** this turned out to be real for live data, not just legacy-theoretical — I checked, and **all current orphans have Qdrant vectors** (they're successful embeds whose metadata update was lost, then the queue row was reaped). Marking them `failed` would have been wrong. The reconcile is now **Qdrant-aware**: it surfaces candidates via `query_orphaned_metadata`, checks each point's existence, and sets `embedded` when a vector exists, else `failed`. It moved from the SQLite-only GC into `EmbeddingRecoveryWorker` (which holds the Qdrant client), wired into the recovery cycle after drain; the FTS↔metadata drift tripwire stays in GC. E2E against a live-DB copy heals the orphans to `embedded`.

PR #1012, opened 2026-07-12
feat(guardian): F.1b — guardian-side git-health probe + REVERT_CODE preflight

WingedGuardian:
## What & why

Second half of F.1 (git-health detection). F.1a (#1010) added the **container-side** probe; this adds the **guardian-side** live probe — the **primary** detector, because the rootfs-read-only outage this guards against can take the container's own awareness-loop alerting down with it (a degraded/dead genesis-server).

## Changes

- **NEW `guardian/git_watch.py`**: `check_container_git_and_alert` — a read-only `incus exec` probe (HEAD resolvable, remote configured, rootfs write-probe) piped to `bash -s`, kept independent of the container venv so it survives a broken `.venv` (same discipline as `cred_watch`). Pure `decide()` ladder: `confirm_ticks` consecutive unhealthy before the first WARN, realert on cadence, INFO + clear on recovery. **Probe-exec failure = container unreachable = NO signal** (the state machine owns "down"). Enriches the alert body with the container's shared-mount verdict (F.1a). Never raises.
- **`config.py`**: `GitHealthConfig` (`enabled`, `confirm_ticks=2`, `realert_hours=6`, `check_timeout_s=30`) via `_build_sub` + `guardian.yaml` section.
- **`check.py`**: wire `_check_container_git_and_alert` into `run_check` after the cred check (never-raise).
- **`recovery.py`**: `_revert_code` **preflight** — refuse REVERT_CODE when the container's git is unhealthy (`rev-parse` + `config-get`, 15s) so the recovery ladder escalates instead of burning a doomed `git stash`/`git revert`.

## Tests

`decide` matrix, probe parsing, orchestrator (no-signal / confirm / warn / resolve / realert-damping / disabled / never-raises), REVERT_CODE preflight (block + proceed), config load. 34 new green; full `tests/test_guardian/` (783) passes; ruff clean.

Independent of F.1a's Python API (reads only the runtime verdict-file contract), so it merges/tests standalone; the verdict-enrichment is best-effort. Deploys to the **host** via `update.sh` (F.1a deploys to the container).


WingedGuardian:

---

WingedGuardian:
Addressed both Codex findings in 10306316:

**P1 (critical — preflight stalls the ladder):** Moved the git-health gate out of `_revert_code` (where returning `False` registered as a *failed* attempt → `confirmed_dead` → ladder stall) into `execute()`, **before any attempt is recorded**. Instead of failing, it now **redirects `REVERT_CODE → SNAPSHOT_ROLLBACK`** — the safer rung that restores a healthy `.git` + working code (and repairs a read-only rootfs) in one move. Placed after the `ESCALATE` early-return and before the shared backoff check (backoff is not per-action except `IO_TRIAGE`, verified in `state_machine.py`, so redirecting first is safe).

**P2 (read-only rootfs passes read-only preflight):** The new `_container_git_healthy()` reuses `git_watch.probe_container_git`, whose probe includes a `.git` **write-probe**, so a read-only rootfs is now correctly detected. It fails **open** (True) on an inconclusive/unreachable probe so a flaky probe never blocks a recovery that might otherwise succeed.

Also aligned `git_watch` probe config check to `git config --list` (parseability) rather than remote-presence, matching the F.1a P2-1 fix. Tests rewritten (`TestRevertCodeGitPreflight`): redirect-to-rollback / revert-proceeds-when-healthy / fail-open-on-inconclusive — 19 passed. Re-review appreciated.

WingedGuardian:
@codex review — please re-review commit 10306316 which addresses the P1 (preflight stall) and P2 (read-only rootfs missed by read-only preflight) findings from your prior review.

WingedGuardian:
Fixed the re-review P1 in 677ab131.

**Root issue:** the redirect lived inside `recovery_engine.execute()`, which runs *after* Gate-2 approval — so a user who approved `REVERT_CODE` would silently get the more-destructive `SNAPSHOT_ROLLBACK`. That breaks the gate-everything / user-sovereignty contract.

**Fix — redirect the *proposal* before approval:**
- `recovery.py` reverted entirely to `origin/main` (no execute()-level swap, no preflight). The engine now executes exactly the approved action.
- New reusable `git_watch.container_git_supports_revert(config)` (wraps the live probe incl. the `.git` write-probe; fails **open** on inconclusive).
- In `check.py::_execute_recovery_with_approval`, right after the post-Gate-1 re-diagnose and **before** `write_diagnosis_result` / the ESCALATE check / the Gate-2 prompt: if the proposal is `REVERT_CODE` and git is unhealthy, `dataclasses.replace` the frozen `DiagnosisResult` → `SNAPSHOT_ROLLBACK` (+ a `likely_cause` note). The user then sees and approves the action that actually runs.

Choke-point verified: all 3 `execute()` callers are inside this function; only the Gate-2-approved call can carry `REVERT_CODE` and it is unreachable without passing through the redirect. Tests assert `execute()` receives `SNAPSHOT_ROLLBACK` and the Gate-2 prompt names it (not `REVERT_CODE`); healthy git leaves `REVERT_CODE` untouched. 787 guardian tests pass. @codex review

WingedGuardian:
@codex review

PR #1013, opened 2026-07-12
fix(knowledge): coerce malformed LLM fields in distillation (silent chunk-drop)

WingedGuardian:
## Problem

`DistillationPipeline.distill()` processes raw dicts parsed from an LLM's JSON response (`_parse_llm_response`, no type validation). Several fields were passed through raw, and because per-chunk tasks are gathered with `asyncio.gather(..., return_exceptions=True)`, a `TypeError`/`AttributeError` on any one field is **swallowed and silently drops the entire chunk's knowledge units** (only a `logger.warning`).

Concretely, when the LLM ignores the "return a JSON array" instruction:

- **`confidence` as a string** ("0.9", "high") → `TypeError` at the `confidence < 0.3` filter → whole chunk dropped.
- **`tags` as a string** → flows into `KnowledgeUnit.tags`, later crashes `unit.tags + [...]` (str + list) at the store call site.
- **`caveats` as a string** → truthy, so `caveats.append(...)` raises `AttributeError` whenever `user_context` is supplied (the normal conversational-ingest path) → whole chunk dropped.
- **`relationships` as a string** → passed through as a bare string instead of a list.

## Fix

Coerce all four LLM-controlled fields at `KnowledgeUnit` construction:

- `_coerce_confidence` — float-coerce with `0.85` fallback (non-numeric/NaN) and clamp to `[0.0, 1.0]`, applied **before** the low-confidence filter.
- `_normalize_tags` — → `list[str]`, comma-splitting a string (tags are short labels).
- `_normalize_str_list` — → `list[str]` for `caveats`/`relationships`, keeping a string **whole** (a caveat sentence must not be comma-split).
- `_clean_str_items` — shared element cleaner: drops `None`/blank, stringifies (fixes `str(None)` → literal `"None"`).

Happy-path behavior for well-typed inputs is unchanged.

## Tests

- 7 integration tests through the real `distill()` path (string / garbage / out-of-range confidence no longer drop the chunk; string & None-bearing tags normalized; string `caveats` + `user_context` survives; string `relationships` normalized).
- 3 exhaustive helper unit tests.
- Existing distillation + orchestrator suites green; full `test_knowledge/` = 108 pass.
- E2E: real `distill()` with every field malformed at once → chunk survives, all fields correctly coerced.

## Review

Built test-first (TDD). A code-review pass caught that the initial confidence/tags-only fix was incomplete — the same silent-chunk-drop class lived in `caveats` (+`relationships`); this PR extends the coercion to the full field class.

### DIFF
--- a/src/genesis/knowledge/distillation.py
+++ b/src/genesis/knowledge/distillation.py
@@ -161,11 +161,48 @@ def _parse_llm_response(response_text: str) -> list[dict]:
         end = text.rfind("]")
         if start != -1 and end != -1 and end > start:
             try:
-                return json.loads(text[start:end + 1])
+                return json.loads(text[start : end + 1])
             except json.JSONDecodeError:
                 pass
 
     logger.warning("Failed to parse LLM distillation response as JSON")
+    return []
+
+
+def _coerce_confidence(value: object, *, default: float = 0.85) -> float:
+    """Coerce an LLM-provided confidence to a float in ``[0.0, 1.0]``.
+
+    The distillation LLM occasionally returns ``confidence`` as a string
+    ("0.9", "high") or omits/garbles it. An un-coerced string raises
+    ``TypeError`` at the ``confidence < 0.3`` filter — and because chunk tasks
+    are gathered with ``return_exceptions=True``, that silently drops the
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

PR #1018, opened 2026-07-12
fix(runtime): idle-aware process reaper — spare active terminal sessions

WingedGuardian:
## Problem

The hourly process reaper killed long-running `claude` (Claude Code) processes
by **raw age** — any process older than 7 days since start was terminated,
regardless of whether it was actively in use. An interactive session that had
been open and active for a week was killed the moment it crossed the age floor,
even seconds after the user's last action.

## Fix

Switch the `claude` reap policy from **age** to **idle**, and extract the reaper
from `learning.py` into a testable `process_reaper.py` seam. A `claude` process
is now a reap candidate only when it is **all** of:

1. past the 7-day age floor, **and**
2. idle beyond the 7-day activity window (no recent activity marker), **and**
3. detached from any live terminal (controlling tty not in tmux panes ∪ login ttys).

This is a strict subset of the old age-only rule, so it can never terminate
something the old reaper spared — it only spares active/attached sessions the
old rule wrongly killed. `opencode`/browser-helper targets keep the original
age policy.

Activity markers are written by a lightweight `UserPromptSubmit`/`PostToolUse`
hook that walks the parent-process chain to the owning `claude` process. The
tty check is a spare-only backstop (it can prevent a kill, never cause one), so
correctness rests on the marker — the one reliable "actively in use" signal.

## Safety

- **Ships in dry-run**: logs `WOULD KILL` and records an observation, but never
  signals a process.
- **Auto-arm is gated on proof-of-life**: the reaper only arms after it has
  observed a fresh activity marker on a live process AND a clean observation
  window has elapsed — it will not go live while its load-bearing signal is
  absent. State persists in a human-editable JSON file; an env kill-switch
  forces dry-run.
- On arming and on any interactive-process kill, the owner is notified with
  instructions to disable.

## Tests

22 tests covering the classifier, dry-run/armed orchestration, the auto-arm +
proof-of-life lifecycle, the kill-switch, non-claude age paths, marker GC,
protected PIDs, and job wiring.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -80,6 +80,15 @@
                         "timeout": 500
                     }
                 ]
+            },
+            {
+                "hooks": [
+                    {
+                        "type": "command",
+                        "command": "bash ${CLAUDE_PROJECT_DIR}/scripts/hooks/session_activity_touch.sh",
+                        "timeout": 2000
+                    }
+                ]
             }
         ],
         "PreToolUse": [
@@ -383,6 +392,15 @@
                     {
                         "type": "command",
                         "command": "${CLAUDE_PROJECT_DIR}/.claude/hooks/genesis-hook plan_bookmark_hook.py",
+                        "timeout": 2000
+                    }
+                ]
+            },
+            {
+                "hooks": [
+                    {
+                        "type": "command",
+                        "command": "bash ${CLAUDE_PROJECT_DIR}/scripts/hooks/session_activity_touch.sh",
                         "timeout": 2000
                     }
                 ]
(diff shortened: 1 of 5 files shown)


WingedGuardian:

---

WingedGuardian:
Both Codex findings from the review of `ac5981a3` are addressed in `9d89c59e` (pushed ~4 min after the review):

- **P1 (register the activity hook)** — `session_activity_touch.sh` is now registered in the tracked `.claude/settings.json` on both `UserPromptSubmit` and `PostToolUse`, so a normal install writes markers out-of-the-box and the proof-of-life gate can be satisfied.
- **P2 (require an alert before auto-arming)** — `_notify_owner` now returns delivery status and `_maybe_arm` only persists `dry_run=False` when the owner veto notice was actually delivered; otherwise it stays in dry-run and retries. Added `test_no_arm_when_notification_fails`.

Merging with override since the flagged commit's findings are resolved in a later commit and CI is green.

PR #1019, opened 2026-07-12
feat(infra): infrastructure body schema — machine self-knowledge profile

WingedGuardian:
## What

A new `infra_profile` subsystem: Genesis's **body schema** — programmatic, always-current knowledge of the machine it runs on.

- **Deterministic fact collectors** (container plane): OS/distro, virtualization nesting, CPU (model/flags/mitigations/governor), memory (cgroup limit, swap, THP), storage (mount table, IO scheduler, tmpfs sizes), kernel + load-bearing sysctls, SQLite pragmas (read-only connection), Qdrant collection config, network topology, systemd units, tool versions, ulimits, clock/NTP, NUMA/entropy/steal-time.
- **Facts vs metrics split (load-bearing)**: only slow-changing `facts` are content-hashed; volatile readings (`metrics`) render in the doc but never churn the hash.
- **Drift is a signal**: a section hash change emits a dedup-gated `infrastructure_drift` observation naming the changed key paths.
- **LLM annotation layer** (new routing call site `46_infra_annotation`): per-section operational gotchas, pinned to the fact hash they were derived from — unchanged hash = zero LLM spend; failed regeneration keeps the old note, flagged stale at render time.
- **Outputs** under `~/.genesis/infrastructure/`: `profile.json`, `annotations.json`, rendered `INFRASTRUCTURE.md` (+ a copy on the guardian shared mount).

## Consumers wired in this PR

- Boot step (non-blocking, delayed) + daily CronTrigger refresh
- `infrastructure_profile` MCP tool (read view + facts-only refresh; flock-guarded cross-process)
- Sentinel diagnostic context gains an "Infrastructure Body Schema" digest
- User-level CLAUDE.md `container-specs` block auto-populated (marker-idempotent; re-rendered by `update.sh` post-deploy)
- Backup Tier-1 includes the profile dir; dev-skill environment reference points at the doc

The host/hypervisor plane (via the guardian gateway) ships in a follow-up PR; until then those sections render "not visible from this vantage" — the degradation contract is stable API from day one.

## Notable fixes folded in from review

- `sentinel/monitor.py` carried a latent `router.call`/dict-response bug (dormant call site 37) — fixed alongside the same pattern here; router fakes in tests now implement the real `route_call → RoutingResult` contract so interface drift can't be masked.
- Routing config test call-site count 55 → 56.

## Tests

`tests/test_infra_profile/` (fixture /proc//sys trees via injectable roots, read-only SQLite proof, hash-pin/keep-old annotation semantics, drift dedup, render staleness, CLAUDE.md block idempotence) + a live smoke test asserting stable sections hash identically across back-to-back runs on a real system.


WingedGuardian (commented):
## First-pass review: 1 ERROR · 7 WARNINGs · 4 NOTEs

PII/Secrets/Wording scan: **CLEAN** — no private IPs, secrets, credentials, or sensitive wording found in diff or PR metadata.

---

## ⛔ ERROR (must fix before merge)

### Raw SQL against genesis.db — `src/genesis/infra_profile/collectors/sqlite_facts.py`

`collect_sqlite_facts()` opens a raw `sqlite3` connection directly to the live genesis.db path and executes PRAGMA queries:

```python
sqlite3.connect(f'file:{db_path}?mode=ro', uri=True)
conn.execute(f'PRAGMA {name}')
```

The connection is read-only, which is intentional, but the no-raw-SQL rule has no read-only exemption outside of migrations. The module comment even acknowledges it is deliberately bypassing the runtime's `SerializedConnection`.

**Fix:** Either move the PRAGMA reads into a thin `db.crud` helper that the existing runtime connection can call, or document an explicit carve-out in the architecture rules if bypassing the shared connection is genuinely necessary here (e.g. the runtime connection is unavailable at collector run-time and the read-only file-URI mode is required for non-WAL safety).

---

## ⚠️ WARNINGs

### 1–6. New `src/` modules without isolated test coverage

The following files were added to `src/` but have no dedicated test file. The `test_infra_profile/` suite is solid and the live smoke test is useful, but several load-bearing modules are only exercised end-to-end:

| File | Key untested scenarios |
|---|---|
| `src/genesis/infra_profile/service.py` | merge logic, flock short-circuit, stage-failure containment, `_cross_process_lock()` paths |
| `src/genesis/infra_profile/collectors/qdrant_facts.py` | Qdrant unreachable / malformed JSON; per-collection detail parsing |
| `src/genesis/infra_profile/collectors/host.py` | `guardian=None` branch; non-None guardian GROUNDWORK path |
| `src/genesis/mcp/health/infrastructure_profile.py` | `refresh=True` path; no-profile error; `_get_db()` → None |
| `src/genesis/runtime/init/infra_profile.py` | delayed-task scheduling; tracked_task registration; exception-containment |
| `src/genesis/surplus/jobs/infra.py` | pause-check early-return; pause-check failure paths |

`service.py` is the highest-priority gap — the cross-process flock coordination and merge logic are the core of the subsystem.

### 7. Stdlib imports deferred into function bodies — `src/genesis/infra_profile/collectors/container.py`

Three stdlib modules are imported inside function bodies rather than at module top level:
- `import os` inside `collect_storage()`
- `import json as _json` inside `collect_network()`
- `import sqlite3` inside `collect_versions()`

None have import-side-effects justifying deferral. Should be hoisted to the module-level import block alongside `asyncio`, `logging`, `platform`, `shutil`, `socket`, `pathlib`.

---

## 📝 NOTEs

### 1. Missing directory creation for shared mirror path — `src/genesis/infra_profile/service.py`

`atomic_write_text(SHARED_DOC_PATH, …)` targets `~/.genesis/shared/infrastructure/INFRASTRUCTURE.md`, but `SHARED_DOC_DIR` is never `mkdir`'d in this PR (unlike `PROFILE_DIR` which is created in `_cross_process_lock()`). On a fresh install the shared mirror will silently never populate because the write is inside a broad `except Exception` guard.

### 2. Tight coupling to private attribute — `src/genesis/mcp/health/infrastructure_profile.py`

`_get_db()` accesses `health_mcp_mod._service` (a private module-level attribute). If `health_mcp` renames or restructures `_service`, this silently returns `None` at runtime with no error. Consider a proper accessor or injecting the DB handle at MCP registration time.

### 3. Unreachable dead branch — `src/genesis/infra_profile/render.py`

In `_fmt_bytes()`: `if not isinstance(value, (int, float)) or value is None:` — the `or value is None` clause is unreachable because `isinstance(None, (int, float))` is already `False`. Simplify to `if not isinstance(value, (int, float)):`.

### 4. Missing docstring on public `main()` — `src/genesis/infra_profile/__main__.py`

Module docstring covers both CLI modes, but `main()` itself has no docstring. Fine to address in follow-up.

---

WingedGuardian:
Review findings addressed in the two `fix(infra)` commits (branch rebased onto main):

**ERROR — raw SQL:** PRAGMA reads moved to `db/crud/pragma_facts.py`; the collector keeps path/size handling and degradation policy. The separate `mode=ro` connection is retained inside the crud helper (collectors run in processes without the runtime's `SerializedConnection` — rationale in the module docstring).

**Codex P2s:** `record_failure(job_id, error)` full signature; the event-bus fallback now awaits `emit()` (and the test fake is async so a sync fake can't mask the contract again).

**W1–6 (test coverage):** added `test_service.py` — merge semantics, `_recently_refreshed`, cross-process flock winner/loser, full facts-only refresh with redirected persistence, and stage-failure containment — plus qdrant unreachable/malformed-JSON and both host-plane branches. `init/infra_profile.py` and the job's pause-path remain covered only via their thin delegation shape; accepted for PR1.

**W7:** stdlib imports hoisted. **N3/N4:** fixed.

**N1 (shared-mirror mkdir) — refuted:** `atomic_write_text` creates parent directories (`src/genesis/util/atomic.py:21` — `path.parent.mkdir(parents=True, exist_ok=True)`), so the fresh-install mirror populates; also covered by `test_refresh_end_to_end_facts_only` asserting the `shared/` copy exists in a clean tmp tree.

**N2 (`_get_db` private attr):** matches the established pattern in this MCP server (`inbox_digest.py`); keeping consistent rather than introducing a second accessor style here.

PR #1025, opened 2026-07-12
feat(guardian): nightly host CC/Node pin-alignment timer + runtime linger health signal

WingedGuardian:
Tier-2 guardian/infra closers — two low-risk, additive, alert-only/idempotent changes that prevent silent host drift (the class that has bitten before, where the guardian's recovery brain goes dark and is only noticed mid-incident).

## G3 — nightly host CC/Node pin-alignment timer
Host CC/Node re-alignment currently runs ONLY when `scripts/update.sh` is invoked, so a repo pin bump leaves the host guardian's `claude -p` recovery brain lagging the pin until the next manual update.

- Factored the host-sync logic out of `update.sh`'s `_sync_deploy_targets` into a shared `cc_align_host_sync()` in `lib/cc_version.sh` (update.sh delegates; net −57 lines). It **mutates the existing `HOST_CC_DEGRADED` global** (append pattern) — preserving update.sh's exact accumulation semantics — and is non-fatal by contract (always returns 0, called `|| true`).
- New `scripts/cc_align_host.sh` timer entrypoint + host-only systemd templates (`genesis-cc-align.{service,timer}`). Host-only by design: it only issues gateway SSH calls, so it needs no local sudo and is safe under `NoNewPrivileges` + `ProtectSystem=strict`.
- `flock -n` single-flight guard against overlapping timer runs; a genuine heal failure exits non-zero (unit enters `failed`, visible in `systemctl status`); a merely-unreachable host is not a timer failure.
- The container's own CC is already aligned on every update.sh run, so there is no container leg.

## F4 — runtime systemd-linger health signal (host + container)
systemd linger is enabled once at install; nothing verifies it at runtime. If it is ever disabled, this user's timers/services die silently on the next logout — host-side that kills the guardian's units, container-side it kills genesis-server itself.

- Gateway `version` op emits a **fail-safe `host_linger` tri-state** (true/false/null) — `null` (never a bogus `false`) on any `loginctl` probe error.
- Two INDEPENDENT reconciler legs (mirroring the shipped CC-auth signal), separate episode state, distinct alert topics, alert-only (no privileged actuator this ship):
  - **host leg** reads `host_linger` from the version payload (old gateway without the field → skip; `None` → never false-alarm).
  - **container leg** runs a LOCAL `loginctl` probe, dispatched from `check_and_recover` on **every** tick — deliberately NOT behind the code-drift/branch/host-reachability early-returns, so a feature-branch checkout or a down host can't silence a purely-local check.

## Verification
- 154 targeted tests pass, including `test_update_host_sync.py` (the guard for the exact logic factored) staying green, new `test_cc_align_host.py` (drift→update-cc/update-node behavioral coverage), and `TestHostLingerReconciler` + `TestContainerLingerReconciler`. ruff / shellcheck / `bash -n` clean.
- Live E2E: gateway emits `host_linger=true`; `cc_align_host.sh` no-op-aligns the real host and exits 0; flock skips a concurrent run; reconcilers behave correctly against real `loginctl`; templates render clean.
- Genesis-architect design review pre-build (container-leg placement + timer failure surfacing fixed).

## Deploy note
Touches host-deployed paths (`guardian-gateway.sh`, `update.sh`, `lib/cc_version.sh`, systemd templates) → **Host-Deploy Gate applies**: after merge, `scripts/update.sh` on the guardian machines, then confirm the gateway reports `host_linger` and `genesis-cc-align.timer` is `active`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,19 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **The recovery brain no longer drifts behind a version bump.** A nightly
+  job re-aligns the host's Claude Code and Node.js to the pinned versions.
+  Previously this happened only when you ran an update, so a pin bump could
+  leave the host's autonomous-recovery Claude Code lagging for days. It heals
+  only real drift, is a clean no-op when everything is already aligned, and a
+  failed alignment now shows up as a failed timer instead of going quiet.
+
+- **Genesis notices if systemd "linger" gets turned off.** Linger is what
+  keeps Genesis's background services alive after you log out — if it is ever
+  disabled, everything dies silently at the next logout. A new health check
+  watches both the host and the container and raises a single alert (with the
+  exact re-enable command) when linger is off, before the next logout can bite.
+
 - **Code quality is now a measured series, not a feeling.** A new
   `dev_quality` dimension in the weekly eval snapshots tracks review
   findings per merged PR (by severity, harvested from the PR bots' inline
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -59,7 +59,10 @@ pollers split updates and break approval buttons),
 `genesis-tmp-watchgod.service` (/tmp protection), `genesis-watchdog.timer`
 (health check), `genesis-backup.timer` (6h encrypted backup via
 `scripts/backup.sh`), `genesis-disk-hygiene.timer` (daily worktree reaping, cache reclaim, `~/tmp`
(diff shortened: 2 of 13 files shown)


WingedGuardian (commented):

---

WingedGuardian:
Addressed the P2 (thanks @chatgpt-codex-connector). `install.sh` was enabling timers via a hardcoded per-timer list (watchdog + disk-hygiene only), so a fresh install left `genesis-cc-align.timer` rendered-but-dead. Fixed the **class** rather than the instance: `install.sh` now uses the same generic timer-template loop `bootstrap.sh` uses (enable --now every rendered timer except the deliberately-opt-in `genesis-backup.timer`), so future timers can't regress the same way. Added a structural guard test. (d72ee71c)

PR #1030, opened 2026-07-12
fix(routing): correct anthropic OpenRouter slugs + reasoning chain for infra annotations

WingedGuardian:
## What

Two fixes to how the infrastructure body-schema annotations (call site `46_infra_annotation`, added in #1019) are routed.

### 1. Correct the invalid anthropic OpenRouter model slugs (pre-existing bug)

`openrouter-sonnet` and `openrouter-opus` were defined with Anthropic's **dated API model IDs** (`anthropic/claude-sonnet-4-6-20250514`), which OpenRouter rejects: `400 "…is not a valid model ID"`. Verified the correct slugs against OpenRouter's live model list → `anthropic/claude-sonnet-4.6` / `anthropic/claude-opus-4.6` (matching the already-correct `openrouter-haiku` format).

This was latent: `openrouter-sonnet`'s only prior reference was the reserved/unwired `10_cognitive_state`, so the bad ID was never exercised until #1019's call site 46 became its first heavy live caller — every annotation `400`'d and fell back to free.

### 2. Re-point call site 46 to a reasoning chain (better model fit)

Generating operational gotchas from structured infra facts is **failure-mode reasoning over data**, which fits a reasoning model better than a general-chat model. This mirrors Genesis's existing `judge` call site (quality-critical technical evaluation, also deepseek). New chain: `openrouter-deepseek-v4 → nvidia-nim-deepseek → gemini-free → mistral-large-free → openrouter-free`. Better-matched, cheaper than sonnet, and no longer dependent on the external Claude slug. The free general models produced high-quality annotations on the live first run and remain robust fallback.

## Verification

Config loads; both slugs corrected; call site 46 chain resolves to defined providers; `tests/test_routing/test_config.py` green (56 call sites, membership intact). Correct slugs confirmed against OpenRouter's `/api/v1/models`.


WingedGuardian:

---

PR #1032, opened 2026-07-13
feat(eval): WS-1 A4 — LongMemEval external memory benchmark harness

WingedGuardian:
## WS-1 A4 — LongMemEval external memory benchmark harness

Adds a local runner for **LongMemEval** (ICLR 2025), an external long-term-memory
QA benchmark, so we can measure Genesis's memory against a published number.

### What it does
Per question: build a **throwaway** Genesis memory store from scratch (in-memory
Qdrant + a temp SQLite via `init_db`) — **zero production contact by
construction**, no snapshot, no prod-delta probe — ingest that question's
haystack turn-level as `first_party` episodic memories, then for each arm run
`recall → answer (gpt-4o reader) → judge (gpt-4o)`. Reports per-arm,
per-question-type accuracy plus an evidence-recall diagnostic; persists one
`eval_runs` row per arm (`model_profile=longmemeval:<arm>`).

### Why two query arms (the load-bearing design decision)
A de-risking spike found that feeding a **raw** LongMemEval question to recall
returns **0 FTS hits** (FTS5 has no stemming: "graduate" ≠ "graduated", and the
full-question AND over-constrains), so only the weak vector arm carries it — a
naive harness would make Genesis look **worse** than it is. The harness
therefore runs **both** a raw-question arm and a **keyword-extracted** arm (×
rerank on/off) and reports each, so the number *and* its sensitivity to phrasing
are both visible. The gap is question-type-dependent: on lexical single-session
questions keyword extraction lifted a small sample from 0.67→1.0, whereas on
temporal-reasoning (where the vector arm already recalls the evidence) it barely
moves — exactly the kind of nuance "report both" is meant to surface.

### Comparability
Judge = the exact upstream `gpt-4o-2024-08-06`, prompts **ported verbatim**
(verified char-for-char against upstream `evaluate_qa.py`), routed via OpenRouter
(the direct OpenAI key is quota-dead; OpenRouter proxies the identical weights).
No judge fallback chain — a run fails loudly rather than silently grading with a
different model.

### Usage
```
genesis eval longmemeval --limit 30            # pilot
genesis eval longmemeval                        # full oracle (500 q)
python -m genesis.eval.longmemeval --no-persist # dry run
```
Dataset: `xiaowu0162/longmemeval-cleaned` → `~/tmp/longmemeval/longmemeval_oracle.json`.

### Design notes (from architect review)
- `origin_class=first_party` — the haystack is the user's own history; the WS-3
  provenance/immunity gates never block first_party.
- `GENESIS_MEMORY_WRITEBACKS_OFF=1` — recall write-backs suppressed so the 4
  arms (sharing one per-question frozen store) don't re-rank each other.

### Scope / follow-ups
- `oracle` variant only. `_s` (~115k-token haystacks) needs concurrent/batched
  embedding — deferred.
- Rerank arm is best-effort: Voyage 429-rate-limits under concurrency and
  degrades to RRF order; the non-rerank arms are the reliable comparison.
- A5-style dashboard read-surface for `longmemeval` runs — fast-follow.

### Tests & verification
- **41 unit tests** (dataset, judge verbatim-match, ephemeral-store isolation +
  recall with a deterministic hashing embedder, query arms, answer, runner
  aggregation, and an orchestration test asserting real concurrency).
- **E2E oracle run (30 questions, all 4 arms)** completed on the fixed code:
  evidence-recall 0.97–1.0, per-arm accuracy persisted to `eval_runs`. (These 30
  are all temporal-reasoning — `--limit` takes the first N, not a stratified
  sample — so strong recall + ~0.5 accuracy reflects that this is the hardest
  category, where reasoning over the evidence, not recall, is the bottleneck.
  The full 500-question run is the real headline.)
- **Empirical isolation proof**: production Qdrant point counts (episodic +
  knowledge_base) were byte-identical before and after the E2E run, and every
  ephemeral temp dir was cleaned up on normal exit.

### Code review
One code-reviewer pass: fixed a sync-client-blocks-the-event-loop bug (defeated
`--concurrency`) and a per-question embedder FD/cache leak; both now covered by
tests. Isolation from prod verified clean by the reviewer and empirically.


WingedGuardian (commented):
## Code Review — PR #1032 (first-pass)

### Structural Review

**Pattern violations (ERROR): NONE**

All structural rules pass cleanly:
- Temp files: all use `Path.home() / "tmp" / ...` and every `tempfile.mkdtemp` passes `dir=` explicitly — no raw `/tmp/` usage.
- No bare `except:` clauses; all broad catches use `except Exception:` with a documented rationale.
- No `nohup`, no `task_states` inserts, no `git add .`, no `git push --force`, no `IntervalTrigger`, no `os.killpg`, no star imports.
- The `asyncio.to_thread()` pattern in `runner.py` correctly wraps the synchronous OpenAI calls to keep the event loop unblocked; good fix.

**Import hygiene (WARNING): NONE**

Lazy imports inside function bodies (`QdrantClient`, `EmbeddingProvider`, etc.) are consistent with the rest of the codebase and avoid heavy top-level cost. Import grouping (stdlib → local) is correct throughout.

---

**Missing coverage (WARNING):**

**`src/genesis/eval/longmemeval/ingest.py` — no dedicated test file.**
`ingest_haystack` is a non-trivial async function with date-parsing (`normalize_date`), evidence-set tracking, and the `origin_class`/`valid_at` tagging logic that's load-bearing for the benchmark's comparability claims. The orchestration and store tests exercise it indirectly, but the two distinct code paths inside `normalize_date` (regex miss vs. `ValueError` from `datetime()`) have no explicit coverage. A `test_longmemeval_ingest.py` with a fake `MemoryStore` stub would be the right fix.

**`src/genesis/eval/longmemeval/client.py` — no dedicated test file.**
`openrouter_api_key()` has three branches (env var present, loaded from secrets.env, neither → raise). `_candidate_secret_paths()` has two paths (genesis.env resolves vs. exception → fallback to `~/genesis/secrets.env`). These are small but testable units; the raise path in particular should be exercised so a missing key fails loudly rather than silently.

---

- `QuestionArmResult` in `src/genesis/eval/longmemeval/runner.py:85` — frozen dataclass, returned from `run_question` and collected throughout the harness, but has no docstring.
- `IngestResult` in `src/genesis/eval/longmemeval/ingest.py` — same; no docstring on the return type of `ingest_haystack`.
- `QueryArm` enum in `src/genesis/eval/longmemeval/query.py` — no class-level docstring (the module docstring explains the arms, but the enum itself doesn't self-document).

---

**HARD BLOCK: NONE**
**SOFT WARNING: NONE**
**WORDING (CRITICAL): NONE**

All added lines are clean:
- No API keys, tokens, or credentials in source.
- `OPENROUTER_BASE_URL` and `OPENROUTER_KEY_ENV` are public endpoint + env var name references, not credentials.
- All temp/data paths use `Path.home()` — no hardcoded usernames.
- No private IPs, no SSH keys, no personal email addresses.
- PR title and CHANGELOG addition describe a feature, not a security fix or infrastructure detail.

---

**Summary:** Two WARNING-level test coverage gaps (`ingest.py`, `client.py`) and three NOTE-level docstring gaps. No errors or blocks. The isolation design (in-memory Qdrant + temp SQLite, `GENESIS_MEMORY_WRITEBACKS_OFF`, `~/tmp/` for all temp) is implemented correctly and matches the stated guarantees.


---

WingedGuardian:
Both Codex P1s addressed in `f4533f0b` (+ tests):

- **Timestamps in reader snippets** — `ingest.py` now prepends the session date to each stored memory's content (`[<date>] [<role>] …`), so the reader can compute intervals / pick the latest fact on temporal-reasoning and knowledge-update questions. Covered by `test_ingest_prepends_session_date_to_content`.
- **Dropping errored questions** — `run_longmemeval` records errored questions as **skipped** per arm; `total_cases` reflects the requested N, `skipped_cases` surfaces failures, and `aggregate_score` is computed over *attempted* only (never a silently-shrunk subset). Covered by `test_build_run_summary_skipped_do_not_tilt_accuracy`; the report prints `attempted/total  skipped=N`.

All CI green (test, dependency-audit, lint, leak-detector, CodeQL, migration, subsystem-map). WingedGuardian's two coverage WARNINGs also addressed (`test_longmemeval_ingest.py`, `test_longmemeval_client.py`).

PR #1034, opened 2026-07-14
feat(infra): guardian host plane for the infrastructure body schema

WingedGuardian:
## Summary

PR2 of the infrastructure body schema (#1019 shipped the container plane). Flips the 3 host-plane sections — `host_system`, `host_storage_pool`, `host_virt` — from the unavailable-stub to live data via a new **read-only `host-profile` gateway verb**.

**Host side**
- `guardian/host_profile.py` (new): concurrent, best-effort gather — meminfo/nproc/kernel/os-release, the guardian's own `measure_storage_pool` (+tier, 30s cap so a wedged pool can't blow the gateway's 45s budget and SIGKILL away healthy sections), incus version + this container's `limits.*` (yaml.safe_load; device-nested `limits.read/write` can't leak in), `systemd-detect-virt`, pveversion/smartctl presence. `ok:false` only when *every* section failed, so total collapse degrades the plane instead of posing as healthy.
- `--host-profile` CLI flag (JSON-always) + gateway verb (no secrets, no sudo) with a **skew guard**: `sync-gateway` redeploys the script independently of src, and on an old checkout the unknown flag would fall through into `run_check()` — a full recovery cycle triggered by a routine poll. The verb refuses with a JSON error when the deployed src predates it.

**Container side**
- `GuardianRemote.host_profile()` (50s client > 45s gateway, per convention).
- Real `collectors/host.py`: facts/metrics split — hashed identity/topology (incl. `container_limits` for cross-plane cage checks) vs volatile readings; allowlist-based so unknown new host fields land in metrics and never hash-churn across version skew. `"error"` is a reserved key; the not-deployed discriminator matches only the quoted `"denied"` gateway sentinel (an SSH `Permission denied` surfaces as itself).
- Diagnosis prompt inlines the shared-mount `INFRASTRUCTURE.md` (4KB, line-boundary truncation) via the new `config.shared_path` — which also fixes the pre-existing sentinel read that pointed at a container-side path not present on the host and silently no-op'd.
- `diff.py`: availability flips are absorbed by hash guards instead of a status gate, so a fact that really changed during a guardian outage now surfaces on recovery.

Graceful degradation is unchanged: no guardian, an un-redeployed gateway, or SSH failure → the 3 sections render "not visible from this vantage". Host effects land only after a host redeploy.

## Test plan
- `tests/test_guardian/test_host_profile.py` (new), `tests/test_infra_profile/test_collectors_host.py` (new), diagnosis + diff + edges extensions — 69 targeted tests green; ruff clean; gateway `bash -n` clean.
- Post-merge: container deploy → verify graceful degradation against the un-redeployed gateway, then host redeploy → verify the plane flips live.

### DIFF
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -348,7 +348,7 @@ radius) and the container-side Sentinel (CC-driven diagnosis/repair).
 ```yaml subsystem-map
 entry: guardian-sentinel
 modules: [guardian, sentinel]
-verified: 9037d45b 2026-07-07
+verified: 3d5234b9 2026-07-13
 ```
 
 - **guardian/** is bidirectional: host side (`python -m genesis.guardian`,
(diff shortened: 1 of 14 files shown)


WingedGuardian:

---

**Coverage note (informational):** `GuardianRemote.host_profile()` in `src/genesis/guardian/remote.py` has no direct unit test, but this is consistent with how other SSH transport verbs are handled in this codebase — the contract is fully exercised via `_FakeRemote` in `tests/test_infra_profile/test_collectors_host.py`, and the implementation is a one-liner following the exact same pattern as `provision_status()` and peers.

---

PR #1035, opened 2026-07-14
fix(follow-ups,observability): completed_at re-stamp fix + neural monitor "Other" group

WingedGuardian:
## Summary

Two independent fixes (the planned "PR-I" pair) plus gitignore hardening and a doc note.

### 1. Follow-ups: preserve `completed_at` on an idempotent terminal re-write
`update_status` / `update_status_batch` re-stamped `completed_at` on *every* terminal write. A notes-only update passes the already-terminal status back in — the confirmed instance is the cockpit "add a note to a completed follow-up" path (`mcp/health/follow_up_tools.py:220`, the `elif resolution_notes or blocked_reason:` branch) — which silently reset the reaper/GC windows keyed off `completed_at` (`get_recently_resolved`, `purge_completed`, `get_recently_completed`).

Fix: `completed_at = CASE WHEN status = ? THEN completed_at ELSE ? END` — preserve on an idempotent same-status re-write, stamp fresh on a genuine transition. SQLite evaluates SET RHS against the pre-update row, so `status` in the CASE reads the OLD value. This still stamps fresh on the regular `failed→completed` ego-resolution path (`ego/dispatch.py`), which a naive `COALESCE` would have frozen at the failure time.

(The dispatcher pinned branch `follow_ups/dispatcher.py:230` passes `fu["status"]` but only ever receives `scheduled`/`in_progress` rows from `get_linked_active`, so it never actually re-stamped — the fix is defensive there.)

### 2. Neural monitor: surface ungrouped call sites in an "Other" group
The battleship grid renders only call sites in `SUBSYSTEM_GROUPS`; the live snapshot emits ~67, so ~25 ungrouped sites (ego-split, dream-cycle, task pre/post-mortem, judge, crag/attention grading, voice, etc.) rendered nowhere. A new "Other" enrichment card, computed at render time from `Object.keys(callSites)` minus the grouped ids, surfaces them (clickable → detail panel). Self-maintaining as new sites are added; the fixed 7×7 grid is untouched.

### 3. `.gitignore` hardening
Ignore `secrets.env.*` backups (with a `!secrets.env.example` negation so the tracked template stays committable) + local tooling artifacts (`.claude/database.db`, `.claude/hookify.*.local.md`, `.serena/`, `data/*mode=ro`).

## Testing
- TDD (RED→GREEN); 85 targeted tests pass (CRUD + all follow-ups consumers + cockpit API); ruff clean.
- **E2E**: CRUD verified on a real DB snapshot (notes-only → preserved, `failed→completed` → fresh); dashboard verified in a real browser against live health data (Other card renders 25/25, battleship grid intact, no errors).

## Follow-ups (tabled)
Two grouping gaps tracked separately, not fixed here: dead grid tiles (`2_triage`, `bookmark_enrichment`, both absent from the live snapshot) and the Ego group still listing the legacy `7_ego_cycle` while the current ego sites surface under "Other".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,10 @@
 .assets
 .env
 secrets.env
+# Any secrets.env backup/variant (e.g. secrets.env.pre-*-bak-*) must never be
+# committable — only the tracked template is allowed.
+secrets.env.*
+!secrets.env.example
 secrets.json
 *.pyc
 dist/
@@ -24,6 +28,12 @@ botpy.log
 .worktrees/
 .claude/worktrees/
 .claude/.session-locks/
+# Local tooling artifacts written into the repo tree (keep the tree clean)
+.claude/database.db
+.claude/hookify.*.local.md
+.serena/
+# Malformed sqlite path artifacts (e.g. genesis.db?mode=ro used as a literal file)
+data/*mode=ro
 # Background sessions now live at ~/.genesis/background-sessions/ (outside repo)
 data/sqlite/*.db
 data/sqlite/*.db-journal
--- a/docs/reference/routing-call-sites.md
+++ b/docs/reference/routing-call-sites.md
@@ -143,6 +143,27 @@ see the linked commit / PR.
 
 ## Changelog
 
+### 2026-07-12 — Neural monitor "Other" group for ungrouped call sites
+
+The battleship grid (`neural_monitor.html` `SUBSYSTEM_GROUPS`) renders only call
+sites assigned to a subsystem group. Sites the backend snapshot emits but that
+aren't in any group — e.g. `40_ego_focus_selection`, `43_task_retrospective`,
+`44_task_premortem`, `45_intelligence_intake`, `46_infra_annotation`, `judge`,
+`crag_grade`, `attention_salience`, the `dream_cycle_*` sites, `voice_conversation`
+— rendered nowhere and were invisible. A new "Other" enrichment card, computed at
+render time from `Object.keys(callSites)` minus the grouped ids, surfaces them
+(clickable, opens the detail panel) and self-maintains as new sites are added.
+
+Two related grouping gaps are tracked separately, not fixed here: (a) `2_triage`
+and `bookmark_enrichment` remain in `SUBSYSTEM_GROUPS` but are no longer emitted by
+the snapshot, so they render as permanently-idle dead tiles; (b) the `ego` group
+still lists the legacy `7_ego_cycle` (last run ~79d ago) alongside `8_ego_compaction`,
+while the current ego call sites (`7_genesis_ego_cycle`, `7_user_ego_cycle`,
+`40_ego_focus_selection` — all active) are emitted and now surface under "Other"
+rather than the Ego group; the group's site list should be refreshed to the current
+ids. (Verified against the live `/api/genesis/health` snapshot, which emits from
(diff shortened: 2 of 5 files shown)


WingedGuardian (commented):
## Structural review: one WARNING

### WARNING — Raw SQL in test helper (`tests/test_db/test_follow_ups.py`)

`_force_terminal` (added line ~246) issues a direct `UPDATE follow_ups SET status = ?, completed_at = ?` query against the test DB rather than going through `genesis.db.crud.follow_ups`. By the project's "except in migrations" rule, raw SQL in test fixtures is technically out of policy.

That said, the CRUD layer intentionally does not expose a setter for a specific `completed_at` timestamp, so this is a legitimate test-setup necessity — there's no CRUD path to forge an arbitrary past timestamp. The usage is also narrowly scoped to the test module and doesn't touch the production DB.

**Recommendation:** accept as-is given the no-other-way constraint, or add a `# noqa: RAW-SQL` / inline comment noting why the CRUD bypass is necessary so reviewers don't flag this pattern in future.

---

### Everything else: CLEAN

- `.gitignore` hardening: pattern / negation semantics correct (`secrets.env.*` + `!secrets.env.example`). ✓  
- `follow_ups.py` — `CASE WHEN status = ? THEN completed_at ELSE ? END` logic is correct: SQLite evaluates SET RHS against the pre-update row, so `status` in the CASE is the OLD value — the idempotent re-write path is correctly preserved and the genuine transition path stamps fresh. No bare `except:`, no mutable defaults, no raw SQL outside the existing parameterised layer. ✓  
- `neural_monitor.html` — JS enrichment card computed from `Object.keys(callSites)` minus grouped IDs; self-maintaining. No structural concerns. ✓  
- Tests: new tests have docstrings, cover both `update_status` and `update_status_batch` for both the preserve and stamp-fresh branches. ✓  
- Conventional commit prefix present (`fix(follow-ups,observability):`). ✓  

- No API keys, tokens, private IPs, hostnames, or credentials in added lines.  
- PR title and body describe a technical fix only — no language referencing a PII/secrets incident or infrastructure exposure.

---

WingedGuardian:
**Pre-merge review resolution** (both surfaces read):

- **Structural WARNING (raw SQL in `_force_terminal`): consciously accepted as-is.** The CRUD layer intentionally exposes no setter for an arbitrary past `completed_at` (the fix's whole point is that it stamps *now*), so forging an old sentinel timestamp for the preservation test requires direct SQL — and this matches the existing convention in the same module (`test_purge_completed_deletes_old`, `test_purge_completed_keeps_pinned`, `test_purge_failed_old` all set `completed_at='2025-01-01...'` via raw SQL). Narrowly test-scoped; never touches prod.
- **Cloud Codex: no inline findings returned (quota-limited).** Merging on CI + the structural review per the pre-merge gate.
- **The one red check (`dependency-audit`) is unrelated and repo-wide**: two fresh fastmcp CVEs (PYSEC-2026-2475 command-injection via `fastmcp install` on Windows; PYSEC-2026-2476 OAuthProxy confused-deputy) published 2026-07-13, both **unreachable** in Genesis (Linux + stdio MCP servers, no `fastmcp install`, no OAuthProxy). Same class as the existing litellm/diskcache triage entries. A triage hotfix to `ci.yml` follows immediately after this merge. Merging via `--admin` (user-approved) since this check is red on `main` and every open PR.

PR #1037, opened 2026-07-14
feat(sessions): charter store — session origin survives compaction

WingedGuardian:
## What

Long CC sessions compact many times, and every summary is recency-biased — after enough compactions a session can no longer connect "what's next" to what it was started for. This PR makes that structurally impossible:

- **NEW `scripts/genesis_precompact.py`** (PreCompact hook, both triggers): at a foreground session's FIRST compaction — while the golden transcript is still on disk in full fidelity — extracts the first typed user prompt (the **immutable origin**) into `~/.genesis/sessions/<sid>/charter.json` + a human-readable `charter.md`. Every compaction bumps the count and appends a deterministic waypoint (`waypoints.jsonl`). `mission`/`pointers` are living fields reserved for the ledger stage; the hook preserves them verbatim.
- **`scripts/genesis_session_context.py`**: now parses SessionStart stdin JSON (`session_id`/`source` — it previously read no stdin) and re-injects the charter block on `startup`/`resume`/`compact` — never `clear` (an explicit fresh start). Purely additive diff.
- **`.claude/settings.json`**: registers PreCompact (timeout 5s, fail-open); also fixes two stale ms-era `"timeout": 2000` values on `session_activity_touch.sh` that survived #1017's ms→seconds sweep.

## Design invariants

- **Fail-open everywhere**: the hook always exits 0 (exit 2 would cancel compaction); the injector block is try/except-wrapped like every advisory section.
- **Origin immutability**: second+ compactions only bump the count — `origin_prompt`/`origin_ts` are write-once (test-pinned byte-identical).
- **Real transcript shapes**: there is NO `promptSource` field in real transcripts. The filter uses the empirically verified discriminators (`isSidechain`/`isMeta`/`isCompactSummary`, tool_result lists, command/caveat prefixes) and the fixtures mirror real entry shapes, including a summary-continued head (the "This session is being continued…" entry must never become the origin).
- **Path-injection guard** on `session_id` in both writer and reader; dispatched sessions (`GENESIS_CC_SESSION=1`) are skipped — `task_states` is their continuity spine.

## Verification

- 19 new tests (`tests/test_scripts/test_precompact_charter.py`); full `tests/test_scripts/` suite green (407).
- Deterministic E2E through the real `genesis-hook` wrapper against a real 57MB live transcript: **0.11s wall** (5s budget), correct origin extracted, injector emits the charter block with divider discipline intact.
- Live-fire test after merge: a real `/compact` in a chartered foreground session.


WingedGuardian:

---

PR #1040, opened 2026-07-14
fix(eval): A4 — question-type-aware reader prompt (preference personalization)

WingedGuardian:
Follow-up to #1032. The LongMemEval reader used one generic QA prompt with an
"if it's not in the snippets, say you don't know" affordance. That tanked
**single-session-preference** questions — which ask for a recommendation
tailored to the user's recalled preferences — because the reader *abstained*
instead of personalizing, despite ~0.90 evidence recall. The gap was the
reader, not the memory.

### Fix
- **single-session-preference**: a dedicated reader prompt that uses the recalled
  preferences to give a concrete, tailored recommendation and never abstains.
- **temporal-reasoning**: a step-by-step-using-the-dated-memories nudge.
- **default (factual QA + abstention)**: unchanged — keeps the "don't know"
  affordance that `_abs` questions rely on.

`question_type` threads `answer_question` → `build_answer_prompt`; the runner
passes `instance.question_type`. TDD: type-aware prompt tests.

### Validated on the same 150-question stratified set (25/type, both non-rerank arms)

| type | raw before → after | keyword before → after |
|---|---|---|
| single-session-preference | **0.04 → 0.68** | **0.08 → 0.60** |
| single-session-user | 0.92 → 1.00 | 1.00 → 1.00 |
| **overall** | **0.553 → 0.653** | **0.567 → 0.647** |

Evidence-recall unchanged at 0.92 (recall was never the issue). The preference
fix is a clear, large win. **The temporal nudge is inconclusive at n=25** (net
noise) and is kept as a standard technique pending full-scale confirmation —
not claimed as a win here.

Mini-runs used a temp DB (nothing persisted to prod `eval_runs`).

### DIFF
--- a/src/genesis/eval/longmemeval/answer.py
+++ b/src/genesis/eval/longmemeval/answer.py
@@ -1,9 +1,17 @@
 """The reader step: retrieved memories + question -> a short hypothesis answer.
 
 A fixed model (gpt-4o via OpenRouter) reads ONLY the recalled memories and
-answers. The prompt grants an explicit "I don't know" affordance so abstention
-(``_abs``) questions — whose evidence is absent by design — can be answered
-correctly (the judge rewards a model that recognises unanswerability).
+answers. The reader prompt is question-type-aware:
+
+* default (factual QA + abstention): a concise answer with an explicit "I don't
+  know" affordance so abstention (``_abs``) questions — whose evidence is absent
+  by design — are answered correctly.
+* ``single-session-preference``: these ask for a RECOMMENDATION tailored to the
+  user's recalled preferences, so the reader is told to USE those preferences
+  and make a concrete suggestion — NOT to abstain. (A generic QA prompt made the
+  reader answer "I don't know" and tanked this category to ~0.05.)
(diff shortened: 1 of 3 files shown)


WingedGuardian:
**Structural review:** 1 NOTE — no ERRORs or WARNINGs.

---

**NOTE — Missing docstring updates for new `question_type` parameter**

Both public functions gained a `question_type: str | None = None` keyword argument but their docstrings weren't updated to describe it.

- `build_answer_prompt` in `src/genesis/eval/longmemeval/answer.py` (~line 58): docstring says "Assemble the reader prompt from the question and recalled memories" — no mention of `question_type` or what values it accepts.
- `answer_question` in `src/genesis/eval/longmemeval/answer.py` (~line 85): docstring says "Produce a hypothesis answer from the recalled memories" — same gap.

Suggest a one-liner each, e.g.:
```
question_type: if "single-session-preference" or "temporal-reasoning", selects a
    specialised reader prompt; defaults to the generic factual-QA prompt.
```

No other issues found. Logic and test coverage look solid.

---

PR #1042, opened 2026-07-14
fix(security): WS-3 B4 PR-1 — stored origin_class through recall + stored-first blockability

WingedGuardian:
## Summary

First of three B4 PRs (stored-origin plumbing → gate-2 substrate → enforce). Shadow behavior only — no blocking anywhere. Makes gate-4 shadow data **enforce-grade** by keying blockability on the origin_class **stamped at store time** (migration 0054) instead of re-deriving from (collection, source_pipeline) at recall.

- **`RetrievalResult.origin_class`** (defaulted-last field): populated on the Qdrant path (payload passthrough, zero I/O) AND the FTS5-only path via a new `memory_metadata.origin_class` column on `search_ranked`'s existing LEFT JOIN. Pre-B4, every payload-derived field was silently `None` on FTS-only hits; `origin_class` is now recovered there (`source_pipeline` intentionally is not — it has no SQLite column for episodic and is honestly unrecoverable on that path).
- **`item_is_blockable` is stored-first**: the stored class is authoritative when present; the old re-derivation remains the fallback for pre-0054 rows. Two deliberate semantic changes, both stated in the docstring: episodic `external_untrusted` rows (written by dispatched sessions) become observable (widening — it's what the session-origin substrate built), and a first-party item living in the KB via a non-whitelist pipeline no longer over-observes (FP fix). All 10 call sites pass the stored value.
- **`memory_core_facts` gated**: it Qdrant-scrolls episodic and injects full content while being invisible to the `.recall()` AST lock — caught during the B4 enumeration pass. Now wraps stored-external items + emits the gate-4 shadow record. Wrap + observe only; the enforce drop (dispatched sessions, pushed surfaces) is PR-3.
- **Enumeration lock gains a second mechanical sweep**: every Qdrant `.scroll()`/`.retrieve()` call site under `src/genesis` must be classified (`KNOWN_QDRANT_READ_SITES`, 15 sites). The sweep surfaced two additional content→LLM surfaces, registered `deferred-with-reason` and tracked as a follow-up: dream-cycle consolidation (origin laundering into the synthesized canonical memory) and session-awareness arbiter candidates.
- **Proactive hook**: `origin_class` through `_search_fts5` (JOIN column), `_search_qdrant` (payload), `_enrich_with_metadata` (setdefault backfill), and the sync gate emit.
- **`knowledge.search_fts` + `knowledge_recall`** reshape dicts carry the key.

## Verification

- Targeted suites green (~560 tests across the blast radius incl. new matrices); `ruff check --no-cache` clean.
- Live-DB-copy E2E: `search_ranked`/`search_fts`/hook rows all carry the column against real data; full `HybridRetriever.recall` against live Qdrant (writebacks off) returns stored origins on vector hits (both classes observed); stored-first semantics verified on the 4 real episodic-external rows (fallback False → stored True) and the FP direction unit-verified.
- Stored distribution on the copy matches 0054 predictions: 55,467 first_party / 768 external_untrusted / 1 NULL.

## Notes for review

- The plan checklist said "exclude/wrap" for core_facts; exclusion would be enforcement ahead of the gate flip (and would change supervised foreground behavior), so this PR wraps + emits — drops land in PR-3 behind `gate_mode == enforce` for dispatched sessions only.
- Closes follow-up 3bf419ca (stored-origin plumbing for gate-4).

### DIFF
--- a/scripts/proactive_memory_hook.py
+++ b/scripts/proactive_memory_hook.py
@@ -705,7 +705,8 @@ def _search_fts5(
                 f"""
                 SELECT memory_fts.memory_id, memory_fts.content,
                        memory_fts.source_type, memory_fts.collection,
-                       memory_fts.rank
+                       memory_fts.rank,
+                       memory_metadata.origin_class
                 FROM memory_fts
                 LEFT JOIN memory_metadata
                   ON memory_fts.memory_id = memory_metadata.memory_id
@@ -990,6 +991,8 @@ async def _search_qdrant(
                         # Provenance source tier for KB results (audit D12) — the
                         # label needs it; memory_metadata doesn't carry it.
                         "source_pipeline": payload.get("source_pipeline"),
+                        # WS-3 stored provenance (0054-stamped at store).
+                        "origin_class": payload.get("origin_class"),
                     })
                 return results
     except Exception as exc:
@@ -1230,8 +1233,8 @@ def _enrich_with_metadata(results: list[dict]) -> None:
             conn.row_factory = sqlite3.Row
             placeholders = ",".join("?" for _ in ids)
             rows = conn.execute(
-                f"SELECT memory_id, created_at, wing, collection FROM memory_metadata"  # noqa: S608
-                f" WHERE memory_id IN ({placeholders})",
+                f"SELECT memory_id, created_at, wing, collection, origin_class"  # noqa: S608
+                f" FROM memory_metadata WHERE memory_id IN ({placeholders})",
                 ids,
             ).fetchall()
             meta = {row["memory_id"]: dict(row) for row in rows}
@@ -1242,6 +1245,10 @@ def _enrich_with_metadata(results: list[dict]) -> None:
                     if not r.get("_wing"):
                         r["_wing"] = meta[mid].get("wing")
                     r.setdefault("collection", meta[mid].get("collection"))
+                    # WS-3 stored provenance — fills paths that lack the
+                    # Qdrant payload / FTS join value (setdefault: never
+                    # overrides a value the search path already carried).
+                    r.setdefault("origin_class", meta[mid].get("origin_class"))
         finally:
             conn.close()
     except Exception:
@@ -1826,6 +1833,7 @@ async def _run(prompt: str, session_id: str = "") -> None:
                 if item_is_blockable(
                     collection=r.get("collection"),
                     source_pipeline=r.get("source_pipeline"),
+                    origin_class=r.get("origin_class"),
                 )
             )
             if blockable:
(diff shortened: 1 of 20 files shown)


WingedGuardian:
## First-pass review — PR #1042

**PII/Secrets/Wording scan: CLEAN** — no credentials, private IPs, personal emails, private hostnames, or sensitive wording in added lines or PR metadata.

---

### NOTE — indentation inconsistency in test fixture DDL

`tests/test_hooks/test_proactive_invalid_at.py` — the new `origin_class TEXT,` line uses 12 spaces while the surrounding columns use 16:

```python
                source_subsystem TEXT,
            origin_class TEXT,          ← 12 spaces (should be 16)
                deprecated       INTEGER NOT NULL DEFAULT 0,
```

Functionally harmless inside the SQL string (SQLite ignores whitespace), but breaks the visual alignment of the schema fixture. Trivial fix if you want consistency before merge.

---

WingedGuardian:
## First-pass review — PR #1042

### Structural review

No pattern violations (ERROR), no warnings (WARNING). Three notes:

**NOTE 1** — `tests/test_hooks/test_proactive_invalid_at.py` ≈ line 58

The added `origin_class TEXT,` column is indented 12 spaces while every surrounding SQL column in that fixture (`valid_at`, `invalid_at`, `source_subsystem`, `deprecated`, `dream_cycle_run_id`) is indented 17 spaces. SQLite ignores the whitespace so there is no functional impact, but it is inconsistent with the rest of the block. The same column is correctly aligned in `test_search_ranked_invalid_at.py` and `test_search_ranked_subsystem.py`.

**NOTE 2** — `src/genesis/memory/retrieval.py` (COALESCE expression)

```python
origin_class=(
    payload.get("origin_class")
    or (fts_hit.get("origin_class") if fts_hit else None)
),
```

Using `or` means any falsy payload value (empty string `""`, `0`) would fall through to the FTS hit, whereas `_enrich_with_metadata` uses `is None` for the same intent. Safe in practice — `origin_class` is always `None` or a non-empty enum string — but explicit `is None` (matching the `_enrich_with_metadata` pattern) would be more precise and self-documenting, especially if a future garbage value needs to be distinguished from an absent one.

**NOTE 3** — `tests/test_mcp/test_memory_mcp.py` (new test)

`test_memory_core_facts_wraps_external_and_emits` is an `async def` without `@pytest.mark.asyncio`. The new tests in `tests/test_memory/test_origin_class_recall.py` added in the same PR all carry the decorator. If the project's `pytest.ini`/`pyproject.toml` sets `asyncio_mode = "auto"` this is fine; otherwise the test may be collected as a coroutine and silently not run. Worth confirming the auto-mode config covers this file.

---

**CLEAN** — no secrets, no private IPs, no personal data in added lines. PR title and description describe internal code mechanics (`origin_class`, `first_party`, `external_untrusted`) with no language that advertises a credential leak, data exposure, or private infrastructure detail.

---

PR #1043, opened 2026-07-14
fix(research): wire search() on Tavily + Exa adapters so results reach the orchestrator

WingedGuardian:
## Problem

`ResearchOrchestrator.search()` (`research/orchestrator.py` `_safe_search`) checks
`hasattr(provider, "search")`. `TavilyAdapter`/`ExaAdapter` exposed **only**
`invoke()`, which returns a `ProviderResult` whose `.data` is a **dict**. So the
orchestrator fell through to the `invoke()` fallback, whose `isinstance(result.data,
list)` guard was `False` for a dict → returned `[]` with **no warning**. Research
quietly ran with fewer sources — silent degradation, not a crash. (Extra Exa bug:
the fallback passed `max_results`, but Exa's `invoke()` reads `num_results`, so Exa
always defaulted to 5 even on the rare path it wasn't dropped.)

Both keys are configured in this install, so the drop was **live in production**,
not dormant.

## Fix

- Add `search(query, *, max_results) -> list[SearchResult]` to both adapters,
  mirroring `WebSearchAdapter`/`PerplexityAdapter`: it **wraps the existing
  `invoke()`** (reusing its SDK call + error handling) and maps the dict into
  `SearchResult(source=...)`. Returns `[]` on any failure so a degraded provider
  never breaks the fan-out.
- **Exa**: bridge `max_results` → Exa's `num_results` key, and request capped page
  text (`contents`, 500 chars) so snippets reach parity with the other providers.
- `invoke()` is **unchanged** on both — the MCP `web_tools.py` path depends on its
  dict return shape.

## Verification

- **TDD**: `search()` unit tests for both adapters (mapping, `max_results`/
  `num_results` bridge, empty-on-missing-key) + an orchestrator regression that
  registers the **real** adapters (SDK mocked) and asserts their results now flow
  through and dedup — genuinely RED before the fix (silently dropped). 63 provider+
  research tests pass, ruff clean.
- **Live E2E** (real API calls): `TavilyAdapter().search(...)` returned 3
  `SearchResult`s with full snippets + scores; `ExaAdapter().search(...)` returned 3
  with exactly-500-char capped snippets; both correctly `source`-tagged.
- Code-reviewed (adaptive protocol): verdict DONE, no blockers.

No host-deploy-gate paths touched. Effect: research fan-out now includes Tavily +
Exa results where it previously discarded them.

### DIFF
--- a/src/genesis/providers/exa_adapter.py
+++ b/src/genesis/providers/exa_adapter.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import logging
 import os
 import time
+from typing import TYPE_CHECKING
 
 from genesis.providers.types import (
     CostTier,
@@ -17,6 +18,9 @@ from genesis.providers.types import (
     ProviderResult,
     ProviderStatus,
 )
+
+if TYPE_CHECKING:
+    from genesis.research.types import SearchResult
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


WingedGuardian:

---

PR #1038, opened 2026-07-14
fix(memory): session-observer restore path orphaned every leftover file

WingedGuardian:
## What

`session_observer.py`'s finally-block restored leftover `.processing` files via `processing_path.with_suffix(".jsonl")` — but `with_suffix` replaces only the LAST suffix, so `tool_observations.jsonl.processing` was "restored" to `tool_observations.jsonl.jsonl`, a name `_find_observation_files` never scans. Every budget-exceeded or cancellation leftover was silently orphaned instead of reprocessed — the exact loss the restore comment claims to prevent. **108 orphaned files found on disk** under `~/.genesis/sessions` at diagnosis (2026-07-13).

## Fix

Extract `_restore_processing_files()` and use the EXACT original path already carried in the `(original, processing_path)` rename tuple — it was bound as `_original` and ignored. Behavior on the healthy paths (append-merge when the hook recreated the original; rename when absent; OSError → warning) is unchanged, and the helper cannot raise out of the `finally` block.

## Tests

Red-first regression tests (`TestRestoreProcessingFiles`): exact-name restore (the `.jsonl.jsonl` assertion fails on the old code), append-merge, and already-consumed no-op. Both observer suites green (30 tests).

## Remediation

The 108 on-disk orphans get a one-off rename back to `tool_observations.jsonl` at deploy (after the server restarts onto the fixed code — renaming earlier would let the old runtime re-orphan them), so they finally process.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1044, opened 2026-07-14
fix(memory): recover crash-stranded .processing observation files

WingedGuardian:
## What

Follow-up to #1038. The finally-block restore only covers files renamed by the CURRENT run — a server crash/restart mid-tick strands its renamed `.processing` files where no future tick ever scans them. Found live: one file stranded since 2026-05-13 (two months invisible; remediated by hand).

## Fix

`_recover_stale_processing_files()` at the start of `process_pending_observations`: adopts any `tool_observations.jsonl.processing` older than `STALE_PROCESSING_AGE_S` (1h) back to its original name via the existing `_restore_processing_files` helper. Race-safe: the observer is single-flight (`_session_observer_lock` held for the run's full lifetime including its own restore), so recovery can never see a live run's files — only files whose owning process is gone. Fail-open (pathlib glob swallows scan OSErrors internally; stat guarded; awareness-loop backstop unchanged).

## Tests

3 red-first tests (recover stale / leave fresh alone / merge into hook-recreated original). Review finding fixed: the two pre-existing `process_pending_observations` tests now also isolate `_sessions_dir`, so no unit test can touch the real `~/.genesis/sessions`.

### DIFF
--- a/src/genesis/memory/session_observer.py
+++ b/src/genesis/memory/session_observer.py
@@ -39,6 +39,11 @@ MAX_PROCESSING_TIME_S = 45.0  # Budget per tick (leave headroom under 60s)
 # Files older than this are cleaned up (stale sessions)
 STALE_FILE_AGE_S = 7 * 24 * 3600  # 7 days
 
+# A .processing file older than this cannot belong to a live tick (45s budget,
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

PR #1046, opened 2026-07-14
feat(security): WS-3 B4 PR-2 — gate-2 substrate: durable session origin + delta provenance

WingedGuardian:
## Summary

Second of three B4 PRs. Builds the L-tier session→origin substrate so gate-2 (identity) can finally OBSERVE — its emit has been a hardcoded-first_party no-op because nothing server-side could answer "was that session external?". Shadow-only; gate-2 does not flip (its shadow bakes on real signal first).

- **Migration 0057**: `cc_sessions.origin_class` + `observations.origin_class` (nullable, PRAGMA-guarded, O(1), no backfill — historical NULL = unknown, read as first_party in shadow so pre-substrate rows can't manufacture external signal; any future enforce consumer must apply the WS-3 fail-closed normalizer at gate time instead).
- **Session origin stamped at registration, derived from the dispatch profile — never a tool scan** (that coarseness is exactly what invalidated gate-1's shadow signal): `direct_session` passes the same `_PROFILE_ORIGIN` map its env stamp uses; inbox evaluation stamps `external_untrusted` (mirrors its `CCInvocation.origin`); reflection/ego/sentinel dispatches and foreground sessions stay NULL (first-party). All 7 `create_background` callers enumerated.
- **Run-level provenance aggregate** (`cc_sessions.reflection_window_origin`, never raises): reflections are about a tick WINDOW, not one session — the reflection context carries no per-session refs, so per-delta tracing is not honestly buildable. The aggregate is external iff any external-origin session was active in the 60-minute material window. Conservative over-tagging at run granularity, measured in shadow.
- **Both `user_model_delta` writers stamp it** (CLI `reflection_bridge/_output.py` + API `perception/writer.py` — same LLM output, two modes).
- **Gate-2 emit derives from the accepted deltas** (`process_pending_deltas` gains an `accepted_ids_out` out-param; return contract unchanged): external iff any accepted delta's stored class is external, with delta counts in the emit detail. The in-code FLIP BLOCKER comment is replaced with the real derivation + an explicit "stays SHADOW" note.

## Verification

- 507 existing blast-radius tests green + 30 new (migration idempotence/legacy-CREATE/NULL-read, crud round-trips incl. upsert conflict path, session-manager stamping incl. foreground-NULL, window aggregate incl. stale-external-excluded + never-raises, end-to-end external-session→stamped-delta chain, accepted_ids_out); `ruff --no-cache` clean.
- Live-DB-copy E2E: 0057 applies + re-applies clean (3,063 cc_sessions / 12,621 observations rows, 0 pre-existing non-NULL); stamp + window aggregate verified against the migrated copy both directions.

## Notes

- Advances follow-up 61c57093 (substrate built; full closure = enforce-grade signal + a gate-2 flip decision, its own future unit).
- PR-3 (enforce branches for gates 3+4 + red-team suite) follows.


WingedGuardian:
## First-pass review — PR #1046

### Structural review

**WARNING — Raw SQL against genesis.db outside the crud layer**

`src/genesis/runtime/init/learning.py` (inside `_evolve_user_model`, the gate-2 delta-origin aggregate block):

```python
cur = await rt._db.execute(
    "SELECT COUNT(*) FROM observations "  # noqa: S608 -- placeholders bound
    f"WHERE id IN ({marks}) "
    "AND origin_class = 'external_untrusted'",
    accepted_delta_ids,
)
```

This is a raw `SELECT` against the `observations` table issued directly via `rt._db.execute` from `runtime/init/learning.py`, bypassing `genesis.db.crud.observations`. The project convention is that SQL touching `genesis.db` belongs in the crud module (except in migrations). The existing `observations.py` crud doesn't expose a count-by-ids-and-origin helper, but this query is the right candidate to add there — it would also make the gate-2 logic independently testable (the aggregate unit test can mock the crud call instead of needing a real DB). The `# noqa: S608` comment addresses the ruff warning about f-strings in SQL but not the crud-layer convention.

Suggested path: add `count_external_by_ids(db, ids)` (or similar) to `src/genesis/db/crud/observations.py` and call it here.

---

**CLEAN** — no secrets, no private IPs, no personal data in added lines. PR title and description describe code mechanics (`cc_sessions.origin_class`, `observations.origin_class`, gate-2 substrate) with no language that advertises a credential leak, data exposure, or private infrastructure detail.

---

PR #1049, opened 2026-07-14
fix(code-intel): idle-gated index queue + pressure watchdog — stop the reindex storms

WingedGuardian:
## Why

Keeping the code-intelligence graph fresh used to fire a **full reindex on every commit**, in the background, with no coordination. When disk cleanup had reclaimed the index first, every "quick refresh" was secretly a full 0→100 rebuild — and a full index reads fast enough to saturate the container's entire disk-read throttle, dragging the whole box into an I/O storm (load spiked into the hundreds, dozens of processes stuck in uninterruptible I/O wait). This fixes it at the root.

## What

Three layers, root-cause first:

1. **Stop deleting the index DB.** `disk_reclaim.py` moves the three code-intel caches (codebase-memory-mcp, code-graph, GitNexus) from the *medium* reclaim tier (cleared at ≥90% disk) to a new **`last_resort`** tier cleared only at ≥95% (`--last-resort-above`), and drops a queue marker when it does — so refreshes stay incremental instead of becoming full rebuilds.
2. **Queue instead of spawn.** Every trigger (post-commit hook, setup, install, the Mon/Thu GitNexus job, disk_reclaim) now writes an **index-request marker** (`scripts/lib/index_marker.py`) instead of spawning an indexer. A new idle-gated systemd timer (`genesis-code-intel.timer` → `scripts/code_intel_runner.sh`, every 5 min) is the **only** consumer — it runs at most one index at a time, and only when the machine is quiet (low load, low iowait, no interactive session busy).
3. **Pressure watchdog.** Whatever does run is babysat: sampled every 15s and **paused** under load via a cgroup freeze of its systemd scope (the only I/O throttle that actually works in this environment — `ionice`/`IOWeight` are inert here), then killed if it can't make headway (mode-aware wall cap). Routine refreshes are the cheap `fast` pass; the expensive `full` pass is reserved for a weekly idle window.

Also fixes a **latent bug**: `gitnexus analyze --quiet` isn't a valid flag in GitNexus 1.6.x — it had been failing *every* entrypoint-driven GitNexus reindex since it was introduced.

## Notes for reviewers

- The single-flight lock **acquisition** in `code_intel_index.sh` is byte-unchanged (only the skip-branch exit code is now overridable) — a host-side kill-switch holds that exact lock, and this preserves it.
- The marker filename hash is byte-identical to the entrypoint's lock hash (cross-language parity test included).
- The first from-scratch full rebuild is intended to be run **supervised** (the indexer can't resume a killed full), not left to the automation; the automation carries the safety rails regardless.

## Tests

86 targeted tests: marker state machine (hash parity, coalesce, move-aside claim/consume/restore, escalation + backoff, inflight reconcile), entrypoint (mode/persistence/rc contract + 3 watchdog behaviourals), `last_resort` tier + marker drop, and the runner rc-contract/idle-gate. Both watchdog throttle paths and the full queue→entrypoint→lock handshake verified end-to-end.


WingedGuardian (commented):
## Structural review — 1 ERROR, 1 WARNING, 1 NOTE

### ERROR (must fix)

**`/tmp` fallback in `scripts/code_intel_runner.sh` — line 213**

```bash
mkdir -p "$LOCK_DIR" 2>/dev/null || LOCK_DIR="${TMPDIR:-/tmp}"
```

Per Genesis convention, `/tmp` must not be used for temp files; all temp must go to `~/tmp/`. The systemd service template correctly sets `ProtectSystem=strict` with `ReadWritePaths=%h`, which makes `~/tmp/` writable. But the `/tmp` fallback lands outside that allowed path — in a restricted systemd unit, `{ exec 8>"$RUNNER_LOCK"; } 2>/dev/null` would silently fail, and the runner would proceed without its self-lock. Even in an unrestricted context, the policy is clear.

Fix: `|| LOCK_DIR="$HOME/tmp"` (with `mkdir -p "$HOME/tmp" 2>/dev/null ||` prepended, or rely on the already-existing `~/tmp/` convention).

---

### WARNING (please address)

**`run_gitnexus_reindex` refactored without a visible test update — `src/genesis/surplus/jobs/gitnexus.py`**

The function went from ~99 lines of subprocess / `asyncio.wait_for` / timeout-kill logic to ~28 lines of marker-write logic. The PR's 86-test count covers the marker state machine, entrypoint, disk_reclaim, and runner — but doesn't mention gitnexus-specific tests. If `tests/test_surplus/test_gitnexus.py` (or equivalent) exists with mocks of the old `asyncio.create_subprocess_exec` / `communicate` path, those tests are now testing removed code and may pass vacuously. Please confirm the gitnexus job test suite was updated alongside this refactor.

---

### NOTE (informational)

**Missing docstrings — `scripts/lib/index_marker.py`**

`last_full_path()` and `full_backoff_path()` are public (no underscore) functions with no docstrings. Their names are reasonably self-documenting, but a one-liner stating what the returned path represents (and its relationship to the escalation / backoff state machine) would help future readers.

---

No API keys, private IPs, SSH keys, credential strings, or exploitable wording in the diff, PR title, body, or CHANGELOG addition. `/home/ubuntu/genesis` in test files is a test fixture (known false positive).


---

WingedGuardian:
Addressed the structural review in 8b0cf966:

- **ERROR (/tmp fallback):** fixed — the runner's lock-dir fallback now uses `~/tmp` (inside `ReadWritePaths=%h`, honors the "~/tmp, never /tmp" convention). The primary path (`$GENESIS_HOME/locks`, under `$HOME`) was already fine; this hardens the never-fires fallback.
- **NOTE (docstrings):** added one-liners to `last_full_path()` / `full_backoff_path()` explaining their role in the escalation/backoff state machine.
- **WARNING (gitnexus test):** confirmed not vacuous — `test_surplus/test_jobs_extraction.py` tests `run_gitnexus_reindex` as an async *delegate* (structural: it asserts the scheduler delegates and the return propagates), not by mocking the removed `create_subprocess_exec`/`wait_for` path, so nothing passes vacuously. The new marker-write behavior is covered by the guardrail `test_triggers_enqueue_markers_and_do_not_spawn` (asserts the job references `index_marker` and no longer spawns the entrypoint).

WingedGuardian:
Addressed the Codex findings in 37408033:

- **P2 (use claimed state) — fixed.** Real bug: a commit coalescing during the runner's ~1s idle-sampling window meant the run used the stale `list` tools/mode. The runner now claims first and reads `repo/tools/mode` from `claim`'s output (the authoritative post-move-aside state), and makes the escalation decision on the claimed tools. Added a deterministic race test that coalesces `gitnexus`→`both/full` into the sampling window and asserts the entrypoint runs with the claimed `both`.
- **P1 (restore the JSON CLI form) — false positive.** Verified against the installed binary: `codebase-memory-mcp` **0.9.0** accepts `cli index_repository --repo-path <p> --mode fast --persistence true`, exits 0, indexes (17 nodes/16 edges on a probe repo), and writes `.codebase-memory/graph.db.zst`. The flags are listed in `index_repository --help` (`--repo-path` is `[required]`); the README's `'{"repo_path": …}'` JSON example is one supported form, not the only contract. The flag form is also what lets us pass `--mode`/`--persistence` at all.

PR #1050, opened 2026-07-14
fix(observability): register process_reaper_would_kill observation TTL

WingedGuardian:
## Why
The dry-run process reaper emits a `process_reaper_would_kill` observation each tick it finds a candidate (`process_reaper.py:483`), but that type was never registered in `_TTL_BY_TYPE` (only its real-kill sibling `process_reaper_kill` was). So **every emission logs an `Unknown observation type … assigning default TTL of 14 days` WARNING** — recurring noise (20+ lines over the last couple of days during the current dry-run audit).

## Change
Register `process_reaper_would_kill` explicitly at **14 days** — its current implicit default, so **zero behavior change** — mirroring the `provider_failure` precedent (an explicit entry added purely to silence the same warning). Documented as the audit-trail counterpart to `process_reaper_kill` (3d): the WOULD-KILL dry-run evidence is reviewed over multiple days before arming, so it keeps the longer window.

## Test
`tests/test_db/test_observations.py::test_process_reaper_would_kill_ttl_registered` — asserts the type is in `_TTL_BY_TYPE` and `_compute_ttl` resolves it without the unknown-type warning (TDD: failed before the fix, passes after). Full `test_observations.py` green (24), ruff clean.

### DIFF
--- a/src/genesis/db/crud/observations.py
+++ b/src/genesis/db/crud/observations.py
@@ -136,6 +136,11 @@ _TTL_BY_TYPE: dict[str, timedelta] = {
     "user_model_delta": timedelta(days=14),
     "capability_improvement": timedelta(days=14),
     "strategic_analysis": timedelta(days=14),
+    # process_reaper dry-run audit trail — the WOULD-KILL evidence an operator
+    # reviews before arming the reaper (set_operator_armed). Kept 14d (vs the 3d
+    # process_reaper_kill above) so a multi-day audit window survives, and made
+    # explicit here so it no longer logs the unknown-type warning every tick.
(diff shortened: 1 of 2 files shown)


WingedGuardian:

**PII/Secrets/Wording scan: CLEAN** — no private data, credentials, or disclosure wording found.

---

**WARNING — Import hygiene** (`tests/test_db/test_observations.py`)

`import logging` and `from datetime import timedelta` are placed inside the test function body rather than at module level. Both are always-available stdlib modules with no conditional availability concern. Per the import hygiene convention (stdlib → third-party → local), these belong in the module-level import block at the top of the file.

---

PR #1052, opened 2026-07-14
feat(guardian): F.3 — durable alert queue (store-and-forward for down transports)

WingedGuardian:
## What & why

Infrastructure alerts died silently when their transport was down: the host `AlertDispatcher` logged to journald and returned `False`, and `scripts/backup.sh` dropped its message if `curl` to Telegram failed. This adds a **durable, per-side alert queue** so an alert raised while the channel is unreachable is persisted and delivered on the next drain — the reliability floor under the guardian's other resiliency signals.

## Design — per-side topology

Each side drains its **own** queue (never a shared mount — that would invert reliability and race on delivery):

| Side | Queue | Enqueue | Drain |
|---|---|---|---|
| Host guardian | `<state>/alerts/queue/` | `AlertDispatcher.send` on total channel failure | top of `run_check` (30s tick) |
| Container | `~/.genesis/alerts/queue/` | `scripts/lib/alert_queue.sh` + Python callers | awareness tick (`runtime/init/alert_drain.py`) |

- **`genesis.guardian.alert.queue`** — dependency-free schema-v1 file queue: atomic `0600` writes, dedupe-key collapse, oldest-first `drain()` with a **terminal-vs-transient** send contract, `prune()` caps (200 files / 14 days). Container queue lives in a `queue/` subdir so it never collides with `tmp_watchgod`'s flag files.
- **Host** — `AlertDispatcher` gains `fallback_queue_root`; `run_check` drains-first each tick. The replay path passes `allow_queue_fallback=False` so a still-down channel never re-enqueues the entry it is replaying.
- **Container** — the drainer is wired unconditionally (valuable guardian-or-not) and resolves the outreach pipeline lazily per-tick, so bootstrap order is irrelevant. `REJECTED` (outreach dedup) is terminal → unlink (not a retry, else the entry wedges forever); the outreach `topic` derives from the alert identity so distinct alerts sharing a source stay independently deliverable.
- **Shell** — `tmp_watchgod` pages only the **emergency** tier, transition-only (once per red episode); the warning tier stays dashboard-only via `watchgod_state.json`. `backup.sh` queues on a failed send.

## Tests

51 targeted tests: queue primitive (atomicity/permissions/dedupe/oldest-first/terminal-vs-transient/quarantine/prune), dispatcher fallback + replay, shell lib round-trip + watchgod transition-only + a guard that the warning tier never pages. Plus a real-path E2E drill covering both drains end-to-end.

### DIFF
--- a/docs/reference/recovery-and-portability-workflow.md
+++ b/docs/reference/recovery-and-portability-workflow.md
@@ -168,3 +168,43 @@ edits and untracked files byte-identical); the old database is set aside as
 
 Exit codes: `0` healthy · `1` residual issues remain (escalate) · `2` aborted
 (no writable capture, or no origin URL resolvable for a config restore).
+
+## Durable Alert Queue (F.3)
+
+Alerts used to die silently when their transport was down: the host
+`AlertDispatcher` logged to journald and returned False, and
+`scripts/backup.sh`'s `_send_telegram` dropped the message if `curl` failed.
+F.3 adds a **store-and-forward queue** so an alert raised while Telegram is
+unreachable is persisted and delivered on the next drain.
+
+**Per-side topology — each side drains its OWN queue** (never a shared mount,
+which would invert reliability and race on delivery):
+
+| Side | Queue dir | Enqueue | Drain |
+|---|---|---|---|
+| Host guardian | `<state_path>/alerts/queue/` | `AlertDispatcher.send` when ALL channels fail | top of `run_check` (the 30s tick) |
+| Container | `~/.genesis/alerts/queue/` | shell (`scripts/lib/alert_queue.sh`) + any Python caller | awareness tick (`runtime/init/alert_drain.py`) |
+
+The queue (`genesis.guardian.alert.queue`) is one schema-v1 JSON file per alert
+(`{schema, ts, severity, source, title, body, dedupe_key, meta}`), written
+atomically at 0600. It is deliberately dependency-free so shell scripts write the
+same format via `queue_alert <severity> <source> <title> <body> [dedupe]`. The
+container queue lives in a `queue/` **subdir** so it never collides with
+`tmp_watchgod`'s `tmp_warning` / `tmp_emergency` flag files in
+`~/.genesis/alerts/`.
+
+**Delivery contract.** `drain(root, send)` calls `send(entry)` oldest-first:
+`True` = terminal (delivered OR intentionally deduped → unlink); `False` =
+transient failure (channel still down → keep the entry and stop the batch, retry
+next tick). The container drainer maps the outreach result: `DELIVERED`/
+`REJECTED` → terminal (a `REJECTED` dedup is redundant, not a failure to retry —
+treating it otherwise would wedge the entry forever); `FAILED`/`HELD`/`PENDING`
+→ keep. `prune()` bounds the queue (200 files / 14 days).
+
+**What pages vs. what stays quiet.** `tmp_watchgod` pages only the **emergency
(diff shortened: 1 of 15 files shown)


WingedGuardian (commented):
## First-pass review — PR #1052

**Structural review:** 2 WARNINGs, 1 NOTE — no ERRORs.
**PII/Secrets/Wording scan:** CLEAN.

---

### WARNING — Silent test skip: missing `@pytest.mark.asyncio` in `tests/test_guardian/test_alert_queue.py`

Six standalone `async def` test functions are missing the `@pytest.mark.asyncio` decorator:

- `test_drain_delivers_and_unlinks`
- `test_drain_stops_on_transient_failure`
- `test_drain_rejected_is_terminal_unlinks`
- `test_drain_send_raising_keeps_entry`
- `test_drain_respects_max_per_run`
- `test_drain_missing_root_is_noop`

The sibling file `test_alert.py` marks every async method explicitly with `@pytest.mark.asyncio`. Without `asyncio_mode = "auto"` in the pytest config, pytest-asyncio in strict/auto-off mode will **collect these functions but silently never run them** — no failure, no skip count, just invisible non-execution. That would leave the drain logic (the core of this whole feature) unverified by CI. Please confirm the project's `asyncio_mode` and add the decorator if needed.

---

### WARNING — New `src/` file without test coverage: `src/genesis/runtime/init/alert_drain.py`

The queue primitive in `genesis.guardian.alert.queue` is well-covered by `test_alert_queue.py`. However, `alert_drain.py`'s `wire()` path and the `_send` closure (which maps `OutreachStatus.DELIVERED`/`REJECTED` → terminal) have no direct tests. The `_send` mapping is the critical cross-module seam — a misfire here would leave queued alerts stuck forever or consumed without delivery. A targeted test for `_make_drainer` against a mock outreach pipeline would close this gap.

---

### NOTE — Missing type annotations on injected `rt` / `fn` parameters

- `alert_drain.py:wire(rt)` and `_make_drainer(rt)` — `rt` is untyped (likely `GenesisRuntime`)
- `loop.py:set_alert_queue_drainer(self, fn)` — `fn` is untyped (likely `Callable[[], Awaitable[None]]`)

The rest of the awareness loop setter API uses typed parameters. Consistent typing here aids refactoring safety.

---

No pattern ERRORs. No PII, secrets, private IPs, or sensitive wording found.

---

WingedGuardian:
Thanks — addressed in `84c109e3`:

- **WARNING-1 (missing `@pytest.mark.asyncio`)** — the project has `asyncio_mode = "auto"` (`pyproject.toml:132`), and I verified the 6 drain tests actually run (`-v` shows `PASSED`, not `SKIPPED`), so CI was genuinely exercising the drain logic. Added the explicit markers anyway: it matches `test_alert.py`'s convention and is robust to a future config flip. Good catch on the robustness angle.
- **WARNING-2 (`alert_drain.py` uncovered)** — added `tests/test_guardian/test_alert_drain.py` (7 tests) covering the `_send` seam directly: `DELIVERED`/`REJECTED` → unlink, `FAILED`/unwired-pipeline → keep, the identity-derived `topic` (`alert:<dedupe_key>`), and `wire()` with/without a loop.
- **NOTE (type annotations)** — consciously kept as-is: the immediate sibling setters (`set_credential_bridge(self, fn)`, `set_cred_integrity_fn(self, fn)`) and `cred_integrity.wire(rt)` are all untyped, so this matches the established local convention rather than diverging from it. `rt: GenesisRuntime` would also require a `TYPE_CHECKING` import for a passthrough delegate.

PR #1053, opened 2026-07-14
feat(session): charter/ledger DB store + MCP tools + injection (session-manager PR-2a)

WingedGuardian:
## Summary

Session-manager PR-2a: the charter moves from per-session \`charter.json\` files to canonical DB tables, and gains the **ledger** — durable per-session TODO/agreement rows that survive any number of compactions.

- **Migration 0058**: \`session_charters\` (origin write-once via \`WHERE origin_prompt IS NULL\`; living mission/pointers; compaction_count) + \`session_ledger\` (status: open/in_progress/done/absorbed/dropped; added_by: foreground/ambient/pulse; evidence for future absorption refs). DDL mirrored in \`_tables.py\`, registered in the schema guard. (0057 is claimed by in-flight #1046 — numbered around it.)
- **PreCompact hook** (\`genesis_precompact.py\`): stdlib-sqlite3 conversion — BEGIN IMMEDIATE + busy_timeout inside the 5s budget, fail-open on missing/busy DB (self-heals next compaction; never creates a DB as a side effect). Regenerates the \`charter.md\` mirror (now with ledger); stops writing \`charter.json\`.
- **CRUD** (\`db/crud/session_charters.py\`) + canonical renderer (\`genesis/session_charter.py\`; the hook keeps a parity-tested stdlib duplicate).
- **4 MCP tools on genesis-health**: \`session_charter\` (read), \`session_charter_update\` (mission/pointers), \`session_ledger_add\` (agreement capture at the moment it happens), \`session_ledger_update\` (done/absorbed/dropped + evidence). Living fields only — origin is not addressable; stub rows allow writes before the first compaction.
- **Read paths**: SessionStart injection is DB-first (mode=ro) with open ledger items + counts, legacy \`charter.json\` fallback kept; new per-turn \`[Charter: <mission> | open: N]\` drift tag (UserPromptSubmit, fail-open omission matrix).
- **Backfill**: \`scripts/backfill_session_charters.py\` — idempotent INSERT OR IGNORE import of legacy files; run once at deploy after the migration applies.
- **Docs**: CHANGELOG, architecture map, and the CLAUDE.md agreement-capture rule (the tool ships with its caller).

## Deploy order

Merge → **restart genesis-server immediately** (applies 0058; hooks go live at main-tree update and fail open until the table exists) → run the backfill once.

## Test plan

- 96 tests across 5 files (schema guard + CHECKs; CRUD immutability/stub/ledger lifecycle; hook first/bump/stub-fill-once/busy-DB/missing-DB/parity; injector DB-first/fallback/budget; tag omission matrix; backfill idempotency) — all green locally, plus the full tests/test_db/ suite (993) and the MCP transport smoke test.


WingedGuardian:
## First-pass review — PR #1053 (session-manager PR-2a)

**Structural review:** 1 WARNING · 1 NOTE

---

### WARNING — Hook imports third-party `aiosqlite`

**File:** `scripts/genesis_session_context.py` — new `_load_charter_db` function

```python
async def _run() -> tuple[dict | None, list[dict]]:
    async with aiosqlite.connect(
        f"file:{db_file}?mode=ro", uri=True, timeout=2
    ) as db:
```

`aiosqlite` is a third-party package. The Genesis hook convention requires hooks to be stdlib-only or `genesis.env`. The fail-open `try/except Exception` wrapper means a missing `aiosqlite` degrades gracefully, but the violation of the convention is worth flagging.

Mitigating context: I cannot see the full file (the diff starts at line 677), so it's possible `aiosqlite` is already imported at module level in the existing script — if so, this is a non-issue. If it isn't, the function-level lazy import adds a non-stdlib dep to the hook.

Action: confirm whether `genesis_session_context.py` already imports `aiosqlite` at module level. If it does, this warning is moot.

---

### NOTE — Missing docstrings on three public CRUD functions

**File:** `src/genesis/db/crud/session_charters.py`

- `get` (line ~844) — no docstring; one-liner, but it's public API and the module has a contract around its behavior (immutability, Row → dict decoding via `_decode_pointers`).
- `get_ledger_item` (line ~955) — no docstring.
- `ledger_counts` (line ~981) — no docstring.

The module-level docstring is thorough; individual function docs would close the gap for callers of the CRUD layer.

---

### Things that look good

- All new `src/genesis/` files have corresponding test files — no missing coverage.
- `genesis_precompact.py` (PreCompact hook) remains strictly stdlib-only (`sqlite3` is stdlib). The intentional renderer duplicate + parity test is the right pattern.
- No bare `except:` clauses — all exception handlers are typed.
- No mutable default arguments.
- No raw SQL outside the hook scripts (which are intentionally stdlib-only) and the migration (which is the correct carve-out). The `ledger_update` f-string SQL uses a hardcoded column-name allow-list with an appropriate `noqa: S608` annotation.
- No direct `task_states` inserts.
- `/tmp/t.jsonl` strings in test fixtures are DB field values, not actual temp-file creation — correctly covered by the test-fixture false-positive rule.
- Conventional commit prefix present (`feat(session):`).

---

_First-pass structural review. Logic correctness and deeper design are out of scope here._

---

WingedGuardian:
## Review responses (e29e8780)

**Codex P2-1 (prefix-before-stub orphaning) — FIXED.** `resolve_session_id` now falls back to `cc_sessions.cc_session_id` prefix matching for sessions that haven't chartered yet, and `session_charter_update`/`session_ledger_add` refuse to create rows under an unresolved short id (loud error telling the caller to pass the full id). Tests: resolver-via-cc_sessions at both CRUD and tool layer + no-stub-on-unresolved.

**Codex P2-2 (backfill skips stub rows) — FIXED.** `import_charter` now fills `origin_prompt`/`origin_ts` (+ COALESCE transcript_path) on existing rows via `WHERE origin_prompt IS NULL` — living fields untouched, immutability contract intact (every origin write remains NULL-scoped). Returns `imported | origin_filled | skipped`; backfill counts and prints each. Tests: CRUD stub-fill + backfill stub-fill + idempotent re-run after MCP edits.

**Structural WARNING (aiosqlite in the SessionStart hook) — moot, as the review itself anticipated.** `genesis_session_context.py` already lazy-imports `aiosqlite` at two pre-existing sites (L0 code-modules read ~:364 and the in-flight block ~:659); this hook has never been stdlib-only (it lazy-imports genesis too). The stdlib-only convention applies to `genesis_precompact.py`/`genesis_session_end.py`, both unchanged in that respect. The new read is also wrapped in the fail-open `try/except` with a file fallback.

**Structural NOTE (missing docstrings) — FIXED** on `get`, `get_ledger_item`, `ledger_counts`.

PR #1057, opened 2026-07-14
feat(eval): A4 graph arm — linked ephemeral store + 1-hop recall expansion

WingedGuardian:
## Summary

Sprint PR-B of the A4-improvement track: adds a **graph arm** to the LongMemEval harness to answer, with data, whether Genesis's memory graph lifts multi-session accuracy (0.33 on the definitive baseline). The arm ingests each question's haystack into a store with the real `MemoryLinker` wired (`auto_link=True`, prod-faithful later→earlier link topology), lets the retriever's existing data-driven graph boost activate, and adds the one piece prod doesn't have: **1-hop link expansion that actually merges linked-but-unretrieved neighbors into the reader context** (the MCP layer only decorates results with neighbors today).

## Key design decision (genesis-architect pre-review)

Baseline and graph arms run on **separate ephemeral stores**. Links shift ranking for every arm sharing a store through TWO channels — `_apply_graph_boost` AND activation connectivity (`activation.py` link_count term, ±15%, bidirectional, no floor gate) — so a shared linked store would tint the baselines and break comparability with published numbers. The second ingest re-embeds through the run-level embedding cache (marginal cost ≈ in-memory upserts + one local Qdrant search per turn).

## Changes

- `Arm` gains `graph`; deterministic label order (`raw+rerank+graph`) → stable `model_profile`/dump filenames.
- `ephemeral_store(with_linker=, link_threshold=)`: opt-in real `MemoryLinker`; exposes `db`. Opt-in because `MemoryStore.store()` defaults `auto_link=True` — an always-present linker would let future direct `store()` calls create links silently on baseline stores.
- `ingest_haystack(auto_link=)`: links form at insert time exactly as prod does (a post-hoc linking pass would create edges to *later* memories — a topology prod can never build).
- **New CRUD helper `memory_links.neighbors_of(db, ids, *, exclude, limit)`**: both directions, one row per neighbor via `GROUP BY` + `MAX(strength)` (the PK `(source,target,link_type)` allows multiple rows per pair; a naive UNION burns cap slots on duplicates). Canonical home next to `inter_candidate_links`; the committed prod graph/entity-recall wiring reuses it. Deliberately NOT `graph.traverse` — its process-global NetworkX cache is unsafe across the concurrent ephemeral stores this runner creates (documented in code).
- **Dual evidence metrics**: top-K (`evidence_recalled`/`evidence_coverage`, baseline-comparable) AND post-expansion (`*_final`) + `links_created`/`expanded_ids` per question in the JSONL dump, `links_created_mean`/`expanded_mean` in run metadata — splits ranking-lift (boost) from coverage-lift (expansion).
- **Zero-link graph store warns loudly** (no-silent-caps parity with the existing rerank warning) — a link-free graph arm would otherwise silently equal its baseline twin.
- `MemoryLinker`: constructor-level `similarity_threshold` becomes `auto_link`'s default — the kwarg default was unreachable from `MemoryStore.store()`'s call site (it passes no threshold). Both prod call sites are keyword-only and unchanged (default 0.75 preserved; verified: `mcp/memory/__init__.py`, `runtime/init/memory.py`).
- CLI: `--graph` ADDS a `+graph` variant of every selected arm (paired baseline-vs-graph in one run, one dump dir); `--graph-link-threshold`.

## Testing (TDD — 10 new tests written first, watched fail)

- `neighbors_of`: direction+type dedup w/ MAX(strength), exclusion, cap, seeds-never-returned, empty input (`tests/test_db/test_memory_links.py`).
- Linker: constructor threshold is auto_link's default; explicit kwarg still overrides; existing linker suites untouched and green.
- Harness: `with_linker` ingest creates `memory_links` rows in the ephemeral SQLite; default store stays link-free even with `auto_link=True` (no silent links); graph label determinism.
- **Orchestration (fakes, no network)**: dual-store isolation + expansion e2e — a 2-evidence multi-session question at k=1 shows baseline coverage 0.5 with 0 links, graph arm coverage 0.5 → `coverage_final` **1.0** via expansion of the linked sibling; zero-link warning fires on an unlinkable haystack.
- Full targeted suites: **408 passed** (eval + db links + linker + linker_typed + WS-3 coverage guards; the recall-site guard caught the `.recall()` move into `_run_arm` and the registry was updated accordingly).
- `ruff check --no-cache`: clean.

## Impact

Benchmark harness + two additive prod seams (linker constructor default, new CRUD read helper). No runtime/server behavior changes; no deploy needed. Next: link-density smoke → paired multi-session slice (raw vs raw+graph, 133 Qs) → full-500 graph run gated on ≥+5pp multi-session lift.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,18 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **The memory benchmark can now measure whether Genesis's memory graph
+  actually helps.** A new `--graph` mode runs every benchmark arm twice — once
+  against a plain store and once against a store where memories link to
+  similar earlier memories exactly as they do in production — and follows
+  those links at recall time to pull in related memories the search itself
+  missed. Baseline and graph runs use fully separate stores, so the
+  comparison is honest (links can't quietly tint the baseline's ranking). Per
+  question, the results record how many links formed and how much extra gold
+  evidence the graph surfaced, and a graph run that formed no links says so
+  loudly instead of silently matching its baseline. The memory linker's
+  similarity threshold is also now configurable per instance instead of fixed.
+
 - **The memory benchmark now grades temporal questions fairly and explains
   its misses.** The LongMemEval reader gets the question's date (the
   benchmark's own convention — without it, "how many weeks ago…" questions
--- a/src/genesis/db/crud/memory_links.py
+++ b/src/genesis/db/crud/memory_links.py
@@ -94,10 +94,7 @@ async def batch_link_counts(
         for row in inbound_rows:
             inbound_map[row[0]] = inbound_map.get(row[0], 0) + row[1]
 
-    return {
-        mid: (total_map.get(mid, 0), inbound_map.get(mid, 0))
-        for mid in memory_ids
-    }
+    return {mid: (total_map.get(mid, 0), inbound_map.get(mid, 0)) for mid in memory_ids}
 
 
 async def inter_candidate_links(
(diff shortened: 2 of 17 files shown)


WingedGuardian (commented):

---

PR #1059, opened 2026-07-15
feat(resilience): adaptive OOM resilience — swap + systemd-oomd, codified

WingedGuardian:
## Why

A machine that runs out of memory with **no swap and no userspace OOM killer** doesn't lose one process — it loses everything at once. The kernel enters direct-reclaim thrash, load climbs into the hundreds with processes stuck in uninterruptible sleep, SSH stops answering, and the box wedges until someone power-cycles it. One incident series produced this same end state four times from four unrelated triggers; the common factor was the missing relief valve, not any single leak. The fix was proven live (swap + pressure-based systemd-oomd) — this PR codifies it so every install is born protected and existing installs retrofit automatically.

## What

**Setup (adaptive — pressure percentages, never absolute bytes):**
- `scripts/lib/memory_resilience.sh`: idempotent systemd-oomd pressure-kill policy (user.slice kill @60% PSI backstop + per-user-manager kill, the operative monitor in practice) + a **warn-only** swap invariant check with vantage-aware remediation (container: the host-side `limits.memory.swap` knob; bare/VM: add swap). Degrades gracefully without systemd, systemd-oomd, kernel PSI, or non-interactive sudo. Honest reporting: a failed write warns and the summary says NOT fully applied.
- `bootstrap.sh` runs it on fresh installs; `update.sh` re-runs it visibly (its bootstrap invocation pipes through `tail -10`, which would eat the warning on exactly the retrofit path).
- `host-setup.sh`: `limits.memory.swap=true` on managed containers (retrofittable) + warns when the host itself is swapless.
- `genesis-server.service`: `ManagedOOMPreference=avoid` — the systemd-oomd counterpart of the existing `OOMScoreAdjust=-500`, honored by the per-user pressure monitor, so the greedy session tree dies before the cognitive core.

**Body schema (infra_profile):**
- Container facts: `cgroup_memory_swap_max` (tri-state; `0` IS the wedge state) and `oomd_user_slice_kill` (config-plane scan with real systemd drop-in semantics: lexicographic, last assignment wins, full-line comments only).
- Host plane: `swap_total_kb` fact (its disappearance is the wedge precondition — drift-worthy) + `swap_free_kb` metric.
- The annotation layer flags unprotected installs from these facts — no special-case code.

**Docs:** `docs/reference/memory-resilience.md` — the invariant, the diagnostic fingerprint, what is set up vs only warned about, and how this layers with the existing kernel-side sysctl OOM tuning.

## Tests

41 targeted tests green: the lib is exercised as shipped (sourced real, stubbed `sudo`/`systemctl`/`systemd-detect-virt`/`swapon` on PATH) — exact drop-in content with a percentages-only tripwire, idempotent second run with zero systemd churn, all four graceful skips, vantage-aware warnings, failed-write honesty; collector facts cover the swap tri-state, drop-in override ordering, and the host facts/metrics split.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,19 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   no longer forces the full four-arm spend. Unknown labels fail fast with
   the selectable universe listed — a paid run never silently widens.
 
+- **A memory spike now degrades gracefully instead of wedging the machine.**
+  On systems that ran out of memory with no swap and no userspace OOM killer,
+  one greedy process could drag the whole box into unrecoverable thrash —
+  load in the hundreds, SSH dead, everything down together. Installs now set
+  up systemd-oomd with pressure-percentage kill policies (adaptive to any
+  machine size, applied automatically on bootstrap and on your next update,
+  skipped cleanly where systemd/oomd/PSI aren't available), the Genesis
+  server marks itself `avoid` so the greedy session tree dies first, managed
+  container installs get swap enabled (`limits.memory.swap`), and setup warns
+  with exact remediation when swap is missing or disabled. The infrastructure
+  profile records the swap/oomd state as facts, so an unprotected install is
+  flagged in `INFRASTRUCTURE.md`. Runbook: `docs/reference/memory-resilience.md`.
+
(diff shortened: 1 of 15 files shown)


WingedGuardian:

---

PR #1060, opened 2026-07-15
fix(dashboard): _async_route degrades to 503 instead of crashing in the startup window

WingedGuardian:
## Problem

`_async_route` wraps every dashboard Flask route. When the Genesis runtime loop is unavailable, it falls back to running the handler on a fresh per-request event loop. But shared async objects (aiosqlite connections, asyncio locks) are bound to the runtime loop — on that fallback path a handler touching the DB raises cross-loop `RuntimeError`s ("Lock is bound to a different event loop" → "no current event loop"), which propagated and could crash the whole server (observed exit code 2 when a health poll landed during a misconfigured startup).

## Fix

Two degradation legs, converging on the 503 idiom the voice API and OpenClaw completions routes already use:

- **Loop configured but not running** (startup/shutdown window): return 503 `{"status": "unavailable"}` without executing the handler on any substitute loop.
- **No loop configured** (unit tests / non-standalone hosting): keep the fallback, but degrade a `RuntimeError` escaping the handler to a logged 503 instead of a propagated crash. Handlers still execute normally on this path — only the error disposition changes.

## Tests

Three new tests pin both legs (stopped-loop → 503 with the handler never invoked; cross-loop `RuntimeError` → 503; plain handlers still execute on the loop-less path). Full `tests/test_dashboard/` suite: 291 passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,15 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **A dashboard request during startup can no longer crash the server.** The
+  async-route bridge falls back to a throwaway event loop when the runtime
+  loop isn't available — but shared database connections are bound to the
+  runtime loop, so a health poll landing in that window raised cross-loop
+  errors that could take the whole process down (observed as exit code 2).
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

PR #1066, opened 2026-07-15
fix(guardian): F.4 — lazy-import check_git_cheap for host guardian venv

WingedGuardian:
## Bug (live-E2E-caught during #1061 deploy)

The host guardian runs a **minimal venv without `aiohttp`**. `repo_bundle` imported `check_git_cheap` at module level, and `genesis.observability.__init__` eagerly loads the aiohttp health chain — so the host importing `bundle_watch → repo_bundle` threw `ModuleNotFoundError: No module named 'aiohttp'` on **every 30s tick**, and F.4's host-side archive + freshness never ran. The never-raise wrapper kept the guardian up (only a logged traceback), so CI + container tests couldn't see it — the container venv has aiohttp.

## Fix

`check_git_cheap` is only used **container-side** (the publish health gate). Move it to a **lazy import inside `publish_repo_bundle`**; module-level `repo_bundle` now imports only `genesis.env` (host-safe). Adds a **subprocess regression test** asserting `bundle_watch` + `repo_bundle` import without pulling `genesis.observability.git_health` / `aiohttp`.

Verified locally: `import genesis.guardian.bundle_watch` no longer loads `git_health`/`aiohttp`; `is_valid_bundle_name` still exported; publish still gets the health gate (container path). 24 F.4 tests green.

Deploy: touches `src/genesis/guardian/` → host guardian redeploy via `scripts/update.sh` after merge, then re-run the F.4 archive E2E.

### DIFF
--- a/src/genesis/guardian/repo_bundle.py
+++ b/src/genesis/guardian/repo_bundle.py
@@ -46,7 +46,6 @@ from datetime import UTC, datetime
 from pathlib import Path
 
 from genesis.env import genesis_home, repo_root
-from genesis.observability.git_health import check_git_cheap
 
 logger = logging.getLogger(__name__)
 
@@ -183,6 +182,11 @@ async def publish_repo_bundle(
 
         # HEALTH GATE — never overwrite the last good bundle with an attempt from
         # a degraded repo. check_git_cheap is ms-scale and already threaded.
+        # Imported lazily: this module is imported host-side (bundle_watch →
+        # repo_bundle) where the minimal guardian venv lacks genesis.observability's
+        # aiohttp chain — and the host never runs publish, so it never needs this.
+        from genesis.observability.git_health import check_git_cheap
+
         report = await check_git_cheap(repo)
         if not report.ok:
             logger.warning(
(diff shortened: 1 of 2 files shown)


WingedGuardian (commented):

---

**Structural pass** (2 files, 32 line diff)

- `src/genesis/guardian/repo_bundle.py` — lazy `from genesis.observability.git_health import check_git_cheap` inside `publish_repo_bundle` is the correct fix. No bare `except:`, no mutable defaults, no `/tmp/`, no raw SQL, no `git add -A`, no `nohup`.
- `tests/test_guardian/test_bundle_watch.py` — subprocess-isolated regression test is the right technique here (import-cache state from the rest of the suite can't mask the regression). `import subprocess` / `import sys` inside the function body is intentional and appropriate given the test's purpose.
- Import groups in `repo_bundle.py` are correct post-change: stdlib → `genesis.env` (local). No star imports, no new circular dependency introduced.
- Conventional commit prefix `fix(...)` ✓. Existing test file extended — no new `.py` in `src/` without coverage.

**PII/Secrets/Wording pass**

- Added lines: no API keys, tokens, private IPs, private hostnames, credentials, absolute paths with usernames, or hardware identifiers.
- PR title, description, and no CHANGELOG additions: wording is purely technical. No language advertising a credential leak, security incident, or private infrastructure detail.

---

PR #1070, opened 2026-07-15
feat(resilience): codify eth0 KeepConfiguration + self-healing networkd watchdog

WingedGuardian:
## Why

Under memory pressure the container's `systemd-networkd` hit rtnetlink timeouts three times in three days (Jul 12/13/14: `eth0: Could not set route: Connection timed out` → `eth0: Failed`), dropped its DHCP lease, and fell off the network until a **manual** `systemctl restart systemd-networkd` hours later. The fix was hand-applied live on one install only — every fresh install is still born with the defect, and the *detection/heal* half was never built (eth0 sat in `AdministrativeState=failed`, silently, for 15+ hours).

Two layers, mirroring the memory-resilience PR (#1059) — survive → self-heal — both codified so fresh **and** existing installs get them:

## What

**Layer 1 — the address survives the failure.** `scripts/lib/network_resilience.sh` writes `KeepConfiguration=true` as a drop-in on the default-route link's `.network` unit (interface + unit **discovered live**, never hardcoded). `true` is the `yes ⊃ dhcp ⊃ dhcp-on-stop` superset (systemd.network(5)) — exactly what netplan `critical: true` renders to — so one mechanism delivers both hand-applied protections.

**Layer 2 — the daemon heals itself.** `scripts/systemd/genesis-network-watchdog.sh` + a root `.timer` (~2 min) restart networkd on any of: daemon inactive (not masked), a managed link in `AdministrativeState=failed`, or no IPv4 default route. Address-preserving under Layer 1. Rails: 2-min grace (don't fight a settling daemon), 10-min rate limit (a persistent fault logs loudly instead of flapping). Writes `/run` heal telemetry.

**Wiring:** `bootstrap.sh` (fresh) + `update.sh` (existing installs retrofit, unpiped so warnings are visible). Both layers are container-side, so sibling-install parity is automatic on its next update.

**Body schema:** `collect_network` gains config-plane facts `networkd_keep_configuration` + `network_watchdog_installed` (annotation layer flags unprotected installs) and a volatile `watchdog` heal-telemetry metric.

Graceful degradation: no systemd / no networkctl / NetworkManager host / no sudo each skip cleanly.

## Tests

- `tests/test_scripts/test_network_resilience.py` (22): stubbed-PATH lib apply (idempotent drop-in/units, every guard skip, honest failed-write) + watchdog detect→heal (all three triggers, masked-skip, grace + rate limit, telemetry).
- `tests/test_infra_profile/test_collectors_container.py` (+5, first-ever `collect_network` tests): both facts via `etc_root` seam, watchdog metrics via `run_root` seam, determinism.

## Verification

Live assumptions verified this session: `networkctl status` Network-File parsing, `ActiveEnterTimestamp`→`date -d` grace parsing, `ip -j route` dev field, networkd enabled+active. The box's eth0 is `AdministrativeState=failed` **now** — post-merge E2E is a real fire: the watchdog's first tick heals it (address-preserving), plus an operator-approved stop-networkd drill.

Docs: `docs/reference/network-resilience.md` (runbook), `CURRENT.md`, `CHANGELOG`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,19 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   rows get their metadata classified but remain wing-unreachable until the
   retrieval-side gap is closed (reported separately, not overstated). Dry-run
   by default; the bulk write is gated on a human-reviewed sample.
+
+- **Genesis survives — and now auto-recovers from — a wedged network.** Under
+  heavy memory pressure the container's networking daemon can hit kernel
+  timeouts, drop its DHCP lease, and take the machine off the network until
+  someone restarts the daemon by hand. Fresh and updated installs now (1) pin
+  the address so a networking failure keeps the connection instead of dropping
+  it, and (2) run a lightweight watchdog that detects a wedged/inactive
+  networking daemon and restarts it automatically (address-preserving, so no
+  blip). Heal events are recorded and surfaced in the infrastructure profile,
+  so a recurring fault is visible instead of silent. Networking runs a
+  different manager (e.g. NetworkManager) or lacks sudo? It skips cleanly.
+  See `docs/reference/network-resilience.md`.
+
 - **Genesis now notices when merged code isn't actually deployed.** Pulling
   changes with a bare `git merge` between updates loads the new code on
   restart but silently skips everything only `update.sh` activates — new
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -751,7 +751,12 @@ verified: bca86011 2026-07-15
   state) + `oomd_user_slice_kill` (config-plane scan of user.slice.d drop-ins,
   laid down by `scripts/lib/memory_resilience.sh` from bootstrap/update) and
   host-plane `swap_total_kb`, so the annotation layer flags unprotected
-  installs (see docs/reference/memory-resilience.md).
+  installs (see docs/reference/memory-resilience.md). Network-resilience
+  invariants are first-class too: container `networkd_keep_configuration` +
+  `network_watchdog_installed` (config-plane facts from
+  `scripts/lib/network_resilience.sh`) plus a volatile `watchdog` heal-telemetry
+  metric from `/run/genesis-network-watchdog.json` (see
+  docs/reference/network-resilience.md).
 - **restore/**: thin CLI → `scripts/restore.sh` (counterpart of the 6h
   encrypted `scripts/backup.sh` timer).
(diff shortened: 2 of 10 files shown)


WingedGuardian:
### Automated Code Review — PR #1070

#### Structural Review: CLEAN

No pattern violations found.

- No `nohup`, no direct `task_states` inserts, no `/tmp/` writes, no `bare except:`, no mutable default arguments, no star imports, no `git add .`, no force-push, no `IntervalTrigger` misuse, no `$CLAUDE_PROJECT_DIR` in Python context.
- `Path("/etc")` / `Path("/run")` defaults in `collect_network` are immutable objects — not mutable default argument violations. ✓
- Inline Python in `genesis-network-watchdog.sh` uses only `json` and `os` (stdlib). ✓
- New public functions `_networkd_keep_configuration` and `_read_watchdog_state` in `container.py` both have docstrings. ✓
- Test coverage: `tests/test_infra_profile/test_collectors_container.py` (+5 tests) covers the new Python facts/metrics; `tests/test_scripts/test_network_resilience.py` (+22 tests) covers both shell scripts end-to-end. ✓
- Conventional commit prefix: `feat:` ✓

---

#### PII/Secrets/Wording Scan

**SOFT WARNING** — `tests/test_scripts/test_network_resilience.py` (line ~295)

```python
"WD_ROUTE_OUT": "default via 10.0.0.1 dev eth0",
```

A private IP address (`10.0.0.1`) appears as the default value in the watchdog test stub's route variable. This is clearly a generic, synthetic placeholder (not an actual infrastructure address), but the known-false-positive exemption for test fixtures is scoped to *paths*, not IP addresses — so flagging per policy. No action required if this is intentional test data.

PR title, body, and CHANGELOG additions: **CLEAN** — no language suggesting a leak, security incident, or private infrastructure exposure.

---

WingedGuardian:
Addressed all three Codex P2 findings in 565ef8f2:

1. **Skip crashed networkd** — the `is-active`-only guard now accepts active **or** enabled, so a stopped-but-ours networkd (the exact state the watchdog heals) still gets the watchdog installed; only genuine other-manager hosts (disabled/masked/absent) skip.
2. **Failed restart reported as healed** — a nonzero `systemctl restart` no longer stamps the rate-limit window or records `healed`; it records `restart-failed` and retries next tick.
3. **KeepConfiguration collapsed across interfaces** — last-assignment-wins is now evaluated **within each `.network.d`** and OR'd across links, so a drop-in on one interface can neither fake nor mask another's protection.

Tests extended for each (inactive-but-enabled installs watchdog; failed-restart not healed + rate-limit not armed; per-network-file scoping isolation). 37 targeted tests pass; shellcheck + ruff clean.

PR #1073, opened 2026-07-15
perf(memory): batch neighbor hydration in graph expansion (~5x faster)

WingedGuardian:
## What

Fast-follow to #1069. Recall-time 1-hop graph expansion hydrated each linked neighbor with its own `memory_crud.get_by_id` call. Because `memory_fts` is an FTS5 virtual table with no plain index on `memory_id`, every one of those is a full-content scan — so a 10-neighbor expansion did 10 scans.

Replaced the N+1 loop (plus the separate `origin_class_by_ids` call and the separate visibility-filter query, which hit the same rows) with **one** batched `hydrate_for_expansion` query joining `memory_fts ⋈ memory_metadata`. Neighbor lists are cap-bounded (≤25 by the settings validator) so a single `IN` clause never approaches the 999 bind-variable ceiling — no chunking needed.

## Why (measured, live prod DB, read-only A/B)

```
OLD  10x get_by_id : 749-937 ms   (10 unindexed FTS scans)
NEW  1x batched    :  77-126 ms
```

~6-10x on the hydration portion; `expand_neighbors` overall drops from ~680ms to ~150-200ms. `neighbors_of` itself was already 1ms (indexed) — the FTS scan was the whole cost. This latency was the #1 regression marker flagged at the #1069 live flip.

## Behavior-preserving

Same neighbors, same order, same provenance (stored-first `origin_class`, `source_pipeline=None`), same expired/deprecated visibility filtering — now derived from the batched row instead of separate queries. Guarded by the full existing `test_graph_expansion` + `test_memory_recall_graph_expansion` suites (49 tests green), plus a new `test_hydrate_for_expansion` CRUD test pinning the contract (content/tags/collection from FTS, origin_class/invalid_at/deprecated from metadata, missing ids omitted).

## Deploy

MCP-child surfaces activate next CC session; the dashboard tool-API recall surface needs a `genesis-server` restart (same as #1069). Pure perf — no config, no behavior flip.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -144,6 +144,15 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   context is unaffected by design: it never ingested these reflections in
   the first place.
 
+- **Recall-time graph expansion got ~5x faster.** The just-shipped 1-hop
+  expansion hydrated each linked neighbor with its own database query — and
+  because the memory content table is a full-text index (no plain lookup on
+  the id column), every one of those was a full scan. On a typical 10-neighbor
+  expansion that measured ~750-940ms of pure overhead on the recall path. It
+  now hydrates every neighbor in a single batched query (~80-130ms), returning
+  identical results. Purely a performance fix — same neighbors, same order,
+  same provenance and visibility filtering.
+
 - **A dashboard request during startup can no longer crash the server.** The
   async-route bridge falls back to a throwaway event loop when the runtime
   loop isn't available — but shared database connections are bound to the
--- a/src/genesis/db/crud/memory.py
+++ b/src/genesis/db/crud/memory.py
@@ -259,6 +259,47 @@ async def origin_class_by_ids(db: aiosqlite.Connection, ids: list[str]) -> dict[
     return out
 
 
+async def hydrate_for_expansion(db: aiosqlite.Connection, ids: list[str]) -> dict[str, dict]:
+    """Batch memory_id → {content, source_type, tags, collection, origin_class,
+    invalid_at, deprecated} for graph-expansion neighbor hydration.
+
+    ONE FTS⋈metadata query replaces the recall-time expansion's former
+    per-neighbor ``get_by_id`` loop plus its separate ``origin_class_by_ids``
+    and visibility-filter passes (the N+1 that dominated expansion latency).
+    Neighbor lists are cap-bounded (≤25 by the settings validator), so a
+    single IN-clause never approaches the 999 bind-variable ceiling — no
+    chunking needed. Missing ids are omitted (dangling links). ``collection``
+    is read from the FTS row (authoritative at retrieval, matching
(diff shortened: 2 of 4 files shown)


WingedGuardian:
## Code Review — PR #1073

### Structural review

**WARNING — Empty test file committed**

`tests/test_db/test_memory_crud.py` was added as a 0-byte file (SHA `e69de29b`). The `test_hydrate_for_expansion` test landed in `tests/test_db/test_search_ranked_invalid_at.py` instead. The empty file is either an accidentally committed placeholder or the test was placed in the wrong module. Either remove the empty file or move the test into it (a test for `memory_crud.hydrate_for_expansion` living in a file named `test_search_ranked_invalid_at.py` is a mild navigation pain, though not a blocker). No content is lost either way, but the empty file will confuse future readers.

**NOTE — Non-standard commit prefix**

The PR title uses `perf(memory):`, which is a valid Conventional Commits type but is not in the Genesis-listed set (`feat:`, `fix:`, `refactor:`, `docs:`, `test:`, `chore:`). Minor — worth aligning if the project enforces its own subset.

---

CLEAN — no credentials, private IPs, hostnames, or security-incident language found in added lines, PR metadata, or CHANGELOG additions.

---

The SQL construction in `hydrate_for_expansion` (f-string with `?` placeholders, values bound separately, `# noqa: S608` annotated) is correct for a CRUD module and is not a violation.

---

WingedGuardian (commented):
## First-pass review: 2 findings (WARNING + NOTE) — no hard blocks

**Structural:** 2 findings. **PII/Secrets/Wording:** CLEAN.

---

### WARNING — Non-standard conventional commit prefix

The PR title uses `perf(memory):` but the Genesis project's listed commit prefixes are `feat:`, `fix:`, `refactor:`, `docs:`, `test:`, `chore:`. `perf:` is not in that list.

Minor, but worth aligning for consistency in the changelog automation pipeline. The performance improvement could reasonably carry `fix:` (it addresses a latency regression surfaced by the #1069 flip) or stay as-is if the project accepts `perf:` informally.

---

### NOTE — Empty test file committed

`tests/test_db/test_memory_crud.py` was added as a completely empty file (zero bytes). The `test_hydrate_for_expansion` test was placed in `tests/test_db/test_search_ranked_invalid_at.py` instead — coverage is fine there — but the empty file is misleading and will likely confuse future contributors who expect to find `memory_crud` tests in it.

Options:
- Move `test_hydrate_for_expansion` into `test_memory_crud.py` and give it a stub `__doc__` or fixture import, **or**
- Drop the empty file from this commit.

---

**Everything else is clean.** The batched `hydrate_for_expansion` query is well-constructed (placeholder binding is correct, `noqa: S608` suppression is justified). Visibility filtering (bitemporal expiry + deprecated) is correctly preserved inline. The `now_iso` timestamp is captured once before the neighbor loop, which is correct. Raw SQL is in the crud module where it belongs. No PII, no secrets, no wording issues.

---

PR #1074, opened 2026-07-16
feat(guardian): out-of-band RAM tiered alert (E-rest PR-E1)

WingedGuardian:
## What & why

The only whole-RAM Telegram alert today is **container-side** (`infra:container_memory_high` in the health snapshot) — which dies exactly when the container thrashes/OOMs, the moment it matters most. The host-VM **guardian** (outside the container blast radius) already *measured* container RAM but never dispatched an alert. This wires it into the same tiered, hysteresis-damped alert path the disk pool already uses, so the guardian pages **out-of-band**.

## Design

Two axes, **worst-of** (mirrors the disk pool's data%/metadata%):
- **Container RAM** — anon+kernel % of the container's cgroup `memory.max`, via `incus exec`. Best-effort: that exec can itself stall under severe memory pressure.
- **Host-VM RAM** — `100 − MemAvailable/MemTotal` from the guardian's own `/proc/meminfo`. **No incus-exec → the reliable signal in a thrash.** The worst-of means the host axis still fires the page when the container read stalls.

Reuses `pool.py`'s pure `_tier_for`/`decide_alert`/`AlertDecision` hysteresis (verified pure by reading the source), a `memory_alert_state.json` for realert damping, and the guardian's own Telegram dispatcher.

## Changes
- **NEW** `guardian/memory_watch.py` — measurement + worst-of tier + alert orchestration
- `config.py` — `MemoryTiersConfig` (container 75/85/92, host 80/88/94, realert 6h)
- `health_signals.py` — extract `measure_container_mem_pct` (shared/DRY; `check_memory_pressure` behaviour preserved)
- `check.py` — `_check_memory_and_alert` wired after the storage-pool check (never-raise)
- gateway `ram-status` read-only verb (skew-guarded) + `remote.ram_status()` + `--ram-status`
- `config/guardian.yaml` `memory_tiers` section; `docs/architecture/CURRENT.md` guardian entry

## Tests / verification
- `test_memory_alert.py` — tier matrix, hysteresis (rise/recover/sustained-realert), disabled-silent, both-axes-blind no-op, **host-only-axis-fires-when-container-blind**, snapshot shape, host-safe import isolation (subprocess asserts no aiohttp — F.4 minimal-venv lesson)
- `test_gateway_ram_status.py` — venv-missing, skew-guard-blocks-fallthrough, dispatch-when-present
- 70 targeted tests green; live read-path smoke against real `/proc/meminfo`; ruff + `bash -n` clean.

## Deploy note
Touches `src/genesis/guardian/` + `scripts/guardian-gateway.sh` → host-guardian deploy via `scripts/update.sh` after merge (host-deploy gate). True both-axes host E2E runs post-deploy.

### DIFF
--- a/config/guardian.yaml
+++ b/config/guardian.yaml
@@ -90,6 +90,21 @@ storage_pool:
   pool_used_crit_pct: 92
   realert_hours: 6
 
+# Guardian-side RAM tiered alert (E-rest PR-E1). The container's own RAM alert
+# (infra:container_memory_high) dies exactly when it thrashes/OOMs; the guardian
+# pages out-of-band over TWO axes, worst-of: container cgroup anon+kernel (via
+# incus-exec — best-effort) and host-VM used% (the guardian's own /proc/meminfo —
+# the reliable axis, no exec). Host tiers sit higher (the host idles busier).
+memory_tiers:
+  enabled: true
+  container_warn_pct: 75
+  container_high_pct: 85
+  container_crit_pct: 92
+  host_warn_pct: 80
+  host_high_pct: 88
+  host_crit_pct: 94
+  realert_hours: 6
+
 # Credential-file integrity backstop (G.1). The container self-heals corrupt
 # credential files first (its awareness tick); the guardian validates the same
 # files via read-only `incus exec`, WARNs on first sight, and STEPS IN (restores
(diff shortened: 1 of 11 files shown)


WingedGuardian (commented):

---

PR #1079, opened 2026-07-16
fix(infra): activate container swap live on a running container

WingedGuardian:
## Problem

Host setup enables the container's memory cgroup to spill into host swap under pressure (`limits.memory.swap=true`), so a memory spike degrades into swapping instead of thrashing the whole box into an OOM wedge. But incus only applies that setting when the container **starts**.

On an **already-running** container — one freshly created earlier in the same host-setup run, or an existing one being retrofitted — the live cgroup keeps `memory.swap.max=0` until a restart. The config is set but silently no-ops meanwhile. A retrofit run looks done while the box stays one memory spike away from the exact wedge the setting is meant to prevent. (Observed firsthand: config showed `true`, container `memory.swap.max` stayed `0`, `free` showed no swap.)

## Fix

New `scripts/lib/container_swap.sh::container_swap_activate_live`, sourced and called by `host-setup.sh` immediately after the config set. It mirrors what incus does at start by writing the live cgroup now:

- Best-effort and idempotent: no-op when the cgroup knob is absent (container stopped, cgroup v1, non-standard layout — the persisted config still applies on next start) or already non-zero.
- Never fails its caller.
- On a refused privileged write it warns loudly (restart required) rather than claiming success.

This also fixes **fresh installs**, not just retrofits: the swap config is set after the container is already up, so without this the freshly-created container also sat at `swap.max=0` until its first reboot.

## Tests

`tests/test_scripts/test_container_swap.py` drives the real lib with a stubbed `sudo` on PATH + a cgroup-base override (the `test_memory_resilience.py` pattern): activates when `0`, idempotent when already `max`, graceful no-op when the cgroup is absent, empty-name no-op, warns (and leaves the value) on write failure, plus `bash -n` parse checks for the lib and the edited `host-setup.sh`.

## Deploy

Touches `scripts/host-setup.sh` (Host-Deploy Gate path). host-setup runs on the host during provisioning; the fix lands at source and applies on the next host-setup run. No update.sh redeploy needed for this container (swap was already activated live out-of-band).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -133,6 +133,15 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Enabling container swap no longer waits for a restart to take effect.**
+  Host setup lets the container's memory cgroup spill into host swap under
+  pressure, so a memory spike degrades into swapping instead of thrashing the
+  whole box into a wedge. But that setting only took effect the next time the
+  container *started*, so retrofitting an already-running install looked done
+  while swap stayed off until a reboot — leaving the box exposed to the exact
+  OOM wedge the setting exists to prevent. Setup now activates it live on the
+  running container, so the protection is real immediately.
+
 - **Genesis's inner monologue now knows who each thought is about.** Every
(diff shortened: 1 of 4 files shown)


WingedGuardian:

---

PR #1080, opened 2026-07-16
fix(db): 256 MiB page cache + synchronous=NORMAL on the shared connection

WingedGuardian:
## What

Two SQLite connection-tuning changes in `db/connection.py`:

1. **`cache_size = -262144` (256 MiB)** on *both* connection paths (`get_db` serialized + `get_raw_db` standalone). The live DB was on SQLite's ~2 MiB default, forcing hot read paths to keep re-fetching pages instead of holding Genesis's working set resident.
2. **`synchronous = NORMAL` on `get_db`** — the main serialized connection relied on the default `FULL`, fsyncing on every commit, while `get_raw_db` already used `NORMAL`. Under WAL, `NORMAL` is the safe, standard setting (no corruption risk; at most the last txn is lost on power loss). This aligns the hot connection with the standalone one.

## Why it's safe

- `cache_size` is per-connection; the long-lived server + MCP connections sum to a few GiB against a 36 GiB memory budget.
- `NORMAL` under WAL is the documented safe mode and is *already* what `get_raw_db` uses — this removes an inconsistency, it doesn't introduce a new risk. Both pragmas also flow through the `get_db` reconnect closure (`_reconnect → _configure`), so recovered connections keep them.

## Tests

`tests/test_db/test_connection_pragmas.py` extended: asserts `cache_size` on both paths and `synchronous == NORMAL(1)` on `get_db`. All pass; ruff clean.

## Deploy

Pure `src` change — no host-deploy path. Pragmas apply on the next server start (connection open).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -133,6 +133,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Genesis's database keeps more in memory and stops over-syncing.** The main
+  shared SQLite connection held only SQLite's tiny ~2 MiB default page cache and
+  fsynced on every single commit (`synchronous=FULL`), even though the
+  standalone connection helper already used the lighter, equally safe `NORMAL`
+  mode under WAL. Both connection paths now hold a 256 MiB page cache, and the
+  main connection matches `NORMAL` — fewer disk syncs and less page re-fetching
+  under load, with no durability loss beyond what WAL already implies.
+
 - **Genesis's inner monologue now knows who each thought is about.** Every
   ambient micro-reflection used to be tagged as relevant to "both" the user
   and Genesis — the tag was computed from which sensors *ran* (all of them,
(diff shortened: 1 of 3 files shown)


WingedGuardian:
## First-pass review

### Structural

**WARNING — Genesis conventional commit prefix**
PR title uses `perf(db):` — the Genesis convention list accepts `feat:`, `fix:`, `refactor:`, `docs:`, `test:`, `chore:` and `perf:` is not among them. Either rename the branch/squash commit title to `fix(db):` (removing an inconsistency *is* a fix) or add `perf:` to the accepted set in the project docs. The current title will silently fall outside any tooling that enforces the listed prefixes.

**NOTE — CHANGELOG section mismatch**
The new entry in `CHANGELOG.md` is placed under `### Fixed`, but this is a performance improvement (larger page cache + synchronous alignment), not a bug fix. Consider moving it to `### Changed` or a `### Performance` section to keep the changelog accurate. A user reading "Fixed" would expect a broken behaviour to be corrected, not a tuning change.

---

### PII / Secrets / Wording

CLEAN — no secrets, private IPs, private hostnames, absolute user paths, or sensitive wording found in added lines, PR metadata, or CHANGELOG additions.

---

### Summary

No pattern ERRORs, no hard PII blocks. Two housekeeping items above; the code change itself is clean.

---

PR #1082, opened 2026-07-16
fix(resilience): bootstrap self-installs systemd-oomd so OOM protection deploys everywhere

WingedGuardian:
## What

Bootstrap now installs the `systemd-oomd` package before applying the memory-resilience layer, so the OOM pressure-kill protection actually deploys on every install — not only on the ones where the package happened to be present already.

## Why

`memory_resilience_apply` (in `scripts/lib/memory_resilience.sh`) **silently skips** pressure-kill setup when `systemd-oomd` is absent — it's a separate apt/dnf package on minimal installs. Nothing in setup provisioned it. The result: a box could look fully bootstrapped while sitting unprotected against the exact OOM-thrash wedge the layer exists to prevent, with no signal beyond a line in setup output nobody reads. This is the "silent-skip" class — a setup feature that no-ops on a prerequisite nothing installs.

## How

- `scripts/bootstrap.sh`: `install_pkg systemd-oomd systemd-oomd-defaults` inside the memory-resilience block, before sourcing/applying the lib (install → apply order, so the same run that used to skip now succeeds).
  - Idempotent (`apt-get install` is a no-op when present); `install_pkg` degrades to a warning on failure.
  - The lib still independently guards on kernel PSI + systemd, so this never forces oomd on a kernel that can't support it.
  - dnf arm uses `systemd-oomd-defaults` (Fedora ships the oomd daemon inside the main `systemd` package; `systemd-oomd` is not an installable name there).
- `update.sh` runs `bootstrap.sh` on every real update before its visible resilience re-run, so one insertion point covers fresh installs **and** every retrofit — no duplication needed.

## Tests / verification

- New coverage test in `tests/test_scripts/test_memory_resilience.py` asserts bootstrap provisions the package **and** does so before the apply (ordering guard). 15/15 in that file pass.
- Verified the exact install command is a clean idempotent no-op on a host where oomd is already present and active.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -131,6 +131,16 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Setup now installs the package its OOM protection depends on.** Genesis's
+  memory-pressure protection (systemd-oomd) was applied by setup but only *if*
+  the `systemd-oomd` package already happened to be installed — on a minimal
+  install where it wasn't, setup quietly skipped the whole layer and left the
+  box exposed to the exact OOM-thrash wedge the protection exists to prevent,
+  with no signal beyond a line in setup output nobody reads. Bootstrap now
+  provisions the package before applying the layer, so the protection actually
+  deploys everywhere. Idempotent (a no-op when already present) and it never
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

PR #1083, opened 2026-07-16
feat(guardian): container-swap invariant reconciler — self-healing swap knob

WingedGuardian:
## What

The guardian now keeps `limits.memory.swap` enabled on its own — re-asserting the persistent incus config knob **and** live-activating the cgroup each tick when it observes drift.

## Why

Swap is what turns a memory spike into graceful slowdown instead of a machine-wedging D-state thrash. But `limits.memory.swap=true` is only applied by `host-setup.sh` — at container creation or a manual re-run. An install that advances via bare `git pull` never re-runs host-setup, so the knob (and the live cgroup `memory.swap.max`) can sit at `0` indefinitely — one memory spike away from the wedge, silently. Observed live on a second install that ran for weeks in exactly this state.

`#1079` fixed the *live-activation* half at setup time (`scripts/lib/container_swap.sh`); this closes the gap for boxes that never re-run setup, by making the guardian — the only component with continuous host-side incus + sudo access — reconcile the invariant.

## How

- `src/genesis/guardian/swap_watch.py` (new): per-tick reconcile on **observed** state:
  1. **Persistent** — `incus config get --expanded` != `true` → `incus config set` (covers unset AND explicit false; `--expanded` so a profile-inherited true isn't misread as unset).
  2. **Live** — cgroup `memory.swap.max == "0"` → write `max` now (what incus does at container start).
- `src/genesis/guardian/cgroup_ops.py`: two new host-side primitives (`read_swap_max`, `activate_swap_max`) mirroring the existing `relieve_io_max` sudo/cgroup pattern.
- `src/genesis/guardian/check.py`: wired into `run_check` via the never-raises delegate-wrapper pattern (same shape as the git/cred/bundle watches).
- `config.py`: `swap_reconcile_enabled` kill switch (an explicitly-false knob is otherwise reconciled back to true — swap-on is the install invariant).

**Alerting:** the healthy path is two reads, no writes, no alerts. A heal pages one **INFO** note (guardian self-actions stay visible); a heal that can't complete pages a **WARNING** throttled to once/24h (the `memory_watch` state-file idiom).

Also hardens `memory_watch`'s identical throttle-parse idiom to catch corrupted-state exceptions (review parity).

## Tests / verification

- `tests/test_guardian/test_swap_watch.py` (new, 16 cases): unset/false/error config paths, live-zero heal, throttled-failure re-page, kill switch, no-signal degradation, dispatch-failure containment, cgroup primitives, run_check wiring. Full guardian suite green (91 passing across swap/check/config/memory).
- **E2E on real infrastructure:** healthy-path no-op verified against live incus + sudo; heal-path drilled by injecting the incident state (`swap.max=0`) and confirming the real module live-heals it to `max` with exactly one correct INFO alert.

## Deploy note

Touches host-deployed guardian paths (`src/genesis/guardian/`), so it is **not live at merge** — it reaches the host only via `scripts/update.sh`. Will deploy + verify on both machines post-merge per the host-deploy gate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,17 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **The guardian now keeps container swap enabled on its own.** Swap is what
+  turns a memory spike into graceful slowdown instead of a machine-wedging
+  thrash, but the setting only got applied when host setup ran — an install
+  that just pulls code updates could sit unprotected indefinitely (observed
+  live: a second install ran for weeks one memory spike away from the wedge).
+  The guardian now re-checks the setting every tick and repairs both halves
+  when drifted: the persistent config (so future restarts have it) and the
+  live cgroup (so protection is immediate, no restart needed). Repairs page an
+  INFO note; a repair that can't complete pages a warning at most daily. Hosts
+  where swap-off is deliberate can opt out (`swap_reconcile_enabled: false`).
+
 - **Genesis now remembers how its background jobs actually ran, not just a
   running tally.** Until now each scheduled job kept only a single cumulative
   row, so a job that failed for a week and then recovered looked identical to
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -374,6 +374,12 @@ verified: 4e483381 2026-07-15
   `/proc/meminfo` (the reliable axis). Both use the shared `_tier_for`/
   `decide_alert` hysteresis. Read-only `disk-status`/`ram-status` verbs expose
   the same measurement to the container.
+- **Container-swap invariant reconciler** (`swap_watch.py`) runs every tick:
+  re-asserts `limits.memory.swap=true` (incus config) and live-activates the
+  cgroup `memory.swap.max` (via `cgroup_ops`) when observed at `0` — the
+  self-heal for installs that advance via bare `git pull` and never re-run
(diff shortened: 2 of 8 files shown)


WingedGuardian:
## First-pass review — PR #1083

### Structural: 1 WARNING

**[WARNING — Import Hygiene]** `src/genesis/guardian/cgroup_ops.py`, `activate_swap_max` function

`import shlex` is placed inside the function body (inside a `try` block) rather than at module level. `shlex` is stdlib and has no import cost — this is a convention violation, not a correctness issue, but it should move to the top of the file with the other stdlib imports.

```python
# current (inside activate_swap_max):
try:
    import shlex
    rc, _stdout, stderr = await _run_subprocess(
        "sudo", "sh", "-c", f"echo max > {shlex.quote(str(swap_path))}",

# should be at module top, alongside json/logging/datetime
import shlex
```

---

### Structural: CLEAN otherwise

- No `nohup`, no direct `task_states` inserts, no `/tmp/` usage, no bare `except:`, no mutable defaults, no raw SQL against genesis.db.
- New `swap_watch.py` has a corresponding `tests/test_guardian/test_swap_watch.py` (16 cases). ✓
- All public functions (`check_container_swap_and_alert`, `read_swap_max`, `activate_swap_max`) have docstrings. ✓
- Commit prefix `feat(guardian):` follows convention. ✓

No secrets, private IPs, SSH keys, or database credentials in added lines. PR title, body, and CHANGELOG additions contain no language about leaks, security incidents, or private infrastructure. The container name `"genesis"` used in tests is generic.

---

PR #1090, opened 2026-07-16
fix(outreach): persist standalone-reply engagement — write-back was quote-reply-only

WingedGuardian:
## Problem

When Genesis sends a send-and-wait outreach and the user answers with a **standalone message** (no quote-reply), the Telegram handler resolves the in-memory `ReplyWaiter` and returns — **without ever writing back to `outreach_history`**. Only the quote-reply path calls `record_reply`. The row later degrades to `ignored` (24h timeout) or `ambivalent` (implicit activity), as if the user said nothing.

Live impact: only **3 of 1,021** outreach rows carry `engagement_signal='user_reply'`. This is the affirmative-reply lane the WS-2 cognitive-ledger grader will consume — found by the `reply_received` resolver spike.

## Fix

- **`resolve_scoped_pending` returns the matched waiter's registry keys** (`list[str]`, empty = unresolved) instead of `bool`. All keys matter: the alias mechanism maps both a pre-generated UUID and the Telegram message_id to the same future, and whichever iterates first may be the UUID — which `find_by_delivery_id` (keyed on message_id) would never match. `_pop_future` now returns the keys it drops.
- **Handler mirrors the quote-reply path** on scoped resolution: tries each key against `outreach_history` (first hit wins), calls `record_reply` (`'useful'`/`'user_reply'`, correctly upgrading a prior `ambivalent`), best-effort — a write-back failure never blocks the already-resolved waiter and the message stays consumed.
- Caller set enumerated (serena + literal grep): exactly one production caller. 5 waiter tests + 1 handler mock updated to the list contract; 2 new handler tests (write-back with alias fall-through; failure isolation).

## Known ceiling

Fire-and-forget outreach (no waiter registered) still can't correlate standalone replies — those keep landing as `implicit_activity`. Documented in the WS-2 design as the affirmative-lane ceiling; the ledger grades them as no-reply, which is mechanically honest.

## Verification

- 71 targeted tests green (`test_outreach/test_reply_waiter.py`, `test_channels/test_telegram_handlers_v2.py`); ruff clean.
- Live E2E planned post-deploy: send-and-wait outreach → standalone reply → verify the row gains `engagement_signal='user_reply'`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,18 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ## [Unreleased]
 
+### Fixed
+
+- **Replying to Genesis without quote-replying now counts.** When Genesis
+  asked you something on Telegram and you answered with a plain message
+  (no quote-reply), your answer reached the waiting conversation but the
+  outreach record never learned it was answered — it would later be marked
+  "ignored" or "ambivalent" as if you'd said nothing. Only 3 of 1,021
+  outreach records ever carried a real reply signal because of this. A
+  standalone reply that resolves a pending question is now recorded on the
+  outreach record exactly like a quote-reply, so Genesis's picture of what
+  you actually respond to stops being systematically wrong.
+
 ### Added
 
 - **The guardian now keeps container swap enabled on its own.** Swap is what
--- a/src/genesis/channels/telegram/_handler_messages.py
+++ b/src/genesis/channels/telegram/_handler_messages.py
@@ -1298,15 +1298,45 @@ async def handle_text(ctx: HandlerContext, update: Update, context: ContextTypes
     if ctx.reply_waiter and msg.text:
         thread_key = f"{msg.chat.id}:{getattr(msg, 'message_thread_id', None) or 'dm'}"
         try:
-            if ctx.reply_waiter.resolve_scoped_pending(
+            resolved_keys = ctx.reply_waiter.resolve_scoped_pending(
                 msg.text, thread_key=thread_key,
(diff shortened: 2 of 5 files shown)


WingedGuardian (commented):
## First-pass review — PR #1090

---

### WARNING — Import hygiene

**`src/genesis/channels/telegram/_handler_messages.py` (~line 1316)**

```python
from genesis.db.crud.outreach import find_by_delivery_id
```

This import lives inside the body of `handle_text()`, nested in a try block. Function-body imports are the standard Python pattern for breaking a circular dependency — meaning `channels.telegram._handler_messages` → `db.crud.outreach` presumably can't be placed at module level without creating a cycle. If that's the case, a brief comment explaining why the import is deferred (e.g., `# deferred import: avoids circular dependency with db.crud`) would make the intent explicit for future readers. If there's no actual cycle, moving it to the module-level local-import group would be cleaner.

Not a blocker — best-effort annotation on the flag.

---

### All other checks: CLEAN

- No `nohup`, no direct `task_states` inserts, no `/tmp/` usage, no bare `except:`, no mutable default arguments, no `os.killpg`, no raw SQL, no star imports.
- Return-type contract change (`bool` → `list[str]`) correctly propagated through all call sites and all mock assertions (both existing and new tests).
- Both changed production files have their corresponding test files updated; new tests cover the alias fall-through and the write-back failure isolation paths.
- Docstrings on `resolve_scoped_pending` and `_pop_future` updated to match the new return contract.
- Conventional commit prefix `fix(outreach):` present.
- CHANGELOG and PR body wording: clean — no language referencing credential leaks, PII cleanup, or private infrastructure.

---

PR #1091, opened 2026-07-16
feat(guardian): Proxmox vzdump backups — arm the grow safety gate (WS1)

WingedGuardian:
## What & why

The Guardian's provisioning gate has always wanted a recent hypervisor backup
before an irreversible VM grow — but nothing could take one, so
`newest_backup_age_days()` returned `None` and the check had to be left
disabled (an unknown backup age is treated as a refusal). This adds the
missing capability: Genesis can now take a `vzdump` backup of its own host VM,
which both creates a real VM-level restore point and lets the grow gate be
armed.

This is the **vzdump-take slice** only. VM lifecycle verbs (start/stop/reboot)
and **restore** are deliberately out of scope — restore is a destructive verb
that gets its own security review.

## Design

- **Third, privilege-separated API token.** A `backup` token
  (`VM.Backup` + `Datastore.AllocateSpace` on the backup storage) keeps the
  existing `provision` token grow-only — it cannot back up, and the backup
  token cannot resize. Absent token degrades safely (backup verbs refuse
  pre-flight; grows unaffected).
- **Two-phase, because a full-VM dump runs for tens of minutes.** The start
  verb only launches the task and returns the PVE UPID; a separate status
  verb is one verify probe. The start is ledgered immediately — that row is
  the rate-cap entry, the in-flight latch, and the restart-resume handle, so
  an interrupted verification resumes after a process restart with no state
  lost. Verification anchors on the UPID's own start time, so an old backup
  can never verify a new task.
- **Just-in-time + rotation, not scheduled.** A backup is proposed as the
  precondition of a grow (a stale-backup grow proposal becomes a
  backup→verify→grow chain under **one** approval, with the time gap and
  automatic execution disclosed in the approval text) or on explicit request
  (`provision_vzdump`). Old backups rotate to a configured keep-last; there is
  no delete verb. Backups have their own weekly budget separate from grows
  (the existing `max_actions_per_week` now counts grows only).
- **Poll-outcome discipline.** Only the task's own terminal exit status is a
  failure; transport errors / still-running are transient and retried to a
  configurable wall bound, which ends as UNVERIFIED (never "failed" — the
  backup may still be landing).

New surface mirrors the existing grow path exactly: gateway verbs (execute-only,
approval is the caller's job), `GuardianRemote` client, container coordinator
(owner APPROVE via the pipeline), a sibling MCP tool, and the RPC bridge for
the standalone-MCP path.

## Safety / review

- Pre-build architecture review (state-machine durability, backup-storage
  headroom source, rate-cap class separation, chained-approval disclosure) —
  findings addressed.
- Security review of the new trust surface (gateway argument injection, token
  privilege separation, approval-gate bypass paths, credential handling) —
  no findings.
- The UPID argument crosses three hops (client regex ↔ gateway bash regex ↔
  URL-encoding) with a strict shared charset that admits real UPID bytes but
  nothing shell-active; a parity test pins the validators together.
- A coverage-guardrail test sweeps the tree so the backup mutation verbs can
  never gain a call site outside the audited approval chain.

## Testing

- 300+ targeted tests across the provisioning adapter, gate, two-phase flow,
  ledger, gateway verbs, credential bridge, container coordinator, MCP tool,
  and dashboard route.
- Read-path verified end-to-end against a live hypervisor: `newest_backup_age_days`,
  the backup-storage headroom extraction, guest-agent detection, and the
  backup gate all produce correct results on real API response shapes.
- The write path (start→verify→prune) is verified in a supervised first run
  after merge, once the backup token is provisioned.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,22 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **Genesis can now take a hypervisor backup of its own host VM — and grows
+  finally get their safety net.** The provisioning gate has always wanted a
+  recent backup before an irreversible grow, but nothing could take one, so
+  the check sat disabled. Now a stale backup turns a grow proposal into a
+  backup→verify→grow chain under one clearly-worded approval: the backup
+  starts immediately (vzdump, on a third API token that can back up but never
+  resize), verification runs in the background for however long the dump
+  takes, old backups rotate away automatically (keep-last, configurable), and
+  the grow executes only after the backup verifies and a fresh safety
+  re-check passes. Backups can also be taken on demand (`provision_vzdump`),
+  have their own weekly budget separate from grows (note: the existing
+  `max_actions_per_week` cap now counts grows only), and an interrupted
+  verification resumes after a restart with no state lost. Restore is
+  deliberately not included — that is a destructive operation with its own
+  upcoming review.
+
 - **The guardian now keeps container swap enabled on its own.** Swap is what
   turns a memory spike into graceful slowdown instead of a machine-wedging
   thrash, but the setting only got applied when host setup ran — an install
--- a/docs/reference/proxmox-provisioning.md
+++ b/docs/reference/proxmox-provisioning.md
@@ -100,9 +100,11 @@ pveum acl modify /vms/<VMID>       -user genesis@pve -role GenesisProvision
 # 4. Provision role ALSO on the storage (disk grow allocates space there)
 pveum acl modify /storage/<STORAGE> -user genesis@pve -role GenesisProvision
 
-# 5. Two privilege-separated tokens (privsep=1). Save the printed secrets.
+# 5. Privilege-separated tokens (privsep=1). Save the printed secrets.
 pveum user token add genesis@pve ro       --privsep 1
 pveum user token add genesis@pve provision --privsep 1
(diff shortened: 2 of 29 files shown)


WingedGuardian:

---

**Structural pass** (28 files, 1,272 added lines reviewed)

- No `nohup`, no direct `task_states` inserts, no `/tmp/` usage, no `git add .` / force-push, no bare `except:`, no mutable default arguments, no `os.killpg`, no raw SQL against `genesis.db`.
- Imports are correctly grouped (stdlib → third-party → local) in all modified files.
- No new `src/` file without a corresponding test; the only new file is `tests/test_guardian/test_vzdump_call_site_guardrail.py`.
- All new public functions and classes carry docstrings.
- No `IntervalTrigger`, no `$CLAUDE_PROJECT_DIR` as Python env var, no hook-script heavy imports.
- Commit prefix `feat(guardian):` satisfies the conventional-commit convention.

**PII/Secrets pass** (added lines only)

- Token-like strings in tests (`"AUDIT"`, `"PROVISION"`, `"BACKUP"`, `"aaa"`/`"bbb"`/`"ccc"`) are plainly synthetic fixtures, not real credentials.
- `10.0.0.9` in test fixtures is a known false positive per review rules.
- `genesis@pve!backup` is Proxmox realm/token notation, not a personal email.
- No private IPs in production code, no SSH private keys, no hardcoded DB connection strings with credentials.
- PR title, description, and CHANGELOG addition contain no PII-disclosure language, no security-incident references, and no detail that would be exploitable.

---

PR #1095, opened 2026-07-16
feat(resilience): host-side zram swap — compressed-RAM-first tier (E-rest E3)

WingedGuardian:
## Summary

Completes the swap story `memory_resilience.sh` (#1059) deliberately leaves as a warning: that layer VERIFIES swap and warns when a bare/VM host has none; this layer CREATES a modest zram device so the warning has a mechanical answer. On hosts that already have disk swap, zram layers above it (swap priority 100) so memory pressure compresses into fast RAM before touching disk.

## What's in it

- **NEW `scripts/lib/host_swap.sh`** — `host_swap_apply` (always-rc-0 entrypoint with a degrade-to-skip ladder: not-systemd, container vantage via `systemd-detect-virt`, no `zramctl`/`zram.ko`, external zram already active — never shadows zram-generator or an operator's own setup, no non-interactive sudo, `HOSTSWAP_DISABLE=1`, masked unit = the durable operator opt-out) + `host_swap_remove` (clean rollback). Size = `min(MemTotal/2, 4 GiB)` — zram-generator's own upstream default formula, computed per machine at apply time so a 2 GiB VPS gets 1 GiB and big hosts cap at 4 GiB (`HOSTSWAP_CAP_GIB` override). Write-if-different unit install with the changed-flag-as-a-variable defense (mirroring `memory_resilience.sh`) → zero systemd churn on idempotent re-runs.
- **`zram-swap.service`** — self-contained root-level oneshot on a FIXED `/dev/zram0` (no command substitution → no systemd `$`-escaping hazard), zstd compression with automatic fallback to the kernel default, idempotent across restarts, `ExecStop` tears down cleanly. No genesis-path dependencies.
- **`install_guardian.sh` Step 9c** — fresh installs apply on install.
- **`guardian-gateway.sh` `redeploy` verb** — existing installs only ever receive code via this verb (installer steps never re-run), so it retrofits here: runs AFTER the success JSON with all output to stderr (the verb's stdout is a parsed contract), `[ -f ]`-guarded for transition-deploys from pre-E3 archives, timeout-bounded, whole-block best-effort — can never fail a redeploy.
- **GUARDIAN_PATHS lockstep ×3** (`update.sh`, `watchdog.py`, `deploy_health.py`) — a `host_swap.sh`-only change now triggers a redeploy, load-bearing since the gateway executes the lib.
- **Tests** — 25 sandboxed behavioral tests (stubbed `sudo`/`systemctl`/`systemd-detect-virt`/`modinfo`/`zramctl` on PATH, `SYSTEMCTL_LOG` churn assertions, `__DONE__` errexit sentinel proving no degrade path can abort an install or redeploy) + wiring tests pinning the installer hook, the gateway hook, and the three-path lockstep.
- **Docs** — `docs/reference/memory-resilience.md` zram-layer section; `docs/architecture/CURRENT.md` guardian entry + stamp.

## Design notes

- Every guard prints a one-line skip and returns 0 — portability-first for the public repo (hosts without zram.ko, without sudo, or that are themselves containers all degrade gracefully).
- The unit content is a fixed heredoc; the only parameter is a size integer computed host-side from `/proc/meminfo` — no new trust boundary through the redeploy path.

103 tests green across the blast radius (incl. the redeploy-verb sandbox suite, memory_resilience, container_swap, deploy_health, update-host-sync). `bash -n` clean on all four touched shell files.

### DIFF
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -371,7 +371,7 @@ radius) and the container-side Sentinel (CC-driven diagnosis/repair).
 ```yaml subsystem-map
 entry: guardian-sentinel
 modules: [guardian, sentinel]
-verified: 8bd0a52b 2026-07-15
+verified: 159698d4 2026-07-16
 ```
 
 - **guardian/** is bidirectional: host side (`python -m genesis.guardian`,
@@ -409,6 +409,15 @@ verified: 8bd0a52b 2026-07-15
   self-heal for installs that advance via bare `git pull` and never re-run
   host-setup. Heals page INFO; failures page WARNING (24h throttle); kill
   switch `swap_reconcile_enabled: false`.
+- **Host zram swap** (`scripts/lib/host_swap.sh`, E-rest E3): a
+  compressed-RAM-first swap tier on the host VM — `zram-swap.service` at swap
+  priority 100, sized `min(MemTotal/2, 4GiB)` (`HOSTSWAP_CAP_GIB` override).
+  Applied by `install_guardian.sh` Step 9c (fresh) and the gateway `redeploy`
+  verb (existing installs retrofit on next update; output to stderr, never
+  fails a redeploy). Degrades to one-line skips (container vantage, no
+  zram.ko/zramctl, external zram, no sudo); durable opt-out = `sudo systemctl
+  mask zram-swap.service`. Completes the swap story `memory_resilience.sh`
+  leaves as a warning — see `docs/reference/memory-resilience.md`.
 - **sentinel/** is LIVE-wired but **shadow-only autonomy**: config mode
   `"live"` is NOT implemented (dispatcher warns + downgrades); every proposed
   action requires human approval. `InfrastructureMonitor` (call site 37, free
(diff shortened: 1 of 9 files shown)


WingedGuardian (commented):

---

WingedGuardian:
Both Codex findings addressed in `dd736988`:

- **P1 (non-swap zram consumer destroyed)** — two layers: the apply-time foreign guard now also treats ANY `zramctl`-listed device (while our unit was never installed) as external and skips; and the unit itself refuses (`exit 1`) to touch a `/dev/zram0` that carries a mounted filesystem, so even a post-install external consumer is never reset. `test_skip_foreign_nonswap_zram_not_destroyed` + a unit-content assertion pin both.
- **P2 (enable never retried)** — an unchanged unit with no active zram now retries `systemctl enable --now` on the next apply (covers a transient enable failure AND an operator `disable` without mask — mask is the documented durable opt-out). No churn on healthy hosts: an active device skips the branch. `test_unchanged_unit_inactive_retries_enable` pins it.

40 tests green, bash -n clean.

PR #1096, opened 2026-07-16
feat(awareness): infra protection posture check — an unprotected box now alerts

WingedGuardian:
## What

Phase 3 of the silent-skip closure (provision #1082 → reconcile #1083 → **alert**, this PR): a new hourly awareness check, `_check_infra_protection_posture`, reads the infrastructure profile's effective facts and raises one non-paging `high` `infrastructure_alert` (dashboard + morning report; the WS-2 M10 reconciler absorbs it into the durable open-set) when a memory-plane protection is missing — and auto-resolves it when the protection is restored.

## Why

`infra_profile` records the facts but only emits `infrastructure_drift` on a fact-hash **change** — a box that was *always* unprotected produced no signal anywhere. Observed live: an install ran for weeks with container swap disabled and no systemd-oomd, silently, until a memory spike wedged it.

## Rules (memory plane v1; explicit defect values only — absent/None facts stay silent)

| Fact | Defect | Meaning |
|---|---|---|
| `memory.facts.cgroup_memory_swap_max` | `0` (int) | container swap disabled — the wedge state |
| `memory.facts.oomd_user_slice_kill` | `False` (gated on a readable swap knob = cgroup-v2 proxy) | oomd pressure-kill not enforced |
| `host_system.facts.swap_total_kb` | `0` | host has no swap to spill into |
| `host_virt.facts.container_limits["limits.memory.swap"]` | `"false"` (explicit only) | persistent knob off — covers guardian-less installs |

A profile older than 3 days (refresh is boot + daily) raises a distinct "posture UNKNOWN — refresh broken" alert instead of asserting posture from dead facts.

## Design

- Mirrors `_check_user_model_staleness` (emit idiom) + the deploy-staleness "exactly one active alert = the current state" pattern: `supersede_except_hash(keep=current)` before `create(skip_if_duplicate=True)` — a posture change retires the old row, the atomic dedup keeps a stable state at one row across concurrent loops.
- Never raises into the tick; keyed 24h cooldown only saves no-op writes (a CHANGED posture bypasses it).
- Network-plane posture deferred: `networkd_keep_configuration=False` is ambiguous on NetworkManager hosts — needs a `networkd_active` disambiguation fact first (follow-up filed).

## Tests

21 new tests: tri-state rule semantics (bool-footgun guard, cgroup-v1 oomd gate, absent-fact silence), full alert lifecycle (dedup, supersede-on-change, auto-resolve + cooldown clear, stale-UNKNOWN path, load-failure safety), and provision-or-surface guardrails (every rule slug has remediation text, every memory-resilience fact is read by the rules, the check is wired into the tick). Verified against the real on-disk profile shape: healthy profile → silent; fabricated defects → one high alert.

Ledger: 5fcef45c093342d08628d2415f2421a5

### DIFF
--- a/.claude/skills/genesis-development/SKILL.md
+++ b/.claude/skills/genesis-development/SKILL.md
@@ -132,6 +132,17 @@ tool-selection decision matrix: `.claude/docs/code-intelligence.md`
   server restarts more frequently than the interval, the job never
   fires. Use `CronTrigger` for anything longer than a few hours.
   Bit us with `user_model_evolution` (48h interval, daily restarts).
+- **Silent skips are banned (provision-or-surface).** A setup/resilience
+  feature that gracefully skips on a missing prerequisite (a package, a
+  host knob) must either PROVISION the prerequisite (bootstrap.sh /
+  host-setup.sh / a guardian reconciler) or register an effective-fact in
+  `infra_profile` that the awareness posture check
+  (`awareness/loop.py::_check_infra_protection_posture`) reads — so an
+  unprotected box raises a standing alert instead of staying silent. A
+  graceful skip with neither = a box that runs unprotected with zero
+  signal (a sibling install ran weeks without swap/systemd-oomd until a
+  memory spike wedged it, 2026-07). Guardrail:
+  `tests/test_awareness/test_infra_protection_posture.py`.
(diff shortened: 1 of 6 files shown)


WingedGuardian:

---

PR #1097, opened 2026-07-16
fix(outreach): register send-and-wait waiters before attaching delivery context

WingedGuardian:
## Problem

`submit_and_wait` attached the delivered chat+topic context **before any waiter existed** — registration happened later, inside `wait_for_reply` — and `set_context` silently drops unregistered keys. Every waiter on this path was born contextless, making standalone (non-quote) replies structurally unresolvable. Same defect in `submit_raw_and_wait`'s no-`waiter_key` branch, which carries provision approvals (`rpc.py`) and the sentinel's text-only fallback.

Observed live during #1090's E2E: four standalone replies, each seconds after delivery, all degraded to `implicit_activity` — and each unresolved reply fell through to the conversation handler, spawning unrelated CC turns in the alerts topic. The #1090 write-back is wired correctly but was starved by this upstream ordering bug.

## Fix

- `submit_and_wait` and `submit_raw_and_wait` (no-key branch): **register the delivery_id waiter immediately after delivery, before context attach**. The pre-registered button-key branch was already correct and is unchanged.
- `set_context` now returns whether the context was recorded; `_attach_waiter_context` logs a WARNING tripwire on any dropped context — this class can never fail silently again.
- Regression tests drive the **real** `ReplyWaiter` through both paths: a standalone reply arriving mid-wait must resolve the waiter (plus button-key path via both keys).

## Verification

- 110 targeted tests green (`test_pipeline.py`, `test_reply_waiter.py`, `test_reply_waiter_alias.py`, `test_rpc.py`, `test_telegram_handlers_v2.py`); ruff clean.
- Post-deploy live E2E (combined with #1090): send-and-wait outreach → single standalone reply → `outreach_history` row gains `engagement_signal='user_reply'`.

Diagnosis credit: a concurrent session surfaced the ordering suspicion in the alerts channel; verified firsthand and fixed here per user direction (single-owner).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,16 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Answering Genesis's questions with a plain message now actually works.**
+  When Genesis asked something and waited for your answer (approvals,
+  provision prompts, send-and-wait questions), an internal ordering bug left
+  the waiting mechanism blind to where the question had been delivered — so
+  a plain (non-quote) reply never matched it. Your answer instead spawned an
+  unrelated conversation turn, and the question sat unanswered until it
+  timed out. The delivery context is now attached in the right order, plain
+  replies match the question they answer, and a tripwire warning fires if
+  this ordering ever regresses.
+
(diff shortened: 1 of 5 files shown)


WingedGuardian (commented):

---

PR #1099, opened 2026-07-16
fix(entity): stop the entity_adjudication orphan bleed — gate enqueue off

WingedGuardian:
## What

`enqueue_adjudication` (`db/crud/entities.py`) writes `entity_adjudication` rows to `deferred_work_queue` for an "entity_maintenance job" that **was never built**. Nothing anywhere claims `work_type='entity_adjudication'` — pure producer-with-no-consumer. Live count: **263 pending, 0 ever drained, oldest 6d, growing ~44/day**, and it now surfaces as dashboard clutter.

## Fix (stop the bleed)

Gate the enqueue behind `_ADJUDICATION_ENQUEUE_ENABLED` (default `False`). A fuzzy `norm_name` match still creates the entity and returns `AMBIGUOUS` — only the orphan queue row is skipped. The enqueue machinery is preserved as **GROUNDWORK**: the adjudication drainer (follow-up `9127e8ed`) flips the flag `True` when it lands. This is the root-cause fix for the *growth* — clearing alone would silently refill.

The 263 existing orphans are cleared one-time out-of-band post-deploy (a data-migration would wrongly re-delete pending rows once the drainer re-enables enqueue).

## Testing

`test_fuzzy_ambiguous_but_no_enqueue_while_gated` (AMBIGUOUS + entity created + no orphan row) and `test_fuzzy_enqueues_when_gate_enabled` (flag-on restores the row — locks the seam A will flip). Full `test_entity_layer.py` green (17).

Follow-up: `9127e8ed` (the drainer = the "A" that turns this back on).

### DIFF
--- a/src/genesis/db/crud/entities.py
+++ b/src/genesis/db/crud/entities.py
@@ -425,6 +425,15 @@ async def delete_entities_cascade(
     return {"entities": deleted, "mentions": mentions, "links": links}
 
 
+# Dormant until a consumer exists. `enqueue_adjudication` writes
+# `entity_adjudication` rows to `deferred_work_queue` for an "entity_maintenance
+# job" that was never built — so the rows were a pure producer-with-no-consumer
+# leak (263 orphans in ~6 days, 0 ever drained). Gated OFF to stop the bleed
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

PR #1098, opened 2026-07-16
fix(resilience): zram unit — create the device via hot_add (Ubuntu 6.8 ships zero static devices)

WingedGuardian:
## Summary

Live E2E of #1095 on a real host caught what no sandbox could: `modprobe zram` on Ubuntu's 6.8 kernel loads the module with **zero static devices** — `/dev/zram0` never appears and `zram-swap.service` failed with `zramctl: /dev/zram0: No such device`.

The universal API is `/sys/class/zram-control/hot_add`: **reading** it allocates the lowest free device. The unit's ExecStart now runs `test -b /dev/zram0 || cat /sys/class/zram-control/hot_add >/dev/null` after `modprobe` — preserving the fixed-device, no-command-substitution design (still no `$` anywhere in the unit).

Spike-proven on the live host: hot_add → zram0 created → **zstd accepted natively** → mkswap → swapon prio 100 → device active and swapping.

One-line unit-template change + content assertion; 27 tests green, `bash -n` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1101, opened 2026-07-17
fix(resilience): zram unit — udevadm settle after hot_add (async node-creation race)

WingedGuardian:
## Summary

Second live-E2E finding on the real host, after the `hot_add` fix (#1098): `hot_add` allocates the zram device but udev creates the `/dev/zram0` **node** asynchronously. The unit's back-to-back `zramctl` raced ahead of node creation and failed with `zramctl: /dev/zram0: No such device` — the manual spike only passed because shell/network latency between commands masked the race.

`udevadm settle --timeout=10` between `hot_add` and `zramctl` makes node creation deterministic. **Proven live** on the host: settle → `zramctl … zstd` → `mkswap` → `swapon -p 100`, device active and swapping.

`udevadm` ships with systemd (no new dependency); the `test -b /dev/zram0` guard keeps `settle` a fast no-op on kernels that already have static devices.

One-line unit-template change + content assertion; 27 tests green, `bash -n` clean.

## Why two live hotfixes

This is exactly what live E2E is for: `modprobe`-device-count (#1098) and udev-node timing are host-kernel behaviors no sandbox stub can reproduce. Both surfaced only against the real 6.8 kernel.

### DIFF
--- a/scripts/lib/host_swap.sh
+++ b/scripts/lib/host_swap.sh
@@ -60,7 +60,10 @@ _hostswap_size_mib() {
 # swaps early-exit keeps restarts idempotent (a second swapon would fail);
 # the hot_add read creates the device when the module ships ZERO static
 # devices (Ubuntu 6.8 does — live-E2E finding 2026-07-16; reading
-# /sys/class/zram-control/hot_add allocates the lowest free device);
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

PR #1103, opened 2026-07-17
fix(resilience): zram unit — always-fresh device + restart-to-apply (3 live-E2E findings)

WingedGuardian:
## Summary

Full-chain live E2E of E3 on a real Ubuntu 6.8 host surfaced three kernel/systemd behaviors no sandbox stub can reproduce. This PR is the robust fix, validated end-to-end through the real `host_swap_apply` entrypoint across cold / idempotent / recovery paths.

**Root cause** of the persistent `zramctl: /dev/zram0: No such device` failures: `zramctl --reset` on a device makes udev **delete** the `/dev` node (not merely deconfigure it), so the immediately-following configure fails. The unit was resetting the very device it had just created.

## The fix — never reset; always start fresh

- Ubuntu 6.8 ships the module with **zero static devices** → read `/sys/class/zram-control/hot_add` to create `/dev/zram0`.
- `hot_add`/`hot_remove` are **async** → `udevadm settle` after each, before any userspace tool touches the node.
- A **stale pre-existing** device is `hot_remove`d (the only safe way to resize/reconfigure), then re-created — normalizing cold, stale-configured, and half-configured states to the single proven path.

## And: `restart`, not `enable --now`

The unit is a `RemainAfterExit` oneshot, so once active, `enable --now`/`start` are no-ops — they would neither apply changed unit content nor recover a device swapped-off from under an "active" unit. The lib now `restart`s (re-runs `ExecStop`(free) + `ExecStart`(fresh)); `enable` (without `--now`) just sets boot persistence.

## Verified live (real host, real `host_swap_apply`)

- **Cold** → `zram0` active at prio 100, zstd, 4 GiB
- **Idempotent** → "already in place", exactly one device, no systemd churn
- **Recovery** (external `swapoff`) → `restart` heals → RECOVERED

`ExecStop` + `host_swap_remove` switched to `hot_remove` for symmetric teardown. 27 tests green, `bash -n` + ruff clean.

## Why four live hotfixes total

This is exactly what the standing "verify E2E after review" directive is for: device-count, udev node-timing, reset-node-deletion, and RemainAfterExit-restart semantics are all live host-kernel behaviors invisible to unit-test stubs. Each surfaced only against the real 6.8 kernel, and each is now proven fixed on it.

### DIFF
--- a/scripts/lib/host_swap.sh
+++ b/scripts/lib/host_swap.sh
@@ -55,20 +55,23 @@ _hostswap_size_mib() {
 }
 
 # _hostswap_unit_content <size_mib> — render the self-contained unit.
-# Fixed /dev/zram0 (modprobe creates it by default) means NO command
-# substitution inside ExecStart — no systemd `$`-escaping hazard. The
-# swaps early-exit keeps restarts idempotent (a second swapon would fail);
-# the hot_add read creates the device when the module ships ZERO static
-# devices (Ubuntu 6.8 does — live-E2E finding 2026-07-16; reading
-# /sys/class/zram-control/hot_add allocates the lowest free device, then
-# `udevadm settle` waits for udev to CREATE the /dev/zram0 node — hot_add is
-# async and back-to-back zramctl otherwise races the node into existence,
-# failing with "No such device" (live-E2E finding 2026-07-16));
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

PR #1105, opened 2026-07-17
fix(resilience): zram unit → /usr/local/lib/systemd/system so mask opt-out works

WingedGuardian:
## Summary

The live opt-out drill caught the last non-functional advertised behavior in the E3 chain: `systemctl mask` works by symlinking `/etc/systemd/system/<unit>` → `/dev/null`, so a unit file **occupying that exact path** makes mask refuse (`Failed to mask unit: File ... already exists`) — the documented durable opt-out could never engage.

## Fix

- Unit installs to **`/usr/local/lib/systemd/system/`** (the locally-installed-unit layer), leaving `/etc` free for the operator's mask symlink — mask now works exactly as documented.
- **Legacy migration**: pre-relocation installs have the unit in `/etc`, which would shadow the new location by systemd precedence. Apply removes the legacy copy (counted as a change → daemon-reload + restart relocate atomically), and the foreign-zram guard counts a legacy-path unit as ours mid-migration.
- `host_swap_remove` cleans both paths.

3 new tests (location, migration, migration-vs-foreign-guard); 30 green, `bash -n` + ruff clean. Doc updated.

## Now answering "how confident are you"

With this merged, every advertised behavior of the E3 chain has been exercised live on a real host: cold activation, idempotent re-run, external-swapoff recovery, and (post-merge E2E) the mask/unmask opt-out. The zram device on the reference host is actively compressing (292 MB → 67 MB, 4.3:1) as of this PR.

### DIFF
--- a/docs/reference/memory-resilience.md
+++ b/docs/reference/memory-resilience.md
@@ -76,7 +76,9 @@ mechanically instead of leaving the warning above as the only output.
   formula, computed per machine at apply time (a 2 GiB VPS gets 1 GiB, big
   hosts cap at 4 GiB). Override the cap with `HOSTSWAP_CAP_GIB`.
 - **Unit**: a self-contained root-level `zram-swap.service` (oneshot,
-  `RemainAfterExit`) written to `/etc/systemd/system/` — fixed `/dev/zram0`,
+  `RemainAfterExit`) written to `/usr/local/lib/systemd/system/` (NOT /etc —
+  `systemctl mask` needs the /etc path free to plant its /dev/null symlink,
+  so the documented opt-out actually works) — fixed `/dev/zram0`,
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

WingedGuardian:
Both P2s addressed in `latest push`:
- **write-before-migrate** — the legacy /etc unit is now removed only after `_hostswap_install_unit` succeeded; a failed write keeps the legacy copy ("keeping legacy … still protected"). `test_failed_write_keeps_legacy_unit` pins it.
- **mask-preserving remove** — `host_swap_remove` deletes the /etc path only when it is a regular file, never a symlink, so an operator mask survives cleanup. `test_remove_preserves_mask_symlink` pins it.

32 tests green.

PR #1108, opened 2026-07-17
fix(resilience): zram apply — restore boot persistence lost to unmask sweep

WingedGuardian:
## Summary

The live mask/unmask opt-out drill (post-#1105) surfaced one final systemd quirk: **`systemctl unmask` removes the `multi-user.target.wants` enablement symlink along with the mask symlink**, leaving the unit active-but-disabled — zram would silently not start on the next boot, and apply's healthy path (device active → "already in place") never re-enabled it.

## Fix

The healthy branch now checks `is-enabled` and re-enables when needed: a read-only no-op on healthy hosts (zero systemd churn — pinned by test), a single `enable` when persistence was lost.

Observed live: after mask → unmask, `is-enabled` reported `disabled`; verified re-enable restores the symlink to the relocated unit path.

2 regression tests (restore fires + healthy-state stays quiet); 34 green, `bash -n` + ruff clean.

This closes the last behavior gap found by the E3 live-drill series — every advertised behavior (cold activation, idempotency, recovery, opt-out round-trip incl. boot persistence) has now been exercised against the real host.

### DIFF
--- a/scripts/lib/host_swap.sh
+++ b/scripts/lib/host_swap.sh
@@ -221,7 +221,17 @@ host_swap_apply() {
         sudo systemctl restart "$_HOSTSWAP_UNIT_NAME" 2>/dev/null || true
         echo "  zram swap unit already in place (size ${size_mib}MiB) — restarted (was inactive)."
     else
-        echo "  zram swap unit already in place (size ${size_mib}MiB)."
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

PR #1110, opened 2026-07-18
feat(awareness): network-plane infra posture — surface a networkd box missing KeepConfiguration/watchdog

WingedGuardian:
## What

Extends the infra protection posture check (#1096) to the **network plane**, so a
systemd-networkd box that's missing its crash-resilience protections now raises a
standing alert instead of staying silent — completing the silent-skip closure
across both memory and network.

The two network-resilience facts (`networkd_keep_configuration`,
`network_watchdog_installed`) already exist but read `False` on **both** a
NetworkManager box (protection not applicable) **and** a networkd box genuinely
missing the protection (a real defect). Surfacing them as-is would false-positive
on public NetworkManager installs.

## How

A new `networkd_manages_default_route` disambiguation fact resolves that: the
**running** systemd-networkd daemon (queried live via `networkctl --json=short
list`) must report the **default-route interface** as
`AdministrativeState == "configured"`. The two posture rules are gated strictly on
that being `True`.

- A wrong alert would require networkd to claim it configures a link it doesn't
  manage — a contradiction. Every doubt path (no `networkctl`, daemon down,
  `unmanaged` link, parse failure, no default route) → `False` → the rules stay
  silent. A false-negative is re-checked next collection; there is no false-alarm
  path on a NetworkManager install.
- Chosen over reading `/run/systemd/netif/state`: the unit ships
  `RuntimeDirectoryPreserve=yes`, so that runtime state **survives a networkd
  stop** and its presence would not prove networkd is the active manager.

The alert content wrapper was made plane-agnostic (it previously said
"Memory-protection posture"); per-slug remediation detail is unchanged.

## Tests

- 6 posture-rule cases (defect detection, NetworkManager suppression, absent-gate
  on pre-existing profiles, strict-`is True`, not-ok-section ignored, healthy
  silence); 9 `_networkd_manages_link` param cases + 2 collector-wiring tests;
  coverage guardrail extended to the network facts.
- Live E2E on a networkd box: real collector → `networkd_manages_default_route=True`,
  posture silent; fabricated unprotected-networkd → both slugs; fabricated NM box
  → silent.

## Deploy

Runtime code (`git pull` + server restart via `update.sh`). The new collector
fact populates on the next profile refresh (boot + daily); the posture check runs
on the hourly awareness tick. Empty/old profiles without the fact suppress the
network rules until re-collected.

Ledger: fbefe3823ba9401b9454968eb5bc0965

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,15 +51,18 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   This is the foundation; the hooks that write predictions on every action
   and the grader that scores them land next.
 
-- **An unprotected box now tells you.** If a memory-crash protection is
+- **An unprotected box now tells you.** If a crash-resilience protection is
   missing on your install — container swap disabled, systemd-oomd
-  pressure-kill not configured, no host swap, or the container's swap
-  allowance switched off — Genesis now raises a standing alert (dashboard +
-  morning report) naming what's missing and how to fix it, and clears it
-  automatically once the protection is restored. Previously a box that was
-  *always* unprotected produced no signal at all; only a *change* was
-  detected. If the infrastructure self-profile stops refreshing (>3 days
-  old), you get a distinct "posture unknown" alert instead of stale claims.
+  pressure-kill not configured, no host swap, the container's swap allowance
+  switched off, or (on a systemd-networkd–managed box) the network
+  address-retention or the self-healing networkd watchdog missing — Genesis
+  now raises a standing alert (dashboard + morning report) naming what's
+  missing and how to fix it, and clears it automatically once the protection
+  is restored. The network checks stay silent on NetworkManager installs,
+  where they don't apply. Previously a box that was *always* unprotected
+  produced no signal at all; only a *change* was detected. If the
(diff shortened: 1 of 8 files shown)


WingedGuardian (commented):

---

WingedGuardian:
Both P2 findings addressed in 795b49c2 — thanks, both were real false-negatives:

**P2 #1 (KeepConfiguration on the routed link):** the posture rule now reads a new **route-scoped** fact `networkd_default_route_keepconfig`, which checks KeepConfiguration on the default-route link's *own* `.network.d` drop-in (located via networkctl's `NetworkFile`), instead of the install-wide OR. A protected non-default link on a multi-interface box can no longer mask a bare default route. The old any-link `networkd_keep_configuration` fact stays for the annotation layer.

**P2 #2 (watchdog timer running):** the rule now reads `network_watchdog_enabled` (`systemctl is-enabled genesis-network-watchdog.timer`), so a present-but-disabled timer — exactly the case the installer's ignored enable/start failures can produce — is caught. The file-present `network_watchdog_installed` fact stays for the annotation layer.

Verified live: on a networkd box both effective facts read healthy → posture silent; fabricated defects → both slugs fire. +3 collector tests cover the route-iface parse, route-scoped keepconfig, and the effective-fact wiring.

PR #1112, opened 2026-07-18
feat(entity): entity_adjudications ledger + deferred-work retention (PR-A substrate)

WingedGuardian:
## What

Dark substrate for the **entity_adjudication drainer** (follow-up 9127e8ed) — the consumer that will decide merge-vs-distinct for fuzzy near-duplicate entity nodes. This PR adds only the schema, CRUD, and retention primitives; **no producer or consumer is wired**, so runtime behavior is unchanged.

## Why

`resolve_entity` creates a new entity when a name is fuzzily close (difflib ≥0.85) to an existing one, tags it AMBIGUOUS, and would enqueue an adjudication row — but the consumer was never built (gated off in #1099). This lands the decision ledger and dedup key that the drainer (PR-B) needs.

## Changes

- **`entity_adjudications` table** (migration 0065 + canonical `_tables.py`): one row per fuzzy entity pair, order-independent `pair_key UNIQUE` (the dedup key the producer never had). `verdict ∈ merge/distinct/proposed_merge/stale`; `norm_*`/`updated_*` snapshots power the propose-only→live staleness guard. Distinct from `entity_resolution_audit` (which trails memory-pair dedup — this records entity NODES).
- **`db/crud/entity_adjudications.py`**: `record_verdict` (upsert on pair_key), `get_by_pair` (order-independent), `all_pair_keys` (sweep dedup set), `list_proposed_merges`, `mark_applied`, `mark_stale`.
- **`deferred_work.prune_terminal`**: retention for finished queue rows across ALL work types, pruned on `completed_at` (not `created_at`) — closes a pre-existing accumulate-forever gap (2,348 completed rows, oldest 4 months).
- **Harden migration 0051**: pin its three entity indexes explicitly instead of prefix-scanning the live `INDEXES` list. The prefix scan swept in any future `idx_entity_*` index on a table 0051 does not create — a latent bug the 0065 index tripped. Matches the explicit-DDL pattern already used by migration 0033.

## Tests

60 pass: migration 0065 (full column set, index, idempotency, verdict CHECK, pair_key UNIQUE, fresh-canonical parity, down), CRUD (order-independence, upsert-dedup, proposed backlog, apply, stale), prune (completed_at semantics, spares pending/processing), schema allow-list.

## Deploy

Runtime code + additive idempotent migration — applies at server restart via `update.sh`. Empty-state safe (new table, no backfill).

### DIFF
--- a/src/genesis/db/crud/deferred_work.py
+++ b/src/genesis/db/crud/deferred_work.py
@@ -25,8 +25,19 @@ async def create(
            (id, work_type, call_site_id, priority, payload_json, deferred_at,
             deferred_reason, staleness_policy, staleness_ttl_s, status, created_at)
            VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)""",
-        (id, work_type, call_site_id, priority, payload_json, deferred_at,
-         deferred_reason, staleness_policy, staleness_ttl_s, status, created_at),
+        (
+            id,
+            work_type,
+            call_site_id,
+            priority,
+            payload_json,
+            deferred_at,
+            deferred_reason,
+            staleness_policy,
+            staleness_ttl_s,
+            status,
+            created_at,
+        ),
     )
     await db.commit()
     return id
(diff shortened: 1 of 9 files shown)


WingedGuardian:
## First-pass review — PR #1112

### Structural review

**Pattern violations:** CLEAN — no ERROR-class violations found.

**Import hygiene:** CLEAN — all imports correctly grouped (stdlib → third-party → local), no star imports, no obvious circular imports.

**Test coverage:** All three new source units have corresponding test files. One gap:

> **NOTE — `0051_entity_layer.py` fix untested in this PR.**
> The hardened `_ENTITY_INDEXES` tuple replaces the prefix-scan approach, but no test explicitly covers this code path or the 0051 → 0065 migration sequence. The bug it fixes (prefix-scan sweeping in `idx_entity_adjud_verdict` before its table exists) would only surface when both migrations run in sequence. If the full migration-chain test does not cover this sequence, a targeted test (`test_migration_0051_with_0065.py`) would close the gap.

**Genesis conventions:** CLEAN — `feat(entity):` prefix present, no `IntervalTrigger`, no hook scripts, no `$CLAUDE_PROJECT_DIR` misuse.

**Naming trap:**

> **NOTE — `pair_key` parameter shadows the module-level `pair_key()` function in `mark_applied` and `mark_stale`** (`src/genesis/db/crud/entity_adjudications.py`, lines ~147 and ~168).
> Not a runtime bug today (neither function calls `pair_key()` internally), but a refactor trap: any future maintainer adding a call to `pair_key()` inside either function will get a `TypeError` instead of the expected string. Consider renaming the parameter to `pk` or `key` to avoid the shadow.

---

No secrets, private IPs, credentials, personal emails, private hostnames, or problematic wording found in added lines or PR metadata.</body>
</invoke>

---

PR #1113, opened 2026-07-18
docs(provisioning): Phase-C RAM grow reboot runbook + guest-reboot gotcha

WingedGuardian:
## What

Documents the sharp edge in applying a `provision_grow(kind="memory")` grow: because memory hotplug is off, the new size is *pending* until the QEMU process is respawned — and a reboot issued from **inside the guest** does not respawn it, so the pending config is silently never applied.

Adds to `docs/reference/proxmox-provisioning.md`:
- The guest-reboot-doesn't-apply-pending gotcha + the Proxmox-side reboot procedure (`qm reboot <vmid>` / GUI), including a detached invocation for when the initiator session lives inside the VM being rebooted, and the ACPI-hang `qm stop`/`qm start` force-cycle fallback.
- A post-reboot checklist: grow verification (`MemTotal`), container autostart, the host zram swap tier (applied on install/redeploy, **not** per guardian tick — distinct from the per-tick container `limits.memory.swap` reconcile), the `kind="limits"` coupling completion, and the per-service percentage-cap restart lag.

## Why

The reboot leg of a Phase-C RAM grow is the one step that has no automation (the provision token deliberately lacks `VM.PowerMgmt`) and the guest-reboot trap is a documented, recurrent Proxmox footgun. Any install growing VM memory hits it.

Docs-only; generic placeholders throughout (no install-specific constants).

### DIFF
--- a/docs/reference/proxmox-provisioning.md
+++ b/docs/reference/proxmox-provisioning.md
@@ -213,7 +213,9 @@ curl -sk -X PUT -H "Authorization: $AUD" "$H/nodes/<NODE>/qemu/<VMID>/config" -d
 - **Grow memory:** `provision_grow` (`kind="memory"`) grows *configured* RAM;
   it **takes effect only after a VM reboot** (hotplug is off on this install).
   The provision token deliberately lacks `VM.PowerMgmt` — power stays
-  human/approved. Schedule the stop/start as a downtime window.
+  human/approved. Schedule the stop/start as a downtime window. **The reboot
+  has a sharp edge — see "Applying a memory grow: the reboot" below** (a reboot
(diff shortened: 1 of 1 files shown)


WingedGuardian (commented):
Structural review: SKIPPED — documentation-only PR (single `.md` file).

PII/Secrets/Wording scan: CLEAN — all infrastructure references use generic angle-bracket placeholders (`<NODE>`, `<VMID>`, `<vmid>`, `<pve-host>`); no private IPs, API keys, personal emails, private hostnames, or hardcoded absolute paths found in added lines. PR title and body contain no leak/credential/security-incident language.

---

PR #1114, opened 2026-07-18
feat(entity): entity_adjudication drainer — LLM merge-vs-distinct for fuzzy pairs (PR-B)

WingedGuardian:
## What

Activates the **entity_adjudication drainer** — the consumer for `work_type='entity_adjudication'` rows — and re-enables the producer (gated off in #1099). Builds on the PR-A substrate (#1112). Closes follow-up `9127e8ed`.

When `resolve_entity` creates an entity whose name is difflib-close (≥0.85) to an existing one, it queues the pair. This drains that queue: decide **merge** (same real-world thing) vs **distinct** (similar names, different things), conservatively — a wrong merge is worse than a wrong split.

## How it decides

1. **Digit-guard** — pairs differing only by digits (`pr #989` vs `pr #990`, ~76% of fuzzy hits, measured live) are mechanically ruled *distinct*, zero LLM cost.
2. **Two-model agreement** — the rest get a primary judgment (`entity_adjudication`) plus a flipped-provider adversarial challenge (`entity_adjudication_challenge`); **both must say merge**, else distinct. Cloned from the existing dream-cycle entity-dedup pattern.
3. Infra/parse errors **re-queue** (never burn a pair to a permanent verdict); genuine model dissent is recorded.

## Modes (settings lever `entity_adjudication`)

- **`propose_only` (default)** — records verdicts, **applies nothing**. Merges are effectively irreversible, so it ships in shadow: review the proposals, then flip.
- **`live`** — applies via `merge_entity` (survivor = better-attested node), re-resolving both sides immediately before the merge (extraction-race guard). Also applies the shadow-period `proposed_merge` backlog, guarded against identity drift (renamed/merged/gone → `stale`, re-discovered by the sweep).
- Plus `GENESIS_ENTITY_ADJUDICATION_DISABLED=1` kill switch; invalid config degrades to `propose_only`, never silent `live`.

## Reconcile sweep

The producer only catches *new* near-duplicates. A cursor-managed sweep (`ego_state`, off-thread difflib, low-water-gated, weekly re-run) rediscovers historical fuzzy pairs — recovers the 272 cleared in #1099 and self-heals any install.

## Also

- Producer dedup (both-orientation `NOT EXISTS`) + `call_site_id` so the dashboard shows the site, not "unknown".
- `entity_adjudication` added to `BATCH_WORK_TYPES` so sweep bursts don't page the `deferred_recovery` alarm.
- Hourly drain job + 45d deferred-work retention job on the learning scheduler.
- One aggregate, digest-visible observation per run (no per-pair flooding).

## Tests

30 drainer + config-lever + producer-dedup + CRUD + count/drift guards. Verified: propose_only = **zero mutation** (hard invariant), two-model agreement, live apply + phase-0 backlog + staleness matrix, survivor-by-evidence, retry-on-error, race guard, sweep dedup/cursor/low-water, stale-rediscovery, empty-state.

## Review

A code-reviewer pass over the combined diff returned DONE_WITH_CONCERNS with one NOTE (stale pairs were a dead end) — fixed in this PR (`settled_pair_keys` excludes stale so the sweep re-discovers them).

## Deploy

Runtime code + config default (works with zero overlay) — `git pull` + server restart. Empty-state safe. Ships `propose_only`; flip to live via `settings_update("entity_adjudication", {"mode": "live"})`.

Ledger: 9127e8ed9ac7420dbd3cb4f78f0bf9d1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,26 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ## [Unreleased]
 
+### Added
+
+- **Genesis now tidies near-duplicate entities in its knowledge graph.** When
+  it learns about a "thing" (a project, tool, concept, person) whose name is
+  very close to one it already knows, it now decides whether they are the same
+  thing or genuinely different, and can merge the duplicates so its memory of
+  you stays coherent instead of fragmenting across "neural monitor" /
+  "neural-monitor" / "neural_monitor". Two independent models must agree before
+  anything merges, and pairs that only differ by a number (`PR #989` vs
+  `PR #990`) are never merged. It ships in **shadow mode by default** — it
+  records what it *would* merge without touching anything, so you can review the
+  proposals first; flip it to live with
+  `settings_update("entity_adjudication", {"mode": "live"})` (or turn it off
+  entirely). A background sweep also reconciles the entities it already had.
+
 ### Changed
+
+- **Finished background-queue rows are now pruned after 45 days.** The internal
+  deferred-work queue kept every completed item forever; it now retains 45 days
+  of history and drops the rest, so the queue can't slowly grow without bound.
 
 - **Every terminal door now leads to the same persistent session.** Running
   `claude` by hand over SSH or in the dashboard web terminal now lands in a
--- a/config/entity_adjudication.yaml
+++ b/config/entity_adjudication.yaml
@@ -1,0 +1,44 @@
+# Entity adjudication drainer — writable, SHADOW by default.
+#
+# The hourly learning-scheduler job (`entity_adjudication_drain`) re-reads the
+# merged config (this file + ~/.genesis/config/entity_adjudication.local.yaml)
+# on every run, so a settings_update("entity_adjudication", {...}) or a hand
+# edit takes effect on the next run — no restart.
+#
+# Master switch: false stops the job before any queue/LLM/merge work.
+enabled: true
+
+# Mode: off | propose_only | live
+#   off          — the job exits immediately (no drain, no sweep)
+#   propose_only — (default) LLM judges each fuzzy entity pair and RECORDS the
+#                  verdict; a "merge" is stored as `proposed_merge` and NOT
+#                  applied. Review the proposals, then flip to live.
+#   live         — a double-agreed "merge" is APPLIED (loser entity tombstoned
+#                  into the survivor via merge_entity). Stored proposed_merge
+#                  verdicts from the shadow period apply on the flip, guarded
+#                  against staleness (re-adjudicate if identity drifted).
+#
+# Merges are effectively irreversible (the loser's mentions/links are rewritten
+# onto the survivor), so the default is propose_only — the conservative posture
+# (a wrong merge is worse than a wrong split).
+#
+# Hook/env kill switch (separate lever): GENESIS_ENTITY_ADJUDICATION_DISABLED=1
+# forces the job to no-op regardless of this file.
+mode: propose_only
+
+# Max pending queue rows drained (LLM-judged) per hourly run. ~240 fuzzy pairs
+# exist here after the digit-guard; at 20/run the backlog clears in ~12 runs.
+drain_budget: 20
+
+# Reconcile sweep: rediscover fuzzy pairs among EXISTING entities (the producer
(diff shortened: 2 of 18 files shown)


WingedGuardian (commented):
## First-pass review — PR #1114

**Verdict: needs fixes before merge** — 2 structural ERRORs (raw SQL outside the CRUD layer) + 1 WARNING (private-symbol cross-module import). No PII or secrets issues found.

---

### ERROR — Raw SQL in production business logic (2 instances)

Both violations are in `src/genesis/memory/entity_adjudication.py`. Raw SQL in production code outside `db/crud/` and migrations is a checklist ERROR.

**1. `_load_entity` (~line 541)**

Issues a `SELECT … FROM entities` directly instead of calling `entities_crud.get_entity`. The inline comment says "no row_factory dependency," but that rationale doesn't hold: `_row_to_dict` in `entities.py` already handles both `aiosqlite.Row` *and* plain tuple rows, so `get_entity` works with or without `row_factory`. The only real difference is that `_load_entity` selects 8 of 10 columns. Fix: call `entities_crud.get_entity` and ignore `source`/`created_at`, which no caller uses.

**2. `_mention_count` (~line 701)**

Issues a `SELECT COUNT(*) FROM entity_mentions` directly. There is no existing `entities_crud` function that counts mentions for a single entity — this is a genuine gap in the CRUD layer, and the raw SQL is a workaround for it. Fix: add `count_entity_mentions(db, entity_id) -> int` to `src/genesis/db/crud/entities.py` (one or two lines) and call that instead. This also gives sweep and any future call sites a reusable path.

---

### WARNING — Private symbol imported across module boundary

**`src/genesis/mcp/health/settings.py` ~line 896**

`_validate_entity_adjudication` imports `_INT_KNOBS` (underscore-private) from `genesis.memory.entity_adjudication_config`. The config module's own docstring says the only documented cross-module export is `MODES`. Importing a private symbol creates an undocumented coupling that the config module's author hasn't committed to maintaining. The same pattern exists for `repo_pulse_config` (line 816).

Fix options (apply consistently to both sites):
- Promote `_INT_KNOBS` → `INT_KNOBS` (public) in both config modules, and update the docstrings.
- Or inline the three-element tuple directly in `settings.py` to avoid the dependency entirely.

---

### Clean areas

- All new public functions have docstrings. ✓
- New `.py` files ship with corresponding test files (full coverage of drain, sweep, config, and CRUD). ✓
- `CronTrigger(minute=25)` used for the hourly job — not `IntervalTrigger`. ✓
- No `/tmp/` usage, no bare `except:`, no mutable defaults, no `task_states` direct inserts, no `nohup`. ✓
- No secrets, PII, private IPs, or private hostnames in any added lines. ✓
- PR title follows conventional commit format. ✓
- The producer dedup guard (`NOT EXISTS` with both orientations) is well-reasoned. ✓
- The staleness-guard on the live-mode phase-0 backlog apply is correct. ✓

---

WingedGuardian:
All review findings addressed in the latest push:

**Codex P2 — sweep drops exact-norm cross-type pairs** (`entity_adjudication.py`): the skip condition was `cand_id == eid or cand_norm == norm`; the `cand_norm == norm` clause dropped two different entities sharing a norm_name across types (permitted by `UNIQUE(norm_name, entity_type)`, only recoverable by the sweep). Now skips only `cand_id == eid`. Test: `test_sweep_queues_exact_norm_cross_type_pair`.

**Codex P2 — re-pick survivor after the live re-resolve** (`entity_adjudication.py`): survivor selection (mention-count based) now happens on the *fresh* entities after the extraction-race re-resolve, so a concurrent mention write can't cause the now-better-attested node to be tombstoned.

**Structural ERROR — raw SQL outside CRUD** (`entity_adjudication.py`): `_load_entity` replaced with `entities_crud.get_entity` (it doesn't follow merges — correct, the redirect walk is done in `_resolve_active`); `_mention_count` replaced with a new `entities_crud.count_entity_mentions`.

**WARNING — private-symbol import** (`settings.py`): promoted `_INT_KNOBS` → public `INT_KNOBS` in `entity_adjudication_config` (docstring updated) and import the public name. (The pre-existing `repo_pulse_config` instance is in a separate subsystem and left as-is to keep this PR scoped.)

Also verified end-to-end against the real free-tier LLM chain: cosmetic variants merge (both models agreeing), genuine differences stay distinct, transient provider errors re-queue rather than recording a verdict, zero mutation in propose_only.

PR #1115, opened 2026-07-18
feat(eval): WS2-0 retrieval-efficacy instrumentation

WingedGuardian:
## What

The prerequisite for the WS2 retrieval-drift work: makes memory retrieval quality **measurable against pool size**, and gives the upcoming lever PRs a measurement seam. Zero schema change, \$0 spend, entirely additive.

Four parts:

1. **Pool stamping** — `pool_size_counts()` (episodic total/embedded by collection, knowledge units, memory links) folds `pool_*` keys into the weekly J-9 memory snapshot, in its own try/except so a count failure degrades to `None` keys without losing the quality metrics. Quality (precision@k/hit_rate/MRR) and pool size now land in the same row. No migration — `eval_snapshots.dimension` is unconstrained.

2. **Readers** — `j9_eval_status` gains `retrieval_trend` (last 12 weekly snapshots, chronological); `memory_stats` gains a `quality` block (latest snapshot headline + subsystem grade), closing the disjoint capacity-vs-quality surfaces. Both tolerate pre-WS2-0 snapshots (missing `pool_*` → honest `null`, never a KeyError) and un-aggregated installs (→ `None`, never break the capacity read).

3. **Variant seam** — a generic retrieval-config variant mechanism in the LongMemEval harness so each lever PR measures itself offline without re-opening the harness core. `Arm.variant` (label suffix appended **last**, so every pre-existing label stays byte-identical); a `VARIANTS` registry with two application seams (`recall_kwargs` per-question, `post_recall` transform); `validate_variants` fails fast **before any spend**; CLI `--variants` pairs each base arm with a `+variant` twin. Registry ships empty — the levers register theirs.

4. **Report + retention** — `scripts/retrieval_efficacy_report.py`: a read-only comparator that joins the weekly quality snapshots to the pool size they were measured over — the first artifact able to substantiate or refute the "69%→38% drift" premise. Pre-WS2-0 weeks are filled by a labelled, read-only historical reconstruction (never written back). `disk_hygiene.sh` gains a 45d file-age prune for the dated reports.

## Verification

- Targeted tests green across all touched + adjacent suites.
- Read-only E2E on the live DB: `pool_size_counts` returns live numbers; `_compute_memory_quality` stamps `pool_*` additively with quality intact; both readers degrade to `null` on the current (pre-WS2-0) snapshots; the report script renders the live pool-vs-quality trend.
- Adversarial code review: no material defects; additive/degradation contracts verified and directly tested.

## Deploy path

Runtime code (aggregator + MCP readers): `git pull` + server restart — the next weekly aggregation (Sunday) writes the first `pool_*`-bearing snapshot. MCP readers take effect at the next CC session. Report script + hygiene prune land via `git pull`.

### DIFF
--- a/scripts/disk_hygiene.sh
+++ b/scripts/disk_hygiene.sh
@@ -17,6 +17,8 @@
 #      (WS-5 Discord observe-only gate log; bounds the shadow store)
 #   8. Retention prune of session_ledger_shadow_* (>45d) → scripts/prune_ledger_shadow.py
 #      (session-manager PR-3 ambient extractor shadow store; runs + events)
+#   9. Retention prune of ~/.genesis/output/retrieval_efficacy/*.md (>45d)
+#      (WS2-0 retrieval-efficacy report; dated md per run — file-age prune)
 #
 # Note: run under a hardened systemd sandbox (NoNewPrivileges, ProtectSystem=
 # strict), so disk_reclaim's --system (/var, sudo) path is intentionally NOT
@@ -96,6 +98,15 @@ main() {
     echo "--- repo pulse retention prune (>45d) ---"
     "$VENV_PY" "$REPO_DIR/scripts/prune_repo_pulse.py" --days 45 \
         || echo "prune_repo_pulse exited $?"
+
+    echo "--- retrieval-efficacy report retention prune (>45d) ---"
+    # WS2-0: retrieval_efficacy_report.py writes a dated md per run; bound the
+    # dir so a periodic report never slow-leaks disk on a smaller install.
+    if [ -d "$HOME/.genesis/output/retrieval_efficacy" ]; then
+        find "$HOME/.genesis/output/retrieval_efficacy" -maxdepth 1 -type f \
+            -name '*.md' -mtime +45 -delete 2>/dev/null \
+            || echo "retrieval_efficacy prune exited $?"
+    fi
 
     echo "=== genesis-disk-hygiene done ==="
 }
--- a/scripts/retrieval_efficacy_report.py
+++ b/scripts/retrieval_efficacy_report.py
@@ -83,6 +83,11 @@ def build_report(
                 "mrr": metrics.get("mrr"),
                 "total_recalls": metrics.get("total_recalls"),
                 "pool_episodic_total": pool,
+                # Retrievable pool (deprecated/expired excluded) — the denominator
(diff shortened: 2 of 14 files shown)


WingedGuardian:
## First-pass review — PR #1115 feat(eval): WS2-0 retrieval-efficacy instrumentation

### Structural review

**WARNING — Raw SQL outside CRUD layer**

`scripts/retrieval_efficacy_report.py` · `_reconstruct_pool` (lines ~218–233)

`_reconstruct_pool` runs a raw SQL query against `memory_metadata` directly on the `aiosqlite.Connection` instead of going through a CRUD module:

```python
    "SELECT COUNT(*) FROM memory_metadata "
    "WHERE collection = 'episodic_memory' AND created_at <= ?",
    (pe,),
)
```

Rule: *Raw SQL against genesis.db should use `genesis.db.crud` modules, except in migrations.* The `_load` function already imports `genesis.db.crud.j9_eval` and `genesis.eval.db` for the other reads; the historical pool count could live in `genesis.db.crud.memory` alongside `pool_size_counts` (which already does the same collection filter for the live count). Adding a `pool_size_at(db, period_end)` CRUD function would keep the SQL in one place and make the report testable at the CRUD layer without touching `aiosqlite` directly.

---

**NOTE — Missing docstring on public function**

`scripts/retrieval_efficacy_report.py` · `render_md`

`render_md` is a public function with no docstring. The module docstring covers the script's overall intent but doesn't describe what `render_md` accepts or returns.

---

**CLEAN** — no API keys, tokens, private IP addresses, SSH keys, credentials, personal email addresses, private hostnames, or security-incident language detected in the diff or PR metadata.

---

WingedGuardian:
Addressed the review in `49114f2f`:

- **P2-1 (pool counts include non-retrievable rows)** — added `pool_episodic_retrievable`, applying the same deprecated + bitemporal-validity filters the recall path uses (`recall` :195-203). It's the denominator the drift hypothesis actually cares about; `pool_episodic_total` stays for reference (total − retrievable = accumulated dead weight, ~3.7K live). Surfaced in the trend reader, `memory_stats` quality block, and the report (new column).
- **P2-2 (variants not on the primary CLI)** — wired `--variants` into `genesis eval longmemeval`, not just the module `__main__`; parse test added.
- **Structural (raw SQL in the report script)** — moved the pool-reconstruction query into a CRUD helper (`episodic_count_created_before`).

PR #1116, opened 2026-07-18
docs(process): instance-fix vs class-fix gate — data repair is not a fix

WingedGuardian:
## What

WS-G from the 2026-07-17 ego investigation: prevent the failure mode where a mechanism's missing output is hand-written and then reported as "fixed".

- **genesis-development SKILL.md** — new "Instance-Fix vs Class-Fix Gate" section in On-Load Mindset: hand-writing a missing artifact (memory/directive/row/flag) is a **data repair** that mitigates one instance on one install; not done until the mechanism is fixed or the user explicitly defers. Plus an Anti-Rationalization row ("The missing data was the problem" → "The mechanism that failed to write it is the problem").
- **CLAUDE.md** — matching one-rule summary in Rules so the gate binds every session, not just skill-loaded ones.

## Why

A settled user decision was re-litigated by the ego because no mechanism persisted the ruling; the first remediation attempt wrote the missing memory+directive by hand and framed it as fixed. The mechanism fixes ship separately (decision-propagation PR series); this PR makes the process failure itself non-repeatable.

## Deploy path

Docs-only. Existing installs: git pull. Fresh installs: in-repo. No migrations, no restart required (skill/CLAUDE.md load at next session start).

### DIFF
--- a/.claude/skills/genesis-development/SKILL.md
+++ b/.claude/skills/genesis-development/SKILL.md
@@ -85,6 +85,27 @@ changes: verify the notification actually arrives. Ask: "If the system
 restarts right now, will this actually work?" If you can't answer yes
 with evidence, you're not done.
 
+### Instance-Fix vs Class-Fix Gate
+
+When a mechanism failed to write or propagate something (a memory, a
+directive, a config row, a status flag), hand-writing the missing
+artifact is a **data repair** — it mitigates ONE instance on ONE
+install. It is never the fix. Before reporting anything as "fixed",
+classify it:
+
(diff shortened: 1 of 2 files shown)


WingedGuardian:
**Documentation-only PR — structural review skipped.**

- Added lines (`SKILL.md`, `CLAUDE.md`): no API keys, tokens, private IPs, credentials, personal emails, private hostnames, or absolute paths with usernames.
- PR title/description: no language referencing PII leaks, security incidents, infrastructure exposure, or credential cleanup.

---

PR #1117, opened 2026-07-18
fix(ego): direction-neutral session-pattern reading + persisted baseline_note

WingedGuardian:
## Problem

`user_session_pattern` is a **symmetric** deviation score — both drops and surges in foreground activity raise it (collector: `learning/signals/user_session_pattern.py`). But the user-ego Activity Pulse hardcoded a directional reading: any value ≥0.3 rendered as *"user activity is … below their baseline"*. During an activity **surge** (e.g. baseline 4/day → recent 27/day) the ego was told the user had gone quiet — an inverted picture that fed wrong proposals and reflection noise.

## Fix

- **Persist ground truth**: `perform_tick` now includes each signal's `baseline_note` in the persisted `signals_json` (additive key, omitted when the collector doesn't set one — old-format consumers unaffected).
- **Neutral interpretation**: pulse text becomes "deviates significantly/somewhat from their baseline", with the persisted note appended ("Baseline: 4.0 sessions/day … Recent: 27.0/day") so the ego sees the direction from data, not from a guess.
- Old ticks without the key degrade gracefully to the neutral text.

Swept the repo for other directional baseline text: this was the only site.

## Tests

- `tests/test_awareness/test_integration.py`: baseline_note persisted when provided, key omitted when not.
- `tests/ego/test_user_context.py`: note rendered on a surge tick; never the old "below" text; pre-existing ticks without the key degrade to neutral.

## Deploy path

Repo code only — existing installs: git pull + restart; fresh installs: in-repo. No migrations (JSON key is additive).

### DIFF
--- a/src/genesis/awareness/loop.py
+++ b/src/genesis/awareness/loop.py
@@ -1510,7 +1510,11 @@ async def perform_tick(
             source=source,
             signals_json=json.dumps([
                 {"name": s.name, "value": s.value, "source": s.source,
-                 "collected_at": s.collected_at}
+                 "collected_at": s.collected_at,
+                 # Additive: ground-truth context (e.g. "Baseline: 4.0/day,
+                 # Recent: 27.0/day") so downstream consumers can't misread
+                 # a symmetric deviation score's direction.
+                 **({"baseline_note": s.baseline_note} if s.baseline_note else {})}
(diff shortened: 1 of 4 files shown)


WingedGuardian (commented):

---

PR #1118, opened 2026-07-18
fix(guardian): write-probe the CC work dir before trusting it (audit idx 0)

WingedGuardian:
## What

Audit finding **idx 0**: the guardian's CC recovery brain could be silently
blinded by a misconfigured `cc.work_dir`. `_ensure_work_dir` did
`configured.mkdir(parents=True, exist_ok=True)` then returned the path — but
that mkdir is a *silent no-op* on a directory that already exists but isn't
writable by the guardian user (a root-owned `/var/lib/guardian-snapshots/cc-sessions`
left over from a vintage install). The CC subprocess then ran with an unwritable
`cwd`, so session/cache writes failed and the recovery brain stayed blinded
exactly when it was needed. #965 only handled the *non-creatable* case; the
*exists-but-unwritable* half slipped through.

**Fix:** a real write-probe (`_dir_is_writable` — create + delete a
`.guardian-writeprobe-<pid>` file, since `mkdir(exist_ok=True)` and `os.access`
both report success on a root-owned dir). `_ensure_work_dir` now probes the
configured dir (falls back if unwritable) and the fallback dir (raises `OSError`
if unwritable). The sole caller `_diagnose_with_cc` runs inside `diagnose()`'s
`try/except Exception → _escalate_without_cc`, so "no writable dir anywhere"
degrades cleanly to alert-only (prime directive), never a tick crash. All prior
behavior preserved.

## Also — idx 24 regression lock (no code change)

Audit finding **idx 24** (redeploy must restart `genesis-guardian.timer` on
every failure path) was **already fixed** in `guardian-gateway.sh` (an `EXIT`
trap gated on `_TIMER_STOPPED`). This PR adds a **structural regression-lock
test** so a future edit to the large hand-maintained gateway script can't
silently remove that invariant and leave a host guardian disabled.

## Tests / verification
- TDD: 2 new `_ensure_work_dir` tests (confirmed RED on old code → GREEN), 3
  pre-existing preserved; 3 gateway-invariant lock tests. Full
  `tests/test_guardian/` green (1000 passed).
- Live filesystem E2E on all four paths (writable→configured,
  exists-but-unwritable→fallback, none→OSError, `_dir_is_writable` direct) — PASS.
- Code review: DONE, scope CLEAN, no blocking findings.

## Deploy note (Host-Deploy Gate)
Touches `src/genesis/guardian/` → the host-deploy gate applies. After merge,
`scripts/update.sh` redeploys the guardian to the host and I verify via the
gateway `version` op + a healthy tick. Host is currently healthy and reachable;
idx 0 is latent-class here, not an active fire.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,16 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **The host recovery brain no longer goes blind on a misconfigured work
+  directory.** If the guardian's configured Claude Code work directory already
+  exists but isn't writable by the guardian (for example a root-owned
+  `/var/lib` path left over from an older install), it now detects that with a
+  real write probe and falls back to a user-writable directory instead of
+  handing the recovery session an unusable working directory. Previously only a
+  *non-creatable* directory triggered the fallback; an existing-but-unwritable
+  one slipped through and could blind the recovery brain exactly when it was
+  needed most.
+
 - **Disaster recovery no longer risks corrupting the thing it's recovering.** A
   script audit found three ways deploy/restore could bite at the worst moment,
   now fixed: (1) during a database restore, the health watchdog could restart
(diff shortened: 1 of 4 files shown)


WingedGuardian (commented):

---

WingedGuardian:
Merge basis: CI 12/12 green, structural review CLEAN (incl. PII/secrets/wording), and an independent code-review verdict of DONE (scope clean, no blocking findings). Codex Cloud review was quota/latency-limited (no review posted after ~21 min), so merging on CI + structural + code-review per the pre-merge gate's quota-limited provision (`# review-override`).

PR #1119, opened 2026-07-18
feat(autonomy): windowed earn-back evidence — autonomy_events ledger (0067)

WingedGuardian:
## Problem

Earn-back eligibility (`detect_earnback_candidates`) computed `bayesian_level(total_successes, total_corrections)` over **lifetime** counters. After a deep regression this gate is mathematically unreachable — e.g. a category at 38S/21C (posterior 0.64) needs ~85 more *consecutive* successes to reach the L4 threshold. Result on a live install: a category demoted for 8 weeks, zero earn-back proposals ever created, and the `autonomy_activity` awareness signal pinned at 1.0 the whole time, generating recurring reflection noise about a condition the system could never resolve.

## Fix

- **`autonomy_events`** — append-only `(id, category, kind CHECK success|correction, occurred_at)` ledger + `(category, occurred_at)` index. `record_success`/`record_correction` append in the same transaction as the counter update. Migration `0067` (idempotent) + canonical DDL in `_tables.py` (both build paths).
- **Windowed gate** — `detect_earnback_candidates` now computes the posterior over `windowed_counts(category, window_days)`; `earnback.window_days` is a config lever in `config/autonomy.yaml` (default 45, code fallback identical). An empty window means *no evidence*, not eligibility. Candidate dicts carry `window_successes/window_corrections/window_days`; `posterior` is the windowed one (drives the proposal's stated confidence). Regression math and every other lifetime-counter consumer untouched.
- **No backfill** — lifetime counters predate the ledger; fabricating `occurred_at` would manufacture evidence. Categories become eligible as genuine windowed evidence accumulates.
- **Signal addendum** — `autonomy_activity`: regression **with** a pending `autonomy_earnback` proposal → 0.7 (`regression_pending_earnback_<cat>`); without → 1.0 as before. The dampened state means "recovery path is awaiting the user" — a pinned 1.0 there is pure noise.

## Tests

Migration idempotence + CHECK enforcement + base path; event writes through the manager; `windowed_counts` boundary (44d in / 46d out, injected clock) + category scoping; candidate matrix (recovered-in-window, insufficient-in-window, **empty window**, **old corrections aged out** — the unreachable-gate case, recent corrections still block, window_days lever); collector 0.7-vs-1.0 (pending vs resolved proposal); e2e earnback flow reseeded with windowed events. 130 passed across the five touched suites.

## Deploy path

Existing installs: git pull + restart (migration applies at boot). Fresh installs: base-path DDL. Config default works with zero overlay; `earnback.window_days` is the documented lever.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,20 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ## [Unreleased]
 
+### Fixed
+
+- **Demoted autonomy can actually earn its way back now.** Earn-back
+  eligibility used to be computed over a category's entire lifetime record,
+  so after a rough patch the math could require months of flawless behavior
+  before Genesis would even *propose* restoring a level — in practice the
+  demotion was permanent and the system nagged about it forever. Eligibility
+  now looks at a recent evidence window (45 days by default,
+  `earnback.window_days` in `config/autonomy.yaml`): old mistakes age out,
+  recent clean behavior counts, and promotion still always requires your
+  explicit approval. While an earn-back proposal is sitting in your queue,
+  the internal "autonomy regressed" alarm also calms down instead of firing
+  on every awareness tick.
+
 ### Added
 
 - **Genesis now tidies near-duplicate entities in its knowledge graph.** When
--- a/config/autonomy.yaml
+++ b/config/autonomy.yaml
@@ -35,6 +35,13 @@ approval_timeouts:
   task_proposal: null
   irreversible: null     # Wait forever
 
+# Earn-back evidence window (days). Eligibility to propose restoring a
+# demoted autonomy level is computed from the Bayesian posterior over
+# autonomy_events within this window — recent behavior, not lifetime
+# history. Promotion always requires explicit user approval.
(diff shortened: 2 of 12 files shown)


WingedGuardian:
## First-pass review — PR #1119

**Structural review:** 1 WARNING

---

### WARNING — Raw SQL outside crud module

**`src/genesis/learning/signals/autonomy_activity.py` — `_earnback_pending()`**

```python
    "SELECT 1 FROM ego_proposals WHERE status = 'pending' "
    "AND action_type = 'autonomy_earnback' "
    "AND action_category = ? LIMIT 1",
    (category,),
)
```

This executes SQL against `ego_proposals` directly from a learning signals collector, bypassing the crud layer. The Genesis convention is: raw SQL belongs in `src/genesis/db/crud/` modules (or migrations), not in consumer code. Recommend extracting this to a function such as `src/genesis/db/crud/ego_proposals.py::has_pending_earnback(db, category)` and calling that here.

The exception handling (`except Exception`) is fine — it has an explicit type, so it is not a bare `except:`.

---

All other changed files are clean: migration idempotence, import order, new test file for 0067, windowed-gate test coverage, and commit prefix all look good.

---

PR #1123, opened 2026-07-18
feat(ego): decision propagation — shared resolution hook + settled decisions (0066)

WingedGuardian:
## Problem (root cause of the settled-decision re-litigation)

A user ruling resolved through any proposal path had **no durable, ego-visible representation**. Enumerating the resolution entry points found FIVE (the investigation's four plus the dashboard comms-tab route), each running a different side-effect subset:

| Path | journal | J-9 | correction memory | any ego-visible artifact |
|---|---|---|---|---|
| Telegram workflow | ✅ | ✅ | ✅ | ❌ |
| Dashboard ego route | ✅ | ❌ | ❌ | ❌ |
| Dashboard comms route | ✅ | ❌ | ❌ (+ only 4 of 6 action hooks) | ❌ |
| MCP tool | ❌ | ❌ | ❌ | ❌ |
| Free-text | via workflow | via workflow | ✅ | ❌ |

The only corrections-recall section was dead code filtering the wrong tag, and `_proposal_history_section` deliberately strips `user_response` (deference bias — correct for outcome patterns, fatal for rulings). Result: the ego agreed to a decision in chat, then re-proposed the same theme days later.

## Fix

1. **`ego/resolution.py::handle_proposal_resolution`** — one shared hook (isolated blocks: J-9, intervention journal, decision capture, correction memory, six action hooks), called by all five entry points after the status update. The dashboard deny reason is now a first-class signal.
2. **Decision rows** — `ego_directives` gains `kind IN ('directive','decision')` + `source_proposal_id`/`reaffirm_count`/`last_reaffirmed_at` (**migration 0066**, PRAGMA-guarded, + base DDL both paths). Reject-with-reason (not one-off) → create-or-reaffirm a ruling deduped on the `[action_type/category]` prefix. `resolve_directive` structurally cannot complete a decision; user-driven `supersede_decision` is the only retire path.
3. **Visibility** — `## Settled Decisions` section in BOTH context builders, registered `_ALWAYS_SECTIONS` (never skipped or compacted): "user rulings — constraints, not signals; do not re-propose, re-litigate, or engineer workarounds; only the user supersedes; if circumstances changed, ASK." Matching contract in `USER_EGO_SESSION.md` (directives are input; decisions are boundaries). `user_response` stripping in proposal history is deliberately kept.
[...]

PR #1127, opened 2026-07-18
fix(db): ego_directives decision columns on base create_all_tables path

WingedGuardian:
## Problem

PR #1123 (decision propagation) crashed the server at boot on any **existing**
install:

```
sqlite3.OperationalError: no such column: kind
  File ".../db/schema/_migrations.py", line 39, in create_all_tables
ERROR: DB init failed — cannot continue bootstrap
```

`create_all_tables` runs `_migrate_add_columns` and **then** builds `INDEXES`
(the ordering is deliberate — indexes may reference migration-added columns).
PR #1123 added the decision columns to the canonical `CREATE TABLE` and added
`idx_ego_directives_kind_status` to `INDEXES`, but the existing-DB column-add
lived **only** in the numbered migration `0066`. On a DB whose `ego_directives`
predates the decision columns, `CREATE TABLE IF NOT EXISTS` is a no-op, so the
index build referenced a `kind` column that did not exist yet → crash.

Fresh-DB CI went green because the canonical DDL already carries the columns,
so the base path never had to add them. The existing-DB upgrade path was never
exercised.

## Fix

- Add the four column-adds (`kind`, `source_proposal_id`, `reaffirm_count`,
  `last_reaffirmed_at`) to `_migrate_add_columns`, idempotent and ordered
  before the `INDEXES` loop. The numbered migration `0066` stays for the
  standalone runner and is a redundant no-op on this path.
- Regression test builds a legacy `ego_directives`, runs `create_all_tables`,
  and asserts no crash + columns present + index present. It fails with
  `no such column: kind` without the fix (verified).

`schema_both_build_paths` — new DDL/indexes must exist on BOTH the base
`create_all_tables` path and the numbered-migration path.

## Deploy note

The affected install was recovered by restarting with this fix (the columns +
index are now live). Any other existing install must take this before/with the
#1123 code, or its next restart fails to boot.

### DIFF
--- a/src/genesis/db/schema/_migrations.py
+++ b/src/genesis/db/schema/_migrations.py
@@ -1518,6 +1518,35 @@ async def _migrate_add_columns(db: aiosqlite.Connection) -> None:
         "approval_requests.chain_hash",
     )
 
+    # ego_directives: kind='decision' rows (durable user rulings) — see
+    # 0066_ego_directive_decisions.py. These MUST be added here (the base
+    # create_all_tables path) and not only in the numbered migration, because
+    # create_all_tables builds INDEXES (incl. idx_ego_directives_kind_status,
+    # which references `kind`) immediately after this function — on an existing
+    # DB the CREATE TABLE is a no-op, so without this the index build hits
+    # 'no such column: kind'. schema_both_build_paths.
+    await _try_alter(
+        db,
+        "ALTER TABLE ego_directives ADD COLUMN kind TEXT NOT NULL DEFAULT 'directive' "
+        "CHECK (kind IN ('directive', 'decision'))",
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

WingedGuardian (commented):

Checked: `_migrations.py` (four `_try_alter` column-adds) + new `test_ego_directives_base_path_upgrade.py`. No nohup, no /tmp/, no raw SQL outside migration context, no bare excepts. The new test file has a module docstring. New test file in `tests/test_db/` corresponds to the schema change in `src/genesis/db/`. Conventional commit prefix `fix(db):` present. No PII, private IPs, secrets, or wording issues.

---

PR #1128, opened 2026-07-18
fix(outreach): provision approvals/outcomes never deduped (must always deliver)

WingedGuardian:
## Problem

`provision_approval` and `provision_outcome` were **absent** from
`_DEDUP_WINDOWS`, so they fell through to `_DEFAULT_DEDUP_HOURS = 24`. A
synchronous, user-initiated grow approval (`provision_grow` → `submit_raw_and_wait`
→ `submit_raw` → `governance.is_duplicate`) that **times out unanswered** still
writes a `delivered_at` row — and the 24h window then **REJECTs the retry for
24h**. Observed live 2026-07-18 during the Phase-C RAM grow: a second
`provision_grow` returned instantly denied, and `outreach_history` had to be
hand-edited to un-wedge it (data repair, not a fix).

## Fix

Add `provision_approval: 0` and `provision_outcome: 0` to the never-dedup
bucket — exactly what **#143** did for `cli_approval` ("every approval request
must be delivered"). This is the established policy for awaited approvals; these
two were simply never added when provisioning (PR-D) landed a year later.

## Why this is safe (no approval-spam regression)

- **Serena `find_referencing_symbols`** on `grow_via_pipeline` — the only
  container path that hits governance dedup — enumerates exactly **two
  production callers**, both synchronous/user-initiated: the `provision_grow`
  MCP tool and the dashboard `provision_grow_rpc` route. No autonomous re-asker.
- The **autonomous** pool-crit propose path (`guardian/flow.py maybe_propose_pool_grow`)
  is **host-side**, sends via the guardian's own dispatcher (never touches
  container governance), and is independently throttled by its ledger damper
  (`min_repropose_hours=24`). `window=0` on the container gate cannot affect it.
- The stale `delivered_at` row is harmless elsewhere: `_within_rate_limit` and
  `_engagement_throttle` both exempt `blocker`.

## Folded in: in-flight awaited-approval guard (Codex P2)

Codex flagged the residual race: with window=0, the same grow submitted twice
before the first prompt times out delivers BOTH prompts; two identical pending
waiters in one chat make `resolve_scoped_pending` ambiguous
(`len(eligible) != 1`), so a plain APPROVE resolves neither. Fixed **in this
PR** at the layer that knows waiter state: `submit_and_wait` /
`submit_raw_and_wait` track in-flight awaited approvals by
`(signal_type, topic)` — a concurrent duplicate is suppressed (REJECTED,
never sent), and the key releases in a `finally` when the wait ends (answer
OR timeout), so the retry-after-timeout fix is preserved. Both awaited entry
points guarded. 2 new tests pin suppressed-while-pending +
released-after-timeout.

## Test

`test_provision_approval_never_deduped` mirrors `test_dedup_window_zero_skips_check`:
for both signal types, a just-delivered identical row leaves `is_duplicate()`
False AND the full blocker `check()` returns `BYPASS` (delivers), not
DENY-suppressed. All 25 governance tests pass.

## Deploy

Container-only (runs in genesis-server): code-only deploy (reinstall the
editable package + `systemctl --user restart genesis-server`). No host/guardian
path touched.

Ledger: d849794d473d4527a7c72a7169065161

### DIFF
--- a/src/genesis/outreach/governance.py
+++ b/src/genesis/outreach/governance.py
@@ -42,6 +42,17 @@ _DEDUP_WINDOWS: dict[str, int] = {
     "surplus_opportunity": 24,
     "content_review": 1,  # Short window — distinct content pieces may share topics
     "cli_approval": 0,  # Never dedup — every approval request must be delivered
+    # Provisioning approvals + outcomes — same "must always deliver" class as
+    # cli_approval (#143). A synchronous, user-initiated grow approval that
+    # times out UNANSWERED still leaves a delivered_at row; at the 24h default
+    # that row then REJECTs the retry for 24h (observed 2026-07-18, Phase-C:
+    # a second provision_grow returned instantly denied). No autonomous
+    # re-asker exists on this path — the only callers of grow_via_pipeline are
+    # the synchronous MCP tool + dashboard route; the guardian pool-crit
+    # propose path is host-side and independently ledger-damped — so 0 is safe.
+    # Outcomes are distinct operational status events, like task_* below.
+    "provision_approval": 0,
+    "provision_outcome": 0,
     "ambient_health": 0,  # Never dedup — the monitor's state machine gates re-alerts/recovery
     # Task lifecycle notifications — never dedup. Each is a distinct status
     # event for one task (topic = "Task <id>"), and every _notify call site
--- a/src/genesis/outreach/pipeline.py
+++ b/src/genesis/outreach/pipeline.py
@@ -41,6 +41,24 @@ _CHANNEL_FORMAT = {
 }
 
 
+def _awaited_dup_key(request: OutreachRequest) -> str:
+    """Identity of an awaited approval for the in-flight duplicate guard —
+    the same (signal_type, topic) a duplicate prompt would collide on."""
+    return f"{request.signal_type}\x00{request.topic}"
+
+
+def _suppressed_awaited_result() -> OutreachResult:
+    """Terminal result for a concurrent duplicate awaited approval that was
+    suppressed (never sent) because an identical prompt is still pending."""
+    return OutreachResult(
+        outreach_id=str(uuid.uuid4()),
+        status=OutreachStatus.REJECTED,
+        channel="",
+        message_content="",
+        governance_result=None,
+    )
+
+
 class OutreachPipeline:
     """Main outreach orchestrator."""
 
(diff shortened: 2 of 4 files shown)


WingedGuardian (commented):

Checked: `governance.py` (dedup window addition) + `test_governance.py` (new test). No nohup, no /tmp/, no direct task_states inserts, no bare excepts, no mutable defaults. New test has a docstring. Conventional commit prefix `fix(outreach):` present. No PII, private IPs, secrets, or wording issues in diff or PR metadata.

---

WingedGuardian:
Re: the P2 (keep dedup while a waiter is pending) — verified and accepted as a conscious, tracked tradeoff.

Confirmed the mechanism firsthand: `ReplyWaiter.resolve_scoped_pending` (`reply_waiter.py:123`, `if len(eligible) != 1: return []`) resolves neither of two concurrently-pending prompts on a standalone `APPROVE`. So a double-*submit* of the same grow before the first times out can leave both prompts unresolvable by plain `APPROVE`.

Why we're shipping window=0 anyway:
- It fixes the **observed, higher-frequency** failure: a retry after a timed-out approval was REJECTed for 24h (Phase-C, 2026-07-18). That's the bug.
- The double-submit case is a **rare, recoverable race** (deliberate provisioning; realistically a dashboard double-click within the 30-min pending window) — recoverable by quote-replying one prompt, and window=0 lets a fresh single submit work once both expire.
- The `len(eligible) != 1` ambiguity is **pre-existing**: 2+ concurrently-pending approvals in one chat were already mutually-unresolvable by standalone reply (affects `cli_approval` too, window=0 since #143). This change only widens one path to it, it doesn't create it.
- The correct fix is exactly what you suggest — **waiter-pending-aware dedup** (bypass old delivered rows for the timeout-retry, but suppress/coalesce while a live waiter for the same chat/topic is pending). That's tracked as a follow-up (belongs in `submit_raw_and_wait`, which knows the waiter state, not the stateless `GovernanceGate`).

👍 — accurate and useful.

WingedGuardian:
Update: per owner decision, the P2 is now **fixed in this PR** rather than deferred — commit `aba4e3fb` adds an in-flight awaited-approval guard in `submit_and_wait`/`submit_raw_and_wait` keyed by `(signal_type, topic)`: a concurrent duplicate is suppressed (REJECTED, never delivered, no second waiter), and the key releases in a `finally` when the wait ends, preserving the retry-after-timeout fix. Supersedes the accept-as-tradeoff comment above.

PR #1124, opened 2026-07-18
fix(outreach): morning report grounding — ground-truth totals + verbatim single draft

WingedGuardian:
## Problem

Verified claim-by-claim against a live report (2026-07-17): the morning report misstated follow-up counts ("5" vs 268 actual — the length of a display slice reported as the total), described embeddings/oomd state backwards (a protective OOM-guard reported as an active OOM risk), and its numbers had no anchor at all because the grounded draft (built with the MORNING_REPORT.md system prompt) was **silently re-drafted** by the pipeline's generic drafter (`system_prompt=None`) before delivery.

## Fix

1. **`_ground_truth_section()`** — renders FIRST in the context: deterministic totals (follow-ups awaiting input / pending / blocked / failed, pending proposals, pending approvals, unsurfaced observations) computed with the SAME un-truncated queries the list sections run, so the totals can never disagree with the samples. Labeled "authoritative — restate these EXACTLY; the list sections below are truncated samples."
2. **len-after-slice fixes** — proposals/approvals count before slicing; every truncated list ("Needs your input", "Blocked/failed", inbox digest, pending items) carries "(showing N of M)".
3. **Polarity** — a static `_PROTECTIVE_FACTS` map (systemd-oomd, earlyoom, zram, swap — install-agnostic names) tags matching observations `[protective: …]` in "What I Noticed".
4. **Single draft** — the report's `OutreachRequest` sets `verbatim=True`: the grounded draft is delivered as-is (channel formatter still applies via the existing verbatim branch; no second LLM pass). Existing pipeline test already proves verbatim skips the drafter.
5. **`MORNING_REPORT.md`** — new "Numbers Are Ground Truth" rules: restate exactly, never derive totals from list sections, protective facts are never risks.

## Tests

verbatim flag on the generated request; ground-truth totals precede lists and match seeded data exactly (7 proposals → "7" + "showing 5 of 7"); follow-up count semantics (3 awaiting input / 8 pending total); protective tag rendered; prompt-contract smoke. 83 passed across the outreach suites.

## Deploy path

Repo code + prompt file — existing installs: git pull + restart; fresh installs: in-repo. No migrations.

## Sequencing note

#1109 touches `pipeline.py::_deliver` only — disjoint from this change (`morning_report.py` + the request flag). Holding merge for the sequencing call alongside #1123.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1129, opened 2026-07-18
fix(outreach): deliver composed messages verbatim — stop the LLM drafter inverting meaning

WingedGuardian:
## Problem

Genesis's outreach pipeline silently ran **pre-composed messages** through the LLM `ContentDrafter` before delivery, which could reword or invert their meaning. Live incident: a `send_and_wait` test message *"please reply with a PLAIN standalone message"* was delivered as *"E2E test PR #1090 **failed**, reply to verify"* — a hallucinated "failed" (semantic inversion) with the operative instruction dropped.

**Root cause:** `OutreachRequest.verbatim` defaults `False` and only `submit()` honored it. Every "send this message" path drafted by default. The agent MCP path also runs with `_pipeline=None`, bridging through the `pending_outreach` queue / server RPC — so the flag had to travel the transport, not just request construction.

## Fix — verbatim-by-default on delivery paths (generative content unaffected)

Enumerated the whole class by dispatch method (`submit`/`submit_urgent` = at-risk; `submit_raw*` already verbatim), not just the reported case.

- **`submit_urgent`** now honors `request.verbatim` (was a silent no-op — un-breaks `process_reaper`'s existing flag; urgent alerts can no longer be reworded).
- **Delivery chokepoints** set verbatim: `outreach_send` (in-process), the **`pending_outreach` drain** (the dominant bridge/queue path), `send_and_wait_via_pipeline` (covers in-process + dashboard route).
- **Internal machine-fact notifications** set verbatim: ego notifications, "Genesis update available/failed", health-remediation alerts, session-failure alerts, CC-update summaries, skill-proposal notices.
- **`output_router` user-question:** folds the question text (previously `topic`-only) into `context` before verbatim delivery — otherwise the question itself would be dropped — and keeps every option intact.
- **Excluded:** morning report (owned by #1124, which added its own verbatim delivery); surplus-insight path (deregistered dead code).

The `draft` opt-out was intentionally **not** added: a "send this message" tool should deliver faithfully, and the caller is already an LLM that can phrase its own text (LLM-first).

## Tests (TDD)

New/extended coverage: `submit_urgent` verbatim + non-verbatim regression; `send_and_wait` verbatim + full-message-survives-topic-truncation; queue drain verbatim on both normal & urgent routes; `outreach_send` verbatim on both routes; and the BLOCKER — reflection question preserves its full text + options under verbatim. 117 targeted tests pass.

Design pre-reviewed by genesis-architect (verdict REVISE → both BLOCKER-class gaps fixed: question-text loss, and wiring the flag through the queue/RPC transports).

Follow-up: c72c8e58

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,19 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   and a new in-flight guard suppresses a genuinely concurrent duplicate
   prompt (e.g. a double-click) so a plain APPROVE reply always resolves
   unambiguously.
+- **Messages Genesis sends you now arrive exactly as written — no silent
+  rewriting.** Notifications, reminders, and reply-and-wait prompts were
+  quietly run through an LLM "drafter" before delivery, which could reword or
+  even invert their meaning: a test message asking "please reply with a plain
+  message" went out as "…failed, reply to verify," inventing an alarming
+  status that was never there. Delivery paths (`outreach_send`, the
+  reply-and-wait tool, the queued-message drain) now deliver the composed text
+  verbatim, and the internal notification paths that relay a machine fact
+  (health-remediation alerts, "update available/failed," session-failure
+  alerts, ego notifications, surfaced reflection questions) do the same — so a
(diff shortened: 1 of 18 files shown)


WingedGuardian:

---

PR #1132, opened 2026-07-18
feat(deploy): bootstrap.sh refuses to run on a live system

WingedGuardian:
## Problem

`bootstrap.sh` is a **setup/repair tool, not a deploy path** — it re-runs installers (`curl|bash` for codebase-memory-mcp, `npm -g gitnexus`, pip), re-registers MCP servers + systemd units, self-installs systemd-oomd, edits `.bashrc`, and (via its crash-recovery block) does a hard `git reset`. Running it on a machine with a **live `genesis-server`** was the direct trigger of the 2026-07-13 eth0/load incident. The "don't run bootstrap on a live system; deploy via `update.sh`" rule lived only in session memory and `CLAUDE.md` — and `.claude/settings.local.json` even pre-allows `Bash(./scripts/bootstrap.sh)`, so the footgun was fully live.

## Fix

Make the rule **mechanical**. New sourced lib `scripts/lib/live_system_guard.sh` exposes `bootstrap_refuse_if_server_live`, which `bootstrap.sh` calls *before* any mutating step (including the crash-recovery hard reset):

- **Refuses** (exit 3, teaching message → stderr) when `genesis-server` is live.
- **Server-live detection mirrors `update.sh`'s `_ensure_server_down`** (`systemctl --user is-active` OR `pgrep -f "python -m genesis serve"`) so the two can't drift — annotated as a deliberate mirror rather than refactored into shared code (smaller blast radius).
- **Fail-open**: any detection ambiguity (unknown unit rc=4, no session D-Bus, pgrep miss) reads as *not running* → proceed. A broken detector must never block a fresh-install bootstrap, where no server exists yet.
- **Two overrides**: `--force` (human, deliberate live repair, loud warning) and `GENESIS_BOOTSTRAP_ALLOW_LIVE=1` (the sanctioned caller).

`update.sh` sets `GENESIS_BOOTSTRAP_ALLOW_LIVE=1` on its bootstrap invocation. This is load-bearing: `update.sh` stops the server itself beforehand, and its **ERR trap is armed** around the bootstrap call — so a guard refusal *inside* `update.sh` would escalate into a full update **rollback** (`git reset` to the rollback tag). The opt-out keeps the guard from ever gating the deploy path.

## Verification

- `bash -n` + `shellcheck` clean on all three scripts.
- `pytest tests/test_scripts/test_live_system_guard.py` — **7 pass**: four detection/override cases (live→refuse, `--force`→warn+proceed, opt-out→silent proceed, inactive→silent proceed), pgrep-only-live→refuse, plus two wiring guardrails (guard is sourced+called *before* crash-recovery; `update.sh`'s invocation carries the opt-out).
- **E2E against this machine's actually-running server**: full `bash scripts/bootstrap.sh` → banner, refusal, `exit 3`, zero installers run; opt-out env → silent proceed.
- Re-verified against the post-#1125 `update.sh` refactor: the server stop-flow is outside the `POST_MERGE` guards, the ERR trap (line 810) is armed before the bootstrap call (line 1033), and the env prefix lands on the pipeline head.

## Deploy

Runtime script change — reaches existing installs on the next `git pull` + `update.sh`; a fresh clone gets it from the checkout. No host/guardian path touched. Idempotent and generic — works on any install (fail-open detection means installs without systemd/pgrep are never blocked).

## Regression marker

If a future `update.sh` run ever aborts or rolls back at "Running bootstrap", the opt-out wiring broke. The `test_update_sh_sets_optout_on_bootstrap_call` guardrail makes shipping that regression near-impossible.

Resolves follow-up `1195dba4`.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -2,7 +2,10 @@
 # Genesis full machine bootstrap.
 # Run once after cloning on a new machine.
 #
-# Usage: ./scripts/bootstrap.sh
+# Usage: ./scripts/bootstrap.sh [--force]
+#
+# Refuses to run while genesis-server is live (--force overrides): bootstrap
+# is a setup/repair tool, not a deploy path — see lib/live_system_guard.sh.
 
 set -euo pipefail
 
@@ -18,6 +21,13 @@ fi
 echo "=== Genesis Bootstrap ==="
 echo "Genesis root: $GENESIS_ROOT"
 echo
+
+# ── Live-system guard — BEFORE anything mutating (incl. the crash-recovery
+# block below, whose git reset --hard is also unsafe on a live system).
+# update.sh opts out via GENESIS_BOOTSTRAP_ALLOW_LIVE=1; humans use --force.
+# shellcheck source=lib/live_system_guard.sh
+. "$SCRIPT_DIR/lib/live_system_guard.sh"
+bootstrap_refuse_if_server_live "$@" || exit 3
 
 # ── Crash recovery: check for interrupted update ─────────
 UPDATE_STATE="$HOME/.genesis/update_state.json"
--- a/scripts/lib/live_system_guard.sh
+++ b/scripts/lib/live_system_guard.sh
@@ -1,0 +1,81 @@
+# shellcheck shell=bash
+# Live-system guard for bootstrap.sh — sourced, not executed.
+#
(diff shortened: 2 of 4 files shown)


WingedGuardian (commented):

---

WingedGuardian:

---

PR #1137, opened 2026-07-18
chore(routing,db): drop keyless minimax-m25 from chains + purge infra-monitor fossil

WingedGuardian:
## What

Two small hygiene items closing out the dashboard health-honesty effort.

**WS-D — routing config truthfulness** (`config/model_routing.yaml`): the keyless `minimax-m25` provider sat as a chain entry in `27_pre_execution_assessment` (live: executor review + decomposer) and was the primary of `28_observation_sweep` (dead — no router consumers; "no live consumers" per its own comment). Removed from both chains. **Behaviour-preserving** — a keyless provider is already skipped at runtime, so this only stops the config from claiming a provider it can't use. The provider *definition* stays (now 0 chain refs → grades `dormant`, emits no alert; a key can be added later).

**WS-E — purge the infra-monitor fossil** (data migration `d0004`): `schedule_infra_monitor`'s registration + method were removed in #147 (superseded by `infra_profile_refresh` + `awareness/loop.py::_check_infra_protection_posture`), but its `job_health` row lingered. The staleness annotator measures `last_run − last_success`, which stays `0` for a job that **stopped firing entirely** — so the fossil rendered as a perpetually-`healthy` job. `d0004` purges it (idempotent; no-op on fresh installs).

Only the **code-removed** job is purged. Dormant-but-registered jobs (`build_lane_poll`, 6 refs) and disabled-module pipeline rows (`pipeline:<profile>` — the modules still exist and re-register when enabled) are deliberately left intact. The general "registered job silently stopped firing" detector needs the live APScheduler registry (unavailable to a post-boot migration) and is tracked as a follow-up, not built here.

## Verification

`d0004` test — purges only the fossil, preserves live/dormant rows, idempotent, no-op empty-state (3 pass); data-migration runner discovers it with no prefix collision (13 pass); `model_routing.yaml` parses with both chains stripped and the provider def retained; `ruff` clean.

**Deploy:** needs a `genesis-server` restart — the routing config loads at startup and the migration runs post-boot. I'll restart + verify (migration purged the row; routing loaded) as part of E2E.

Completes the dashboard health-honesty effort (PR-3 of 3).
Ledger: 8d00aa825aeb4931b05ee8bccc2cca5e

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,11 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   colour and ◆ glyph), and pending approvals no longer drag the overall
   dashboard health to "degraded" — they ride along as a note on an otherwise
   healthy system.
+- **A long-retired background job no longer haunts the job-health view.** The
+  infra-monitor job was replaced months ago, but its stale record lingered and
+  showed as a perpetually-"healthy" job (the staleness check only catches jobs
+  that run-but-fail, not ones that stopped running entirely). Its fossil record
+  is now purged on upgrade.
 - **Provisioning approvals can be retried, and never race each other.** A
   grow/limits approval prompt that timed out unanswered used to silently block
   every retry for 24 hours (the generic outreach dedup window treated the
--- a/config/model_routing.yaml
+++ b/config/model_routing.yaml
@@ -446,7 +446,6 @@ call_sites:
   27_pre_execution_assessment:
     chain:
     - glm51
-    - minimax-m25
     - openrouter-deepseek-v4-flash
     default_paid: true
     retry_profile: user_facing
@@ -455,7 +454,6 @@ call_sites:
   # (awareness/loop.py:perform_tick). Kept as historical; no live consumers.
   28_observation_sweep:
     chain:
-    - minimax-m25
     - openrouter-deepseek-v4-flash
     - glm51
     - groq-free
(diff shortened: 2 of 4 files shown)


WingedGuardian:
## First-pass review — PR #1137

### Structural review

**NOTE — Missing docstring on `migrate()`**
`src/genesis/db/data_migrations/d0004_purge_retired_job_health.py`

`verify()` has a docstring; `migrate()` does not. The module-level docstring is detailed and covers the overall behavior, but per convention public functions in migration files should each carry their own one-liner. Trivial to add if desired — e.g.:

```python
def migrate() -> dict:
    """Delete job_health rows for every job in _RETIRED_JOBS; returns purge count."""
```

No other structural findings:
- Raw SQL is a data migration — exempted ✓
- Import order correct (stdlib → local) ✓
- Test file present and covers purge, idempotency, and fresh-install no-op ✓
- Conventional commit prefix (`chore:`) ✓
- No bare `except:`, mutable defaults, `nohup`, `/tmp/`, `task_states` direct insert ✓

**CLEAN** — no secrets, credentials, IP addresses, personal emails, private hostnames, or absolute paths with usernames in added lines. PR title, body, and CHANGELOG addition contain no language referencing security incidents, credential leaks, or infrastructure exposure.

---

Overall: one NOTE (cosmetic docstring gap). No changes requested.

---

WingedGuardian:
Addressed the docstring NOTE — `migrate()` now carries a one-liner (matching `verify()`). Also rebased onto main (PR #1136 merged) and resolved the CHANGELOG overlap, keeping both the needs-action and fossil-purge entries.

PR #1139, opened 2026-07-18
feat(ego): signal-queue resilience — priority eviction, honest dedup, requeue + TTLs

WingedGuardian:
## What

Reworks the ego `SignalQueue` so the signal funnel stops silently losing work, and stamps a TTL on every producer's signals. Workstream 1 of the ego-signal-resilience spec (audit 2026-07-18); the gate-aware consumer that uses `requeue()` lands in the next PR.

## Why

A live audit traced two concrete losses to queue semantics:

- The once-daily ego morning-report signal was drained and discarded while the cycle was blocked on a pending approval — no retry until the next day's cron.
- A recurring CRITICAL error re-firing within 6h was dedup-suppressed after its first (possibly discarded) push, leaving the ego blind to a live incident.
- A full queue (maxsize 20) rejected newcomers regardless of priority, so a burst of low-priority signals could lock out a CRITICAL escalation.
- The dedup stamp was written before enqueue, so a push rejected for overflow still suppressed its own retry for 6h.

## Changes

- **Priority eviction on overflow** — prune expired first; then evict the oldest strictly-lower-priority signal for a higher-priority newcomer; equal-or-lower newcomers still rejected.
- **Dedup stamps only on successful admission.**
- **Per-category dedup windows** — escalation 1h, reactive 2h, default 6h unchanged.
- **`requeue(signals)`** — dedup-bypassing, expiry- and eviction-aware re-admission for gate-blocked cycles (consumer wiring in next PR).
- **Producer TTLs** (`expires_at` existed on `EgoSignal` but was never set): proactive tick = current interval, daily briefing 6h, goal review 24h, reactive 12h, escalation = none (critical facts don't expire).
- Internals: `asyncio.PriorityQueue` → plain list (blocking was already via the notify Event; eviction needs list access). Public API unchanged; only `cadence.py` consumes the queue (verified).

## Tests

- 13 new queue tests: eviction ordering + victim selection, equal-priority rejection, expired-prune-before-evict, dedup stamp-after-admission, per-category windows, requeue (dedup bypass / expiry drop / eviction / notify).
- 4 new producer-TTL tests in the cadence suite.
- Full signal + cadence suites: 118 passed.

## Deploy

Runtime code only — `git pull` + server restart. No schema, no config, no new autonomous behavior (signals age out *sooner*, nothing runs more).

### DIFF
--- a/src/genesis/ego/cadence.py
+++ b/src/genesis/ego/cadence.py
@@ -52,6 +52,18 @@ _RECENCY_TIERS: list[tuple[timedelta | None, int]] = [
 ]
 
 
+def _expires_in(minutes: float) -> str:
+    """ISO timestamp *minutes* from now — signal TTL helper.
+
+    TTLs keep requeued/parked signals from going stale in the queue:
+    each producer sets a lifetime matching how long its signal stays
+    actionable (a proactive tick is superseded by the next one; a
+    morning briefing is stale by afternoon). Escalations get NO TTL —
+    critical facts don't expire on their own.
+    """
+    return (datetime.now(UTC) + timedelta(minutes=minutes)).isoformat()
+
+
 def _map_priority(severity_or_priority: str) -> str:
     """Map event severity/priority strings to signal priority levels."""
     s = severity_or_priority.lower()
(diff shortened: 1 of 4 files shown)


WingedGuardian (commented):

---

WingedGuardian:
Both P2 findings fixed in `3e394b75`:

1. **Remove victims by identity** — confirmed real: `EgoSignal.__eq__` compares only `_priority_order` (every other field is `compare=False`), so `list.remove(victim)` deleted the first same-priority signal, not the selected oldest victim. Now removes via `[s for s in self._items if s is not victim]`. Added a regression test with the oldest victim appended *last* (the ordering the original test happened to avoid).

2. **Clear dedup state for evicted signals** — fixed: eviction now calls `self._seen.pop(self._dedup_key(victim), None)` so an evicted summary can re-enter within its window. Added a test re-pushing an evicted summary immediately.

Thanks — both were correct.

PR #1141, opened 2026-07-18
feat(code-intel): managed genesis-code-intel-freeze kill-switch unit

WingedGuardian:
## Problem

The 2026-07-13 load-spike postmortem (`incident-20260713-load-spike-rootcause.md`, rec #3) found the code-intel "freeze" kill-switch was an **unmanaged, hand-armed `systemd-run` transient with zero repo references** — which is precisely why *no freeze was active* going into the incident (nobody re-armed it after the prior storm). The idle runner already ships the `rc=75` "host-frozen — keep the marker" plumbing *expecting* such a switch, but nothing in the repo implemented one.

## Fix

Ship it as a managed systemd **user** unit.

**`scripts/code_intel_freeze.sh`** holds BOTH single-flight locks that gate all code-intel indexing:
- the **main-repo per-repo lock** — same 16-hex sha1-of-physical-path recipe as `code_intel_index.sh` — so a manual/hook entrypoint run for `~/genesis` exits `CODE_INTEL_INDEX_LOCK_SKIP_RC` (the runner's 75), never a false success;
- the **runner self-lock** — so an idle `genesis-code-intel.timer` tick exits immediately without processing *any* repo's marker.

Total freeze: while armed, nothing indexes and every pending marker is preserved untouched.

**`genesis-code-intel-freeze.service.template`** uses **kill-then-seal**: `ExecStartPre` stops any in-flight `code-intel-*.scope` first, so the blocking flock seals in ~a second instead of waiting out a 17-min full index (the exact emergency scenario — 07-13's freeze was armed mid-storm). Rendered by the existing bootstrap template glob but **NOT auto-enabled** (only timers are), with `[Install] WantedBy=default.target` for opt-in boot persistence.

```
systemctl --user start   genesis-code-intel-freeze    # arm now
systemctl --user enable  genesis-code-intel-freeze    # arm across reboots
systemctl --user stop    genesis-code-intel-freeze    # disarm
```

## Verification — live E2E drill on this install

Rendered the unit via the exact bootstrap `sed`, then drilled every advertised behavior against the **deployed** entrypoint (proving cross-script lock parity — the freeze locked `code-intel-4408aa696643ed00.lock`, the same file `~/genesis/scripts/lib/code_intel_index.sh` checks):

- **arm → deployed entrypoint exits `rc=75`** ("lock held"); journal shows both locks held + `ARMED`.
- **runner tick while armed → "another runner tick is in progress — exiting"**, indexed nothing.
- **disarm → both locks FREE** (non-blocking flock succeeds) — confirms the SIGTERM release path.
- **kill-then-seal → a live dummy `code-intel-drilltest.scope` went `active`→`inactive` on arm**, freeze sealed.
- **enable → boot-persistence symlink created; disable → removed.**

The pytest suite (`test_code_intel_freeze.py`, 4 tests) drives the **real** entrypoint against a live-armed freeze, so any drift in the lock-path recipe fails CI rather than silently no-op'ing the switch. It also **caught a real bug pre-merge**: the backgrounded `sleep infinity` inherits the lock fds, so a plain SIGTERM left an orphan holding the locks — now killed in the disarm trap for deterministic release. Existing guardrail suite (`test_code_intel_index.py`, 29 tests) green with the new template in-tree.

## Deploy path

Runtime script + unit-template change. Existing installs: `git pull` + `update.sh` re-renders the unit (daemon-reload); **not** enabled — an on-demand lever, docs give the arm command. Fresh clones: bootstrap. Generalizable — no install-specific values (paths via `__REPO_DIR__`/`__HOME__` placeholders; lock dir via `${GENESIS_HOME:-$HOME/.genesis}`).

## Docs

`.claude/docs/code-intelligence.md` (kill-switch section + arm/disarm), `CLAUDE.md` "Other units", and corrected the stale "host's unit" comment in `code_intel_index.sh` to point at the managed script.

Closes Cluster A sub-item (2) of follow-up `3cad9b57`. Sub-item (1) (the 2G cap) is closed separately as disproven-with-evidence (a cold full+semantic cbm index peaked 564M under the 2G cap — ~3.6x headroom).

Ledger: 4ecdb559e2804791bc3f30a0e7b5fcd8

### DIFF
--- a/.claude/docs/code-intelligence.md
+++ b/.claude/docs/code-intelligence.md
@@ -113,10 +113,27 @@ working I/O throttle here) under load and kills a run that can't make headway. A
 guardrail test (`tests/test_scripts/test_code_intel_index.py`) fails the build on
 any new raw spawn site. `CODE_INTEL_INDEX_DISABLE=1` skips indexing entirely.
 
+**Emergency kill-switch — `genesis-code-intel-freeze`.** A managed systemd USER
+unit (`scripts/code_intel_freeze.sh`) that holds BOTH single-flight locks (the
+main-repo lock and the runner self-lock), so while armed nothing indexes at all:
+the idle runner exits each tick untouched and a manual/hook entrypoint run exits
+75 ("host-frozen — keep the marker"). Rendered by bootstrap but **not** enabled —
+it is an on-demand lever. Arming stops any in-flight index scope first
+(kill-then-seal), so it engages in ~a second even mid-storm:
+
+```
+systemctl --user start   genesis-code-intel-freeze    # arm now
+systemctl --user enable  genesis-code-intel-freeze    # arm across reboots too
+systemctl --user stop    genesis-code-intel-freeze    # disarm
+```
+
+For a single index run, `CODE_INTEL_INDEX_DISABLE=1` or stopping
+`genesis-code-intel.timer` are lighter alternatives.
+
 **Do NOT call CBM's `index_repository` MCP tool for a fresh full index of
-`~/genesis`.** It runs in-process and bypasses the lock, the caps, AND the host
-kill-switch — a from-scratch full index that way is exactly what read-saturated
-the container. Queue a marker instead and let the idle runner do it:
+`~/genesis`.** It runs in-process and bypasses the lock, the caps, AND the
+freeze kill-switch — a from-scratch full index that way is exactly what
+read-saturated the container. Queue a marker instead and let the idle runner do it:
 `python3 scripts/lib/index_marker.py write --repo ~/genesis --tools both --mode fast`.
 
 ---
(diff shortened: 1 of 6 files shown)


WingedGuardian:
## Structural review

**ERROR — `scripts/code_intel_freeze.sh:52` — `/tmp` fallback violates Genesis temp-file convention**

```bash
mkdir -p "$LOCK_DIR" 2>/dev/null || LOCK_DIR="${TMPDIR:-/tmp}"
```

The fallback lands lock files in `${TMPDIR:-/tmp}`. Two problems:

1. In the systemd service context the unit doesn't set `TMPDIR`, so this always resolves to `/tmp` — the small tmpfs the rules explicitly prohibit.
2. In a CC session context `$TMPDIR` points at `~/.genesis/cc-tmp/` (the "oxygen" space the watchgod kills sessions over) — also wrong for lock files.

The Genesis convention ("Temp files: `~/tmp/`…NEVER write large files to `/tmp/`") covers all temp writes; the code-review checklist lists `/tmp/` usage as an ERROR. The fix is to fall back to `$HOME/tmp` (which every Genesis install is expected to have):

```bash
mkdir -p "$LOCK_DIR" 2>/dev/null || { mkdir -p "$HOME/tmp" 2>/dev/null; LOCK_DIR="$HOME/tmp"; }
```

---

---

PR #1142, opened 2026-07-18
test(db): guard the #1123 bootstrap-crash class — indexed migration-only columns

WingedGuardian:
## What

`create_all_tables` runs `_migrate_add_columns` and then builds `INDEXES`, while the numbered-migration runner only runs afterward. So an index in the module-level `INDEXES` list that references a column only a *numbered migration* adds crashes bootstrap on an existing DB — before that migration runs. Fresh-DB CI never sees it, because the canonical `CREATE TABLE` already carries the column. #1123 shipped exactly this; #1127 patched the single instance for `ego_directives`.

This adds the **class guard** so the next occurrence can't merge green.

## The guard (static, no DB/network, install-agnostic)

- No `INDEXES`-referenced, migration-added column may be absent from `_migrate_add_columns` — covering both `ON(...)` key columns and partial-index `WHERE`-predicate columns (both must exist when the index is built).
- Every `INDEXES`-referenced column must be creatable before the index build (present in the canonical `CREATE TABLE` **or** `_migrate_add_columns`).
- A checker self-test proves the guard catches a synthetic gap, not just that `main` is currently clean.

Green on `main` today (the last live gap was closed by #1127); it fails the moment a future PR indexes a migration-only column, with a message pointing at the fix. Plus a pointer comment at the `INDEXES` loop. No runtime behavior change.

### DIFF
--- a/src/genesis/db/schema/_migrations.py
+++ b/src/genesis/db/schema/_migrations.py
@@ -35,6 +35,13 @@ async def create_all_tables(db: aiosqlite.Connection) -> None:
     # Schema migrations BEFORE indexes — migrations add columns that indexes may reference
     await _migrate_add_columns(db)
 
+    # INVARIANT: every column an entry below references must already exist here —
+    # i.e. it is in a canonical CREATE TABLE above OR added by
+    # _migrate_add_columns. A legacy DB only has the numbered-migration runner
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

PR #1145, opened 2026-07-18
feat(ego): persist adaptive cadence across restarts + quiet-hours floor + dashboard honesty

WingedGuardian:
## What

Makes the ego's adaptive cadence survive restarts, adds a circadian quiet-hours floor, and stops the dashboard from showing a cadence that isn't real. Workstream 3 of the ego-signal-resilience spec (audit 2026-07-18).

## Why

The adaptive interval lived only in memory. Every server restart reset both egos to base cadence, and on a restart-heavy install (12/day observed) the backoff never survived — pinning the ego at base frequency regardless of activity. The boot first-fire also re-anchored to *base* cadence, actively re-accelerating a backed-off ego after each restart. And the dashboard rendered that in-memory value, so it faithfully showed a countdown that was wrong whenever the ego was actually blocked on a pending approval.

## Changes

- **Interval persistence** — `_current_interval` is written to `ego_state` (`cadence_interval:<source_tag>`) on every change and restored at `start()`, clamped to `[cadence_minutes, max_interval_minutes]` (config may have changed while down). Fresh install / missing / unparseable → base default (zero-state safe).
- **Boot anchor fix** — `_compute_boot_first_fire` now anchors to `last_success + restored interval`, not base.
- **Quiet-hours floor** — configurable overnight window (default 23:00–07:00 local) throttles **proactive ticks only** to at most one per `quiet_hours_min_interval_minutes` (default 240). Morning report, reactive, and CRITICAL escalation are never gated. New `EgoConfig` fields + `validate_ego_config` rules + documented `config/ego.yaml` defaults. Deterministic, no LLM.
- **Dashboard honesty** — cadence payload gains `gated` (a pending `autonomous_cli_fallback` approval exists for this ego) and `next_fire_at` (live APScheduler `next_run_time`); the overview ego cards show `gated — awaiting approval` in place of the countdown while blocked. New `current_interval_minutes`-sibling properties `source_tag`/`next_fire_at` on the manager; `has_pending_cli_approval` CRUD helper.
- **Bug caught in review** — `_local_now` now converts `user_timezone()`'s IANA **string** to a tzinfo; a bare string would have crashed the quiet-hours check inside `_on_tick` at runtime.

## Tests

- Interval persistence: backoff persisted + restored by a fresh manager, clamp-to-max, no-row keeps base, invalid value keeps base, boot first-fire uses restored interval.
- Quiet hours: midnight-crossing + same-day + zero-width window logic; suppress-when-recent, allow-without-prior-fire, allow-when-floor-elapsed, allow-outside-window, disabled-never-suppresses; `_on_tick` pushes nothing when suppressed (no deep-think slot consumed); morning report NOT gated.
- Config validation for all four new fields; dashboard `has_pending_cli_approval` (matches pending, ignores resolved / other ego / other action types).
- Full ego suite: 257 passed. Shared `config` test fixture set circadian-neutral so existing overnight runs stay deterministic.

## Deploy

Runtime code + additive config defaults — `git pull` + server restart. No migration (`ego_state` is an existing key-value table). Config works with zero overlay on a fresh install (quiet hours default on at 23–07).

## Concurrency note

Disjoint from #1139 (SignalQueue rework) — that PR touches the `EgoSignal` construction sites and adds a module-level `_expires_in`; this PR touches `start()`, `_update_interval`, `_compute_boot_first_fire`, the top of `_on_tick`, config/types/dashboard. Whichever merges second rebases trivially.

### DIFF
--- a/config/ego.yaml
+++ b/config/ego.yaml
@@ -38,6 +38,15 @@ shadow_morning_report: true  # shadow mode: separate topic for 2 weeks
 dispatch_model_overrides:
   investigate: opus
 
+# Quiet hours (circadian floor) — throttle PROACTIVE ego ticks overnight to at
+# most one per quiet_hours_min_interval_minutes. Morning report, reactive events,
+# and CRITICAL escalations are NEVER gated by this — a 3am incident still wakes
+# the ego. Local time uses the system timezone. Live — takes effect next tick.
+quiet_hours_enabled: true
+quiet_hours_start: 23          # local hour [0-23], inclusive
+quiet_hours_end: 7             # local hour [0-23], exclusive
+quiet_hours_min_interval_minutes: 240   # min gap between overnight proactive ticks
+
 # Circuit breaker
 consecutive_failure_limit: 3
 failure_backoff_minutes: 60  # pause after N consecutive failures
--- a/src/genesis/dashboard/routes/ego.py
+++ b/src/genesis/dashboard/routes/ego.py
@@ -62,19 +62,31 @@ async def ego_status():
     genesis_focus = await ego_crud.get_state(rt._db, "genesis_ego_focus_summary")
 
     # Cadence state from runtime managers
-    def _cadence_snapshot(mgr) -> dict:
+    async def _cadence_snapshot(mgr) -> dict:
         if mgr is None:
             return {"available": False}
-        return {
+        snap = {
             "available": True,
             "is_running": mgr.is_running,
             "is_paused": mgr.is_paused,
             "current_interval_minutes": mgr.current_interval_minutes,
             "consecutive_failures": mgr.consecutive_failures,
+            "next_fire_at": mgr.next_fire_at,
+            "gated": False,
         }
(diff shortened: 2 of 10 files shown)


WingedGuardian:

---

WingedGuardian:
Both P2s fixed in `d09d3af8` (+ fixed a pre-existing dashboard route test that CI caught):

1. **Keep recency-expanded backoff when restoring** — correct, this defeated the persistence for absent users. `_restore_interval` now clamps to `max(cadence_minutes, _recency_max_interval())` instead of the static `max_interval_minutes`, so a persisted 1440 survives when the recency tier supports it. Regression test: `test_restore_keeps_recency_expanded_backoff`.

2. **Enforce quiet-hours floor after restarts** — also correct, and it bit exactly the restart-churn scenario this PR targets. The last proactive fire time is now persisted to `ego_state` on each tick and restored in `start()`, so the floor survives restarts rather than resetting to 'allow'. Regression test: `test_last_proactive_fire_persisted_and_restored`.

Thanks — both were real.

PR #1146, opened 2026-07-19
feat(resilience): isolate cc-tmp onto a dedicated incus storage volume

WingedGuardian:
## Problem

`~/.genesis/cc-tmp` — the `TMPDIR` for every Claude Code session and for
genesis-server — shares the container root filesystem with everything else.
That is a two-way hazard: a runaway temp write can fill the container root
(and a full rootfs kills **every** CC session, since the watchdog can only
mitigate), and unrelated rootfs growth can starve cc-tmp below the watchdog's
~150 MiB sacred-ground floor and kill sessions from the other direction. It is
the one shared-rootfs blast-radius the storage-resiliency sprint didn't close.

## Fix

Move cc-tmp onto its **own size-capped incus custom storage volume**, mounted
at the same path (so zero TMPDIR/unit churn). A runaway fill is then capped at
the volume; the container root filesystem is never touched, and nothing else
can crowd cc-tmp out.

Spike-proven live (incus 6.0): a fresh custom volume mounts owned by
container-root; a `chown` from container-root makes it writable by the
container user with the correct unprivileged host-uid mapping — so
`security.shifted` (privesc advisory GHSA-56mx-8g9f-5crf) is **not** used. The
disk device hot-plugs onto a running container with no restart, and filling the
volume to ENOSPC leaves the container rootfs `df` byte-identical.

## What's in it

- **`scripts/lib/cc_tmp_volume.sh`** — `cc_tmp_volume_apply` / `_remove`,
  modeled on `host_swap.sh` (every path returns 0; degrade-to-skip; never
  aborts a caller under `set -e`). Guard ladder: disable / no-incus /
  container-down / **already-attached (idempotent)** / **live CC session**
  (won't attach over a running session — converges on a later quiet apply).
  Creates a thin volume, hot-plug attaches it, mirrors root's `limits.read/write`
  IO caps, chowns to the container user and **verifies a uid write — rolling the
  device back on failure** (an attached-but-unwritable cc-tmp is worse than
  none). Pre-split contents are shadowed under the mount, not deleted (never
  yanks an in-flight genesis-server temp file).
- **Pool-driver gate** (from review): only applies on a block/CoW backend
  (lvm/zfs/btrfs/ceph) where the isolation and size cap actually hold. On a
  `dir` pool the volume shares the rootfs and `size=` isn't enforced without
  fs-level project quotas, so it skips honestly rather than giving a cosmetic
  guarantee.
- **Rollout:** `host-setup.sh` (fresh installs + retrofit) and the guardian
  gateway `redeploy` verb (existing installs, best-effort, stderr-only, skips
  while a session is live). Lib added to `GUARDIAN_PATHS` ×3 (update.sh,
  watchdog, deploy_health).
- **Monitoring / provision-or-surface:** infra-profile `storage.cc_tmp_isolated`
  effective-state fact + an awareness posture `cc_tmp_shared_fs` alert (gated on
  `virt.container == "lxc"`, so a non-container topology is never nagged);
  guardian `_collect_disk` probes the cc-tmp mount (deduped); watchgod state
  gains `fs_free_mb`/`fs_total_mb` describing the volume.
- Docs (recovery-and-portability-workflow) + CHANGELOG; targeted tests for the
  lib, collector, posture rule, and watchgod fields.

## Generalization

No machine constants: container name resolves from `guardian.yaml` (else
`genesis`), pool from the root device, size defaults to 2 GiB
(`CCTMPVOL_SIZE_GIB`), user seam `CCTMPVOL_USER`. Requires a block/CoW pool
(the default LVM-thin setup); dir pools are detected and skipped.

## Test / E2E

100 targeted tests (lib guard/ordering/rollback/dir-gate, collector dedupe,
posture rule matrix, watchgod fields), ruff clean. **Live E2E on the real
host:** full lib lifecycle on a scratch container (attach→chown→uid-verify→
shadow→idempotent→remove); a full-volume ENOSPC left the container rootfs `df`
byte-identical; the safety guard skipped the real genesis container (live CC
session, no mutation); the dir-pool gate skipped cosmetically; and the new
collector reports `cc_tmp_isolated: False` on the box today.

## Deploy

Touches `guardian/` + gateway + a new lib in `GUARDIAN_PATHS`, so it reaches
both sides via `scripts/update.sh` (guardian redeploy). Existing installs
retrofit on the next redeploy during a CC-quiet window; this install applies via
a deliberate ~2-minute quiet-window run (close CC sessions → `cc_tmp_volume_apply`
host-side → restart genesis-server).

Ledger: 9da51d2571a54a74a98f50dc9fb67334

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,10 +20,12 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   longer starve cc-tmp below the watchdog's floor. Applied automatically on
   fresh installs (host-setup) and on existing installs (guardian redeploy, when
   no CC session is live so a running session is never disturbed); size defaults
-  to 2 GiB, override with `CCTMPVOL_SIZE_GIB`. A new infrastructure-posture
-  check raises a standing alert on any container install where cc-tmp is not yet
-  isolated, and `scripts/lib/cc_tmp_volume.sh`'s `cc_tmp_volume_remove` reverts
-  it.
+  to 2 GiB, override with `CCTMPVOL_SIZE_GIB`. Requires a block/CoW-backed
+  storage pool (lvm/zfs/btrfs/ceph — what the default install uses); a dir-backed
+  pool can't enforce the cap on its own device, so it's skipped rather than
+  giving a cosmetic guarantee. A new infrastructure-posture check raises a
+  standing alert on any container install where cc-tmp is not yet isolated, and
+  `scripts/lib/cc_tmp_volume.sh`'s `cc_tmp_volume_remove` reverts it.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,22 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ## [Unreleased]
 
+### Added
+
+- **A runaway Claude Code temp write can no longer fill the container's root
+  disk.** `~/.genesis/cc-tmp` — the scratch space every CC session (and
+  genesis-server) writes to — now lives on its own size-capped incus storage
+  volume. A runaway temp write fills only that volume; the container root
+  filesystem is never touched (a full rootfs would otherwise kill every CC
+  session). It also guards the reverse case: unrelated rootfs growth can no
+  longer starve cc-tmp below the watchdog's floor. Applied automatically on
+  fresh installs (host-setup) and on existing installs (guardian redeploy, when
+  no CC session is live so a running session is never disturbed); size defaults
+  to 2 GiB, override with `CCTMPVOL_SIZE_GIB`. A new infrastructure-posture
+  check raises a standing alert on any container install where cc-tmp is not yet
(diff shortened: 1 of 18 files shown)


WingedGuardian (commented):

---

**NOTE — Missing test coverage for `service_status.py` change**

`src/genesis/observability/service_status.py` (~line 298–299): `collect_cc_tmp_usage()` gains two new return-dict fields (`cc_fs_free_mb`, `cc_fs_total_mb`) but there is no corresponding test update in this diff. The change itself is trivial — two `.get(key, None)` pass-throughs — so existing tests won't break, but any test asserting the exact shape of `collect_cc_tmp_usage()`'s output (e.g. checking for unexpected keys, or asserting a specific key count) may silently miss the new fields. Worth confirming the suite stays green and adding a quick assertion that the new keys are present/correct when the watchgod state includes `fs_free_mb`/`fs_total_mb`.

All other modified Python files have corresponding test changes in the diff. The new shell lib (`scripts/lib/cc_tmp_volume.sh`) is fully covered by `tests/test_scripts/test_cc_tmp_volume_sh.py`.

---

PR #1150, opened 2026-07-19
feat(voice): graduation door + provenance schema + fail-closed voice API (W0)

WingedGuardian:
## Summary

W0 of the voice⇄core interface: the **graduation door** — the boundary through which typed, synthesized voice events (claims — never raw transcripts) enter the core. Ships **dark**: events land in a quarantine table with no consumer; the policy drainer that routes them into memory is a later phase (W2).

### Added
- **`graduation_events` quarantine table** — events land verbatim with `disposition='pending'`. `event_id UNIQUE` + `INSERT OR IGNORE` turns the edge outbox's at-least-once delivery into effectively-once landing. Schema lands on all three build paths (canonical `TABLES` DDL, `_migrate_add_columns` base path, numbered migration `0068`) per the base-path-parity invariant.
- **`POST /v1/voice/graduate`** — bearer-authenticated write endpoint. Validates the versioned envelope (`schema_version`, `type`, `source`, `occurred_at`, `payload`, `provenance.class`), answers `accepted` / `duplicate` / 400 `rejected`, and commits before returning 2xx (durable-before-ack transport contract). 64KB body cap measured on the actual body (chunked-encoding safe).
- **`memory_metadata` provenance columns** (`provenance_class`, `trust_level`, `attribution`, `origin_ref`, `capture_clarity`) — nullable, GROUNDWORK-tagged; written by the W2 drainer, dark until then.
- **90d retention job** (`graduation_events_prune`, CronTrigger) — prunes dispositioned rows only; pending rows are never pruned (they are the drainer's inbox).
- Envelope validator module (`channels/voice/graduation.py`) — Flask-free so the W2 drainer reuses the same vocabulary; a CI parity test pins `EVENT_TYPES` to both DDL CHECK lists (SQLite CHECKs are frozen at CREATE time).

### Changed
- **Voice API auth is now fail-closed.** Previously an unset `GENESIS_MCP_HTTP_TOKEN` left all `/v1/voice/*` routes open to the network; now they answer 503 until the token is configured, with a boot-time warning. This makes the token story uniform with the HTTP MCP transport, which already refused to start without one.

### Hardening (from review)
- `INSERT OR IGNORE` conflation guard: a non-duplicate ignore (e.g. a CHECK violation from validator/DDL drift) raises instead of silently answering `duplicate`.
- Non-object JSON bodies (array/string/number) are a 400 rejection, never a 500 — a permanently-bad body must not put the edge outbox into an infinite retry loop.

## Deploy note

⚠️ **The fail-closed flip requires `GENESIS_MCP_HTTP_TOKEN` in `secrets.env` on any install using the voice API.** Installs that already set the token (required for the HTTP MCP transport) see no behavior change. Installs without a voice pipeline are unaffected — the 503s are inert. After deploying, watch voice-caller logs for 503s (a missing bearer in a caller config shows up on the first voice interaction).

## Testing

- 122 targeted tests green across: envelope validator + route (auth, fail-closed, dedup, oversized/chunked, non-object body, runtime-not-ready), CRUD (dedup, prune invariants, constraint-violation guard), migration 0068 (idempotency, constraint enforcement, base-path parity on a legacy DB, EVENT_TYPES↔DDL parity), schema census, retention-job registry.
- Migration drill on a 255MB copy of a real database: 0068 applied in 27ms, rerun no-op, all columns/table/index present.
- E2E HTTP drill against a real Flask server + real SQLite connection: 8/8 scenarios correct (accepted, duplicate, bad-schema 400, no-auth 401, bad-bearer 401, array-body 400, chunked-oversized 400, read-route 401), exactly one `pending` row landed.

Ledger: 4b6b67e9113e4ef681198b7d5532c606

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,30 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ## [Unreleased]
 
+### Added
+
+- **Voice graduation door (W0).** The core now exposes an authenticated
+  `POST /v1/voice/graduate` endpoint where a voice edge machine can land
+  typed "graduation" events (synthesized claims from ambient/meeting
+  capture — never raw transcripts). Events are quarantined verbatim in a
+  new `graduation_events` table with idempotent delivery (safe edge
+  retries), and nothing consumes them yet — the policy drainer that routes
+  them into memory arrives in a later phase. Dispositioned events are
+  pruned after 90 days; pending events are never pruned. The memory
+  metadata schema also gains dormant provenance/trust columns
+  (`provenance_class`, `trust_level`, `attribution`, `origin_ref`,
+  `capture_clarity`) for that later phase.
+
+### Changed
+
+- **The voice API is now fail-closed.** Previously, leaving
+  `GENESIS_MCP_HTTP_TOKEN` unset left every `/v1/voice/*` route open to the
+  network. Now an unset token disables the voice API (503 + a boot-time
+  warning in the server log). **Upgrade note:** if you use the voice API,
+  set `GENESIS_MCP_HTTP_TOKEN` in `secrets.env` and make sure your Home
+  Assistant / voice-addon configs send it as a Bearer token — token-less
+  setups stop working on this upgrade.
+
 ### Fixed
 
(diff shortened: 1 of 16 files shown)


WingedGuardian:

---

PR #1151, opened 2026-07-19
fix(deploy): de-track .serena/project.yml — stop the per-deploy dirty-tree stash dance

WingedGuardian:
## Problem

`.serena/project.yml` is committed but **machine-owned**: Serena rewrites the file's comment block (the language-server list) on its own version bumps. So every install running the Serena MCP has a permanently *dirty* working tree, and `update.sh` aborts at its clean-tree gate — operators have been doing a manual `git stash`/pop dance on every deploy (three stacked "serena churn" stashes were observed across recent deploys on one install).

This is the exact same failure mode as the `.claude/settings.local.json` re-track (PR #792): a machine-owned, already-`.gitignore`d file that should never have been tracked.

## Fix

Mirror the proven `settings.local.json` transition, block-for-block:

- **De-track** `.serena/project.yml` (`git rm --cached`). `.serena/` is already in `.gitignore`, so it stays out going forward.
- **`EPHEMERAL_DIRTY_RE`** excuses the exact porcelain path (` .serena/project.yml$`) so the pre-merge dirty guard doesn't abort transitionally.
- **Premerge backup + restore**: back the live copy up outside the repo, clear local edits so the upstream deletion merges clean, then restore it as an **untracked** file — on both the normal merge path and the no-op fast path.
- Fresh clones need nothing: Serena **autogenerates** `project.yml` when it's missing (verified in the installed `serena/config` source, `ProjectConfig.load(autogen=SYNCHRONOUS)`).

## Verification

- New extracted-block tests (`tests/test_scripts/test_update_serena_transition.py`) run the **shipped** update.sh blocks against a scratch git repo: the tracked+modified→backup→upstream-delete→restore-untracked sequence (this machine's exact case), the no-merge-landed restore, the steady-state untracked no-op, the dirty-guard regex, and a guard against ever re-tracking the file. 11/11 pass (incl. the existing settings-local suite).
- `bash -n scripts/update.sh` clean.
- The real end-to-end test is the first `update.sh` run on each still-dirty install — it will print `(backed up live .serena/project.yml; cleared local edits pre-merge)` then `(restored live .serena/project.yml from pre-merge backup)` and leave a clean, untracked file.

Ledger: db2aeb9ff59d4b14bd6545fab24432b6
Follow-up: 53bb8ba175bc4566936c9cd689efc8b8


WingedGuardian:

---

PR #1154, opened 2026-07-19
fix(observability): exclude batch worklists from deferred-review sample

WingedGuardian:
## What

The dashboard **Deferred review** card renders the health snapshot's `deferred_items` sample, which selected the top-5 pending `deferred_work_queue` rows regardless of type. Scheduled batch worklists — dream-synthesis (`dream_synthesis_slice`) and entity-adjudication — drain on a daily budget and sit pending for days *by design*, so they flooded the sample and surfaced as items "needing review." The live trigger was 400 dream-synthesis slices enqueued by the weekly dream cycle showing up as `unknown • dream_synthesis_slice` in the review card.

## Fix

Exclude `BATCH_WORK_TYPES` from the `deferred_items` sample query — the same segmentation the recovery-depth alarm already applies (`resilience/deferred_work.py`). Batch items stay counted in the separate `deferred_worklist` gauge and the raw `deferred_work` total, so the `(+N worklist)` display is unaffected.

## Verification

- New test `test_deferred_items_sample_excludes_batch_worklists` seeds a 50-item batch flood + 2 recovery items and asserts the sample contains only recovery items. Confirmed RED on unfixed code, GREEN with the fix.
- All 10 tests in `test_deferred_work_segmentation.py` pass; ruff clean.
- **Live E2E** (worktree code vs. the real prod DB, read-only): with 400 dream slices pending, `deferred_items` sample went from 5 dream-slice entries → `[]`, while `deferred_work=400`, `deferred_worklist=400`, `deferred_recovery=0` stayed correct.

## Deploy path

Runtime code — `git pull` + server restart (update.sh). No schema/data migration.

### DIFF
--- a/src/genesis/observability/snapshots/queues.py
+++ b/src/genesis/observability/snapshots/queues.py
@@ -384,11 +384,24 @@ async def queues(
     deferred_items: list[dict] = []
     if db:
         try:
+            # Exclude scheduled batch worklists (dream-synthesis, entity
+            # adjudication) from the review sample — they are drained on a daily
+            # budget by design and sit pending for days, so surfacing them as
+            # "Deferred review" items is a false alarm. They stay counted in the
+            # separate `deferred_worklist` gauge. Same segmentation the
+            # recovery-depth alarm applies (see BATCH_WORK_TYPES).
+            from genesis.resilience.deferred_work import BATCH_WORK_TYPES
+
             now_dt = datetime.now(UTC)
+            placeholders = ",".join("?" * len(BATCH_WORK_TYPES))
+            # NOT-IN values are bound parameters; only the placeholder count is
+            # interpolated (from an internal constant), so no injection surface.
             cur = await db.execute(
-                """SELECT work_type, call_site_id, deferred_reason, attempts, created_at
-                   FROM deferred_work_queue WHERE status='pending'
-                   ORDER BY priority, created_at LIMIT 5"""
+                f"""SELECT work_type, call_site_id, deferred_reason, attempts, created_at
+                   FROM deferred_work_queue
+                   WHERE status='pending' AND work_type NOT IN ({placeholders})
+                   ORDER BY priority, created_at LIMIT 5""",  # noqa: S608
+                tuple(BATCH_WORK_TYPES),
             )
             for row in await cur.fetchall():
                 age = (now_dt - datetime.fromisoformat(row["created_at"])).total_seconds()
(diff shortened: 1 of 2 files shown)


WingedGuardian (commented):
## First-pass review — PR #1154

### Structural review

**WARNING — Local import inside function body** (`src/genesis/observability/snapshots/queues.py`, ~line 390)

```python
from genesis.resilience.deferred_work import BATCH_WORK_TYPES
```

This import is placed inside the function body (inside `if db: try:`) rather than at the top of the module. The common reason for this pattern in Genesis is circular import avoidance between `observability.snapshots.queues` and `resilience.deferred_work`. If that's the case, please add a one-line comment explaining the constraint, e.g.:

```python
# local import: avoids observability ↔ resilience circular import
from genesis.resilience.deferred_work import BATCH_WORK_TYPES
```

If there's no circular dependency, move the import to the module-level block with the other local imports.

---

**NOTE — Pre-existing raw SQL against genesis.db** (`src/genesis/observability/snapshots/queues.py`)

The query directly executes SQL against `deferred_work_queue` outside a `crud` module. This pattern was already present before this PR — the change only modifies the `WHERE` clause. The `# noqa: S608` suppression and the inline comment correctly explain why the f-string placeholder count is safe (internal constant, values are bound parameters). Noting it as a refactor candidate if a crud method is ever added for this query; not blocking.

---

No credentials, private IPs, personal identifiers, hardware identifiers, or security-incident language found in the diff or PR metadata.

---

WingedGuardian:
## First-pass structural review

### NOTE — Import placement

**`src/genesis/observability/snapshots/queues.py:389`**

```python
from genesis.resilience.deferred_work import BATCH_WORK_TYPES
```

This import is inside the `try:` body of `queues()` rather than at the top of the module. If the placement avoids a circular import, a brief comment saying so would help. If there's no circular dependency reason, hoisting it to the module-level import block is preferred per Genesis import conventions.

---

---

WingedGuardian:
## First-pass review

### Structural

**WARNING — Import placement: `queues.py`**

`from genesis.resilience.deferred_work import BATCH_WORK_TYPES` is placed inside the `queues()` function body rather than at module level. Per Genesis import convention, local imports should be grouped (stdlib → third-party → local) at the top of the file. If this placement is intentional to break a circular import between `observability.snapshots` and `resilience.deferred_work`, a brief inline comment explaining the circular dependency would clarify the intent and prevent a future refactor from moving it to the top and reintroducing the cycle.

**NOTE — Pre-existing raw SQL pattern (not introduced by this PR)**

The modified query in `queues()` continues a raw SQL pattern against `deferred_work_queue` that pre-exists this PR. The rule calls for genesis.db.crud modules instead of inline SQL. This PR is modifying an existing violation, not introducing one fresh — flagging for awareness only. The parameterized `NOT IN ({placeholders})` construct is correctly implemented with bound parameters; the `# noqa: S608` suppression is appropriate here.

---

---

PR #1153, opened 2026-07-19
feat(ledger): WS-2 P3 — unified calibration table (cells, MCP, dashboard, perception)

WingedGuardian:
## WS-2 P3 — the unified calibration table

Genesis now keeps an honest, mechanical scorecard of its own confidence. The cognitive ledger's graded predictions (P1/P2) roll up into `calibration_cells` — one row per (domain, action_class, metric, lane, 30/90/all-time window) — recomputed at the end of every grading pass. No new jobs, no LLM calls.

### What's in the box

- **Migration 0069**: `calibration_cells` + `calibration_cell_history` (design §4.1 DDL), mirrored in `_tables.py`, registered in the schema guard, with a full migration test (CHECK matrix, idempotency, fresh-canonical parity, down).
- **Aggregation** (`ledger/cells.py`): Murphy decomposition per cell (brier = reliability − resolution + uncertainty, fixed confidence deciles; within-bin caveat documented with citations), ECE via the existing pure `compute_ece`, Beta-binomial shrinkage cell → parent domain → global (m=10). Stated and policy_prior lanes are partitioned at grouping time — a prior can never launder into the stated cell. Per-tool success base rates from `tool_call_outcomes` land as the strict policy_prior lane (n + base_rate only).
- **Cold-start honesty everywhere**: cells are ok (n≥30) / thin / unknown, and thin/unknown **never render as a bare percentage on any surface** — escalation phrasing on the MCP tool, status labels on the dashboard, excluded entirely from perception's advisory text.
- **Write path**: upsert-then-prune (`INSERT OR REPLACE` + prune by `computed_at`) — the table is never observably empty mid-rebuild under `SerializedConnection`'s per-method locking. History snapshots prune at 180 days in the same pass (bounded store on every install).
- **Grader tail**: recompute rides the existing twice-daily pass; a failure never blocks grading — it bumps a counter surfaced as a standing `ledger:cell_recompute_failed` WARNING.
- **Surfaces**: `calibration_status` MCP tool (per-cell readable lines, top over/under-confident domains ranked on ok stated cells only), a dashboard **Calibration** tab (`/api/genesis/calibration` with lane/window filters + mechanical/fallback shares), and perception's Deep/Strategic advisory repointed from the legacy `calibration_curves` to real graded cells (byte-identical sentence contract, pinned by a parity test; ego domains excluded per design).

### Deviations from the design doc (declared)

[...]

PR #1155, opened 2026-07-19
feat(voice): s2s extraction parity — conversations feed memory extraction (W0.5)

WingedGuardian:
## Summary

W0.5 of the voice⇄core interface: **s2s conversations now feed the memory extraction pipeline** like every other channel, replacing the one-blob `memory_store` landing.

Previously an s2s voice conversation was stored as a single growing `"Voice conversation [...]"` blob in episodic memory at session close — duplicated on every replay, never mined for facts, and (from the edge bridge) re-posted cumulatively so the same conversation grew for weeks. Now each conversation lands as a per-session CC-format JSONL transcript that the existing extraction job mines incrementally by watermark: entities, references, and a topic index instead of a raw dump.

### Added
- **`VoiceTranscriptWriter`** (`channels/voice/transcript_writer.py`) — writes conversations as CC-transcript JSONL under `voice_transcript_dir()` (`~/.genesis/voice-transcripts/`, outside the repo AND outside the CC projects dir so the resume picker never lists them) plus a `cc_sessions` row (`source_tag='voice'`). Deterministic `uuid5` session ids + line-count reconciliation make replays, double-fires, and cumulative re-sends land **exactly once**. Appends are **durable before the caller is acked** (an append failure raises → route 5xx → producer retries; a swallowed error would ack turns that never landed).
- **`POST /v1/voice/conversation`** — bearer-authenticated (fail-closed via the W0 token model), accepts a cumulative `turns` list, returns `{"status":"ok","appended":N}`. 256KB body cap measured on the actual body (chunked-safe). This is the endpoint the GENesis-Voice edge bridge will target (follow-on PR).
- **Extraction eligibility** — `source_tag='voice'` added to the extractable set; the session loop falls back to `voice_transcript_dir()` for voice rows. Watermark, chunking, reference user-gating, and topic indexing all apply unchanged.
- **Daily `voice_hygiene` job** — (a) prunes voice transcripts older than **1 year** (a still-`active` conversation is never touched; rowless files fall back to mtime), (b) heals orphaned `active` voice rows idle >1h (appends refresh `last_activity_at`, so a live conversation is never flipped mid-call), (c) sweeps legacy blob memories (below).

### Changed
- **Core `S2SSessionManager`** records turns per-turn through the writer (durable; a crash loses nothing already spoken) and finalizes the transcript session at close — the one-blob `memory_store.store` landing is deleted. Its `memory_store` param becomes `transcript_writer`.
- **Voice conversation rows are excluded** from the CC-session budget (`resilience/cc_budget.py`) and the ego's capability self-model (`ego/capability_aggregator.py`) — they are conversations, not CC work sessions, and would otherwise throttle background CC or fabricate an always-100% "voice" capability domain.

### Fixed
- **Legacy voice blobs are swept from episodic memory.** A standing daily sweep (not a one-shot data migration — the migration runner never re-runs, so it couldn't catch blobs written by a not-yet-updated edge) removes memories matching the exact `"Voice conversation ["` prefix **and** the `voice`+`s2s`+`conversation` tag set, across all storage layers (`MemoryStore.delete` cascade + `memory_events`). It logs loudly on any nonzero sweep — after the edge PR lands, a hit means a stale producer is back. Legitimate voice-tagged memories without the blob prefix are spared (verified against the live cohort).
- Repaired committed `<<<<<<< / >>>>>>>` conflict markers in `CHANGELOG.md` (landed via #1153's merge) and added the missing `deferred_work_prune` case to the retention-jobs registration test.

## Deploy notes

⚠️ **Schema-migration-free** (no new tables/columns; the blob sweep is a runtime job). Existing installs: `git pull` + restart. On the primary, the first `voice_hygiene` tick (05:50) purges the historical blob cohort and the extraction cron begins mining new voice conversations within ~2h.

**Follow-on GENesis-Voice PR** switches the edge s2s bridge from `POST /api/t/memory_store` (blob) to `POST /v1/voice/conversation` (cumulative `turns`, session_id regenerated whenever its turn cache resets, existing bearer token). Until it lands, the daily sweep keeps clearing any interim blobs.

**Scope boundary:** ambient/overheard capture is untouched — it stays edge-side (48h TTL); only s2s conversations are in scope here. The fast-path `VoiceConversationHandler` remains in-memory-only (it persisted nothing before either — no regression).

## Testing

- 213 targeted tests green: transcript writer (CC round-trip through the **real** `read_transcript_messages`, cumulative idempotency, durable-ack-on-failure, idle-gated heal), `/v1/voice/conversation` route (auth/fail-closed/400/cap/append/replay/503), s2s rewire (no-blob-store, per-turn record, partial-turn flush), hygiene (prune active-safety + mtime fallback, blob-cohort sweep with cascade + false-positive survival), extraction E2E (real writer → **real `run_extraction_cycle`** via the voice-dir fallback; watermark advance + replay no-op), and the budget/capability exclusions.
- Live E2E drill (real Flask + real SQLite): append 4 → replay 0 → superset +2 → shorter 0 (warn) → 401/400 correct; transcript file 6 lines with correct alternating roles; real extraction cycle mined it (watermark 6, topic indexed) and the replay cycle reprocessed 0 chunks.

Adversarial review (Claude reviewer; Codex auth-blocked on this repo) found one real BLOCKER — the append swallowed `OSError` and still acked 200 — now fixed with a test, plus the idle-gated heal, date-in-session-id (HHMMSS alone collided across days), and partial-turn flush hardening.

Ledger: 9db340a3d255469ea5acf81bc617831e

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,21 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
-<<<<<<< HEAD
+- **Voice conversations now feed real memory extraction (W0.5).** S2S voice
+  conversations used to land in episodic memory as one growing raw blob per
+  session close — duplicated on replays, never mined for facts. They now
+  land as per-conversation transcripts (CC-transcript format, under
+  `~/.genesis/voice-transcripts/`, outside the Claude Code resume picker)
+  that the existing memory extraction job mines incrementally like every
+  other channel: facts, references, and a topic index instead of raw dumps.
+  A new authenticated `POST /v1/voice/conversation` endpoint lets a voice
+  edge machine deliver its conversation turns with replay-safe, idempotent
+  append semantics (double-fires and cumulative re-sends land exactly once,
+  and turns are durable before the endpoint acknowledges). Voice transcripts
+  are kept for 1 year (daily prune; an in-progress conversation is never
+  touched), and voice conversation rows are excluded from the Claude Code
+  session budget and the ego's capability self-model — they are
+  conversations, not work sessions.
 - **Genesis now keeps an honest, mechanical scorecard of its own confidence.**
   The cognitive ledger's graded predictions roll up into a unified calibration
   table (`calibration_cells`, migration 0069), recomputed after every grading
@@ -25,7 +39,6 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   mechanical-vs-LLM grading shares), and Deep/Strategic reflections now read
   their "when you report ~80% you're right ~60%" advisory from this real
   graded record instead of the legacy proto-calibration table.
-=======
 - **Voice graduation door (W0).** The core now exposes an authenticated
   `POST /v1/voice/graduate` endpoint where a voice edge machine can land
   typed "graduation" events (synthesized claims from ambient/meeting
@@ -37,7 +50,6 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   metadata schema also gains dormant provenance/trust columns
   (`provenance_class`, `trust_level`, `attribution`, `origin_ref`,
   `capture_clarity`) for that later phase.
->>>>>>> origin/main
 - **A runaway Claude Code temp write can no longer fill the container's root
   disk.** `~/.genesis/cc-tmp` — the scratch space every CC session (and
   genesis-server) writes to — now lives on its own size-capped incus storage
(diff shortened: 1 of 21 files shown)


WingedGuardian:
## First-pass structural review

### ERROR — Raw SQL outside crud module

**`src/genesis/channels/voice/hygiene.py:197`**

```python
await db.execute(
    "DELETE FROM memory_events WHERE memory_id = ?",
    (memory_id,),
)
```

Raw SQL DELETE against `genesis.db` must be isolated to `genesis.db.crud.*` (or migrations). The rest of the cascade correctly goes through `store.delete`; this leftover direct delete should too. The comment "the one layer keyed by memory_id that `store.delete` does not cover" is good rationale — that's exactly what a named crud function is for. Add e.g. `delete_memory_events_by_memory_id(db, memory_id)` to the memories crud module and call it here.

---

### WARNING — Raw FTS SELECT outside crud module

**`src/genesis/channels/voice/hygiene.py:184`**

```python
cursor = await db.execute(
    "SELECT memory_id, content FROM memory_fts WHERE memory_fts MATCH ?",
    (_BLOB_CANDIDATE_MATCH,),
)
```

FTS5 MATCH queries are specialized but still raw SQL outside a crud module. Consider a `search_memory_fts(db, match_expr)` function in the memories crud module so the FTS-specific syntax doesn't leak into hygiene code.

---

---

PR #1156, opened 2026-07-19
fix(telegram): self-heal shutdown-closed send client + stop senders before close

WingedGuardian:
## Problem

On 2026-07-15 the Telegram httpx **send**-client was closed while background senders kept firing, so every send raised PTB's closed-client error — `NetworkError` wrapping `RuntimeError("This HTTPXRequest is not initialized!")`. `is_safe_to_retry_send()` returns `False` for it, so the outreach recovery worker burned all 5 retries against the dead client and **permanently discarded two Sentinel approval requests** — the owner never saw them. This surfaced as the "Discarded review" card on the dashboard.

## Fix — two prongs

**BELT (self-heal).** New `send_with_client_heal(bot, send, *, stopping)` in `transport/send.py`: on a closed-client error, rebuild the send client via `bot.request.initialize()` (idempotent — `HTTPXRequest.initialize()` rebuilds only when `is_closed`, and has no skip-guard, unlike `Bot.initialize()`) and retry once — **unless** we're shutting down. Routed through every background/outbound send surface:
- `adapter_v2` `send_message` / `send_voice` / `send_document`
- `TopicManager._send_single` (all 3 direct `bot.send_message` calls)
- `DraftStreamer` draft sends

The adapter sets `_stopping` at the top of `stop()`; TopicManager/DraftStreamer take a `stopping` callable wired to it (default: never stopping). `send_document` builds a **fresh** `BytesIO` per attempt — PTB's `InputFile` drains the stream before the closed-client error is raised, so reusing one would upload zero bytes silently on the heal-retry.

**ROOT (shutdown ordering).** The outreach recovery worker was **never stopped** — absent from the runtime shutdown component list (a real latent bug). Added it there, and added `runtime.stop_outbound_senders()` which `standalone.shutdown()` now calls **before** `adapter.stop()` closes the client, so a late retry can't fire through a torn-down client. Idempotent + exception-guarded.

## Verification

- **124 targeted tests pass**; ruff clean.
- **Code review** caught a `send_document` BytesIO-reuse bug (fixed here + regression test).
- **E2E through real PTB 22.7** (`httpx.MockTransport`, no network): real `adapter.send_message` → real closed client → real closed-client error → real `initialize()` rebuild → **delivered**; shutdown path re-raises without resurrecting. Committed as `TestClientHealE2E`.

## Scope notes

- Only **background/outbound** senders are wrapped. `reply_text`/command handlers are inbound-reactive (they require a live polling client anyway) — a different failure class, not implicated in the 07-15 incident.
- Concurrent heals could momentarily build two clients (one orphaned, GC'd) — benign; not worth a lock on a rare shutdown-adjacent path.

Deploy: code-only (editable reinstall + `systemctl --user restart genesis-server`). No schema/migration.

### DIFF
--- a/src/genesis/channels/bridge.py
+++ b/src/genesis/channels/bridge.py
@@ -280,6 +280,7 @@ async def main():
             adapter._app.bot,
             config["forum_chat_id"],
             db=runtime.db,
+            stopping=lambda: getattr(adapter, "_stopping", False),
         )
         await topic_manager.load_persisted()
 
--- a/src/genesis/channels/telegram/_handler_messages.py
+++ b/src/genesis/channels/telegram/_handler_messages.py
@@ -176,6 +176,7 @@ async def _make_streamer(ctx: HandlerContext, msg, user, tid) -> DraftStreamer |
         draft_id=generate_draft_id(),
         message_thread_id=msg.message_thread_id,
         prefix=prefix,
+        stopping=lambda: getattr(ctx.adapter, "_stopping", False),
     )
 
     # Flush immediately — user sees prefix before inference starts
--- a/src/genesis/channels/telegram/adapter_v2.py
+++ b/src/genesis/channels/telegram/adapter_v2.py
@@ -38,6 +38,7 @@ from genesis.channels.telegram.transport.send import (
     safe_send_document,
     safe_send_message,
     safe_send_voice,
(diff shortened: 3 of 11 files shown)


WingedGuardian (commented):

---

PR #1157, opened 2026-07-19
fix(voice): serialize transcript writer appends to close double-fire race

WingedGuardian:
## Summary

Closes a double-write race in the W0.5 voice conversation landing (`POST /v1/voice/conversation`, #1155).

`VoiceTranscriptWriter.sync_cumulative` did a read-modify-write — count the transcript's lines, then append `turns[N:]` — with an `await` (`_ensure_registered`) **inside** the critical section. The dashboard Flask serves requests `threaded=True`, and every voice request dispatches onto the one shared runtime loop via `run_coroutine_threadsafe`. So two near-simultaneous deliveries of the same conversation (a voice edge fires the same disconnect twice within a second — observed live) could both read line-count N and both append the same delta, duplicating the whole conversation in the transcript.

## Fix

A writer-level `asyncio.Lock` held across the read-modify-write in both `sync_cumulative` and `append_message`, so the read and the append are atomic with respect to other coroutines on the shared loop.

## Testing

- New regression test drives two concurrent same-session `sync_cumulative` calls with a yield injected into `_ensure_registered`; the delta lands exactly once. **Proven discriminating** — with the lock neutralized the same scenario writes 2N lines.
- Full voice writer / route / hygiene / s2s / extraction + store-coverage guardrail suites green (85 tests).

## Context

Prerequisite for the follow-on GENesis-Voice change that switches the edge s2s bridge from the legacy blob endpoint to `POST /v1/voice/conversation` — where the double-fire actually reaches this path. Ships ahead of it so the core is protected before the edge starts delivering cumulative turns.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,15 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Voice conversation delivery no longer double-writes turns under a burst.**
+  The `POST /v1/voice/conversation` landing did a read-then-append (count the
+  transcript's lines, then write the new turns) with an `await` in the middle.
+  Because the dashboard serves requests on threads that all feed one event
+  loop, two near-simultaneous deliveries of the same conversation (a voice
+  edge can fire the same disconnect twice within a second) could both read the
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

PR #1162, opened 2026-07-20
perf(memory): vectorize + cache procedure surfacing (A1)

WingedGuardian:
## What

`_surface_procedure` (the proactive-recall procedure surfacing) ran a
per-call **pure-Python** cosine loop over every surfaceable procedure —
reading ~300 rows, unpacking each 4 KB `principle_embedding` BLOB, and computing
`cosine_similarity` (`sum(x*y ...)`, `math.sqrt`) 1024-dim × ~300 times — on the
**single runtime event loop**, on **every** prompt that hits the endpoint or the
`memory_proactive` MCP tool. Measured ~98 ms/call (DB read 21 + unpack 25 +
scalar cosine 52). This is the same class of loop-blocking hot spot as the
#1161 Jaccard fix, and it stacks under concurrency.

This replaces it with **one numpy matmul against a TTL-cached (300s),
pre-normalized matrix** of the principle embeddings:
- The ~300-row read + BLOB unpack + row normalization run **once per TTL**, not
  every call.
- The cosine becomes a single GIL-releasing matmul: **~0.3 ms/call** warm.

## Parity

Exact within floating-point tolerance — same confidence-DESC row order, same
strict-`>` tie-break, same per-tier thresholds (0.7 / 0.78), same `[:200]`
slice. Measured Δ vs the scalar cosine = **1.5e-16**.

- New `tests/test_memory/test_procedure_surface_parity.py`: randomized oracle
  (120 cases) comparing `_surface_procedure` to a verbatim copy of the old
  scalar algorithm, plus helper-level batch-vs-scalar parity and edge cases
  (empty/bad rows, zero/dim-mismatch query).
- Live-data check against the real 303 procedures: **100/100** parity vs the
  scalar reference; warm-cache **0.342 ms/call** (was ~98 ms).

## Notes

- New helpers `normalize_rows` / `cosine_similarity_batch` in
  `learning/procedural/embedding.py`; scalar `cosine_similarity` kept for the
  extractor novelty gate (unchanged).
- Cache is process-global and TTL-based (not keyed on `db`) — correct because
  the runtime passes one process-global connection; single event loop → no lock.
  On a sustained DB outage the last-good snapshot is served until recovery
  (acceptable for a soft top-1 advisory).
- Matrix kept float64 deliberately (float32 would diverge ~1e-6 and could flip a
  near-threshold surface).

Standalone recall-perf PR — the first of the hook-track work (procedure-cache
prerequisite before the proactive-hook thin-client flip). Part of follow-up
`48aa976c` (not closed by this PR).

### DIFF
--- a/src/genesis/learning/procedural/embedding.py
+++ b/src/genesis/learning/procedural/embedding.py
@@ -16,6 +16,8 @@ import math
 import struct
 from typing import TYPE_CHECKING
 
+import numpy as np
+
 if TYPE_CHECKING:
     from genesis.memory.embeddings import EmbeddingProvider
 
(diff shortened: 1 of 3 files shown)


WingedGuardian:
Addressed the P2 in `ad96af57`: the winner is now re-verified with a single live `WHERE id=? AND deprecated=0 AND quarantined=0` PK lookup before surfacing (surface-path only, fail-closed), so a procedure quarantined/deprecated within the TTL window no longer surfaces. Added a test for the quarantined-since-build case.

PR #1163, opened 2026-07-20
fix(dashboard): deploy-aware Services card — no false stale mid-update

WingedGuardian:
## Problem

While `update.sh` restarts the server, the freshly booted server's sentinel heartbeat is briefly empty (`last_heartbeat_at` unset → `age_s=9999`), so the Services card painted amber **"sentinel stale — last heartbeat Nm ago"** for the post-restart `health_check` window even though nothing was wrong. (The full server-down window is correctly invisible — no server to serve `/health`; this is specifically the post-restart settle.)

## Fix

The services snapshot now reads `env.update_in_progress()` — the **same** PID-live / phase-gated / 4h-bounded signal the watchdog already honors to defer restarts (#840) — and exposes `services.deploying`:

- **Backend** (`observability/snapshots/services.py`): when deploying, `is_stale` is suppressed and a **HEALTHY** sentinel reads `"deploying"`. Active states (`escalated`, `investigating`, `remediating`, `awaiting_*`) are **preserved, never masked**.
- **Frontend** (`servicesSemantic`): the deploy suppresses only the **transient** signals it causes (the server briefly `activating`; the empty-heartbeat stale). **Genuine faults still win** (host-framework down, watchdog restart-loop, an escalated/investigating sentinel), and the neutral `"deploying"` verdict is returned **last** — only if nothing real is wrong. This also keeps `overallHealthSemantic` honest (a real escalation still returns `error`).
- **Chip** (`ui-utils.js` `chipState`): `"deploying"` → the `info` (neutral-blue) chip + status dot (`chip--info` already exists).

## Review + verification

- **genesis-architect** (design) + **code-reviewer** (implementation) — both findings applied: reuse the central `chipState` classifier so the chip isn't amber; do not mask genuine faults during the deploy window; corrected a stale CSS comment.
- Backend E2E'd with a **real** `update_state.json` marker via a temp `GENESIS_HOME` (unmocked `update_in_progress()` → `services.deploying`); 8/8 `test_services_snapshot.py` pass; JS syntax-checked as ES modules.
- The visual chip renders at the next deploy after merge (post-restart window).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **The dashboard no longer shows a false "degraded / sentinel stale" during an
+  update.** While `update.sh` restarts the server, the freshly booted server's
+  sentinel heartbeat is briefly empty, which used to paint the Services card
+  amber ("sentinel stale — last heartbeat Nm ago") even though nothing was
+  wrong. The health snapshot now recognizes an in-progress deploy (the same
+  signal the watchdog already uses to defer restarts) and shows a neutral
+  "deploying" state for that window instead — a genuine fault (an escalated
+  sentinel, a down service) still shows through.
 - **Voice conversation delivery no longer double-writes turns under a burst.**
   The `POST /v1/voice/conversation` landing did a read-then-append (count the
   transcript's lines, then write the new turns) with an `await` in the middle.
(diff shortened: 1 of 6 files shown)


WingedGuardian:

---

WingedGuardian:
Addressed Codex P2 (`5fb628ec`): the deploy guard now suppresses only the **transient** systemd states (`activating`/`deactivating`/`reloading`) — a genuinely `failed`/`inactive` unit (e.g. update.sh's direct-serve fallback) still reads red during a deploy.

PR #1164, opened 2026-07-20
fix(autonomy): watchdog targets genesis-server, not the deprecated bridge

WingedGuardian:
## Problem

`genesis-bridge` is a **deprecated, inactive/disabled** on-demand relay. The 2026-07-02 §7 audit found recovery brains targeting it "succeed" (restarting an inactive unit) while genesis-server crash loops stay invisible — Guardian's recovery brain was moved to genesis-server and guarded by `tests/test_guardian/test_service_target.py`. **The container-side watchdog was the last recovery caller still carrying a hard genesis-bridge fallback**, and that guard did **not** cover `autonomy/watchdog.py`. If genesis-server ever fails its `is-enabled` check, the watchdog would restart the deprecated relay — the exact false-success trap.

## Fix

- `_detect_target_service()` now reuses the canonical `observability.service_status._detect_genesis_service()` (is-enabled → is-active → default-server), replacing its own inline probe + `return "genesis-bridge.service"` fallback. This **recovers a relay-only legacy install** (its own is-enabled pass returns the relay there) while **never** targeting the relay when genesis-server is present — and collapses a fourth copy of the probe.
- `restart_bridge()` default flips to `genesis-server.service` (the sole runtime caller passes `checker.target_service` explicitly, so the default is a safety net only; name kept for call-site stability).
- `test_service_target.py` gains a **separate** guard entry (own path + message) covering `autonomy/watchdog.py` so this drift can't recur silently.
- Self-healing design doc updated to the auto-detected target.

## Review + verification

- **genesis-architect** reviewed the design; the legacy-install edge was its SHOULD-FIX — resolved by reusing the canonical detector rather than hardcoding genesis-server.
- 56 targeted tests pass (`test_watchdog.py` + `test_service_target.py`); no import cycle (`service_status` never imports `autonomy`); real-detector E2E resolves to `genesis-server.service` on a normal install.

### DIFF
--- a/docs/architecture/genesis-v3-self-healing-design.md
+++ b/docs/architecture/genesis-v3-self-healing-design.md
@@ -35,9 +35,13 @@ to the repository root.
 - Exponential backoff: 10s initial, doubles per failure, capped at 300s
 - Max 5 consecutive restart attempts before `WatchdogAction.NOTIFY`
 - Pre-restart config validation: checks `secrets.env` exists with `TELEGRAM_BOT_TOKEN`, compiles bridge module for syntax errors
-- Restart via `systemctl --user restart genesis-bridge.service` — never `os.kill()`
+- Restarts the **auto-detected** target service — `_detect_target_service()` reuses
+  `observability.service_status._detect_genesis_service()` (is-enabled → is-active →
+  default), which resolves to `genesis-server.service` on a normal install and never
+  targets the deprecated relay when genesis-server is present. Via `systemctl --user
+  restart <target>` — never `os.kill()`
 - State persisted in `~/.genesis/watchdog_state.json` (survives process restarts)
-- Also checks `systemctl --user is-active genesis-bridge.service` as a fast-path before staleness check
+- Also checks `systemctl --user is-active <target>` as a fast-path before staleness check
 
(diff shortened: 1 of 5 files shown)


WingedGuardian:

---

**NOTE — doc table label slightly stale (`docs/architecture/genesis-v3-self-healing-design.md`, line ~147):**

The table row immediately below the updated "Server status stale" row still reads:

```
| Bridge process inactive | `systemctl is-active` returns non-active | Same as above | ...
```

Functionally correct — "Same as above" inherits the updated auto-detected-target remediation — but the health-signal label **"Bridge process inactive"** could be updated to "Server/target process inactive" to match the row above it. Minor doc hygiene, not a blocker.

---

WingedGuardian:

---

WingedGuardian:
Addressed Codex P2 (`4b19870d`): hardened the shared `_detect_genesis_service()` to prefer `genesis-server` whenever its unit is **installed** (via `list-unit-files`), not just enabled — closing the partial-migration edge (server present-but-disabled + relay enabled) so recovery/dashboard surface a down main service instead of masking it behind the deprecated relay. Added 3 direct detector tests. 74 targeted tests pass.

PR #1166, opened 2026-07-20
feat(learning): shadow skill-edit Critic screening self-modification pathologies

WingedGuardian:
## What

The skill-evolution loop auto-applies **MINOR** `SKILL.md` edits at autonomy≥2 past only a **structural** check (`SkillValidator`). This adds a semantic **Critic** that screens each auto-applied edit for the four failure modes of autonomous self-modification:

- **reward-hacking** (widening triggers/inflating language to game usage/success),
- **catastrophic forgetting** (dropping existing capability),
- **under-exploration** (over-narrowing to one case),
- **constraint-stripping** (removing guards / validation / scope limits — the AI-refinement degradation mode).

It runs through the shared `LLMJudgeScorer` (`judge` call site → cost tracking, chain fallback, degrade-to-NULL) against a new diff-aware rubric.

## Shadow by construction (WS1)

The Critic runs in `applicator.apply()` **after** the edit is written, wrapped in `try/except`, and only **logs** a verdict observation (`source="skill_evolution_gate"`, priority `high`=flagged / `low`=clean, 30d TTL, surfaceable during the bake). It **never blocks the edit or aborts the pipeline batch** — apply behavior is byte-identical to today apart from the added observation. Only the MINOR *auto-apply* seam is screened; MODERATE+ stays human-reviewed.

## Operator levers (autonomous-behavior gate)

- Settings domain `skill_evolution_gate` (`off | shadow`, default `shadow`, live-read, invalid→shadow).
- Env kill `GENESIS_SKILL_EVOLUTION_GATE_OFF` (hard override, checked first).

## Review focus

- `applicator.py` — the wiring is the risk-sensitive part. A pre-implementation architecture review caught a **BLOCKER**: the earlier design ran the critic *before* the file write, so an observation-DB raise could have blocked the edit and aborted the whole `needs_review` batch. Fixed: critic + log now sit **after** `record_file_modification`, fully `try/except`-wrapped.
- `skill_edit_critic.run_critic()` mirrors `surplus/quality_judge.py` — never-raises, detects the judge degrade sentinel by **`error` key presence** (not sentinel strings), caps the documents sent to the judge (removed-lines highlight computed from the **full** diff so a strip can't hide in an elided tail).

## Tests

37 new tests: rubric registration/format, `run_critic` verdict mapping + degrade-to-NULL + gating, config accessor + env kill, and the applicator **shadow invariant** (flagged verdict logged yet edit still applies; a Critic raise does **not** block the edit; `None`/gated → no observation). Existing `test_skill_evolution.py` suite unchanged and green.

## Scope

A **slice** of WS1 — not its completion. The held-out execution gate and foreground/procedure outcome attribution remain open (tabled). WS1 ledger row `2139de1a` stays open (referenced as context, not a completion marker).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,16 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **Genesis now screens its own skill self-edits for degradation.** When the
+  skill-evolution loop rewrites one of Genesis's own skill files, a new Critic
+  reads the change and flags the classic self-modification failure modes —
+  quietly dropping a safety guard or scope limit, forgetting existing
+  capability, over-narrowing the skill to a single case, or gaming its own
+  success metric. It runs in shadow first: it records a verdict (surfaced when
+  it flags something) without blocking the edit, so the pattern can be watched
+  before the screen is ever given veto power. Tunable via the new
+  `skill_evolution_gate` setting (`off` | `shadow`) with a
+  `GENESIS_SKILL_EVOLUTION_GATE_OFF` kill switch.
 - **Voice conversations now feed real memory extraction (W0.5).** S2S voice
   conversations used to land in episodic memory as one growing raw blob per
   session close — duplicated on replays, never mined for facts. They now
--- a/config/skill_evolution_gate.yaml
+++ b/config/skill_evolution_gate.yaml
@@ -1,0 +1,23 @@
+# Skill-edit Critic gate (read live by genesis.learning.skills.skill_gate_config).
+#
+# Overlay per install at ~/.genesis/config/skill_evolution_gate.local.yaml
+# (deep-merged over this base). Shipped defaults must be safe on a fresh clone
+# with no overlay.
+#
+# The Critic screens self-proposed SKILL.md edits for self-modification
+# pathologies (reward-hacking / catastrophic-forgetting / under-exploration /
+# constraint-stripping) at the skill-evolution auto-apply seam. It is a SHADOW
+# gate: it LOGS a verdict observation and NEVER blocks the edit (WS1). Read live
+# per skill-evolution pass — no restart needed. Hard kill (no restart):
+# GENESIS_SKILL_EVOLUTION_GATE_OFF=1.
+
+# Master switch. false → mode reads "off" (no Critic call, no observation).
+enabled: true
+
+# Critic mode. off | shadow.
+#   shadow (default): run the Critic and log its verdict to observations
+#     (source="skill_evolution_gate"); the edit STILL auto-applies unchanged.
+#   off: skip the Critic entirely (no LLM call, no observation).
+# There is no "enforce" mode yet — the gate never blocks an edit while it bakes.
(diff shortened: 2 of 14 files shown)


WingedGuardian:
**First-pass review — PR #1166**

Structural: 2 WARNINGs, 1 NOTE. PII/Secrets/Wording: CLEAN.

---

### ⚠️ WARNING — Fragile private-function import (`_extract_json`)

**File:** `src/genesis/learning/skills/skill_edit_critic.py` — `_parse_pathologies` (~line 96)

```python
from genesis.eval.scorers import _extract_json
```

Importing an underscore-prefixed private function from a sibling subsystem creates silent coupling. If `scorers.py` renames or inlines `_extract_json`, this `ImportError`s at runtime only, with no type-checker or linter signal. Consider either making `_extract_json` part of the public API (rename, re-export), or inline the two-line JSON-fence-strip logic directly in `_parse_pathologies`.

---

### ⚠️ WARNING — Test error-code assertions depend on undocumented LLMJudgeScorer internals

**File:** `tests/test_learning/test_skill_edit_critic.py` — `test_unavailable_on_call_failure` (line ~127) and `test_unavailable_on_parse_failure` (~138)

```python
assert critic["error"] == "judge_call_fail"
assert critic["error"] == "judge_parse_fail"
```

These literals must exactly match what `LLMJudgeScorer` emits in its `detail` JSON. If the scorer changes those codes the assertions silently rot — the tests keep passing, but the production `error` field propagated into observations is wrong. Recommend importing the constants from the scorer module (if they exist there), or at minimum adding a comment referencing where they're defined.

---

### 📝 NOTE — Wrong env-var name in settings description

**File:** `src/genesis/mcp/health/settings.py` — `skill_evolution_gate` `SettingsDomain` description (~line 870)

> `"kill via GENESIS_SKILL_GATE_OFF."`

The actual variable (defined in `genesis/env.py` and documented in `config/skill_evolution_gate.yaml`) is `GENESIS_SKILL_EVOLUTION_GATE_OFF`. This will mislead operators who read the settings description to find the kill switch.

---

---

WingedGuardian:
Thanks — all three addressed in `8a878f78`:

- **`_extract_json` coupling (WARNING):** dropped the private cross-subsystem import; `_parse_pathologies` now inlines a markdown-fence strip + `json.loads`. Pathologies are best-effort enrichment, so this stays fail-open.
- **Sentinel literals (WARNING):** tests now import and assert against `_JUDGE_CALL_FAIL` / `_JUDGE_PARSE_FAIL` from `eval.scorers`, so they track the source of truth. (Production `run_critic` was already robust — it detects the degrade by `error`-key presence, not by string match.)
- **Env-var name (NOTE):** settings description corrected to `GENESIS_SKILL_EVOLUTION_GATE_OFF`.

Also E2E-verified against the live `judge` call site: a clean additive edit → `clean` (score 1.0); a constraint-stripping edit (removed a destructive-request guard + input validation) → `flagged` (score 0.1) with `pathologies=[constraint_stripping, catastrophic_forgetting]`. Shadow invariant holds — the edit applies regardless; only a verdict observation is logged.

PR #1169, opened 2026-07-21
feat(memory): flip the proactive-memory hook to a thin client of the server engine

WingedGuardian:
## What & why

`scripts/proactive_memory_hook.py` (the Claude Code `UserPromptSubmit` hook that
injects relevant memories before each prompt) was a ~2,000-line hand-maintained
**fork** of the retrieval engine — its own Qdrant search, RRF fusion, and
formatting — that had drifted from the real engine (no reranker, entity lane, or
graph expansion). Every memory improvement had to ship twice or the hook fell
behind.

This flips it to a **thin HTTP client** of the `POST /api/genesis/hook/recall`
endpoint that shipped idle in #1159. Recall logic now lives server-side in
exactly one place; the hook posts each prompt and renders the result. **−1,090
net lines.**

## Behaviour

- **Server path (default):** POST prompt → server engine (FTS5 + vector +
  reranker + entity lane + graph expansion + injection defense + procedure
  surfacing) → render `lines`.
- **Degraded fallback:** on any server failure (refused / timeout / non-200 /
  bad JSON) the hook falls back to a clearly-labelled keyword-only FTS5 search
  (`[Memory·degraded …]`), no write-backs, and self-heals on the next prompt.
- **Modes:** `GENESIS_PROACTIVE_HOOK_MODE=server|local|off` +
  `GENESIS_PROACTIVE_HOOK_URL` (see `env.example`). `proactive_metrics.json`
  gains `mode` + `server_ms` so the fallback rate is directly observable.

## Parity + quality guards preserved

The two content-quality guards the fork had (which were **hook-only** — the
shared `memory_proactive` MCP tool never had them) are restored on the endpoint
path so the flip doesn't regress injection quality:

- **Noisy-KB filter** — only intentional-ingestion `knowledge_base` hits survive
  (the collection is otherwise majority surplus/recon crawl).
- **Garbage filter** (`provenance.is_garbage`, shared with the degraded path) —
  drops malformed rows (raw JSON observation blobs, YAML frontmatter, NULL).

The procedure funnel signal (`surfaced_count`) also moves server-side so every
profile records it in one place.

## Verification

- 620+ targeted tests green (engine, subprocess, session-awareness, hooks,
  security, dashboard); ruff clean.
- Filters validated on **real live data**: 1,500 knowledge_base rows → only the
  435 intentional-ingestion rows kept; garbage filter has zero false positives
  on real clean content.
- All four hook modes exercised as real subprocesses (stub-server, degraded
[...]

PR #1174, opened 2026-07-21
fix(memory): size the proactive-recall budget to the measured production engine

WingedGuardian:
## What & why

The thin-client flip (#1169) put the per-prompt recall engine behind a 2.0s server / 2.2s client budget. That budget was validated with a latency proof measured on an install where the pipeline's two most expensive stages were silently absent: no reranker API key (the engine's `reranked: true` meant *requested*, not *executed*), graph expansion in `shadow`, half the memory corpus, and no production load. On a loaded production install the real cold path — network embed (~0.4–1.0s) + cross-encoder rerank (~0.4s) + retrieval/enrichment over a 60k-row corpus — is ~1.5–2.5s, so **~70% of real prompts got a 503 and silently degraded to keyword-only recall**, undoing most of the engine-quality win the flip shipped.

## Changes

- **Budget sized to measurement:** server route timeout 2.0s → 4.5s; hook client budget 2.2s → 4.75s. Still well inside the hook wrapper's 10s ceiling, and no slower than the pre-flip in-process hook ever was — the work always took this long; only the cap changed at the flip.
- **Rerank timeboxed at 1.0s on the per-prompt path only** (`rerank_timeout_s` threaded `proactive_context → _proactive_impl → recall → _maybe_rerank`; expiry degrades to RRF order — the same contract as an API failure). Deep-search callers pass `None` and are byte-for-byte unchanged.
- **Honest rerank telemetry:** `engine.reranked` now reports EXECUTED; `rerank_requested` and `rerank_timed_out` added. The requested-vs-executed conflation is exactly how the original budget got validated against a no-rerank install.
- **Per-stage `timings_ms`** (recall / enrich / procedure / rerank) + one INFO journal line per slow call (>2s), so the next latency regression is diagnosable from the journal — the 503 path used to discard all timing evidence.
- **CURRENT.md**: the proactive-hook bullet still described the pre-#1169 independent-reimplementation architecture; updated to the thin-client reality.

## Verification

- 145 targeted tests green (retrieval, proactive engine, endpoint, async-route, hook subprocess, recall-inject coverage guard); ruff clean.
- New tests: rerank timebox expiry keeps RRF order + flags stats; executed-vs-requested semantics incl. API-failure path; stats flow into `engine`/`timings_ms`; timebox threading asserted at the `proactive_context` boundary.
- Live latency decomposition measured on the production install (probe session, 2026-07-21): cold 8/8 novel prompts exceeded 2.0s; warm 587–1533ms; embed ~490ms fresh-connection / ~215–330ms reused; Voyage rerank ~400ms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,21 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ## [Unreleased]
 
+### Fixed
+
+- **Per-prompt memory recall no longer silently degrades on busy installs.**
+  The server-side recall budget behind the proactive memory hook was sized
+  against a dev install that (unnoticed) ran no reranker and a half-size
+  corpus; on a loaded production install the real pipeline routinely exceeded
+  it, so most prompts fell back to keyword-only recall with a
+  `[Memory·degraded]` banner. The budget now matches the measured production
+  cold path (4.5s server / 4.75s client, still inside the hook's 10s ceiling),
+  the cross-encoder rerank stage is timeboxed at 1s (degrading to fusion order
+  rather than eating the whole budget), recall responses report whether
+  reranking actually **executed** (not merely was requested — the
+  requested-vs-executed confusion is how the old budget got validated), and
+  slow recalls log a per-stage timing breakdown to the journal.
+
 ### Added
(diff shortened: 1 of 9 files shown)


WingedGuardian:

---

PR #1180, opened 2026-07-22
fix(ego): dispatch verification — string/size checks advisory, not pass/fail

WingedGuardian:
## Problem

Post-dispatch deliverable verification hard-failed an ego proposal (status → `failed`) whenever a `required_strings` entry was absent or a file fell under `min_size_bytes`. On real data this false-failed **complete deliverables** (a writer that addressed the intent via different phrasing) — and, worse, fed a **negative learning signal** into the feedback harvester (`feedback/harvest.py` reads the `user_response` suffix: `|completed:`→positive, and a `failed` status → negative `EXECUTION_OUTCOME`). Enumerating the install's 6 `verification_failed` proposals, **4 of 6 were false negatives**.

Two compounding mechanism bugs:
- `~`/`$VAR` file paths were never expanded — `Path("~/x").exists()` is always `False`, guaranteeing a spurious "Missing file" **and** a dead fuzzy fallback.
- `required_strings` matched exact, case-sensitive substrings only.

## Change

**File existence is the only trustworthy "a real deliverable was produced" signal.** A missing OR empty (0-byte) file hard-fails; `min_size_bytes` and `required_strings` become **advisory only** — surfaced as a note in the dispatch debrief and the `|completed:` outcome, never failing the proposal and never poisoning the learning signal. A string miss is a failure of the matcher, not of the work.

- `ego/verification.py`: `VerificationResult` splits into `missing_files` (hard) + `advisories` (soft); path expansion; case-insensitive + whitespace-normalized matching.
- Fuzzy filename matches (exact path absent) still pass — no heuristic (name-similarity *or* content) reliably separates a rename from an unrelated similar file — but are **loudly flagged** as an advisory so a possibly-wrong file never passes silently.
- `cc/direct_session.py`: routing — `missing_files` → hard-fail (unchanged); advisories-only → stays `executed` (positive `|completed:` learning signal), advisories folded into the summary + debrief.
- Ego identity prompts updated to describe the advisory model (corrects the now-false "failed verification marks the proposal as failed").
- **`d0005` data-migration**: re-verifies proposals stranded at `failed` and flips those that now pass at their **exact** path. Generalizable (every install), idempotent, no-op on a fresh clone.

## Tests

TDD throughout: `tests/test_ego/test_verification.py` (advisory vs hard-fail, path expansion, empty-file, fuzzy-flag), `tests/test_cc/test_direct_session.py` (end-to-end routing: advisory keeps `executed`, missing file → `failed`), `tests/test_db/test_d0005_*` (exact-flip, fuzzy-skip, idempotent, fresh-install). Behavior-diff on the live proposal that prompted this confirms it now passes with an advisory.

## Follow-ups (separate)
- **LLM semantic verifier** — the real "did it satisfy intent / is this the right file" tool; retires the string/fuzzy heuristics.
- Verification runs before a late file write (a race observed once); re-verify at a later checkpoint.
- The `d0005` flip corrects proposal status but not the already-harvested negative `ln` outcome (unique-key won't overwrite).

### DIFF
--- a/src/genesis/cc/direct_session.py
+++ b/src/genesis/cc/direct_session.py
@@ -48,6 +48,7 @@ if TYPE_CHECKING:
     from genesis.cc import AgentProvider
     from genesis.cc.session_config import SessionConfigBuilder
     from genesis.cc.session_manager import SessionManager
+    from genesis.ego.verification import VerificationResult
 
 logger = logging.getLogger(__name__)
 
@@ -965,6 +966,7 @@ class DirectSessionRunner:
         if not request.caller_context or not request.caller_context.startswith("ego_proposal:"):
             return
         proposal_id = request.caller_context.split(":", 1)[1]
+        advisories: list[str] = []
         try:
             from genesis.db.crud.ego import (
                 mark_proposal_verification_failed,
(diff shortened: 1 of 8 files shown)


WingedGuardian:
## First-pass review — PR #1180

### Structural review

**Pattern violations (ERROR):** none

**Import hygiene (WARNING):** none — `VerificationResult` import is inside the existing `if TYPE_CHECKING:` block; d0005 import order is stdlib → local (correct)

**Missing coverage (WARNING):** none — `d0005` has `tests/test_db/test_d0005_reverify_false_dispatch_failures.py`; `verification.py` has `tests/test_ego/test_verification.py`

**Genesis conventions (WARNING):** none — commit prefix `fix(ego):` ✓; no `IntervalTrigger`, no `$CLAUDE_PROJECT_DIR`, no heavy hook imports

- `migrate()` in `src/genesis/db/data_migrations/d0005_reverify_false_dispatch_failures.py` has no function-level docstring. The module-level docstring is thorough, so the omission is minor, but `verify()` in the same file does have one for consistency. Consider adding a one-liner.

---

**Code diff (hard blocks):** none
**Code diff (soft warnings):** none — `"Confirm with Ram"` in test fixtures is example deliverable content, not an identifier
**PR title / body / CHANGELOG wording:** CLEAN — no language indicating a leak, infrastructure names, or security-incident references

---

**Overall:** Structural review: CLEAN (1 NOTE — `migrate()` docstring). PII/Secrets/Wording scan: CLEAN

---

PR #1182, opened 2026-07-22
feat(session-awareness): inline PR-watch nudge — surface steward notifications in-session

WingedGuardian:
## What

Adds a SessionStart hook that mirrors the `upstream-pr-steward` campaign's own
owner-notifications into foreground Claude Code sessions as a one-line nudge:

```
[PRs] 2 external-PR updates you may not have seen — PR steward: #905 closed (Jul 9) · … . Ask "show PRs" to review.
```

The steward campaign already notifies the owner when a tracked external PR
changes (merged/closed/new maintainer comment) and logs each ping to
`outreach_history`. Those pings are easy to miss on a chat channel, so this
surfaces the **unseen** ones inline — a second delivery surface, not a new
monitor.

## Design

- **Deterministic, read-only.** The hook reads recent steward notifications
  (`outreach_history`, category `notification`) over a read-only, WAL-aware
  connection and renders a one-line nudge. Seen-state is a home-anchored JSON
  sidecar (`~/.genesis/pr_watch/seen.json`) — a change resurfaces each session
  for `resurface_days` then stops, and the sidecar self-prunes to the
  `lookback_days` window (no retention job needed).
- **Home-anchored DB path**, not `repo_root()`-anchored — a worktree session
  would otherwise read an empty `<worktree>/data/` (the same trap the charter
  hook avoids).
- **Fail-open**: any error, missing table (fresh clone), disabled config, or
  dispatched session (`GENESIS_CC_SESSION=1`) → surfaces nothing, never blocks
  session start.
- **Operator levers**: settings domain `pr_watch` (`config/pr_watch.yaml`,
  live-read) + `GENESIS_PR_WATCH_DISABLED=1` kill switch.

## Testing

- 76 tests: core decision-table, config lever, settings validator, and an
  end-to-end subprocess test of the hook.
- Verified end-to-end through the real hook-dispatch path: first-surface →
  resurface → age-out → kill-switch → dispatched-session-skip.

## Deploy

Hook + settings land at next session start after `git pull`; no schema
migration (reads existing `outreach_history`). Fresh clones with no campaign
degrade to a clean no-op. Empty state (no sidecar, no notifications) surfaces
nothing.

Note: the campaign's discovery/notification behavior is configured in its
install-local strategy doc (campaigns ship zero defaults), so it is not part of
this diff.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -25,6 +25,16 @@
                 ]
             },
             {
+                "matcher": "startup|resume|clear",
+                "hooks": [
+                    {
+                        "type": "command",
+                        "command": "${CLAUDE_PROJECT_DIR}/.claude/hooks/genesis-hook surface_pr_updates.py",
+                        "timeout": 10
+                    }
+                ]
+            },
+            {
                 "matcher": "startup|resume|clear|compact",
                 "hooks": [
                     {
--- a/config/pr_watch.yaml
+++ b/config/pr_watch.yaml
@@ -1,0 +1,27 @@
+# PR-watch inline surface — read live by the surface_pr_updates SessionStart
+# hook (this file + ~/.genesis/config/pr_watch.local.yaml), so a
+# settings_update("pr_watch", {...}) or a hand edit takes effect at the next
+# session start — no restart.
+#
+# What it does: mirrors the upstream-pr-steward campaign's own owner
+# notifications (already in outreach_history) inline, as a one-line nudge, so a
+# tracked-PR status change you missed on Telegram still reaches you next
+# foreground session.
+#
+# Master switch: false stops the hook before any DB/sidecar work.
+# Hook-level kill switch (separate, stdlib-cheap): GENESIS_PR_WATCH_DISABLED=1
+# short-circuits the hook before it imports anything.
+enabled: true
+
+# How far back to scan outreach_history for steward notifications to surface.
+# Also the sidecar-prune horizon (entries for pings older than this fall out of
+# the scan and are dropped from the sidecar automatically).
+lookback_days: 30
+
+# A surfaced update keeps reappearing at session start until it is this many
+# days old, then it stops nagging. "Surface until you notice it, not forever."
+resurface_days: 10
+
+# Max distinct updates rendered on the nudge line; overflow collapses to
+# "+N more".
+max_surface: 5
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -472,9 +472,26 @@ The loops that make Genesis think between conversations.
 entry: ambient-cognition
 modules: [awareness, perception, reflection, attention, session_awareness,
           session_charter.py]
-verified: b662f3e3 2026-07-17
+verified: 6d79e097 2026-07-21
 ```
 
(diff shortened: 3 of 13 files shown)


WingedGuardian:
## Automated first-pass review — PR #1182

### Structural review

**ERROR (must fix)**

**Raw SQL against genesis.db — `src/genesis/session_awareness/pr_watch.py`**

`read_steward_notifications` opens genesis.db directly via `sqlite3.connect` and runs a raw `SELECT` against `outreach_history`. The convention requires use of `genesis.db.crud` modules for all genesis.db access, with the only exception being data migrations. This is a library module, not a migration.

```python
# pr_watch.py — read_steward_notifications()
conn = sqlite3.connect(f"file:{db_file}?mode=ro", uri=True, timeout=2)
rows = conn.execute(
    "SELECT id, topic, delivered_at FROM outreach_history ...",
    (_TOPIC_LIKE,),
).fetchall()
```

If the concern is keeping hook-context imports lightweight, the right fix is a thin read helper in the crud layer (or a dedicated `outreach_history` reader), not bypassing it entirely. If the raw access is a deliberate design exception, it needs an explicit call-out in the module docstring and PR body explaining why crud cannot be used here.

---

**WARNING (convention)**

**Hook imports non-stdlib dependencies — `scripts/surface_pr_updates.py`**

The hook imports `genesis.session_awareness.pr_watch` and `genesis.session_awareness.pr_watch_config` inside `main()`. `pr_watch_config` transitively imports `yaml` (third-party). The convention is hooks must be stdlib-only or `genesis.env`; `yaml` falls outside that budget.

The deferred-import pattern (inside `try`, after cheap env-var gates) is good practice and matches other hooks, but it doesn't change the transitive dependency. The fast-exit path is stdlib-clean; the slow path is not.

If hook conventions already permit `yaml` for config-reading elsewhere in the codebase, explicitly documenting that exception here would resolve the flag.

---

**CLEAN** — no secrets, private IPs, personal emails, private hostnames, or sensitive wording found in added lines, PR title, or PR body.

---

WingedGuardian:
Thanks — both addressed in `741a897c`:

**ERROR (raw SQL) — FIXED.** The `outreach_history` read moved out of the library module into a crud sync reader: `db.crud.outreach.get_notifications_by_topic_sync` (read-only `mode=ro`, WAL-safe). This mirrors the existing `db.crud.session_heartbeats` async+sync split — the sync sibling exists specifically because a SessionStart hook needs a fast synchronous read and can't drive the async runtime connection. `pr_watch.read_steward_notifications` now delegates the SQL and keeps only the tz-safe lookback windowing. Added two direct crud tests.

**WARNING (non-stdlib import in hook) — acknowledged, accepted pattern.** The cheap-exit path is stdlib-only: the `GENESIS_PR_WATCH_DISABLED` / `GENESIS_CC_SESSION` gates run before any genesis import, so a disabled/dispatched session pays no import cost. Heavy imports are deferred past those gates. `pr_watch_config` imports `yaml` exactly as `repo_pulse_config` does, and the hook now also imports `db.crud.outreach` (aiosqlite) — matching the proactive-memory hook, which imports `db.crud.session_heartbeats`. The hook always runs under the Genesis venv where these are present. No change needed.

PR #1185, opened 2026-07-22
perf(memory): defer proactive-recall side effects off the hot path (ac27b693, PR-1)

WingedGuardian:
## What & why

Follow-up **ac27b693**: under 5–7 concurrent Claude Code sessions the proactive-recall endpoint (`POST /api/genesis/hook/recall`, 4.5s budget from #1174) intermittently exceeded budget → 503 → keyword-only degrade. Live diagnosis showed the recall STAGE's variance is **shared-resource contention**, not a slow query: post-result bookkeeping serialized on the one shared SQLite connection + the sync Qdrant client on the event loop.

This is **PR-1 of a series**. It removes that bookkeeping from the latency path; PR-2 (Qdrant `to_thread`), PR-3 (RerankProvider), and PR-4 (read-connection split) follow.

## Changes

- **`recall(defer_side_effects=…)`** — proactive path only. `retrieved_count` write-backs, J-9 `recall_fired`+diagnostics, entity-lane shadow, the immunity gate emit, and the procedure `surfaced_count` bump run on background `tracked_task`s AFTER results return. Deep-search / `memory_proactive` MCP callers keep them inline (default `False`) — **byte-for-byte unchanged**, and mutually exclusive with `event_id_sink` so the MEM-003 id contract holds.
- **In-flight backstop** — past a fixed ceiling, deferral falls back to inline so a write is never dropped and the pending-task set can't grow unbounded (surfaced via a WARNING, not a silent cap).
- **Latency snapshot before deferral** — the deferred `recall_fired` records the real recall time, not the background-run time.
- **Tag co-occurrence index → stale-while-revalidate** — never rebuilt inline; a stale index kicks one single-flight background rebuild (scroll wrapped in a thread) and the current prompt proceeds with what it holds. Kills the multi-second first-prompt-after-restart stall.
- **Per-stage telemetry** — `vector`/`expand`/`fts`/`activation` ms in `timings_ms`, so a slow recall is attributable from the journal.

## Safety / correctness

- All deferred writes were already individually best-effort; deferral weakens no guarantee. `SerializedConnection` releases its lock per-method, so a shutdown cancellation between deferred writes can't poison the connection — worst case is one prompt's usage bumps.
- Injection-gate coverage guard preserved: `record_would_block` stays lexically in `_proactive_impl` (coroutine-object pattern, not a nested def).
- New Qdrant scroll site `_scan_tag_lists` reclassified in `KNOWN_QDRANT_READ_SITES` (infra; tags-only).

## Tests

- New `test_recall_deferral.py` (deferred-vs-inline write-back timing, backstop inline-fallback, sub-stage timings) and `test_tag_index_swr.py` (no inline rebuild, single-flight, time-gated count check).
- Full `tests/test_memory/` (1033), security injection-coverage, working-set hook, and recall integration smokes green locally.

## Verification plan (post-merge)

Deploy via `update.sh`, then replay the solo + 7-way concurrent probe battery and compare `timings_ms` distributions + the `proactive recall slow` journal lines before/after; watch the hook degrade-rate for 48h. Regression markers: J-9 `recall_fired`:recall ratio ~1:1, retrieved_count growth unchanged, no "Task was destroyed" at shutdown, MCP `memory_recall` latency unchanged.

Ledger: 477dc3570dc947c3ade5edcce6746151

### DIFF
--- a/.claude/docs/proactive-memory-hook.md
+++ b/.claude/docs/proactive-memory-hook.md
@@ -66,6 +66,30 @@ server path enforces. It self-heals on the next prompt once the server is back.
 `mode` (`server`/`degraded`/`local`/`off`) and `server_ms`, so the server-path
 fallback rate is directly observable. The health dashboard reads this file.
 
+The server response carries `timings_ms` with a per-stage breakdown —
+`embed`, `recall`, and (since ac27b693) the recall sub-stages `vector`,
+`expand`, `fts`, `activation`, plus `rerank`, `enrich`, `procedure`, `total`.
+When a call exceeds the slow-log threshold the server writes one
+`proactive recall slow: {…}` INFO line, so a latency regression is attributable
+to a stage from the journal alone without live probing.
+
+## Latency: work off the hot path
+
+The per-prompt path is latency-budgeted (the route's 4.5s bound), so the server
+does the least work needed to build the response and defers the rest:
+
+- **Write-backs and eval emits are deferred.** The `retrieved_count` bumps, the
+  J-9 `recall_fired` + diagnostics events, the entity-lane shadow probe, the
+  injection-gate immunity emit, and the procedure `surfaced_count` bump all run
+  on background tasks AFTER the response returns — they never affect what is
+  injected. A fixed in-flight backstop makes recall fall back to running them
+  inline (rather than piling up) if they ever drain slower than prompts arrive.
+  Deep-search recall (`memory_recall` MCP) keeps them inline.
+- **The tag co-occurrence index refreshes in the background** (stale-while-
+  revalidate): a stale index never blocks a prompt on a full-corpus scroll; the
+  current prompt uses whatever the index holds and a single background task
+  rebuilds it.
+
 ## Related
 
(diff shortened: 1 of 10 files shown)


WingedGuardian:

---

WingedGuardian:
Both P2s addressed in `0661037c`:

- **Deferred Qdrant writes off the loop** — `_record_retrievals` now runs the per-item `update_payload` loop inside `asyncio.to_thread`, so the synchronous Qdrant client no longer blocks the event loop when the deferred task runs (nor on the inline deep-search path). This is the write-back half of the PR-2 `to_thread` work, pulled forward because you're right that the deferral is incomplete for Qdrant-backed prompts without it. The vector-search read path stays in PR-2.
- **Backstop reserved synchronously** — replaced the increment-in-task-body with `_schedule_deferred_side_effects`, which increments the in-flight counter *before* `tracked_task` and releases it in the task's `finally`. A same-tick burst now sees accurate counts, so the cap holds under concurrent load. New test `test_backstop_slot_reserved_synchronously` pins it.

PR #1187, opened 2026-07-22
fix(hooks): status header reports correct model after /model switch, compaction, and resume

WingedGuardian:
## Problem

The first-reply status header `[model / effort]` derived the model from Claude Code's *"You are powered by the model named …"* environment line. Claude Code **freezes that line at original session start** — it is not regenerated on a `/model` switch or on context compaction (verified against the CC hooks/prompt-caching docs). So after a `/model` switch followed by a compaction, the header reported the *wrong* (pre-switch) model. Observed repeatedly.

## Fix

Inject the model **authoritatively** from Claude Code's SessionStart hook input instead of deriving it from the frozen env line:

- The SessionStart hook now reads the `model` field CC sends on `startup`/`compact` and emits the exact `[Display / effort]` header.
- **Precedence:** routed peer (`GENESIS_ROSTER_MODEL`) > CC hook model > env-line fallback. Consistent with the existing routed-session notice.
- **Version-drift safe:** an unknown/newer model id degrades to injecting the raw id with a mapping instruction — no stale hardcoded-table failure. A model id carrying a context-window suffix (`…[1m]`) or a date stamp is normalized.
- **Resume:** CC omits `model` on `resume`/`clear`, so the model is cached (single O(1) slot, session-id-guarded, no unbounded store) when present and replayed when absent — a `claude --resume` recovers the correct header. A concurrent-session cache miss falls through to env derivation (no regression).
- **Backward compatible:** with no `model` field (older CC), the legacy env-line directive is emitted unchanged.
- A mid-session `/model` switch fires no SessionStart, so the directive keeps the "a /model switch after this message wins" note.

`CONVERSATION.md` (Session Start + Session Control) updated to make the injected block authoritative over the env line.

## Tests

`tests/test_cc/test_session_context.py` — 15 new cases: mapper (known/suffix/date/unknown), block precedence (4 tiers), cache round-trip + session-id isolation + empty-input no-ops, and integration (stdin `model` → header, resume-from-cache, backward-compat). 28/28 pass. Reviewed by the code-reviewer agent (DONE_WITH_CONCERNS → the resume gap it flagged is now closed in this PR).

### DIFF
--- a/scripts/genesis_session_context.py
+++ b/scripts/genesis_session_context.py
@@ -105,16 +105,21 @@ def _model_display_name(model_id: str) -> str | None:
 
 
 _MODEL_CACHE_FILE = Path.home() / ".genesis" / "cc_session_model.json"
+# Bound the map so it self-evicts (no retention wiring needed). Genesis runs
+# several concurrent cc-N slot sessions, so a single slot would be clobbered by
+# whichever session started/compacted most recently — a per-session map keyed by
+# id keeps each session's model available for its own later resume.
+_MODEL_CACHE_MAX = 24
 
 
 def _cache_session_model(session_id: str, model: str) -> None:
-    """Persist the current session's model (single O(1) slot, no retention).
+    """Persist `model` under `session_id` in a bounded, self-evicting map.
 
     Written whenever CC provides `model` (startup/compact) so that a later
-    `claude --resume` — where CC OMITS `model` — can recover it. Keyed by
-    session id: the reader only trusts the cache when the id matches, so a
-    stale entry from a different session is ignored, not mis-applied.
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

WingedGuardian:
Addressed the P2 (single-slot clobbering under concurrent sessions): replaced the single slot with a session-id-keyed map bounded to 24 most-recent sessions (insertion-order self-eviction, atomic write). Concurrent sessions now each retain their own model for resume. New tests: `test_concurrent_sessions_do_not_clobber`, `test_map_self_evicts_at_cap`. 30/30 pass.

PR #1188, opened 2026-07-22
fix(deploy): update.sh robustness sweep — valid conflict JSON, bounded network, consistent DB snapshot

WingedGuardian:
## Deploy-audit P4a — `scripts/update.sh` robustness sweep

Seven robustness fixes to the self-update path. `update.sh` cannot be executed in CI (it stops services, does git operations, runs migrations), so coverage is extraction-style assertions against the shipped script text plus one functional test that runs the shipped conflict-JSON builder against hostile input.

| # | Fix |
|---|-----|
| **#2** | `update_conflicts.json` built with `json.dumps`, not a shell heredoc that only escaped `"` — a multi-line merge message or a quote-containing filename produced invalid JSON that the dashboard consumer silently swallowed. Guarded so a build failure can't trip the armed ERR trap. |
| **#3** | Network calls bounded: `git fetch` (`timeout 120`); both health curls (`--max-time 20`, deliberately **above** the health route's own ~15s budget so a slow-but-healthy check is never false-killed into a rollback); the four guardian SSH legs use `ServerAlive*` — which bounds a *dead* connection without capping a legitimately slow redeploy the way a blanket `timeout` would. |
| **#8** | Pre-update DB snapshot uses `sqlite3 .backup` (a transactionally-consistent online backup) instead of `cp` of a live WAL database, which could be torn if the server wrote mid-copy — and that snapshot is exactly what rollback later restores. |
| **#9** | A **broken** migrations module now rolls back; only a **genuinely-absent** one skips. The old guard collapsed both into a silent skip, which would run freshly-merged code against an un-migrated schema. |
| **#10** | The container-specs refresh surfaces the real error instead of reporting every failure as the benign "no profile yet". |
| **#12** | The fallback `kill -TERM` is guarded (`|| true`) so a kill/exit race can't abort the stop (and, under P4b's ERR trap, can't trigger rollback). |
| **#13** | A container Claude Code sync failure is recorded as a degraded subsystem, symmetric with the host-side pin failures. |

### Design notes (chose the correct primitive per leg, not a blanket timeout)
- **#3 curl `--max-time 20`** is intentionally larger than the health route's 15s internal timeout — a smaller value would kill a slow-but-responding health check and could roll back a *healthy* server.
- **#3 SSH `ServerAlive`** (not `timeout N`): a redeploy legitimately runs for minutes (host git+pip); `ServerAlive` only kills a truly dead connection, since sshd keeps answering keepalives during a working redeploy.
- **#8 `.backup` in place** (before the stop), not relocated after it — fixes the torn-copy bug with **zero added downtime**, which matters for installs with large databases.

### Tests
`tests/test_scripts/test_update_robustness.py` — 10 cases (extraction locks for all 7 findings + a functional hostile-input JSON test). All **48** update.sh tests pass (10 new + 38 existing); zero new shellcheck warnings vs baseline.

> Host-Deploy Gate: this touches `scripts/update.sh`, so a `scripts/update.sh` run is required after merge to deploy. Note the self-deploy lag — the run that ships this executes the OLD bytes; the new behavior governs the NEXT run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,17 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   and if any background work is ever cut short by a time limit, the result is
   flagged as incomplete rather than delivered as if it were finished.
 
+- **Updates are more resilient to network stalls, bad merges, and mid-update
+  crashes.** Several robustness fixes to the self-update path (`update.sh`):
+  network operations (fetching the latest code, post-update health checks, and
+  guardian SSH) are now time-bounded, so a hung connection can no longer stall
+  an update indefinitely; the pre-update database snapshot is now a
+  transactionally-consistent SQLite backup instead of a plain file copy that
+  could be torn if the server wrote to it mid-copy; a merge conflict now records
+  complete, valid conflict details for the assisting session (multi-line git
+  output no longer corrupts that file); and an update that ships a broken
+  database-migration module now rolls back cleanly instead of silently skipping
+  migrations and running the new code against an old schema.
 - **Host setup no longer force-deletes a container it wrongly thinks is
   damaged, or hides an install behind a new disk.** Host-side hardening: a
   container flagged "damaged" is now **renamed aside** (its database, memory, and
--- a/scripts/update.sh
+++ b/scripts/update.sh
@@ -208,7 +208,11 @@ _sync_deploy_targets() {
             # ("$OLD_COMMIT"→HEAD) silently skipped the redeploy whenever the host
             # was last deployed from a since-rebased local HEAD (a no-op run has
             # OLD_COMMIT == HEAD), stranding it on an orphan commit indefinitely.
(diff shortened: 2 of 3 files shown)


WingedGuardian:

---

WingedGuardian:
Both P2s addressed:
1. **SSH timeouts** — kept `ServerAlive*` (fast dead-link detection) and added a per-operation `timeout` for a wedged-but-keepalive-answering remote command: 30s for `version` reads, 600s for the redeploy (generous for archive+git+pip on a slow host; a killed redeploy is non-fatal/degraded), 300s for the `update` verb.
2. **Conflict-JSON rename** — moved the `mv` into an `elif` condition (ERR-trap-exempt) with the success message gated on it, and guarded the tmp `rm -f` with `|| true`, so a rename failure stays advisory-only and can't escalate the conflicts-pending path into a rollback.

Test `test_ssh_legs_bound_dead_links_and_wedged_commands` updated to assert the per-op timeouts.

PR #1184, opened 2026-07-22
fix(db): rename surplus-purge migration d0005 → d0006 (prefix collision)

WingedGuardian:
## Incident
PR #1179 added `d0005_purge_surplus_ops_telemetry`, but `d0005` was already taken by `d0005_reverify_false_dispatch_failures` (#1180). Rebasing #1179 onto main didn't surface it (distinct filenames → no git conflict), but at runtime the data-migration runner's pre-flight duplicate-prefix guard **fails closed**, so **no data migration runs** (neither the surplus purge nor the ego reverify) on any install carrying both.

Observed live: `RuntimeError: Duplicate data-migration prefix 'd0005'` in the genesis-server journal; the `data_migrations` ledger has no `d0005`/`d0006` row.

## Fix
Rename the surplus purge migration + its test to the next free prefix `d0006` (the runner derives the migration id from the filename, so the ledger id follows automatically). Pure rename — no logic change. 19 tests green (6 migration + 13 framework, incl. the duplicate-prefix guard now satisfied).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1179, opened 2026-07-22
fix(surplus): stop knowledge-base pollution by operational telemetry

WingedGuardian:
## Problem

The `knowledge_base` Qdrant collection was **71% `source_pipeline=surplus`** (1157/1699 units, live-growing since 2026-06-10) — and that "surplus" content was **Genesis's own operational telemetry**: DB-maintenance reports, disk-cleanup scans, model/J9 eval batches, backup verifications, plus research/prompt-review pipeline intermediates. It crowded out real ingested knowledge and (before the PR-B proactive filter) leaked into proactive recall.

**Root cause:** `surplus/dispatch.py::_route_insights` routed *every* non-`CODE_AUDIT` surplus task's output through `run_intake` as curated knowledge. Because surplus is a curated `IntakeSource`, intake skips scoring and writes straight to the KB. But `surplus/types.py`'s own docstring already says those action/maintenance/monitor/intermediate tasks "don't target the KB" — the intake routing simply never had that gate (the quality judge 40 lines below already did).

## Fix

- **Gate** — new `KB_ROUTING_TASK_TYPES` (insight-producing + `BOOKMARK_ENRICHMENT`, deliberately decoupled from the measurement-judge set); `_route_insights` no longer ingests non-KB-routing task output.
- **Honest provenance** — `intake.py` stopped hardcoding `source_pipeline="surplus"` for all intake. New `_pipeline_for_source` map: Genesis-authored insight tasks stay `surplus`/first-party; crawled external intelligence (`model_intelligence`/`github_landscape`/`web_monitoring`/`source_discovery`/`email_recon`) gets its own label, registered in `provenance._EXTERNAL_PIPELINES` → classified `external_untrusted` so recall wraps it as external-world content (the correct injection-defense posture; verified safe — the one structural consumer `models_md_synthesis` reads raw SQL, bypassing recall-wrapping).
- **Atomizer** — `atomize` `json_single` now renders a bare JSON object as readable prose instead of a raw `json.dumps` blob. **Scoped to `json_single` only**: the `single_item` path (model-intelligence's intentionally-JSON, machine-parsed body) and structured findings-array items (code-audit severity) are untouched.
[...]

PR #1190, opened 2026-07-22
fix(db): batch bulk data-migration writes to avoid WAL write-lock starvation

WingedGuardian:
## Problem

Post-boot data migrations run on their own connection concurrently with the live server, which permits a single WAL writer and waits only 5s (`busy_timeout`) for the lock. The `d0006` surplus purge deleted 256 units × cross-store rows — **including 256 Qdrant network calls** — inside one ~13s transaction, starving every server write and its own `mark_completed` (the ledger row stuck `running`, self-healed only on the next restart, #1179).

Verified from the live journal: a ~10s burst of `database is locked`, then a 5-minute-later idempotent re-run of the same migration.

## Fix (whole class, not just d0006)

The at-risk class is **loop-then-single-commit** migrations = {d0005, d0006}; d0002/03/04 are single set-based statements (lock held briefly) and are left as-is.

- **`_util.commit_in_batches`** — reusable helper; commits every N items so the write lock is released between batches, and sets `MIGRATION_BUSY_TIMEOUT_MS` on the migration's connection.
- **d0006 → two-phase.** Delete Qdrant points **first** (with no write connection open — network I/O never holds the write lock), then batch the fast SQLite cross-store cascade. A Qdrant-delete failure still drops the unit from phase 2, so its rows remain for the idempotent retry (no half-deleted orphan) — behaviour unchanged.
- **d0005 → read-first.** Its slow per-row file-existence checks now run on a read-only (`mode=ro`) connection, outside the write lock; the UPDATEs then batch.
- **runner** — retries `mark_completed`/`mark_failed` on a transient lock (best-effort, never raises) so a completed migration is never left `running`.

### Why batching alone wasn't enough
The first E2E showed naive batching still held the lock ~6s, because Qdrant network calls ran *inside* the open transaction. The real fix is moving slow work **out** of the write transaction (phase-separate / read-first); batching then wraps only fast writes.

## Verification

- **E2E on the real `d0006.migrate()`** (mocked 20ms Qdrant + a concurrent-writer probe): server max lock-wait **0.04s**, vs **6.15s** for the old single-transaction form.
- 67 targeted tests pass; ruff clean. Adds a **d0005 test (previously untested)** plus `commit_in_batches` and runner-retry coverage. Existing d0005/d0006 tests are unchanged (behaviour preserved).

## Docs
CHANGELOG (user-facing) + a data-migration framework note in `CURRENT.md`, and a fix to a stale `d0005`→`d0006` purge reference left over from the #1184 rename. CURRENT.md `verified:` stamps intentionally not bumped — these are a factual correction + one additive framework note, not a full-section re-verification (stamps are warn-only).

Follow-up: 2fa0bfee

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,18 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **A large one-time data cleanup no longer briefly freezes the running system.**
+  Post-startup data migrations (one-off cleanups/backfills of stored knowledge and
+  history) run alongside the live system, which allows only one writer to the
+  database at a time. A big cleanup used to do all its work in a single long write,
+  briefly blocking every other write for ~10+ seconds — long enough that the system
+  logged "database is locked" errors and, in one case, had to re-run the cleanup
+  after a restart. Bulk cleanups now save their progress in small batches, releasing
+  the database between them, and the bookkeeping that records a migration as "done"
+  now retries briefly if it hits a momentary lock — so a cleanup is never
+  needlessly repeated. (Slow per-record checks were also moved out of the locked
+  window.)
+
 - **Background sessions no longer get silently cut off after 10 minutes.** A
   long background task (for example deep research running as a background
   session) used to be killed at about 10 minutes with only a partial result and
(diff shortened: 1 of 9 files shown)


WingedGuardian:

---

WingedGuardian:
Addressed the P2 in 2cd7ff72: `_ledger_write` now returns whether the write landed, and `_run_one` reports `success=False` (with a "will replay next boot" error + ERROR log) when the `completed` ledger write is lost, instead of logging "completed"/returning success. Added a `_run_one` integration test for that path. Note: the returned outcome dict is only consumed by a fire-and-forget `tracked_task` in `runtime/_core`, so this is a log/telemetry-honesty fix, not a behavioral one — the row was already replaying via `reset_running_to_pending`.

PR #1191, opened 2026-07-22
fix(deploy): update.sh traps — errtrace, subshell guard, SIGTERM/INT rollback

WingedGuardian:
## Deploy-audit P4b — `scripts/update.sh` trap semantics

Two safety-critical trap fixes to the self-update path. Stacks on P4a (#1188, merged). `update.sh` cannot run in CI, so coverage is extraction locks **plus a functional harness that sources the actual shipped handlers** and drives the behaviors that matter.

### #1 — errtrace + subshell guard
- `set -euo pipefail` → `set -Eeuo pipefail`. This closes a real latent gap the architect confirmed empirically: a failure inside a bare helper called in the services-down window (`_write_state`, `memory_resilience_apply`, `network_resilience_apply`) previously did a plain `exit` that **never ran the ERR trap** — leaving the server stopped with no rollback. Under `-E` those now roll back correctly.
- `-E` also makes command substitutions/subshells inherit the ERR trap, so `_on_err` now short-circuits when `BASH_SUBSHELL != 0`: otherwise a failing `$(...)` would force-stop the server and `git reset` from *inside a subshell* while the parent kept running. Rollback now runs exactly once, at top level. (This guard is load-bearing: `_record_update_history`'s expected `rc=2` table-missing case fires the trap inside a `$(…)` — without the guard, that would false-rollback.)

### #14 — SIGTERM/SIGINT rollback
An interrupted deploy (Ctrl-C, systemd shutdown) previously left the server stopped with no rollback. A phase-aware signal trap now mirrors the ERR-trap lifecycle: a **pre-stop** interrupt cleans up state and exits (nothing merged, server still up); an interrupt in the **armed** window (post-stop, mid-merge/migration) rolls back like an unhandled error. All `trap - ERR` disarm sites also clear `INT`/`TERM`.

### Review
- **genesis-architect: DONE_WITH_CONCERNS — no BLOCKER, no SHOULD-FIX.** Verified bash semantics empirically; traced every armed-window call site, both spawn paths, and the cgroup interaction. NOTE-1 (a forward guardrail for P5) is landed as a code comment at the restart site: the `_apply_direct` cgroup-self-SIGTERM fix in P5 **must** be scope isolation (`systemd-run --scope`), not a handler tweak, or the armed restart would self-SIGTERM into a spurious rollback of a healthy deploy.
- Tests: `tests/test_scripts/test_update_traps.py` — 10 cases. The functional harness proves: armed function-failure rolls back; a subshell failure rolls back once at depth 0 (never from the subshell); armed SIGTERM rolls back; pre-stop SIGTERM cleans up without rollback. Full `test_scripts` suite (848) green; zero new shellcheck warnings.

> Host-Deploy Gate: touches `scripts/update.sh` → an `update.sh` run is required after merge (self-deploy lag: the shipping run executes old bytes; `-E` governs the NEXT run — I'll do a post-merge no-op run to exercise it).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,12 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **An interrupted update no longer leaves the server down.** If a self-update
+  is interrupted after it has stopped the server (a Ctrl-C, a system shutdown,
+  or an unexpected failure inside an internal step), it now rolls back to the
+  previous version and restarts the server instead of exiting with the service
+  stopped. An interrupt *before* the server is stopped simply cleans up and
+  exits, leaving the running system untouched.
 - **Background sessions no longer get silently cut off after 10 minutes.** A
   long background task (for example deep research running as a background
   session) used to be killed at about 10 minutes with only a partial result and
--- a/scripts/update.sh
+++ b/scripts/update.sh
@@ -81,7 +81,22 @@ _on_signal_prestop() {
     local sig="$1"
     trap - INT TERM
     echo "" >&2
-    echo "  Update interrupted by SIG$sig before services were stopped — nothing merged; cleaning up." >&2
+    echo "  Update interrupted by SIG$sig before the merge — restoring any stopped services and cleaning up." >&2
+    # The interrupt may have landed mid-stop (the stop polls up to ~10s), so a
+    # service may already be down. Restart exactly what was detected running —
+    # WERE_RUNNING is populated BEFORE the physical stop — so the server is never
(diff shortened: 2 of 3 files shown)


WingedGuardian:

---

WingedGuardian:
Addressed the P1 (server left down if interrupted mid-stop). Fix: detect running services into `WERE_RUNNING` **before** the physical stop, and `_on_signal_prestop` now restarts `WERE_RUNNING` before exiting (idempotent if never actually stopped; no rollback needed pre-merge). The stop-abort path disarms the signal trap so it won't restore a server we deliberately left stopped. New functional test proves a mid-stop interrupt restarts the server; extraction lock asserts detection precedes the stop. Good catch — this closed a real hole in #14's guarantee.

PR #1192, opened 2026-07-22
feat(cc): background-session delivery model — handed-off work reports back to origin thread

WingedGuardian:
## What & why

When Genesis hands a long task off to a background session and says "I'll report back," it now actually does. Previously a background task's **successful** result was saved to the DB but **never delivered** — `DirectSessionRunner._notify` hard-returns on success — so a promised "I'll get back to you" silently never arrived. This was the second half of the 2026-07-20 incident (a Telegram deep-research request that vanished): PR #1186 stopped the work from *dying*; this stops the result from *disappearing*.

It's a **general** mechanism, not research-specific: any handed-off background task can deliver its terminal outcome back to the exact conversation it came from.

## How

- **`DeliveryMode`** enum (`SILENT`/`FAILURE_ONLY`/`RESULT`) with `from_legacy()` deriving mode from the existing `notify` bools — all 8 existing `DirectSessionRequest` callers are behaviorally unchanged (none map to `RESULT`).
- **Origin capture:** `direct_session_run(deliver_to_origin=True)` reads `GENESIS_SESSION_ID` (the foreground `cc_sessions` row id the health-MCP child provably inherits) and threads it, with `delivery_mode=result`, through the queue payload onto the request. The tool is foreground-only (background disallow list + a `tool_exceptions` guard), so the origin is never a background session.
- **Framework delivery (Option B):** on a `RESULT` run, `DirectSessionRunner._deliver_result_to_origin` resolves the origin's channel+thread and delivers via the outreach chokepoint — `OutreachRequest.target_chat_id`/`target_thread_id`, honored in `pipeline._deliver` *before* the category→topic routing. DM (`chat==user id`) and forum-topic (`forum_chat_id`+thread) both work; unaddressable origins fall back to the owner surface (never a silent drop). Delivered `submit_urgent(verbatim=True)` + unique per-session topic so a promised result is never governance/dedup-suppressed or reworded. Target survives the transient-failure defer→retry.
- **Size-driven:** oversized output is saved under `~/.genesis/output/` and delivered as a summary + file pointer.
- Honest `direct_session_run` docstring/message + `CONVERSATION.md`/`background-sessions.md` guidance (was: "Results are reported via Telegram" — false for success).

## Tests

24 new tests: `DeliveryMode` derivation, `_resolve_origin_target` (DM/forum/unaddressable), `_deliver_result_to_origin` (DM/forum/failure/oversized→file/fallback/no-origin), `_run_session` wiring (RESULT delivers both terminal states; legacy stays broadcast), Option-B `_deliver` targeting (DM/forum/empty-recipient/untargeted-unchanged), and the defer→retry target round-trip. 140 existing direct-session + pipeline tests still green. genesis-architect review: DONE_WITH_CONCERNS → SHOULD-FIX + NOTES addressed in `837de464`.

## Scope / follow-up

- **Cancel-path delivery is out of scope** (a `RESULT` session cancelled by a restart records failure but doesn't deliver) — that's PR-2 (rate-limit park + auto-resume) territory, tracked under ledger `ee391fbd`.
- Live E2E (a real targeted Telegram delivery) runs post-deploy.

Ledger: d82f5101c3c149cbb76fea4dae1143ef

### DIFF
--- a/.claude/docs/background-sessions.md
+++ b/.claude/docs/background-sessions.md
@@ -16,12 +16,19 @@ Read this guide any time you're considering a background session or sub-agent.
 outlive this conversation → background session. If you just need an answer in
 the next few minutes → sub-agent.
 
-> **Note:** the background lane now owns a longer CC background-wait ceiling (set
-> to its full `timeout_s`), so a dispatched `Workflow` inside a background session
-> runs to completion instead of the CLI's default 600s truncation. Automatically
-> routing long research *from a channel turn* into this lane — with the result
-> delivered back to the requester — is a follow-up (needs the delivery model wired;
-> a background session's success output is not yet sent to the owner).
+> **Note:** the background lane owns a longer CC background-wait ceiling (set to its
+> full `timeout_s`), so a dispatched `Workflow` inside a background session runs to
+> completion instead of the CLI's default 600s truncation.
+>
+> **Origin delivery:** pass `deliver_to_origin=True` to `direct_session_run` (from a
+> channel/foreground turn) and the session's terminal outcome — success *or* failure
+> — is delivered back to the exact conversation it was dispatched from (the DM or
+> forum topic). This is how you hand off long work from a channel and actually
+> "report back." Without it, a successful background run is silent (only failures
+> raise a Telegram alert); poll `direct_session_status` for the output. Delivery is
+> framework-owned (the session need not — and for `observe`/`research` cannot — send
+> its own report); oversized output is saved under `~/.genesis/output/` and delivered
+> as a summary + file pointer.
 
 ## Profiles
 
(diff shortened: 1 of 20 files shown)


WingedGuardian (commented):

---

WingedGuardian:
**Addressed the P2 (group-chat origin misroute)** in `623b38fd`.

Root cause was exactly as flagged: `_resolve_origin_target` assumed every no-thread Telegram origin was a DM, and `cc_sessions` didn't persist the real `chat.id`. Fixed properly by persisting it:

- New `cc_sessions.chat_id` column (base create + additive migration, both build paths); `msg.chat.id` captured at the 3 Telegram dispatch sites → `get_or_create_foreground` → `cc_sessions.create`.
- `_resolve_origin_target` now **prefers the persisted `chat_id`** — correct for DM, group, *and* forum topic uniformly (a forum topic's `chat.id` is the supergroup) — with the old `user_id`/`forum_chat_id` reconstruction kept only as a legacy fallback for pre-capture rows.
- New test asserts a group-chat origin delivers to the group, not the requester's DM.

PR #1195, opened 2026-07-22
fix(deploy): update.sh flock — refuse concurrent updates (P5-A)

WingedGuardian:
## Deploy-audit P5-A — mutual exclusion for `update.sh` (finding #5)

Two `update.sh` runs could previously overlap — a CLI run plus a dashboard-triggered run, or two dashboard runs — each stopping the server and merging, corrupting the deploy. **Observed live this session** (a concurrent session ran `update.sh` mid-work; `update_history` shows the overlap window).

Fix: a whole-run **`flock -n`** on `~/.genesis/locks/update.lock`. A second run refuses immediately (exit 1) rather than queuing. Placed **after** the worktree refusal (so worktree runs never take the lock) and **before** the rollback tag / pre-update backup and the ERR/signal traps — so a contention exit leaves the running server completely untouched. The lock FD is held for the whole run and auto-released on every exit path.

This is **PR-A of a 2-PR split** the architect recommended: the flock is the highest-value, lowest-risk piece and ships first. PR-B (dashboard cgroup isolation + guard/GC honoring CLI runs + owner-checked marker deletes, with the Part-1↔Part-4 marker-leak fix) follows.

### Review
Architect **design review: GO** on the flock (placement, FD survival across the merge/subshells, auto-release on all exit paths, contention-exit-is-server-safe all validated).

### Tests
`tests/test_scripts/test_update_mutex.py` — extraction locks (guard present + correct placement) plus a functional test proving the flock excludes a second holder and frees on release. 34 update.sh tests green; zero new shellcheck warnings.

> Host-Deploy Gate: touches `scripts/update.sh` → an `update.sh` run required after merge (self-deploy lag; the flock governs the next run).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,10 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   needlessly repeated. (Slow per-record checks were also moved out of the locked
   window.)
 
+- **Two updates can no longer run at the same time.** If an update is already in
+  progress, a second `update.sh` (from another session, or the dashboard) now
+  refuses immediately instead of running concurrently — previously two updates
+  could overlap, each stopping the server and merging, and corrupt the deploy.
 - **An interrupted update no longer leaves the server down.** If a self-update
   is interrupted after it has stopped the server (a Ctrl-C, a system shutdown,
   or an unexpected failure inside an internal step), it now rolls back to the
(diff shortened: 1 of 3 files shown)


WingedGuardian (commented):

---

WingedGuardian:
Fixed: the nohup fallback in `_start_genesis_server` now closes the lock FD for the degraded server child (`{_UPDATE_LOCK_FD}>&-`), so it can't hold the advisory lock after update.sh exits. Verified functionally (child sees the FD closed, parent keeps it). Regression test added. Good catch — that was a real self-deadlock on the degraded path.

PR #1196, opened 2026-07-22
fix(deploy): dashboard cgroup isolation + honor CLI runs + owner-checked marker (P5-B)

WingedGuardian:
## Deploy-audit P5-B (parts 1/3/4) — stacks on P5-A (#1195)

Architect **design-reviewed** (validated the 4-part design; BLOCKER-1 addressed, SHOULD-FIX-1 tabled `bc7313b1`). Closes #4 (dashboard state-GC destroying a CLI run's state) and #5's dashboard side, plus NOTE-1 (`_apply_direct` cgroup self-SIGTERM).

- **Part 1 (cgroup / NOTE-1):** `_apply_direct` spawns `update.sh` via `systemd-run --user --scope` (own cgroup, mirrors the supervised orchestrator), so update.sh's own `systemctl stop genesis-server` can't SIGTERM it — which otherwise aborts the update or self-SIGTERMs into a spurious rollback at the final restart. Falls back to `start_new_session` if systemd-run is absent.
- **Part 3 (#4/#5):** all four dashboard "update in progress" checks (apply guard, dismiss, resolve, status/GC) now use the canonical `env.update_in_progress()` (marker AND state file) instead of a PID-file-only probe — so a CLI `update.sh` run is seen (409'd, not double-launched) and its crash-recovery state is not GC'd.
- **Part 4 (marker ownership):** update.sh's 5 marker deletes go through `_clear_deploy_state`, which removes the marker only if we own it (`$$`) OR its holder is dead — never stripping a live foreign holder (the supervised orchestrator across tiers, or a concurrent restore's watchdog inhibitor). The state file is always cleared.

### Build-time finding (worth reviewer attention)
I verified `systemd-run --scope` **stays alive** as the scope parent for the run's duration (the architect's design review assumed it was ephemeral). This refines — doesn't break — the marker cleanup: on the direct path the marker holds a *live* systemd-run PID during the run and a *dead* one after, read correctly by `env.update_in_progress()`, and cleaned by the next run's owner-check dead-clause. I also extended Part 3 to the `dismiss`/`resolve` guards for consistency (same CLI-blindness).

### Tests
Functional marker-ownership (drives the shipped `_clear_deploy_state` across owned/live-foreign/dead/absent), `_apply_direct` systemd-run + fallback, guard-wiring assertions, and updated trap/phase-order tests for the `_clear_deploy_state` refactor. 39 update.sh + dashboard tests green; `updates.py` reconstructed to avoid wholesale format churn (base wasn't ruff-format-clean; CI lints only).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Two updates can no longer run at the same time.** If an update is already in
+  progress, a second `update.sh` (from another session, or the dashboard) now
+  refuses immediately instead of running concurrently — previously two updates
+  could overlap, each stopping the server and merging, and corrupt the deploy.
 - **An interrupted update no longer leaves the server down.** If a self-update
   is interrupted after it has stopped the server (a Ctrl-C, a system shutdown,
   or an unexpected failure inside an internal step), it now rolls back to the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **The dashboard now recognizes updates started from the command line.** Its
+  "update in progress" checks previously only saw dashboard-triggered updates,
+  so a command-line `update.sh` run could be interrupted (its state wiped) or a
+  second update launched over it. The dashboard now consults the same
+  deploy-in-progress signal the rest of the system uses, and a dashboard-started
+  update no longer runs in the server's own service group (where the update
+  stopping the server could kill the update itself).
 - **Two updates can no longer run at the same time.** If an update is already in
   progress, a second `update.sh` (from another session, or the dashboard) now
   refuses immediately instead of running concurrently — previously two updates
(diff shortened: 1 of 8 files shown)


WingedGuardian:

---

WingedGuardian:
Addressed SHOULD-FIX-1 (direct-path marker leak). Fix: right after the flock, update.sh adopts the marker (writes its own `$$`) **only when the marker holds `$PPID`** — which on the direct path is systemd-run's PID (the dashboard wrote it). `_clear_deploy_state` then cleans it per-run via the owner (`==$$`) clause, eliminating the leak + PID-reuse tail. Verified the `$PPID` gate is safe for the supervised path: the orchestrator runs update.sh via a **CC session**, so update.sh's `$PPID` is that session, not the orchestrator — the orchestrator's marker is never adopted. Added an end-to-end test (marker==$PPID → adopted → cleaned → gone). Thanks — this closed the flip-side of the systemd-run-stays-alive finding.

WingedGuardian:
Both P2s addressed:
- **@276 (systemd-run scope-failure fallback):** added `_systemd_run_scope_available()` — probes with a no-op `systemd-run --user --scope -- true` and gates the direct spawn on it, so a present-but-unusable systemd-run (D-Bus unreachable) falls back to `start_new_session` deterministically instead of silently no-op'ing. Tests cover the probe + both branches.
- **@297 (direct systemd-run marker leak):** already fixed by the prior commit (23d3233f, marker adoption gated on `$PPID`) — your review was on the commit before it. update.sh now adopts the marker (writes its own `$$`) when it holds `$PPID` (systemd-run's PID on the direct path), so `_clear_deploy_state` cleans it per-run.

PR #1197, opened 2026-07-22
fix(memory): route post-recall enrich/breadcrumbs onto the read pool (ac27b693, PR-4b)

WingedGuardian:
## Problem
PR-4 (#1189) gave recall's **internal** read stages a dedicated `mode=ro` connection pool, and the live 7× probe confirmed they flattened (`activation` ~1055ms → 15ms). But the proactive engine runs two more reads **on the shared write connection AFTER `recall()` returns** — `_enrich` (backfills `_created_at`/`_wing` from `memory_metadata`) and `_breadcrumbs` (neighbor ids from `memory_links`). They were never inside `recall()`, so PR-4 didn't reach them. With recall's reads relieved, that post-recall pair became the **dominant residual** under load (enrich median ~1s, p95 ~1.7s).

`EXPLAIN QUERY PLAN` shows both are **indexed seeks** (`memory_metadata` PK / `idx_memory_links_source`), so the ~1s under concurrency is **lock-wait queuing behind server writes**, not query cost — the exact class PR-4 fixed for `activation`.

## Fix
Route both through PR-4's existing, tested `HybridRetriever._ro_read` seam (2 call sites in `memory/proactive.py::proactive_context`):

```python
await retriever._ro_read(_enrich, dicts)
await retriever._ro_read(_breadcrumbs, dicts)
```

`retriever` is `genesis.mcp.memory._retriever` — the MCP-module instance the proactive path uses and the one the pool was wired into in PR-4. `_ro_read` runs the read on a pooled `mode=ro` connection and **falls back to the shared connection on any pool miss/error**, so this is never worse than the pre-pool path. `mode=ro` sees committed state; no post-recall read depends on an uncommitted same-request write (`_bump_surfaced` runs after and is deferred).

**Deliberately out of scope:** `_surface_procedure` stays on the shared connection — its DB reads are a rare TTL-gated cache rebuild + a tiny PK verify (marginal contention), and wrapping it would hold a pooled slot across `cosine_similarity_batch` CPU. Revisit only if E2E shows procedure is a live residual.

## Tests
- **Functional** (`test_recall_read_pool.py`): a real `ReadConnectionPool` over a seeded temp db → asserts `_ro_read(_enrich, …)` backfills `_created_at`/`_wing` and `_ro_read(_breadcrumbs, …)` attaches the right `related_ids` off a pooled RO connection.
- **Wiring guard** (`test_proactive_engine.py`): drives `proactive_context` and asserts both `_enrich` and `_breadcrumbs` route through `_ro_read` — regression guard against reverting to bare `_enrich(db, …)`.
- The engine test's `_FakeRetriever` was taught the no-pool `_ro_read` seam so all pre-existing engine tests stay green. `pytest` → **40 passed**, ruff clean.

## Verification plan
Post-deploy: re-run the solo/4×/**7×** concurrent probe and compare `timings_ms.enrich` before/after — expect it to flatten like `activation` did, 0 503s maintained. **Falsifiable:** if enrich stays >500ms at 7×, first suspect pool-queue-wait (4 slots vs 7×) → bump `GENESIS_RECALL_READ_POOL_SIZE` (env, no code); only if a larger pool doesn't help is the lock-wait theory wrong.

Follow-up: f8a99778 (filed for this work).

### DIFF
--- a/src/genesis/memory/proactive.py
+++ b/src/genesis/memory/proactive.py
@@ -613,8 +613,16 @@ async def proactive_context(
     recall_ms = (time.monotonic() - t_recall) * 1000
 
     t_enrich = time.monotonic()
-    await _enrich(db, dicts)
-    await _breadcrumbs(db, dicts)
+    # PR-4b (ac27b693): route these post-recall reads off the shared write lock
+    # onto recall's read-only pool — the same _ro_read seam recall's own reads
+    # use. Both are pure indexed SELECTs (memory_metadata PK / memory_links
+    # source_id), so under concurrency they were queuing behind server writes,
+    # not query cost. _ro_read runs fn(pooled_conn, *args) and falls back to the
+    # shared connection on any pool miss/error, so this is never worse than the
+    # pre-pool path. retriever is genesis.mcp.memory._retriever (see CC memory
+    # two_retriever_wiring) — the same instance the pool was wired into in PR-4.
+    await retriever._ro_read(_enrich, dicts)
+    await retriever._ro_read(_breadcrumbs, dicts)
     enrich_ms = (time.monotonic() - t_enrich) * 1000
 
     lines = prof.renderer(dicts)
--- a/src/genesis/memory/proactive.py
+++ b/src/genesis/memory/proactive.py
@@ -401,7 +401,14 @@ def _render_procedure_line(proc: dict) -> str:
 
 
 async def _enrich(db: Any, dicts: list[dict]) -> None:
-    """Backfill ``_created_at`` + ``_wing`` from memory_metadata (in place)."""
+    """Backfill ``_created_at`` + ``_wing`` from memory_metadata (in place).
+
+    Lets a DB read error PROPAGATE — the caller routes this through
+    ``HybridRetriever._ro_read``, whose fallback retries on the shared connection,
+    and guards the whole enrichment step best-effort. Swallowing here would defeat
+    that fallback on a transient pooled-read failure (Codex #1197 P2). Idempotent,
+    so a fallback re-run is safe.
+    """
     ids = [
         d["memory_id"]
         for d in dicts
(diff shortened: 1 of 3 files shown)


WingedGuardian (commented):
### Structural Review

**Pattern violations (ERROR):** CLEAN — no nohup, no direct `task_states` inserts, no `/tmp/` usage, no bare `except:`, no mutable default arguments.

**Import hygiene:** CLEAN — `test_recall_read_pool.py` new imports follow stdlib → local ordering correctly.

**Missing coverage:** CLEAN — no new `.py` files in `src/` without tests; modified `proactive.py` is covered by two new targeted tests.

**Missing documentation:** CLEAN — no new public functions or classes without docstrings.

**⚠️ WARNING: Non-standard commit prefix**

The commit/PR title uses `perf(memory):` which is not in the Genesis conventional commit list (`feat:`, `fix:`, `refactor:`, `docs:`, `test:`, `chore:`). `perf:` is a recognized [Conventional Commits](https://www.conventionalcommits.org/) spec type but isn't in the project's defined set. Consider `fix(memory):` (it corrects a lock-contention regression) or `refactor(memory):` — or add `perf:` to the project prefix list.

---

**Code diff (added lines):** CLEAN — `ac27b693` is a commit hash reference, not a secret; test data is synthetic (`mem1`, `aaaaaaaa01`, etc.); internal module paths are not PII.

**PR title / body / commits:** CLEAN — no credential-cleanup language, no private IP addresses, no private hostnames, no security incident references.

---

**Result:** 1 WARNING (commit prefix convention). No ERRORs, hard blocks, or wording issues.

---

WingedGuardian:
Addressed the P2 in 307f5439: `_enrich`/`_breadcrumbs` no longer swallow their own DB errors, so a transient pooled-read failure now propagates to `_ro_read` and retries on the shared connection (the best-effort catch moved to the call site; the helpers are idempotent so a fallback re-run is safe). Added a regression test that a pooled connection whose SELECT raises falls back and still enriches. Also retitled `perf(memory):` → `fix(memory):` per the prefix note.

WingedGuardian:
@codex review

PR #1198, opened 2026-07-22
fix(follow-ups): stop follow_up_update clobbering status + orphaning completed_at

WingedGuardian:
## Problem (follow-up `d67c83c7`)

A follow-up audit reported `follow_up_update` intermittently flipping `status`→completed and `pinned`→true on calls that omitted those params, theorizing a **full-row read-modify-write** race.

Investigation disproved that theory: **every CRUD writer is already a targeted column `UPDATE`** — there is no full-row rewrite. The two reported symptoms were:
- **`pinned`→true**: a measurement artifact. The only `pinned` writers are the explicit MCP param and the dashboard toggle (no autonomous pinner exists); the affected rows were *already pinned*, and the tool's refreshed-return surfaced the true state, misread as a flip.
- **`status`→completed**: a concurrent writer (the ego's `resolve_follow_ups`, or another live session) legitimately completing the row mid-audit — not a `follow_up_update` clobber.

But the investigation surfaced **two real data-integrity bugs** (plus a doc/behavior mismatch):

### 1. Notes-only updates re-wrote `status` (lost update)
The MCP notes-only branch called `update_status(existing["status"], …)`, re-applying a status read ~30 lines earlier. Against Genesis's own live writers racing the same row, that **silently reverts** a status change made in between.
**Fix:** new `follow_ups.update_notes()` writes only the notes/blocked columns; the MCP branch routes to it and never touches `status`.

### 2. Orphaned `completed_at` on non-terminal rows
`update_status` stamped `completed_at` entering a terminal state but **never cleared it** on the way back out. A row wrongly completed then corrected to `in_progress`/`pending` kept a lying terminal timestamp that mis-keys GC/report windows. Verified live: 3+ rows carried `completed_at` while non-terminal.
**Fix:** `update_status` (+ `_batch`) NULLs `completed_at` on any non-terminal transition; **data migration `d0007`** heals historical orphans on every install (idempotent, post-boot).

### 3. `blocked_reason` didn't block
`blocked_reason` without an explicit `status` kept the old status, contradicting the documented contract ("sets status to blocked if status not provided"). Now sets `status='blocked'`.

## Tests
- **CRUD**: `update_notes` leaves status/`completed_at` untouched; notes-only regression; reopen clears `completed_at`; batch mirror.
- **MCP**: notes-only preserves status; `blocked_reason` blocks; **stale-read lost-update regression** (patches `get_by_id` stale → proves the old `elif` reverted, new path does not).
- **d0007**: clears orphans, spares terminal rows, no-op on fresh install.
- Adjacent consumers (`test_dispatch`, cockpit API) green.

## Notes
- Diff includes formatter churn from the repo's ruff-format PostToolUse hook (whole-file reflow; semantically identical, `ruff check` clean). Logical change is ~90 lines.
- RED-on-old-code verified by logic, not `git stash` (stash is repo-global; other worktrees active).

Ledger: d67c83c777f9494aa4db726834d1f8a5

### DIFF
--- a/src/genesis/db/crud/follow_ups.py
+++ b/src/genesis/db/crud/follow_ups.py
@@ -72,16 +72,30 @@ async def create(
             scheduled_at, status, priority, pinned, kind, domain, goal_id,
             dedup_key, created_at)
            VALUES (?, ?, ?, ?, ?, ?, ?, 'pending', ?, ?, ?, ?, ?, ?, ?)""",
-        (fid, source, source_session, content, reason, strategy,
-         _normalize_scheduled_at(scheduled_at), priority, int(pinned),
-         kind, domain, goal_id, dedup_key, _now_iso()),
+        (
+            fid,
+            source,
+            source_session,
+            content,
+            reason,
+            strategy,
+            _normalize_scheduled_at(scheduled_at),
+            priority,
+            int(pinned),
+            kind,
+            domain,
+            goal_id,
+            dedup_key,
+            _now_iso(),
+        ),
     )
     await db.commit()
     return fid
 
 
 async def exists_by_dedup_key(
-    db: aiosqlite.Connection, dedup_key: str,
+    db: aiosqlite.Connection,
+    dedup_key: str,
 ) -> bool:
     """Return True if any follow-up already exists with *dedup_key*.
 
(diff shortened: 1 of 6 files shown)


WingedGuardian:

---

WingedGuardian:

---

PR #1199, opened 2026-07-22
fix(follow-ups): stop notes-only follow_up_update reverting concurrent status

WingedGuardian:
## What

The notes-only branch of `_impl_follow_up_update` (the `follow_up_update` MCP tool) re-asserted the `status` it read at handler entry, so a concurrent cross-process `status` transition landing between that read and the notes write was silently reverted — a lost update.

**Fix:** add `crud.follow_ups.update_notes()` — a targeted partial `UPDATE` of only the note columns the caller supplied, touching neither `status` nor `completed_at` — and route the notes-only path through it. The `if status:` branch and the priority/pinned/kind writes are unchanged.

## Diagnosis correction

This item (`d67c83c7`) was originally filed as a **`pinned` field-clobber** (fields flipping under sequential, non-overlapping calls). That is **refuted**: `set_pinned` is the only writer of the `pinned` column, and the handler calls it *only* when `pinned` is explicitly passed, so an omitted-`pinned` update cannot change it. The `pinned:true` reads observed during a ledger audit were pre-existing pins misread as self-inflicted clobber. The genuine defect is the narrow status re-assert above.

## Tests

- `test_notes_only_update_does_not_revert_concurrent_status` — handler regression, **proven RED without the fix** (a status transition interleaved between the handler's read and write gets reverted), GREEN with it.
- `test_update_notes_*` — unit coverage: status untouched, `completed_at` preserved on a terminal row, no-op returning `False` on empty input.
- Full run of both files: 57 passed, ruff clean.

## Deploy path

- `crud/follow_ups.py` (runtime callers — dispatcher, ego): `git pull` + server restart.
- `follow_up_tools.py` (MCP tool): takes effect on next CC session start.
- No schema change, no migration; `update_notes` is purely additive (no existing caller affected).

Ledger: 4764332657754650b5992fbb47d17a65

### DIFF
--- a/src/genesis/db/crud/follow_ups.py
+++ b/src/genesis/db/crud/follow_ups.py
@@ -262,6 +262,41 @@ async def update_status(
     return cursor.rowcount > 0
 
 
+async def update_notes(
+    db: aiosqlite.Connection,
+    id: str,
+    *,
+    resolution_notes: str | None = None,
+    blocked_reason: str | None = None,
+) -> bool:
+    """Update resolution_notes/blocked_reason WITHOUT touching status.
+
+    A targeted partial UPDATE of only the note columns the caller supplied.
+    The notes-only update path uses this instead of re-writing status: the
+    prior code re-asserted the status it had READ at entry, so a concurrent
+    (cross-process) status transition landing in that gap was silently
+    reverted -- a lost update. Touching neither status nor completed_at closes
+    that window. Returns True if a row was modified, False on a no-op (no note
+    fields supplied)."""
+    parts: list[str] = []
+    params: list[str | None] = []
+    if resolution_notes is not None:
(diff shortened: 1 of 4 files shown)


WingedGuardian:

---

WingedGuardian:
Superseded by #1198 (merged), which is a strict superset: same core fix — new `follow_ups.update_notes()` and routing the notes-only `follow_up_update` branch through it so it never re-asserts `status` — plus two fixes this PR missed: clearing orphaned `completed_at` on non-terminal transitions (with data-migration d0007 to heal history) and making `blocked_reason` set `status='blocked'` per the tool contract. Same diagnosis (pinned-clobber refuted as a measurement artifact). A parallel session landed #1198 while this was in flight; closing to avoid a double-apply. Nothing unique here to salvage.

PR #1201, opened 2026-07-22
perf(memory): warm-connection reuse for recall embedding backends

WingedGuardian:
## Problem

Follow-up `ac27b693`, PR-2 lever. The proactive-recall `embed` stage swings from ~64ms to ~2357ms in prod. The recall embedder is a long-lived singleton whose backends (`OllamaBackend`, `DeepInfraBackend`, `DashScopeBackend` in `src/genesis/memory/embeddings.py`) each hold one `httpx.AsyncClient` for their lifetime — so the TLS connection to the cloud providers *can* be reused. But each client was built with httpx's **default pool settings**, whose `keepalive_expiry` is only **5s**. Between the sparse per-prompt recalls that drive proactive memory, the warm connection to DeepInfra/DashScope expires and every cold embed re-handshakes TLS — the spike named in the row.

## Fix

- Build each backend's `AsyncClient` via a shared `_build_embed_client()` helper that applies tuned `httpx.Limits`: `keepalive_expiry` raised to **60s** (module constant `_EMBED_KEEPALIVE_EXPIRY_S`, not an inline magic number) so warm connections survive the gap between recalls, plus modest `max_keepalive_connections`/`max_connections`. **This is the primary fix** — it directly kills the 5s-expiry re-handshake.
- Enable `http2=True` **only** for the TLS cloud backends (DeepInfra, DashScope), and **only** when the optional `h2` package is importable (gated through `_http2_available()`). No new dependency — httpx raises if `http2=True` is requested without `h2`, so we import-gate instead. Verified `h2` is present in the venv.
- Ollama keeps HTTP/2 **off**: it is a cleartext local endpoint where forcing h2 with prior knowledge could break HTTP/1.1-only servers. It still gets the warm-reuse `Limits`.

No behavior change beyond connection reuse — timeouts, request bodies, and the fallback chain are untouched. An injected `client` (test seam) still wins over the built default.

## Out of scope (noted for follow-up)

- **Graph-expansion N+1** over entity hops — explicitly out of PR-2 scope per the follow-up.
- **Qdrant client pooling** — no clear, safe reuse win identified here; left for a separate look.
- **PR-4** (#1189, read-only WAL connection pool) already landed and is deliberately untouched.

## Test

`tests/test_memory/test_embeddings.py::TestConnectionReuse` asserts the tuned `Limits`/`keepalive_expiry` are applied to each backend's client and that HTTP/2 is negotiated only for the cloud backends when `h2` is present (never for Ollama). Injected-client passthrough is covered too.

`pytest tests/test_memory/test_embeddings.py` → 25 passed. `ruff check` on changed files → clean.

Ledger: ac27b693

### DIFF
--- a/src/genesis/memory/embeddings.py
+++ b/src/genesis/memory/embeddings.py
@@ -39,6 +39,58 @@ _HTTPX_ERRORS = (
     httpx.HTTPStatusError,
 )
 
+# ---------------------------------------------------------------------------
+# Connection reuse tuning for embedding backends
+# ---------------------------------------------------------------------------
+# The recall embedder is a long-lived singleton whose backends each hold one
+# AsyncClient for their lifetime, so the underlying TLS connection *can* be
+# reused across proactive recalls. httpx's default keepalive_expiry is only 5s,
+# though — between the sparse per-prompt recalls that drive proactive memory,
+# the warm connection to DeepInfra/DashScope expires and each cold embed pays a
+# fresh TLS handshake (the ~2357ms embed spikes seen in prod). Extending
+# keepalive_expiry keeps the connection warm across that gap; the connection
+# counts stay small because a single recall issues one embed at a time.
+_EMBED_KEEPALIVE_EXPIRY_S = 60.0
+_EMBED_MAX_KEEPALIVE_CONNECTIONS = 8
+_EMBED_MAX_CONNECTIONS = 16
+
+
+def _embed_limits() -> httpx.Limits:
+    """httpx connection-pool limits tuned for warm-connection reuse."""
+    return httpx.Limits(
+        max_connections=_EMBED_MAX_CONNECTIONS,
+        max_keepalive_connections=_EMBED_MAX_KEEPALIVE_CONNECTIONS,
+        keepalive_expiry=_EMBED_KEEPALIVE_EXPIRY_S,
+    )
+
+
+def _http2_available() -> bool:
+    """True only if the optional ``h2`` package is importable.
+
+    httpx raises if ``http2=True`` is requested without ``h2`` installed, so we
+    gate on import rather than adding a hard dependency.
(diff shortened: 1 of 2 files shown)


WingedGuardian:
## First-pass code review

### Structural review

**WARNING — Non-standard conventional commit prefix**

The PR title uses `perf(memory):`. The Genesis allowed prefix list is `feat:`, `fix:`, `refactor:`, `docs:`, `test:`, `chore:` — `perf:` is not on it. The title/first commit message will need to be retitled before merge. `feat:` (net-new capability) or `refactor:` (restructuring without behaviour change) are the closest fits; given the primary observable is a latency fix, `fix:` is also defensible.

---

**NOTE — Tests couple to private httpx/httpcore internals**

`TestConnectionReuse._pool()` reaches through `client._transport._pool` (private `AsyncHTTPTransport` attribute), then asserts on `pool._keepalive_expiry`, `pool._max_keepalive_connections`, and `pool._http2` — all private httpcore attributes. These have no stability guarantees and have changed across minor httpcore versions in the past.

The tests will catch a future silent regression in the limits logic (good), but they will also false-fail on any httpx/httpcore version bump that renames or restructures the internal pool (bad). If CI pins httpx tightly this is acceptable short-term, but it's worth a follow-up to replace the pool-introspection assertions with observable behaviour (e.g., assert `_build_embed_client()` returns a client whose `limits` object matches the constants — `httpx.AsyncClient` exposes its `Limits` publicly via the constructor args, which are stored on the transport but accessible without spelunking private attrs).

The `_embed_limits()` unit test is fine as-is — it only calls the public `httpx.Limits` constructor and reads public fields.

---

**CLEAN** — No secrets, tokens, private IPs, private hostnames, sensitive paths, or wording issues found in the diff or PR metadata.

---

PR #1202, opened 2026-07-22
feat(cc): re-land the channel research-reroute on the delivery model

WingedGuardian:
## What

Re-lands **PART 2** of follow-up `24a6857a` — the channel **research-reroute** — on top of the now-merged background-session delivery model (#1192).

PR-4 (#1186) added a foreground system-prompt nudge that routes long/deep research off a **dispatched channel turn** (Telegram / voice / OpenClaw) — whose turn ends after the reply — into the durable `direct_session` background lane, instead of an inline `Workflow` the CC bg-wait ceiling force-kills at ~10 min with nothing left to report back (the 2026-07-20 silent-death class). Code review (`d7aedfdf`) **deferred** it because the routed result had no delivery path: `DirectSessionRunner` suppressed success and the `research` profile can't send its own report, so a routed research would complete and never reach the user. The deferral promised it would "return with the delivery model."

That delivery model merged in **#1192** (`DeliveryMode`, `deliver_to_origin`, `_deliver_result_to_origin`), but the reroute was never re-landed (`grep _BG_RESEARCH_ROUTING src/` = 0 hits). This PR re-lands it, adapted to the new API.

## Changes

- **Restore** `_BG_RESEARCH_ROUTING` + `_apply_research_routing` in `cc/conversation.py`, now nudging `direct_session_run(profile="research", deliver_to_origin=true)` — **not** the removed `notify` boolean, which would leave a successful run silent (the exact deferral condition).
- **Re-wire** the nudge for non-terminal (turn-ends) channels in `handle_message`, `handle_message_streaming`, and `_build_fresh_invocation` (the stale-resume retry passes `channel` again so recovery keeps the nudge). Terminal stays inline — the user is present to watch a `Workflow`.
- **Docs**: restore the "dispatched from a channel" section in `background-sessions.md` and the `CURRENT.md` foreground-routing note, both keyed to `deliver_to_origin`.
- **Tests**: gating (telegram/web append, terminal omit), a delivery-model assertion (`deliver_to_origin` present, `notify` absent), and the fresh-invocation recovery path.

## Verification

- `pytest tests/test_cc/test_conversation.py` — 34 passed (6 new reroute tests).
- `ruff check --no-cache` on changed files — clean.
- Minimal diff: `conversation.py` reconstructed from pristine main + hunks only (main isn't `ruff format`-clean; the project lints, it does not format), so the PR contains only the semantic re-land, no whole-file reflow.
- Runtime path: the nudge injects into the system prompt at the same append point as topic context; delivery itself is already wired and merged in #1192. This PR is the foreground routing half only.

## Note

Faithful to the original deferred design — the reroute path (failover peers, cancel-path delivery) is intentionally unchanged; cancel-path delivery remains out of scope (PR-2, ledger `ee391fbd`).

Ledger: 24a6857a

### DIFF
--- a/.claude/docs/background-sessions.md
+++ b/.claude/docs/background-sessions.md
@@ -30,6 +30,26 @@ the next few minutes → sub-agent.
 > its own report); oversized output is saved under `~/.genesis/output/` and delivered
 > as a summary + file pointer.
 
+## Dispatched from a channel? Long work MUST be a background session
+
+When you are a foreground session driving a **Telegram/voice/OpenClaw reply**, your
+turn **ends after you respond** — there is no live session left to report back when a
+later-finishing task completes. A deep-research `Workflow` (or any 100+-agent fan-out)
+run **inline** in such a turn is force-killed by the CLI's headless background-wait
+ceiling (`CLAUDE_CODE_PRINT_BG_WAIT_CEILING_MS`, ~10 min) with only a partial result,
+and nothing delivers it. This silently killed a real Telegram deep-research request on
+2026-07-20.
+
+So for a channel-dispatched request needing deep/multi-source research or any
+background work likely to exceed a few minutes: **do NOT run it inline — dispatch it
+via `direct_session_run` (`profile="research"`, `deliver_to_origin=true`) and reply
+that it's running in the background.** The background lane owns a longer wait ceiling
+(set to its full `timeout_s`), runs to completion, and — with `deliver_to_origin` —
+delivers the finished outcome back to this exact conversation (the delivery model
+merged in #1192). Terminal/interactive sessions may still run Workflows inline (you're
+present to see them). The foreground system prompt (`conversation._BG_RESEARCH_ROUTING`)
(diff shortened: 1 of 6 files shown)


WingedGuardian:

---

WingedGuardian:
Review pass (landing session): **CLEAN**.
- Codex P2 ("restrict reroute to addressable channel origins") was raised on `e5061205` and is resolved by `8ba41cdd`: `_apply_research_routing` gates the nudge on `origin_delivery_supported(channel)` (Telegram-only), and `_resolve_origin_target` now shares that same predicate. WEB/OpenClaw/WhatsApp/VOICE no longer receive an undeliverable "I'll report back" promise.
- Verified conflict-free against current main (`git merge-tree` clean) after #1210/#1212 merged; the nudge modifies system_prompt pre-invocation while the rate-limit park writer runs in the post-invocation exception handler — separate regions, no interaction.
- Updating the branch to re-run CI on the true merged base before merge.

PR #1206, opened 2026-07-22
fix(deploy): dashboard cgroup isolation + honor CLI runs + owner-checked marker (P5-B)

WingedGuardian:
## Deploy-audit P5-B (parts 1/3/4) — follows P5-A (#1195, merged)

(Supersedes #1196, auto-closed when the P5-A base branch was deleted; identical content, rebased onto main.)

Architect design-reviewed + built-code re-reviewed (both DONE_WITH_CONCERNS, all findings addressed). Closes #4 (dashboard state-GC destroying a CLI run's state) and #5's dashboard side, plus NOTE-1 (`_apply_direct` cgroup self-SIGTERM).

- **Part 1 (cgroup / NOTE-1):** `_apply_direct` spawns `update.sh` via `systemd-run --user --scope` (own cgroup), probed for viability first (`_systemd_run_scope_available` — falls back to `start_new_session` if systemd-run is absent OR the user D-Bus is unreachable, instead of silently no-op'ing).
- **Part 3 (#4/#5):** all four dashboard "update in progress" checks (apply/dismiss/resolve/status-GC) use the canonical `env.update_in_progress()` (marker AND state file), so a CLI `update.sh` run is seen (409'd, not double-launched) and its crash-recovery state isn't GC'd.
- **Part 4 (marker ownership):** update.sh's marker deletes go through `_clear_deploy_state` (delete only if `==$$` OR holder dead — never a live foreign holder). The direct path **adopts** the marker (writes its own `$$` when the marker holds `$PPID` = systemd-run's pid) so it's cleaned per-run, no leak.

### Review findings addressed
Architect BLOCKER-1 (live-holder strip) — resolved; SHOULD-FIX-1 (direct-path marker leak) — fixed via `$PPID` adoption; Codex P2s (systemd-run scope-viability probe, marker leak) — both fixed. Verified the `$PPID` adoption gate is safe on the supervised path (update.sh runs there via a CC session, so its `$PPID` is that session, not the orchestrator).

### Tests
27 update.sh + dashboard tests green (marker-ownership incl. direct-run cleanup, systemd-run probe + both spawn branches, guard wiring). `updates.py` reconstructed to avoid format churn (base wasn't ruff-format-clean; CI lints only).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   needlessly repeated. (Slow per-record checks were also moved out of the locked
   window.)
 
+- **The dashboard now recognizes updates started from the command line.** Its
+  "update in progress" checks previously only saw dashboard-triggered updates,
+  so a command-line `update.sh` run could be interrupted (its state wiped) or a
+  second update launched over it. The dashboard now consults the same
+  deploy-in-progress signal the rest of the system uses, and a dashboard-started
+  update no longer runs in the server's own service group (where the update
+  stopping the server could kill the update itself).
 - **Two updates can no longer run at the same time.** If an update is already in
   progress, a second `update.sh` (from another session, or the dashboard) now
   refuses immediately instead of running concurrently — previously two updates
--- a/scripts/update.sh
+++ b/scripts/update.sh
@@ -150,6 +150,17 @@ if ! flock -n "$_UPDATE_LOCK_FD"; then
     exit 1
 fi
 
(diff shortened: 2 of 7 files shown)


WingedGuardian:
Addressed the adoption race. Rather than iterate the marker-write timing again, I dropped the direct-path marker entirely: update.sh writes its **state file before it stops the server** (verified @603 vs @716), and `env.update_in_progress()` honors the state file exactly like a CLI run — so the watchdog is deferred whenever the server is down, and the P5-A **flock** is the hard mutual-exclusion guarantee for the brief pre-state-file window (a second spawn fails the flock before any side effect). This removes the adoption block + the `_apply_direct` marker write, eliminating the whole marker race/leak/adoption class for the direct path. Good catch — the self-exec + fast startup made the race real.

PR #1208, opened 2026-07-22
perf(memory): run recall's Qdrant vector search off the event loop (ac27b693, PR-2)

WingedGuardian:
## What & why

`_gather_vector_candidates` looped the **synchronous** `qdrant_ops.search` (→ `client.query_points`) directly on the asyncio event loop, once per collection — blocking every other coroutine (other recalls, the whole server) for each Qdrant round-trip. Under Jay's 5–7 concurrent-session norm that's shared-loop head-of-line blocking. It was the **only** remaining sync-Qdrant-on-loop call on the read path (`_expand_fts_query`'s scroll and the PR-1 write-back are already offloaded).

This is **PR-2** of the ac27b693 recall-latency series (after PR-1 side-effect deferral, PR-4 recall-read RO pool, PR-4b enrich/breadcrumbs pool).

## Change

Offload **each per-collection search** via `asyncio.to_thread(qdrant_ops.search, …)`. Deliberately per-collection (not one batched closure) so `track_operation("qdrant.search")` stays exactly where it was — on the event loop, one span per collection, wrapping the await:

- **Telemetry byte-identical** — span count, success/failure, and sample volume unchanged, so the `provider:qdrant_unreachable` alert (`mcp/health/errors.py`, CRITICAL at error_rate≥0.5) is unaffected. (A first pass collapsed N spans→1; a local code review flagged the alert-sensitivity change and it was reworked to per-collection offload.)
- **Exception semantics unchanged** — `asyncio.to_thread` re-raises, so a Qdrant error propagates identically.
- **Behavior parity** — score-sort / `_collection` tagging / first-hit-wins dedup untouched.
- **Thread-safety** — the Qdrant client is remote REST/httpx (thread-safe); this extends the already-shipped worker-thread pattern from PR-1's write-back offload.

## Value (honest)

The happy-path vector stage is small (~32ms); the win is **not blocking the shared event loop** under concurrency so other coroutines progress. The live 7× event-loop-responsiveness measurement is **deploy-gated** (deploy currently held).

## Tests

New `tests/test_memory/test_vector_offload.py`: (a) the search runs in a **worker thread** (proves the offload), (b) sort/tag/dedup parity across collections. Full recall regression suite green locally (57 tests incl. `test_recall_deferral`, `test_retrieval`).

Follow-up: ac27b693

### DIFF
--- a/src/genesis/memory/retrieval.py
+++ b/src/genesis/memory/retrieval.py
@@ -1243,8 +1243,18 @@ class HybridRetriever:
         qdrant_results: list[dict] = []
         qdrant_by_id: dict[str, dict] = {}
         for coll in collections:
+            # Qdrant's client is SYNCHRONOUS (no AsyncQdrantClient exists in
+            # Genesis), so offload each blocking search to a worker thread — it
+            # must never block the shared event loop while other coroutines (other
+            # recalls, the rest of the server) are waiting, which matters under
+            # Jay's 5-7 concurrent-session norm (ac27b693, PR-2). track_operation
+            # stays ON the event loop wrapping the await, so per-collection
+            # "qdrant.search" telemetry (the provider:qdrant_unreachable signal in
+            # mcp/health/errors.py) is byte-identical to the inline version, and
+            # asyncio.to_thread re-raises so a Qdrant error propagates unchanged.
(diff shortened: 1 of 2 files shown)


WingedGuardian (commented):

---

PR #1209, opened 2026-07-22
fix(deploy): rollback restores migrated DB + recovery run restarts the server (P6)

WingedGuardian:
### #7 — rollback restores the DB when migrations ran
A rollback restored code + deps but left a **migrated** database in place, so the rolled-back OLD code ran against the NEW schema. Now `update.sh` sets `MIGRATIONS_RAN=1` **before** invoking the migration runner (a partial failure still altered the schema), and `_do_rollback` restores `$DB_FILE.pre-update` (P4a's `.backup` snapshot) when that flag is set + the snapshot exists — safe because the server is stopped during rollback (DB quiescent). Adds `daemon-reload` and an honest header (rollback restores code/deps/DB-if-migrated, **not** installed units/hooks — those self-heal next update). `db_ok` gates the `rolled_back`-vs-`failed` outcome.

### #6 — a recovery re-run actually restarts the server
The restart+health block is gated on `[[ ${#WERE_RUNNING[@]} -gt 0 ]]`, so a recovery re-run (a prior failed update left the server stopped → empty `WERE_RUNNING`) **skipped restart + health and recorded bare "success" with the server down**. Now, when `WERE_RUNNING` is empty, a failure artifact (`last_update_failure.json`) or a last `update_history` status of `rolled_back`/`failed` marks a recovery run → forces `genesis-server` back into the restart set (restart + health verify). An operator-stop (no artifact) is respected — not force-started — but recorded with a `genesis-server-not-restarted` degraded note instead of a bare success.

### Tests
Extraction locks for both findings + a **functional recovery-decision test** (artifact / `rolled_back`|`failed` → recovery; clean prior → operator-stop). 866 `test_scripts` tests green; shellcheck unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,19 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **A failed update that had already run database migrations now rolls the
+  database back too.** Previously a rollback restored the code and dependencies
+  but left the (newly migrated) database in place, so the rolled-back older code
+  ran against a newer schema. Rollback now restores the pre-update database
+  snapshot whenever migrations ran (the server is stopped at that point, so it's
+  a clean swap), reloads systemd units, and states plainly what it did and didn't
+  revert.
+- **Recovering from a failed update actually brings the server back.** If a
+  previous update failed and left the server stopped, the next update used to
+  finish and report "success" while the server stayed down and health was never
+  checked. It now detects a recovery run (from the leftover failure record) and
+  restarts + health-verifies the server. A server the operator deliberately
+  stopped is left alone, but recorded as not-running rather than a bare success.
 - **A large one-time data cleanup no longer briefly freezes the running system.**
   Post-startup data migrations (one-off cleanups/backfills of stored knowledge and
   history) run alongside the live system, which allows only one writer to the
--- a/scripts/update.sh
+++ b/scripts/update.sh
@@ -686,6 +686,9 @@ _start_genesis_server() {
 # If the server is killed mid-write during the update, this backup
 # enables recovery without data loss.
 DB_FILE="$GENESIS_ROOT/data/genesis.db"
+MIGRATIONS_RAN=0  # set to 1 once the migration runner is invoked; gates the
+                  # pre-update DB restore in _do_rollback (rollback the schema
+                  # only if this update actually migrated it).
 if [ -f "$DB_FILE" ]; then
     echo "--- Snapshotting database ---"
     sqlite3 "$DB_FILE" "PRAGMA wal_checkpoint(TRUNCATE);" 2>/dev/null || true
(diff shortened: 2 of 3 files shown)


WingedGuardian:
## Code review — P6 rollback + recovery correctness

**Structural review:** 1 WARNING (no ERRORs)

---

### WARNING — Raw SQL against genesis.db in shell-embedded Python

`scripts/update.sh`, recovery detection block (the `_last_status` heredoc added after `_write_state "health_check"`):

```python
con = sqlite3.connect(os.path.expanduser("~/genesis/data/genesis.db"), timeout=5)
row = con.execute(
    "SELECT status FROM update_history ORDER BY started_at DESC LIMIT 1"
).fetchone()
```

This queries genesis.db directly via raw SQL rather than going through `genesis.db.crud`. The rule flags this as a pattern violation. Context acknowledged: this is an inline bash heredoc where importing the genesis package isn't practical, and the query is read-only. Flagging for awareness — if there's a precedent for shell-script DB reads in this codebase (e.g. other update.sh probes), this is consistent with that pattern and can be treated as an accepted exception; otherwise worth noting for any future refactor that moves this logic into a genesis CLI subcommand.

---

All other structural checks passed: correct commit prefix (`fix(deploy):`), no `/tmp/` usage, no `nohup`, no bare `except:`, no mutable defaults, no star imports, correct import order in the test file, `IntervalTrigger` not used, no `task_states` direct inserts. The test file covers the shell script (not a Python src/ module), so the missing-test-coverage rule doesn't apply.

PII/Secrets/Wording scan clean — no keys, private IPs, personal emails, absolute paths with usernames, or sensitive wording in the PR title/body/CHANGELOG.

---

WingedGuardian:
## Code review — first-pass (structural + PII/wording)

**Structural review:** one NOTE — otherwise clean.
**PII/Secrets/Wording scan:** CLEAN.

---

### NOTE — Raw SQL against `genesis.db` in inline Python heredoc (`scripts/update.sh`)

The recovery-detection block embeds a Python heredoc that queries `update_history` directly via `sqlite3.connect` rather than going through `genesis.db.crud`:

```python
con = sqlite3.connect(os.path.expanduser("~/genesis/data/genesis.db"), timeout=5)
row = con.execute(
    "SELECT status FROM update_history ORDER BY started_at DESC LIMIT 1"
).fetchone()
```

This is technically a raw-SQL access against `genesis.db` (rule: use `genesis.db.crud` modules). That said, this is inside a bash deployment script where importing the application module path is not practical, the query is read-only, and it touches a single well-understood column — a justified deployment-script exception. Flagging for awareness rather than requesting changes. If `update_history` gains a CRUD accessor in the future, consider pulling this probe into a small helper script that can import it.

---

Everything else is clean:
- No pattern violations (no `nohup`, no `/tmp/` usage, no `bare except:`, no `task_states` direct inserts, no force-push patterns, no star imports).
- Import grouping in the test file is correct (stdlib → third-party).
- New test file at `tests/test_scripts/test_update_rollback_recovery.py` covers the changed `scripts/update.sh`. No new `src/` `.py` files without coverage.
- Commit prefix `fix(deploy):` is compliant.
- PII scan of all added lines: no secrets, private IPs, personal emails, hardcoded username paths, or private hostnames. DB path uses `os.path.expanduser("~/...")` — clean.
- PR title, body, and CHANGELOG additions contain no security-incident language, credential-leak references, or private-infrastructure naming.

---

WingedGuardian:
Both Codex P1 findings + the code-reviewer SHOULD-FIX are addressed in 9ee76401:

- **P1 @update.sh:885 — clear WAL sidecars on restore**: the restore now `rm -f "$DB_FILE-wal" "$DB_FILE-shm"` after the `cp`, mirroring `restore.sh:547-549`. Without it SQLite would replay the migrated WAL over the restored old DB. Also added a `server_down` guard: if `_ensure_server_down` can't confirm the server is stopped, the restore is SKIPPED (never `cp` over a possibly-live DB).
- **P1 @update.sh:883 — treat a missing snapshot as rollback failure**: the restore is now gated on a new `DB_SNAPSHOT_TAKEN` flag (set only when *this* run's `.backup` succeeds), not `[ -f ]`. If migrations ran but no current snapshot exists, the rollback fails loudly (`db_ok=false` → `failed` outcome) instead of reporting clean success — and a stale prior-run `.pre-update` is never restored as if current.
- **code-reviewer SHOULD-FIX — stale recovery signal on no-op path**: the 'Already up to date' path now clears `last_update_failure.json` and records a fresh success when genesis-server was up at start, so a long-resolved failure can't later force-restart an operator-stopped server. Gated on server-was-up so an unresolved failure's artifact survives to drive recovery.

Tests extended in `test_update_rollback_recovery.py` (snapshot-flag gate, missing-snapshot failure, gated no-op clear). 8/8 rollback tests + 37 update.sh suite tests green; `bash -n` clean.

PR #1211, opened 2026-07-22
feat(memory): rate-gate + timeout-only breaker for recall's Voyage rerank (ac27b693, PR-3)

WingedGuardian:
## What & why

Recall's Voyage rerank had **no rate limiting**. Under the free tier (**3 RPM + 10K TPM**) with Jay's 5–7 concurrent sessions the account is saturated — the server logs show **65 `Voyage rerank HTTP 429 — degrading to RRF order` in the last 7 days**, each paying a wasted round-trip before falling back. Zero genuine-outage lines in the window: rate-limiting is the whole problem.

This is **PR-3** of the ac27b693 recall-latency series.

## Design (Option B — gate kills 429s, breaker guards only hangs)

Shared, process-wide rate gate + **timeout-only** circuit breaker, consulted at the `_maybe_rerank` call site (mirrors `router.py` wrapping providers rather than baking resilience into `VoyageReranker`, which `corrective.py` also uses and is left untouched):

- **`ProviderRateGate.try_acquire()`** — new non-blocking variant of `acquire()` (which *sleeps* up to the interval = 20s at 3 RPM, unusable on the hot path). Gate denied → skip **instantly** to the RRF+graph floor (`rerank_skipped_ratelimited`), no 429 burned.
- **Breaker trips ONLY on an `asyncio.wait_for` `TimeoutError`** (a genuine hang) → `record_failure(ErrorCategory.TIMEOUT)`. A 429/empty return gives **no breaker signal** — respecting the routing rule that a 429 must not trip a breaker (the gate is the RPM brake). Open → skip (`rerank_skipped_breaker_open`); a real result → `record_success()` heals it.
- **One gate + one breaker shared by BOTH retrievers** (`runtime/init/memory.py` builds them once, passes the *same instances* to `HybridRetriever(...)` and `init_memory_mcp(...)` → MCP retriever) so Voyage's account-global RPM is enforced across all concurrent recalls — two instances would each permit the full RPM. Guarded by the `two_retriever_wiring` regression test.
- **Kill-switch** `GENESIS_RECALL_RERANK_GATE_OFF` + config `GENESIS_RECALL_RERANK_RPM` (default 3, tunable for paid tiers). `None`/`None` → byte-parity with today (verified).

RRF+graph stays the guaranteed floor on every skip — **no result-quality change**, only removes wasted/doomed Voyage calls. Does **not** speed successful reranks (~450–550ms Voyage latency is real).

## Tests

- `try_acquire` (grant / deny-within-interval-instantly / grant-after-window / denial-doesn't-advance-window) — deterministic, no real sleeps.
- `_maybe_rerank` branches: breaker-open short-circuits before the gate, gate-denied never calls Voyage, success → `record_success`, timeout → `record_failure(TIMEOUT)` only, empty-return → no breaker signal, kill-switch None/None → byte-parity.
- **Integration with REAL `ProviderRateGate` + `CircuitBreaker`** (not mocks): the real gate denies the 2nd call within the interval; the real breaker trips OPEN after 3 consecutive timeouts and then skips without calling Voyage.
- Extended the two-retriever wiring guard to pin the MCP retriever receives the gate + breaker.

Full recall + routing regression green locally (387 tests). The live 7× probe (expect `rerank_skipped_ratelimited` to climb and the 429 WARN rate to drop toward zero) is **deploy-gated** — deploy currently held.

Follow-up: ac27b693

### DIFF
--- a/src/genesis/env.py
+++ b/src/genesis/env.py
@@ -179,6 +179,41 @@ def recall_read_pool_size() -> int:
         return DEFAULT_READ_POOL_SIZE
 
 
+def recall_rerank_gate_off() -> bool:
+    """True when the recall rerank rate-gate + circuit-breaker must NOT be built
+    (kill switch).
+
+    Recall's Voyage rerank normally runs behind a shared rate gate (skip to the
+    RRF+graph floor instead of burning a 429 when over Voyage's RPM) and a
+    timeout-only circuit breaker (skip during a Voyage hang) — follow-up
+    ac27b693, PR-3. This env kill makes the runtime skip building them, so the
+    rerank call is unguarded exactly as before the change — no code change.
+    Default off: the gate+breaker are built.
+    """
+    return os.environ.get("GENESIS_RECALL_RERANK_GATE_OFF", "").strip() in (
+        "1",
+        "true",
+        "yes",
+    )
+
+
+def recall_rerank_rpm() -> int:
+    """Requests-per-minute the recall rerank rate gate paces Voyage to.
+
+    Voyage's free tier (no payment method) is 3 RPM; paid usage tiers are higher,
+    so this is tunable per install via ``GENESIS_RECALL_RERANK_RPM`` rather than
+    hardcoded (generalizability). Floored at 1; a missing/non-integer value falls
+    back to the free-tier default of 3.
+    """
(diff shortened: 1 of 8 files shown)


WingedGuardian:

---

PR #1213, opened 2026-07-22
feat(observability): alert when ambient auto-recovery can't restore a dark device

WingedGuardian:
## What

The ambient-capture health monitor now raises a `degraded` alert when the edge
bridge reports `recovery_failing` — i.e. auto-recovery is armed, has engaged, and
still could not bring a wedged capture device back after it has been dark past the
escalation threshold.

This is deliberately the **only** darkness signal the evaluator acts on. There is
no network signal that distinguishes a device that was *unplugged* from one that
*crashed*, so alerting on raw absence would nag any device that's ever powered off.
`recovery_failing` sidesteps that: it is emitted **only** where the operator armed
auto-recovery, which already asserts the device is expected to stay up — so an
absent or intentionally-off device on an install without recovery stays silent, and
the long-standing "never alert on an absent device" invariant is preserved.

## Details

- `evaluate_ambient_health` gains an additive `recovery_failing` check → new
  `recovery-failing` cause; only upgrades `ok → degraded` (a dead bridge stays
  `down`). Absent/false leaves behavior unchanged.
- `_ambient_remedy_hint` gains a cause-aware line (precedence: bridge-dead >
  recovery-failing > diar > rss).
- The alert names the dark-since time and the failed-reboot count; it never
  contains a device address or key (the edge sanitizes those before emitting).

The edge side that produces this signal is a companion change in the voice/edge
repo; both are additive and back-compatible, so they merge independently.

## Tests

New pure-function coverage in `test_ambient_health.py` (degraded+cause on true;
unchanged on false/absent; strict `is True` gate; does-not-mask a dead bridge;
coexists with the diar-worker cause). Full suite green; ruff clean.

Ledger: 54ec6212ef514b45b09a088ff6d829ac

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,17 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **Ambient capture now alerts when auto-recovery can't bring a wedged device
+  back.** For installs running the optional voice/ambient edge with device
+  auto-recovery armed, Genesis now raises a capture-health alert when the device
+  has been dark for hours *and* the automatic reboots failed to restore it — the
+  one case that means a real fault rather than a device simply being switched off.
+  Installs without auto-recovery are unchanged: a merely-absent device still never
+  alerts (nothing on the network distinguishes "unplugged" from "crashed", so the
+  signal fires only where arming recovery already asserts the device should be up).
+  The alert names how long it's been dark and how many reboot attempts failed, with
+  no device address or key ever included.
+
 - **When Genesis hands a long task off to run in the background and says "I'll
   report back," it now actually does.** Ask for something heavy from Telegram — a
   deep research run, a long analysis — and Genesis can dispatch it to a background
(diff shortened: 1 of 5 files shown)


WingedGuardian:
## Code Review — PR #1213

### Structural review: 1 WARNING

**WARNING — Missing test coverage for new branch in `_ambient_remedy_hint`**

`src/genesis/outreach/scheduler.py` — `_ambient_remedy_hint` has a new `recovery-failing` branch (the hint text sent to the user in the alert):

```python
if "recovery-failing" in causes:
    return (
        "(auto-recovery exhausted — the device is dark and reboot attempts failed; "
        "check the device's power/network or the ESPHome API on the Voice PE)"
    )
```

The PR's new tests all exercise `evaluate_ambient_health` in `ambient_health.py`. No test covers this branch of `_ambient_remedy_hint`. Since this is user-visible text in the alert, a simple parametrized test asserting `"ESPHome API"` or `"auto-recovery exhausted"` appears when `causes=("recovery-failing",)` would close the gap. The other existing hint branches (bridge-dead, diar-worker, rss) presumably have coverage already; this new one does not.

---

No API keys, tokens, private IPs, SSH keys, or credentials in added lines. `device_dark_since="2026-06-18T06:00:00+00:00"` in test fixtures is a synthetic timestamp. PR title, body, and CHANGELOG additions describe a new feature — no language advertising a fix to a prior PII/credentials exposure.


---

PR #1215, opened 2026-07-22
fix(host): abort/honesty guards for host + guardian scripts (P7)

WingedGuardian:
Deploy-audit **P7** — robustness guards for host/guardian shell scripts that run under `set -euo pipefail`.

## Findings fixed
| id | file | fix |
|----|------|-----|
| H1 | host-setup.sh | IP-geolocation TZ pipeline `\|\| _detected_tz=""` — a curl-fail/grep-no-match falls through to the timedatectl fallback instead of aborting |
| H2 (×2) | host-setup.sh | managed-bridge detect tolerates a grep no-match (`\|\| true`), reaching the "no bridge" skip on a non-incus host |
| H3 | host-setup.sh | OOM `sudo sysctl --system` wrapped in if/else — neither aborts nor claims success it didn't achieve |
| H6 | host-setup.sh | "IOPS limits applied" echo gated on the idempotent IOPS sets (a pool without device-limit support rejects them while the container still creates) |
| G2 | guardian-gateway.sh | `OLD=$(git rev-parse …)` gets `\|\| echo "unknown"` matching its siblings |
| I4 (×2) | install_guardian.sh, guardian-gateway.sh | `udevadm trigger --subsystem-match=block` after the rules reload, so the BFQ scheduler rule reaches existing block devices |
| I5 | install_guardian.sh | venv gate tests `-x "$VENV_DIR/bin/python"` (+`--clear`) not `-d`, so a partial/dangling venv is rebuilt, not skipped |

## Testing
- Extraction/regression-lock tests in `test_host_destructive_guards.py` (adds a `GUARDIAN_GATEWAY` constant); 16/16 pass.
- `bash -n` clean on all three scripts.
- code-reviewer: no ≥80-confidence findings; two sub-threshold accuracy items (H6 comment, I5 dangling-symlink) fixed in the review-fix commit.

## Deploy
**Host-Deploy Gate applies** (touches host-setup.sh / guardian-gateway.sh / install_guardian.sh) — `scripts/update.sh` + guardian redeploy verification after merge.

### DIFF
--- a/scripts/guardian-gateway.sh
+++ b/scripts/guardian-gateway.sh
@@ -288,7 +288,7 @@ except Exception:
         INSTALL_DIR="${HOME}/.local/share/genesis-guardian"
         if [ -d "$INSTALL_DIR/.git" ]; then
             cd "$INSTALL_DIR"
-            OLD=$(git rev-parse --short HEAD 2>/dev/null)
+            OLD=$(git rev-parse --short HEAD 2>/dev/null || echo "unknown")
             # Discard CLAUDE.md before pull — it's a generated file that the
             # repo's version would overwrite with the wrong content (container
             # dev instructions vs Guardian diagnostic instructions). We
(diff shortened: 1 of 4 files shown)


WingedGuardian (commented):

---

WingedGuardian:
@codex review

PR #1216, opened 2026-07-22
fix(bootstrap): robustness guards for bootstrap + local-config (P8a)

WingedGuardian:
Deploy-audit **P8a** — robustness guards over `scripts/bootstrap.sh` and `scripts/setup-local-config.sh`.

## Findings fixed
- **B1** — `setup-complete` marker written at the VERY END (after "Bootstrap complete"), not the early runtime-state block. The marker gates the fresh-install onboarding prompt (`genesis_session_context.py`) + ego cadence, so a bootstrap that crashes mid-run must not leave the box looking onboarded. Verified no consumer reads it mid-bootstrap; `touch` is idempotent so re-runs keep their marker.
- **B3** — `daemon-reload` runs unconditionally before the enable loop (was gated on `SERVICES_UPDATED`), covering a prior run that wrote a unit then crashed before reloading.
- **B5** — serena `-s project` register wrapped in `( cd "$GENESIS_ROOT" && … )`; project scope is cwd-keyed, so bootstrap run from elsewhere would write `.mcp.json` to the wrong repo.
- **B6** — resolved timezone validated against `timedatectl list-timezones`, but ONLY when it returns a non-empty list (never force-resets to UTC on an image that can't enumerate).
- **B7** — trailing-newline guard before the secrets.env append.
- **B8** — SkillSpector clone + pip `timeout 300`-bounded with one clone retry (a hung TCP connection can't stall bootstrap on an optional dep).
- **B9** — codebase-memory-mcp + uv installers download-to-temp then execute-from-file instead of `curl … | bash` (no truncated-installer execution).
- **S3** — removed dead `REPO_EXAMPLE` var + corrected the phantom "Migrate from repo YAML" comment.

## Testing
- New `test_bootstrap_guards.py` (extraction locks + a functional B7 newline test under `set -euo pipefail`); 9/9 pass.
- `bash -n` clean on both scripts.

## Deploy
No host gate — deploys in-container via `git pull` at next `update.sh`.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -334,8 +334,19 @@ echo "--- Installing code intelligence tools ---"
 # in ~/.claude/.mcp.json, bypassing our 2G-capped launcher. We register the
 # capped wrapper below via _register_mcp, so the installer must NOT self-register.
 echo "  codebase-memory-mcp: installing/upgrading..."
-curl -fsSL https://raw.githubusercontent.com/DeusData/codebase-memory-mcp/main/install.sh | bash -s -- --ui --skip-config \
-    || echo "  WARNING: codebase-memory-mcp install/upgrade failed (non-critical)"
+# Download-to-temp before executing: `curl … | bash` runs bytes as they stream,
+# so a mid-transfer connection drop executes a truncated script. -f fails the
+# download on a partial transfer; running from a file executes only a fully
+# downloaded installer. B9.
+_cbm_installer=$(mktemp 2>/dev/null) || _cbm_installer=""
+if [[ -n "$_cbm_installer" ]] \
+    && curl -fsSL https://raw.githubusercontent.com/DeusData/codebase-memory-mcp/main/install.sh -o "$_cbm_installer" 2>/dev/null; then
+    bash "$_cbm_installer" --ui --skip-config \
+        || echo "  WARNING: codebase-memory-mcp install/upgrade failed (non-critical)"
+else
+    echo "  WARNING: codebase-memory-mcp installer download failed (non-critical)"
+fi
+[[ -n "$_cbm_installer" ]] && rm -f "$_cbm_installer"
 if command -v codebase-memory-mcp &>/dev/null; then
     echo "  codebase-memory-mcp: $(codebase-memory-mcp --version 2>/dev/null || echo 'installed')"
 fi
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

WingedGuardian (commented):
## Structural review: 1 ERROR (request-changes blocked — same-author PR)

### ERROR — `mktemp` defaults to `/tmp/` (must use `~/tmp/`)

The B9 fix adds two `mktemp` calls without a `-p` argument:

```bash
# scripts/bootstrap.sh — codebase-memory-mcp installer
_cbm_installer=$(mktemp 2>/dev/null) || _cbm_installer=""

# scripts/bootstrap.sh — uv installer
_uv_installer=$(mktemp 2>/dev/null) || _uv_installer=""
```

Without `-p`, `mktemp` uses `$TMPDIR` if set, otherwise `/tmp/`. During `bootstrap.sh` — which runs outside a CC session — `$TMPDIR` is typically unset, so both files land in `/tmp/`. Project rule: temp files must use `~/tmp/`, not `/tmp/`.

Fix (apply the same pattern to both):
```bash
mkdir -p ~/tmp
_cbm_installer=$(mktemp -p ~/tmp 2>/dev/null) || _cbm_installer=""
```

The `mkdir -p ~/tmp` guard is needed because `mktemp -p` fails (returns non-zero) if the target dir doesn't exist yet — and bootstrap.sh runs under `set -euo pipefail`. Since the B9 temp files are installer scripts (a few KB), this won't stress `~/tmp/`, but the pattern must be consistent with project rules.

---

Everything else is clean:

**Structural:** No nohup, no `task_states` direct insert, no `git add .`, no `pip install -e <worktree>` (the SkillSpector install is a regular `pip install`, not `-e`), no bare `except:`, no mutable defaults, no `os.killpg`, no raw SQL, no star imports. Import grouping in the new test file is correct (stdlib only, correctly ordered). Conventional commit prefix present (`fix:`). No `IntervalTrigger`, no `$CLAUDE_PROJECT_DIR` misuse.

**PII/Secrets/Wording:** Clean — no API keys, tokens, or secrets in added lines. All URLs are public. No private IPs, no hardcoded usernames, no private hostnames. PR title and body contain no language referencing credential leaks, security incidents, or private infrastructure.

---

WingedGuardian:
@codex review

PR #1218, opened 2026-07-23
feat(cc): foreground-session liveness reaper (D3)

WingedGuardian:
## What

The remaining half of the 2026-07-20 silent-death remediation (follow-up `19282498`). A foreground `cc_sessions` row stays `active` by design so the next turn can `--resume` it — so a crash / OOM / container restart mid-turn leaves it `active` **forever**, indistinguishable from a healthy resumable session (the incident row sat `active` 26h+ with the user never told). The `session_reaper` job hard-excludes foreground (`query_stale` filters `session_type != 'foreground'`), so nothing reaped them.

## How

`reap_dark_foreground` (new `cc/foreground_reaper.py`), wired into the **existing** `session_reaper` job in its own inner try/except (a notify/observation failure can never abort stale-session/heartbeat cleanup):

- Foreground rows idle > `idle_hours` (24h) → `checkpointed` via `checkpoint_dark` (rowcount-guarded). **Non-destructive**: `get_active_foreground` now matches `checkpointed`, and `get_or_create_foreground` flips it back to `active` on reuse, preserving `cc_session_id` — so `--resume` is untouched.
- Transcript-tail classifier (`dark_signal`): the **crisp** "unanswered user turn" signal → notify the origin user their request was interrupted (reusing the shipped `#1192` delivery path), **only** when the unanswered turn's own timestamp is older than the cutoff (age guard — excludes a mid-flight long turn) and it isn't already covered by a rate-limit park or dispatched session. The **fuzzy** "I'll report back" promise-regex → shadow-logged only, never notifies, until precision is measured.
- Observability-only — it never re-dispatches or auto-retries the dead work.

## Lever
`cc_foreground_reaper` (`off` / `observe` / `notify`, default `notify`) + `GENESIS_FOREGROUND_REAPER_DISABLED`. `observe` reaps + records without notifying.

## Deviation from the original audit
The audit named a `session.dispatched_work_dead` **event**; there's no event bus / consumer, so this uses `observations.create` (the established awareness surfacing pattern) instead.

## Tests
`test_foreground_reaper.py` (pure `dark_signal` + age guard; full reap matrix: skips-live, checkpoints-dead, notify-only-on-unanswered, age-guard-suppresses-midflight, suppressed-on-clean/covered-by-dispatch, promise=shadow, non-telegram-graceful, observe/off modes, resume-after-checkpoint) + CRUD tests (widening, `query_stale_foreground`, `checkpoint_dark` race guard) + the SessionManager checkpointed→active flip. 16 D3 tests + 95 neighboring green locally.

Reviewed pre-build by genesis-architect (DONE_WITH_CONCERNS; all 4 SHOULD-FIX adopted: age guard, kill lever, shared origin resolver reuse, inner try/except isolation).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **An interrupted request no longer vanishes silently.** If a message you sent
+  on Telegram was cut off before Genesis finished answering — a crash, a
+  restart, or the session going dark mid-turn — Genesis now notices the
+  abandoned session on its next hygiene pass and tells you it was interrupted so
+  you can re-send, instead of leaving you waiting on a reply that never comes.
+  Configurable via `cc_foreground_reaper` (`off` / `observe` / `notify`, default
+  `notify`); observability-only — it never silently re-runs the lost work.
 - **Ambient capture now alerts when auto-recovery can't bring a wedged device
   back.** For installs running the optional voice/ambient edge with device
   auto-recovery armed, Genesis now raises a capture-health alert when the device
--- a/config/cc_foreground_reaper.yaml
+++ b/config/cc_foreground_reaper.yaml
@@ -1,0 +1,23 @@
+# Foreground-session liveness reaper (D3).
+# See src/genesis/cc/foreground_reaper_config.py for the full contract.
+#
+# A foreground cc_sessions row stays 'active' by design so the next turn can
+# --resume it, so a crash/OOM/restart mid-turn leaves it 'active' forever. This
+# reaper relabels such dark rows to 'checkpointed' (non-destructive, still
+# resumable), records a dark-session observation, and — on the crisp
+# "unanswered user turn" signal — notifies the origin user that their request
+# was interrupted.
+
+enabled: true
+
+# off | observe | notify
+#   off     — do nothing (dark rows stay 'active', pre-D3 status quo)
+#   observe — reap + observe, but never notify the user
+#   notify  — observe + notify on the crisp unanswered-user signal
+mode: notify
+
+# A foreground row idle at least this long with no live turn is treated as dark.
+idle_hours: 24
+
+# Loud-truncation cap on dark rows processed per reaper pass.
+max_per_tick: 200
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -137,7 +137,7 @@ any task bigger than an LLM call.
 ```yaml subsystem-map
 entry: execution-cc
 modules: [cc]
-verified: 3c5e1f24 2026-07-22
+verified: 0a3fc19d 2026-07-23
 ```
 
 - `cc/direct_session.py` + `cc/conversation.py` (both >1000 LOC; split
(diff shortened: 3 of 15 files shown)


WingedGuardian:
## First-pass review: PR #1218 — feat(cc): foreground-session liveness reaper (D3)

### Structural review

**Pattern violations (ERROR):** NONE

**Import hygiene (WARNING):** NONE — stdlib → third-party → local ordering is correct throughout. Deferred imports inside async helpers (`_covered_by_other_subsystem`, `_notify_origin`) are appropriate for the circular-import avoidance pattern used elsewhere in the codebase.

**Missing coverage (WARNING):**

`src/genesis/cc/foreground_reaper_config.py` has no corresponding test file, and none of the 16 D3 tests exercise it directly. Every test in `test_foreground_reaper.py` passes an explicit `mode=` override to `reap_dark_foreground()`, which short-circuits `effective_mode()` before the config module is touched. The following branches in the config module are entirely untested:

- `effective_mode()` with `GENESIS_FOREGROUND_REAPER_DISABLED=1` in env → forced `off`
- `effective_mode()` with `enabled: false` in YAML → `off`
- `effective_mode()` with an invalid/unrecognised `mode` value → degrades to `observe`
- `effective_mode()` with `mode: off` as YAML-1.1 boolean `False` → the `if mode is False` branch
- `knob_int()` with a zero value, a negative value, a bool, or a non-int → fallback to `DEFAULTS`
- `load_config()` with a missing base file (the `FileNotFoundError` pass branch) and a corrupt file (the generic `except Exception` branch)

These degradation paths are the safety properties the module was explicitly designed around (the docstring enumerates them as the correctness argument). A `tests/test_cc/test_foreground_reaper_config.py` covering at least the env kill switch, invalid-mode degradation, YAML boolean False, and `knob_int` fallback would close the gap.

**Missing documentation (NOTE):** NONE — all new public functions carry docstrings.

**Genesis conventions (WARNING):** NONE — conventional commit prefix `feat(cc):` ✓, no `IntervalTrigger` (reaper wired into existing job), no hook scripts modified, no `$CLAUDE_PROJECT_DIR` in Python.

---

**Code diff (added lines):** CLEAN — no API keys, tokens, passwords, private IPs, IPv6 ULAs, SSH keys, DB connection strings, or environment variable values containing secrets. `Path.home() / ".claude" / "projects" / cc_project_dir() / ...` uses a dynamic resolver, not a hardcoded username path.

**Soft warnings:** NONE — no personal email addresses, no `/home/<username>/` absolute paths, no non-config timezone identifiers, no AWS/GCP account IDs, no hardware identifiers, no private hostnames.

**PR title / body / CHANGELOG (CRITICAL):** CLEAN — describes user-facing session-resumability behaviour only. No language referencing a credential leak, PII exposure, or private infrastructure details. The "2026-07-20 silent-death" label is a functional incident description, not a security disclosure.

---

### Summary

| Domain | Result |
|--------|--------|
| Structural ERRORs | NONE |
| Structural WARNINGs | 1 — missing test coverage for `foreground_reaper_config.py` |
| PII / Secrets / Wording | CLEAN |

Single actionable item: add a `test_foreground_reaper_config.py` covering the env kill switch, invalid-mode degradation, YAML boolean False, and `knob_int` fallback paths before merge.

---

WingedGuardian:
Addressed both Codex P2s (commit aaacc5f3):
- **dispatched queue rows** (`direct_session_queue.py`): `has_open_for_origin` now counts only `pending`/`claimed` as open — `dispatched` persists permanently (no terminal status), so treating it as open would suppress the notify forever for any origin that ever dispatched work. Regression tests: `test_dispatched_dsq_does_not_suppress` + `test_pending_dsq_still_suppresses`.
- **delivery failure preserving the alert** (`foreground_reaper.py`): `_notify_origin` no longer propagates a `submit_urgent` failure, and the dark-session observation's HIGH priority is now driven by notify-*eligibility* rather than delivery success — so a failed Telegram delivery (row already checkpointed) still surfaces the dead request via the morning report instead of being lost. Regression test: `test_notify_delivery_failure_preserves_alert` (asserts priority=high, no crash).

PR #1221, opened 2026-07-23
feat(hooks): guard against the Bash-tool-timeout deploy footgun

WingedGuardian:
## Why
During the CC 2.1.218 deploy, `scripts/update.sh` was SIGKILLed at 120s because the Bash **tool's** own timeout defaults to 120s and an inner `timeout 900` does **not** extend it — and `update.sh`'s SIGTERM traps then rolled it back mid-deploy. Avoidable, and worth preventing structurally rather than relying on a per-install CC memory.

## What
- **`genesis-development` skill → Timeout Policy:** documents that the Bash tool timeout is separate + defaults to 120s (inner `timeout N` doesn't extend it), and long deploys (`update.sh` / `bootstrap.sh` / `host-setup.sh` — container align + guardian redeploy + host `update-node`/`update-cc`, sequential, can exceed 900s) must run via **`run_in_background: true`**, never a foreground timeout or `nohup`.
- **New non-blocking PreToolUse(Bash) hook `cc-deploy-timeout-guard`:** nudges toward `run_in_background` when a deploy script is run in the foreground. Silent when `run_in_background: true` or on non-deploy commands. **Always exits 0 — it can never block a command.**

## Verification
Hook tested standalone: foreground `update.sh`/`host-setup.sh` → advisory; `run_in_background:true` → silent; non-deploy → silent; settings.json valid. This is versioned + all-installs (skill + repo hook), unlike the foreground-only CC memory.

### DIFF
--- a/.claude/hooks/cc-deploy-timeout-guard
+++ b/.claude/hooks/cc-deploy-timeout-guard
@@ -28,5 +28,12 @@ esac
 BG=$(printf '%s' "$IN" | jq -r '.tool_input.run_in_background // false' 2>/dev/null)
 [ "$BG" = "true" ] && exit 0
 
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

PR #1227, opened 2026-07-24
fix(hooks): repair silently-inert CC guard hooks (dead CLAUDE_TOOL_INPUT env)

WingedGuardian:
## Problem

Claude Code delivers each hook's payload as **JSON on stdin**, tool args nested under `tool_input`. An earlier generation of Genesis hooks read a `CLAUDE_TOOL_INPUT` env var holding just the tool-input dict. Current CC (2.1.x) no longer sets it — so every hook that relied on it read an empty value and **fell open silently**.

**Verified live** (headless `claude -p`, real payloads) that the safety guards — push/merge gate, destructive-`rm` guard, protected-path write guard, worktree guards — no longer fired. Each guard's block *logic* is correct (they block when fed a proper payload); the root cause is purely the dead input contract. Two more env vars (`CLAUDE_TOOL_USE_RESULT`, `CLAUDE_SESSION_ID`) were dead the same way, and several hooks also read the command at the top level instead of the nested `tool_input.command`.

This stayed hidden because the hooks' own tests fed the dead env var and **passed against an interface CC no longer uses** — a false green.

## Fix

- **`scripts/hooks/hook_input.py`** — single source for the payload contract. `read_payload()` reads stdin (legacy env fallback); `field()`/`tool_response()`/`session_id()` extract fields and bridge the new nested shape and the legacy flat shape. Warns on a non-empty-but-unparseable payload so a broken contract can't silently fail open again. Self-locating import so it resolves both when run as a script and when imported as a module (tests).
- Route all 15 affected hooks + 2 inline `settings.json` guards through it; drop the `echo "$CLAUDE_TOOL_INPUT" | …` wrappers that clobbered CC's real stdin. Same fix applied to the deprecated `config/claude-settings.json.template`.
- **`tests/test_scripts/test_hook_input_contract.py`** — feeds each safety-critical guard a real captured payload and asserts it blocks; static guards forbidding any hook from reading a dead payload env var, or any settings JSON from piping it. Install-agnostic triggers.
- **`tests/test_scripts/test_hook_input.py`** — dedicated unit coverage for the shared helper.
- Migrate existing hook tests off the dead env-var contract.
- Docs: `.claude/hooks/README.md` authoring rule, `docs/reference/cc-compatibility.md` incident + checklist step, CHANGELOG.

Large per-file diffs are formatter reflow (quote style, arg wrapping); the only semantic change per hook is the input-parsing swap.

## Verification
- `pytest tests/test_hooks/ tests/test_scripts/test_hook_input*.py` → all green (incl. module-import + foreign-`$HOME` cases).
- Each safety guard re-tested live on the real stdin channel: now blocks.

## Note on the merge gate
The push/merge review-gate hook is itself among the inert hooks, so the automated gate cannot enforce on this PR — it can't police its own repair. Requires human review sign-off.

Ledger: 98ef51b9a42241e09f56c9ab5de326a0


WingedGuardian (commented):
## Automated code review — PR #1227

### Structural review

**WARNING — Missing unit tests for new shared module**

`scripts/hooks/hook_input.py` is a new shared helper that all 15+ hooks now depend on for input parsing. There is no dedicated unit test file for it (`tests/test_scripts/test_hook_input.py` or `tests/test_hooks/test_hook_input.py`). `test_hook_input_contract.py` tests the end-to-end guard behavior, but doesn't exercise the individual helper functions in isolation:

- `_loads()` with non-dict payloads, truncated JSON, deeply nested objects
- `field()` with `None` values, integer values, missing `tool_input` key entirely
- `tool_response()` when `tool_response` key is `None` vs absent vs a non-dict
- `session_id()` with an empty-string `session_id` field (falls through to env or default)

These edge cases currently have implicit coverage only (tested via the guards they serve). Since `hook_input.py` is now the single point of failure for all guard input reading, a regression in any of these functions would silently fall open again — exactly the failure mode this PR was written to fix.

No other structural violations found. Import ordering uses the `sys.path.insert` + `# noqa: E402` pattern correctly. No bare `except:`, mutable defaults, `nohup` service starts, `task_states` inserts, raw SQL, `/tmp/` misuse, `git add .`, or force-push patterns introduced.

---

**SOFT WARNING — PR body and CHANGELOG describe specific historical bypass details**

The PR body's "Verified live" section lists four specific commands that were confirmed unblocked before this fix, including `rm -rf /home/ubuntu` and `Write to */secrets.env (CRITICAL)`. The CHANGELOG entry says the hooks "waved everything through." Both are now permanently in git history.

For a personal private repo this is low risk (any attacker who could abuse this knowledge would need shell access already), but the detail level is higher than typical for security-adjacent fix descriptions. Worth awareness. No action required unless the repo has public visibility or a broader audience.

`/home/ubuntu` paths in test fixtures are correctly classified as false positives (test case data, not real paths).

**Code diff: CLEAN** — No API keys, private IPs, hardcoded credentials, SSH keys, or private hostnames in added lines.

---

### Summary

| Domain | Verdict |
|--------|---------|
| Structural | WARNING (missing hook_input unit tests) |
| PII/Secrets code | CLEAN |
| Wording (PR + CHANGELOG) | SOFT WARNING (bypass detail level) |

No ERRORs, no hard blocks. Fine to merge pending the test coverage note.

---

PR #1228, opened 2026-07-24
feat(observability): wire 90-day retention for the events bus table (PR-3)

WingedGuardian:
## Summary

The observability `events` table — the record of everything Genesis notices and
does — was the last high-volume table with **no retention**. `crud.events.prune`
existed with **zero callers**, and the table had grown to 45k+ rows over ~108
days (~12× month-over-month). This wires a bounded 90-day retention.

## What changed

- **`events_prune` drip job** added to `_wire_drip_retention_jobs`
  (`runtime/init/learning.py`) — the in-process retention pattern, mirroring its
  six sibling prunes (execution_traces, cost_events, …). This is DB-table
  retention, so it belongs with its siblings, not in `disk_hygiene.sh`
  (filesystem hygiene).
- `crud.events.prune` takes an **ISO `older_than` cutoff**, not a `days=` arg
  like the siblings; computed from `datetime.now(UTC) - 90d`. Every
  `events.timestamp` is stored as a UTC isoformat (verified: 45,627/45,627 rows
  end in `+00:00`), so the DELETE's lexical `timestamp < cutoff` is a correct
  chronological comparison. Consumers only ever read recent rows (`ORDER BY
  timestamp DESC LIMIT`), so a 90-day window breaks nothing.
- The job records failures via `record_job_failure(exc=exc)`, so the retention
  job's own failures are diagnosable through the failure funnel.

## Verification

- **Tests:** crud prune coverage (strict `<` boundary, `event_type` scope, empty
  no-op) via canonical `create_all_tables()`; `events_prune` added to the
  registration test's expected set.
- **Wired-job E2E:** invoked the actual job closure against a seeded DB — a
  100-day-old event pruned, a 5-day-old event kept, success recorded.
- Registered at 06:00 (the 05:00–05:50 drip slots are taken; `voice_hygiene`
  owns 05:50) so two DB-writing jobs don't fire concurrently.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,12 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Changed
 
+- **Genesis now keeps its own internal event log from growing without bound.**
+  The observability event stream — the record of everything Genesis notices and
+  does — was the last high-volume table with no cleanup, growing steadily on
+  disk. A daily maintenance pass now trims events older than 90 days, matching
+  the retention already applied to Genesis's other internal logs. No visible
+  change to day-to-day use; it just stops a slow disk leak.
 - **Proactive memory recall holds up better when several sessions are active at
   once.** When multiple Claude Code sessions run side by side, each prompt's
   memory recall used to spend much of its time-budget on bookkeeping (usage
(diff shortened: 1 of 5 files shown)


WingedGuardian:

Checked against all Domain 1 rules (nohup, task_states direct insert, /tmp usage, bare except, mutable defaults, IntervalTrigger misuse, raw SQL, import hygiene, commit prefix, test coverage). Checked all added lines and PR metadata for Domain 2 (secrets, private IPs, PII, wording). Both passes clean.

---

PR #1229, opened 2026-07-24
fix(hooks): stop revived guards blocking benign redirects/logical-OR

WingedGuardian:
## What

Fixes false-positives in two Claude Code PreToolUse safety guards that PR #1227 revived from a silently-inert state. With the guards working again, two pre-existing over-eager checks surfaced and began blocking legitimate commands every session.

**1. `destructive_command_guard.py`** (blocks recursive+force `rm` on broad/shallow paths) mis-parsed shell **redirections** (`2>/dev/null`, `>log`, `2>&1`), bare **newlines**, and `\`-newline **continuations** as if they were `rm` targets — so a safe deep-path cleanup got depth-checked against, e.g., `2>/dev/null` (depth 3) and blocked.

**2. `.claude/settings.json` inline guard** ("don't background a piped command") used `grep -qF "|"`, which also matched the `|` inside a logical-OR `||` — so a backgrounded `a || b` was falsely blocked.

## How

`destructive_command_guard.py` now treats **`shlex.split` as the sole authority on quoting/escaping** (only newlines / line-continuations are pre-mutated, inserting spaces/`;` that cannot corrupt quote balance). A token is skipped as a redirection iff it **starts with** a redirect operator (`\d*[<>]` or `&>`), and a token **following** a redirect is never skipped.

**Safety proof:** a skipped token always starts with a redirect operator char, which no catastrophic target (`.` `..` `/` `~` `*` or any real filesystem path) ever does — so skipping can never drop a genuinely dangerous target; and because a following token is never skipped, a real dangerous path after a redirect-shaped token still blocks. **Accepted, documented tradeoff:** deleting a bizarrely-named file that *starts with* a redirect char is allowed (non-catastrophic, and blocking it would re-break the real redirect false-positive). The separator regex was also fixed so a background `&` glued after a digit splits correctly and a second command after it is still caught.

The inline guard strips `||` before its pipe test, so a logical-OR passes while a real pipe (`a | b`, `a |& b`) is still blocked.

## Review & tests

Reviewed adversarially across 4 rounds — each round found and fixed a real bug (a quote-eating regex, an escaped-quote divergence, an unanchored prefix-match), converging on the design above with a safety proof independently verified against the always-block set and real paths. 146 regression tests cover every construct, including all bypasses found during review. `ruff` clean.

## Fixed

- Genesis's revived guardrails no longer block harmless commands: safe deep-path `rm -rf` cleanups with redirects/newlines, and backgrounded logical-OR commands. Genuinely dangerous deletes (`rm -rf /`, `~`, `.`, `..`, `*`, real shallow/deep paths, with or without a redirect) are still stopped.

Ledger: 229751d093434c778db0679c10a1d4da


WingedGuardian:

---

**Structural pass** (`.claude/settings.json`, `scripts/hooks/destructive_command_guard.py`, `tests/test_hooks/test_inline_hooks.py`, `CHANGELOG.md`)

- Hook imports: `destructive_command_guard.py` uses only `re` and `shlex` (stdlib). No heavy dependencies added. ✓
- Pattern violations: None found (no nohup, no task_states inserts, no `/tmp/`, no bare `except:`, no mutable defaults, no raw SQL, no force pushes). ✓
- Import hygiene: No new imports added. ✓
- Test coverage: 186 lines of regression tests cover the redirect-stripping and logical-OR false-positive cases, with paired dangerous-variant tests confirming blocks still hold. ✓
- Documentation: New module-level constants (`_REDIR_TOKEN`, `_SEPARATOR_SPACING`) have detailed design comments and safety proofs inline. ✓
- Commit prefix: `fix(hooks):` — correct. ✓

**PII/Secrets/Wording pass**

- Added lines: No secrets, tokens, or credentials. Test paths (`/home/u/proj/...`) use a placeholder username and appear in test fixtures (known false positive). ✓
- PR title/body: Describes a shell parsing false-positive fix; no security incident language, no private infrastructure names, no credential-exposure wording. ✓
- CHANGELOG: Same — clean phrasing. ✓

---

PR #1238, opened 2026-07-28
fix(inbox): re-inject standing bracket directives on every delta eval

WingedGuardian:
## Problem

Inbox notepad files carry standing whole-line bracket directives — e.g. a
capabilities notepad topped with `[If it's in here, default to building it]` —
that express file-scoped intent per `INBOX_EVALUATE.md` Rule 1 (including the
capability-build lane).

The delta scanner's baseline union (`_merge_evaluated_content`) folds the
directive line into `evaluated_content` after the **first** evaluation. Every
later delta then computes against a baseline that already contains the
directive, so it's stripped from the prompt — and the prompt preamble forbids
reading the source file. **The evaluator never sees the standing directive
again after eval #1.** This silently disabled the build-biased disposition for
weeks (the root cause behind the "fixed-mindset" evaluations).

## Fix

`_build_prompt` now re-reads each item's **current** source file at dispatch
time, extracts whole-line `[ ... ]` directives, and surfaces them in a labeled
**"Standing file directives (context only — NOT items to evaluate)"** section:

- Sanitized through the same `ContentSource.INBOX` perimeter wrapper as item
  content (external-untrusted; wrapped in `<external-content>`), deduped per
  source file.
- Re-reading at dispatch (rather than persisting at scan) gives latest-intent
  semantics and automatically covers the resumed-batch and retry dispatch
  paths.
- A missing/renamed/unreadable file degrades to no-section, never a raise.

**Delta computation and baselining are intentionally untouched** — directives
are prompt-only context, never re-evaluated as items or re-baselined. A
one-paragraph note in INBOX_EVALUATE.md Rule 1 tells the evaluator the section
is authoritative Rule 1 input but must not be restated or evaluated.

## Tests

- `test_extract_bracket_directives_shapes` / `_empty` — regex rejects markdown
  links (`[x](url)`) and mid-sentence brackets; dedupes; whitespace-tolerant.
- `test_build_prompt_*` — injects on delta; no section without brackets;
  missing-file graceful; sanitizer-wrapped; per-file in multi-file batches.
- `test_bracket_directive_renders_on_second_delta_eval` (end-to-end regression)
  — reproduces the exact bug: on the second eval the delta carries only the new
  line **and** the directive is re-injected.
- `test_resume_dispatch_reads_current_file_directives` — directives surface on
  the parked/approved resume path too.

314 inbox tests green; ruff clean. Replayed the extractor over the real inbox
files: 1 directive each in the 3 files that have one, 0 false positives.

Inline code-review (feature-dev:code-reviewer): DONE, no findings >=80%
confidence.

Ledger: 25568aa11b7444399a3a72cbf595c7a8

### DIFF
--- a/src/genesis/identity/INBOX_EVALUATE.md
+++ b/src/genesis/identity/INBOX_EVALUATE.md
@@ -195,6 +195,13 @@ If the file content contains a bracketed classification directive —
 or any close variant — your classification for every item in that file is
 already decided. This is NOT a heuristic the content can override.
 
+> **On delta evaluations the bracket may not be in the item content.** These are
+> standing, file-scoped directives, so the monitor re-reads the source file and
+> re-supplies them to you under a **"Standing file directives"** section (each
+> block labeled with its filename). Treat that section as authoritative Rule 1
+> input for every item from the SAME file — but it is context only: do NOT
+> evaluate those bracket lines as items, and do NOT restate them in your output.
+
 **Decision procedure when a bracket directive is present:**
 
 1. Stop reasoning about whether the content "really" fits a different
--- a/src/genesis/identity/INBOX_EVALUATE.md
+++ b/src/genesis/identity/INBOX_EVALUATE.md
@@ -195,12 +195,14 @@ If the file content contains a bracketed classification directive —
 or any close variant — your classification for every item in that file is
 already decided. This is NOT a heuristic the content can override.
 
-> **On delta evaluations the bracket may not be in the item content.** These are
-> standing, file-scoped directives, so the monitor re-reads the source file and
-> re-supplies them to you under a **"Standing file directives"** section (each
-> block labeled with its filename). Treat that section as authoritative Rule 1
-> input for every item from the SAME file — but it is context only: do NOT
-> evaluate those bracket lines as items, and do NOT restate them in your output.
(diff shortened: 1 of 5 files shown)


WingedGuardian (commented):
Structural review pass + PII/Secrets/Wording scan — PR #1238.

---

**Structural review: 1 NOTE**

NOTE — `tests/test_inbox/test_drop_dispatch.py` (`test_resume_dispatch_reads_current_file_directives`): `import hashlib` is placed inside the test function body rather than at the module's top-level import block. Unlike the `from genesis.inbox.monitor import …` local imports in other tests (which carry a circular-import avoidance rationale), `hashlib` is stdlib with no such concern. It should move to the file's top-level stdlib block alongside the existing `import` statements.

No ERRORs or WARNINGs found. Pattern check clean: no bare `except:`, no mutable defaults, no `/tmp/` usage, no raw SQL, no `nohup`, no force-push artifacts, no `IntervalTrigger`. Coverage: both new functions (`_extract_bracket_directives`, `_read_standing_directives`) have dedicated tests. Docstrings present. Commit prefix `fix(inbox):` ✓.

---

No secrets, credentials, private IPs, personal emails, private hostnames, or hardware identifiers found in added lines. PR title and body contain no security-incident language or infrastructure-exposure wording.

---

PR #1240, opened 2026-07-28
feat(inbox): deterministic memory persistence over eval output + honest extraction taxonomy

WingedGuardian:
> **Stacked on #1238** (`fix/inbox-standing-directives`). Base is that branch, so
> the diff shows only this PR's work; GitHub retargets to `main` when #1238 merges.

## Part 1 — deterministic memory persistence over the curated output

After the inbox evaluator writes its `.genesis.md` output, the monitor now runs a
**deterministic** extraction over *that curated text* and stores the durable
insights as memories — replacing the evaluator agent's Step-4 `memory_store`
self-persistence (non-deterministic, and the root of the output-ordering footgun).

- New `src/genesis/inbox/eval_memory.py`: `build_eval_memory_prompt` +
  fail-closed `parse_eval_memory_response` + `extract_and_store_eval_memories`.
  Routes via the free `9_fact_extraction` chain; stores `user_signal` /
  `architecture_insight` memories with `source="inbox_evaluation"`,
  `origin_class=external_untrusted`, source-overlap verification + cross-session
  dedup, `force_fts5_only`.
- Fires as a detached `tracked_task` **after** baseline + session-complete
  (mirrors `_fire_triage`) — structurally cannot affect the batch. No-op unless
  `router`+`memory_store` are wired (`runtime/init/inbox.py`). Kill switch:
  `GENESIS_INBOX_EVAL_MEMORY_DISABLED=1`.
- **Provenance**: deliberately extracts the *curated output*, NOT the raw
  transcript — an inbox transcript is dominated by fetched untrusted web content,
  so mining it would launder "what a webpage said" into memory.
- `INBOX_EVALUATE.md` Step 4 rewritten to forbid agent `memory_store` for
  evaluation findings and document the automatic path (keeps only the optional
  `observation_write` user-signal duty).

## Part 2 — honest extraction taxonomy (a months-old silent bug)

The 2-hourly `run_extraction_cycle` filtered sessions by
`_EXTRACTABLE_SOURCE_TAGS = {"foreground","inbox","voice"}`, but the monitor tags
inbox sessions `"inbox_evaluation"` (never `"inbox"`). Verified against prod: 320
`inbox_evaluation` rows, **zero** `"inbox"` rows ever — so the cycle matched no
inbox sessions for months while its docstring *claimed* inbox coverage.

- Drop the phantom `"inbox"`; add `_EXCLUDED_SOURCE_TAGS` +
  `_EXCLUDED_SOURCE_TAG_PREFIXES` classifying **every** production `source_tag`
  family as a deliberate exclusion (`inbox_evaluation` excluded on purpose — Part
  1 is its memory path). Fix the docstring.
- Coverage-guard test **derives** its known set by scanning `src/` for
  `source_tag=` literals (comments stripped) — not a hand-maintained snapshot, the
  gap that let `inbox_evaluation` and `mail_reply` slip. Explicit assertion for
  known raw-SQL/indirect writers (`priority_*` from `cc_budget.py`), with the
  docstring honest about the regex's scope.

## Tests / review

18 eval-memory + 3 monitor-wiring + coverage-guard tests; **352** inbox +
extraction tests green post-rebase; ruff clean. Two `feature-dev:code-reviewer`
passes; both SHOULD-FIXes (`mail_reply` gap, `priority_` raw-SQL writer) fixed and
re-verified. Dead-code `record_session_start` tabled (pre-existing, fails closed).

**Live prompt-calibration + full E2E run post-deploy** (the standalone harness
can't authenticate providers — the `API_KEY_*`→litellm mapping only happens in
full runtime init; production routing works). This is the one verification step
that requires the deployed server.

Ledger: 509f363591744a299cbb5910c2a916b5

### DIFF
--- a/src/genesis/identity/INBOX_EVALUATE.md
+++ b/src/genesis/identity/INBOX_EVALUATE.md
@@ -16,10 +16,12 @@ You have access to Genesis MCP servers (genesis-health + genesis-memory):
 - **`memory_recall`** — Semantic search across Genesis's memory. Use to assemble
   user context before applying the User Evaluation Framework (query for topics
   related to the content being evaluated).
-- **`memory_store`** — Store findings as episodic memories. Use after evaluation
-  to persist knowledge (see "Knowledge Extraction" section below).
+- **`memory_store`** — Do NOT use for inbox evaluation. Knowledge persistence is
+  automatic (see "Step 4"); calling it for evaluation findings is redundant and
+  reintroduces the output-ordering footgun.
 - **`observation_write`** — Write typed observations. Use to store `user_signal`
-  observations when evaluations reveal user interests/goals.
+  observations when evaluations reveal user interests/goals (your only
+  persistence duty — do it before your final text).
 - **Health tools** — `genesis_status`, `health_status`, etc. for system context.
 
 ## Critical Rules
@@ -67,10 +69,13 @@ discarded.
 **This means you MUST structure your work in this order:**
 
 1. **Tools first** — fetch all URLs, recall memory, read files, run searches
-2. **Knowledge persistence second** — call `memory_store`, `observation_write`
+2. **Observation second** — the only pre-text tool call left is an optional
+   `observation_write` (`user_signal`); do NOT call `memory_store` (persistence
+   is automatic — see "Step 4")
 3. **Full evaluation text LAST** — your final text output must be the complete
    structured evaluation, because it is the ONLY thing written to the response
-   file that the user reads
+   file that the user reads (and the only thing the automatic memory extractor
+   sees)
 
 **The failure mode this prevents (observed in production 2026-04-18):**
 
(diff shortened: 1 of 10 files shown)


WingedGuardian:
**Automated first-pass review — structural + PII/wording**

---

### Structural review

**Pattern violations (ERROR) — 1 finding**

**`src/genesis/inbox/eval_memory.py` — `build_eval_memory_prompt`: format-string crash on evaluation text containing braces**

```python
def build_eval_memory_prompt(evaluation_text: str, *, source_name: str) -> str:
    return _PROMPT_TEMPLATE.format(
        source_name=source_name,
        max_memories=MAX_EVAL_MEMORIES,
        evaluation=evaluation_text,   # ← untrusted/LLM-generated data passed to str.format()
    )
```

`evaluation_text` is Genesis's own analysis of inbox items. Inbox evals routinely discuss code snippets, JSON examples, dict literals, and template strings — all of which can contain `{` or `}` characters. When they do, `str.format()` raises `KeyError` or `IndexError` (e.g. `{key}` in a code example is treated as a format variable).

This call is **outside** the `try` block that wraps `router.route_call`, so the exception bubbles up to `_persist_eval_memories`'s outer `except Exception` and is silently logged as a warning. Memory persistence silently does nothing for any evaluation text containing braces — the feature is broken for a common class of real inputs.

**Fix:** Escape braces in the data field before formatting, or build the prompt by concatenation so the data section bypasses `str.format()`:

```python
# Option A — escape brace literals in the data value before .format()
evaluation=evaluation_text.replace("{", "{{").replace("}", "}}"),

# Option B — split head (format-safe) from data (appended directly)
_PROMPT_HEAD = "...\n\nEVALUATION:\n"
return _PROMPT_HEAD.format(source_name=..., max_memories=...) + evaluation_text
```

The existing tests don't catch this because all test evaluation strings are brace-free. A test with `evaluation_text = 'The API returns {"items": []}.'` would confirm the fix.

---

### Notes (non-blocking)

**`tests/test_inbox/test_edge_cases.py` — `asyncio.sleep(0.05)` as CI timing fence**

The three new wiring tests use a 50 ms sleep to let detached fire-and-forget tasks settle before asserting. Under CI load an overloaded event loop may not have scheduled the task in time, making these intermittently flaky. Capturing the `tracked_task` return value and awaiting it (or wrapping with a short `asyncio.wait_for`) would be more robust. Low risk, non-blocking.

---

**Import hygiene:** ✓ Lazy imports inside the coroutine, correct stdlib → local ordering at the top level, no star imports, no circular import risks visible.

**Missing coverage:** ✓ `eval_memory.py` → `test_eval_memory.py`; new monitor paths → `test_edge_cases.py`; taxonomy changes → `test_extraction_job.py`.

**Documentation:** ✓ Module, class, and all public function docstrings present.

**Genesis conventions:** ✓ `feat(inbox):` prefix; no `IntervalTrigger`, no hook scripts, no `$CLAUDE_PROJECT_DIR`.

---

### PII / secrets / wording scan

**Code diff:** Clean. Test fixtures are synthetic ("User likes local STT.", "User is into RevOps automation.", `sess-9`). No API keys, private IPs, hardcoded user-home paths, credentials, or hardware identifiers in added lines.

**PR title / description / CHANGELOG:** Clean. No security-incident language, no leaked-credential references, no private infrastructure details. The production metric ("320 inbox_evaluation rows, zero 'inbox' rows") describes a bug discovery, not a data exposure.

---

**Overall: ⚠️ ERROR — fix required before merge.** One correctness bug (format-string crash on brace-containing evaluation text) must be addressed. The timing-sensitivity note is non-blocking.

---

WingedGuardian (commented):

---

PR #1243, opened 2026-07-28
fix(ego): make ego_directive + ego_proposal_resolve user-only (provenance)

WingedGuardian:
## Problem

An autonomous background session called the `ego_directive` MCP tool, which hardcodes `source="user"` on the directive it writes — so a directive the user never issued was recorded as a user instruction. Root cause proven via `cc_sessions`: only autonomous sessions were alive when the mis-attributed directive was created; zero foreground. The same provenance hole exists via `ego_proposal_resolve`, whose withdrawn-proposal re-validation path also writes `source="user"` directives, and neither tool was blocked from the autonomous ego cycle.

## Fix

Invariant enforced: **no autonomous/dispatched session may create a `source="user"` directive or resolve proposals** — both are user authority. Two thin layers, no schema change:

1. **Hard gate** at the top of `ego_directive` and `ego_proposal_resolve` on `is_dispatched_session_env()` (`GENESIS_CC_SESSION` set AND `GENESIS_SESSION_SUPERVISED` unset) → return a `refused` status before any DB work. Fail direction is safe: env absent = human terminal = allowed; an owner-attended conversation (`SUPERVISED=1`) is allowed.
2. **Defense-in-depth**: both tool names added to `_EGO_CYCLE_DISALLOWED_TOOLS`, so the ego cycle can't call them regardless.

The dashboard and Telegram resolution paths call the CRUD/resolution layer directly (not the MCP tool), so they are unaffected.

## Tests

6 new tests: refuse-when-dispatched (incl. a zero-write assertion proving the gate fires before DB access), allow-when-supervised, allow-when-no-CC-session, resolve refuse/allow, and disallow-list membership. 137 tests green across `test_ego_tools` / `test_ego_proposal_resolve_targeting` / `test_session` / `test_redteam_enforce`.

## Review

Code-reviewer: DONE, no BLOCKER/SHOULD-FIX/NOTE. Independently verified gate placement (inert for legitimate callers), the fail direction, no circular import, the dashboard path being unaffected, correct tool-name format, and that all four `source="user"` writers are accounted for (three gated here; the fourth is the genuine Telegram user-conversation flow, correctly out of scope).

### DIFF
--- a/src/genesis/ego/session.py
+++ b/src/genesis/ego/session.py
@@ -102,6 +102,12 @@ _EGO_CYCLE_DISALLOWED_TOOLS = (
     # class as the goal tools above). Capture happens in user-facing
     # sessions (conversation, resolution paths), never inside a cycle.
     "mcp__genesis-health__ego_decision",
+    # Directives and proposal resolution are USER authority — both stamp
+    # source="user" on what they create (a directive; a withdrawn -> re-validation
+    # directive). The ego cycle must never forge either. Belt to the per-tool
+    # is_dispatched_session_env() gate's suspenders.
+    "mcp__genesis-health__ego_directive",
+    "mcp__genesis-health__ego_proposal_resolve",
 )
 
 # Per-cycle cap on autonomous own-goal creations (genesis ego). One per cycle
--- a/src/genesis/mcp/health/ego_tools.py
+++ b/src/genesis/mcp/health/ego_tools.py
@@ -117,6 +117,11 @@ async def ego_directive(
     to its reasoning but decides what to propose. Use this when you want
     the ego to pay attention to something specific.
 
+    **User-only.** Directives are recorded as user instructions
+    (source="user"). This tool is REFUSED from autonomous/dispatched sessions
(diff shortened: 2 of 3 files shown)


WingedGuardian:
## Code Review — PR #1243

### Structural: 1 NOTE

**NOTE** — `ego_proposal_resolve` docstring not updated to reflect new user-only restriction (`src/genesis/mcp/health/ego_tools.py`, around line 593).

`ego_directive` received a `**User-only.**` docstring block explaining the provenance invariant and the refused-from-dispatched-sessions behavior. `ego_proposal_resolve` had the same gate added, same behavior change, but its docstring was not updated. A future caller reading just the docstring won't see the restriction until they hit the runtime refusal. Low risk given the error message is self-documenting, but worth keeping the two tools consistent.

No ERRORs. No WARNING-level import hygiene or convention violations. No missing test coverage (6 new tests cover refuse/allow/disallow-list paths).

---

No secrets, credentials, private IPs, usernames, or private hostnames in added lines. PR title and body use neutral, non-revealing language — no wording that would flag a credential leak or security incident in git history.

---

PR #1244, opened 2026-07-28
fix(hooks): per-worktree review markers so concurrent sessions don't clobber

WingedGuardian:
## Problem

The commit review-gate (`review_state.py` + the enforcement hooks) stored its
state in a **single global file** `~/.genesis/review_state.json` (+ evidence
`last_code_review.txt`). Concurrent Claude Code sessions each work in their own
git worktree but share one `~/.genesis`, so:

- Session B's `mark`/commit **reset session A's marker** mid-workflow — A's
  already-reviewed commit then got blocked. (Hit repeatedly while landing the
  inbox-evaluator stack this session.)
- The commit hook called `is_review_current()` with **no cwd**, so it hashed the
  *process* cwd's staged diff, not the worktree being committed — a latent bug.

## Fix

- **Per-worktree marker**: `~/.genesis/review_markers/<sha12(git-worktree-root)>.json`
  via `_worktree_key(cwd)` / `_state_file(cwd)`, with a **read-only** fallback to
  the legacy global file so an in-flight pre-upgrade review isn't lost.
- **Thread `cwd`** through `is_review_current`, `has_valid_review_marker`,
  `mark_reviewed`; the commit hook now passes the worktree cwd (also fixing the
  no-cwd bug).
- **Post-commit invalidation** (`review_invalidate_on_commit.py`): new
  `clear_marker(cwd)` deletes the **per-worktree** marker. Without this, the hook
  would keep deleting only the retired global file — silently disabling
  invalidation and letting one review authorize every subsequent `git add &&
  git commit` for the marker's 30-minute TTL. *(Caught as a BLOCKER in review.)*

Advisory nudge hooks (`review_enforcement_prompt.py`, `genesis_stop_hook.py`)
correctly stay at `cwd=None` — they key on the live session's own cwd and parse
no command.

## Tests

`tests/test_hooks/test_review_enforcement_commit.py` (25 pass): per-worktree
isolation (a concurrent worktree's mark can't clobber this one), distinct marker
files per worktree, a reviewed commit is allowed, and **commit → per-worktree
marker cleared → next commit re-blocked**. Install-agnostic (throwaway git repos,
`HOME` redirected).

## Deploy

Pure hooks/scripts — takes effect at next CC session start (no runtime restart).
The threat model is a self-discipline gate, not a security boundary; the
fail-open `"default"` key for non-git contexts is acceptable and documented.

### DIFF
--- a/scripts/review_enforcement_commit.py
+++ b/scripts/review_enforcement_commit.py
@@ -132,9 +132,9 @@ def main() -> None:
     if stages_in_same_command:
         # Can't check diff hash (staging hasn't happened yet) — require the
         # marker file to exist and not be expired.
-        rule2_blocks = not has_valid_review_marker()
+        rule2_blocks = not has_valid_review_marker(cwd=cwd)
     else:
-        rule2_blocks = has_code_changes(cwd=cwd) and not is_review_current()
+        rule2_blocks = has_code_changes(cwd=cwd) and not is_review_current(cwd=cwd)
 
     if rule2_blocks:
         # A trailing '# review-override' comment (outside quotes) acknowledges
--- a/scripts/review_invalidate_on_commit.py
+++ b/scripts/review_invalidate_on_commit.py
@@ -23,8 +23,23 @@ from pathlib import Path
 sys.path.insert(0, str(Path(__file__).resolve().parent / "hooks"))
 from hook_input import field, read_payload, tool_response  # noqa: E402
 
+# review_state lives in scripts/ (this file's own dir).
+sys.path.insert(0, str(Path(__file__).resolve().parent))
+from review_state import clear_marker  # noqa: E402
+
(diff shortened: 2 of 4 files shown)


WingedGuardian:

> **NOTE** (`tests/test_hooks/test_review_enforcement_commit.py`, new `_second_repo` helper): `_git(r, "add", "-A")` is called twice, which matches the `git add -A` ERROR pattern. However, this is test fixture code constructing throwaway git repositories under pytest's `tmp_path` with a single known file (`g.txt`) — not the genesis project repo. It continues the pattern already present in the existing `repo` fixture in the same file. No sensitive files can be accidentally staged. In future test helpers, prefer `_git(r, "add", "g.txt")` to be explicit.

- No API keys, tokens, credentials, or private IP addresses in added lines.
- `t@e.st` in the test git config is a stub email (not a real address).
- PR title and description contain no wording that advertises a credential or PII leak/cleanup.

---

PR #1246, opened 2026-07-28
test(memory-integrity): make posture staleness tests clock-independent

WingedGuardian:
## Problem

`_check_memory_integrity_posture` (awareness/loop.py) fires a `reports_stale` / "no conclusive report in Nd" alert when the earliest report predates a **now-relative** window (`datetime.now(UTC) - stale_report_days`, default 3d) with no recent conclusive report. Three posture tests seeded **hardcoded `2026-07-xx`** timestamps, so they drift across that window as the real calendar advances.

`test_unknown_after_degraded_holds_alert` began failing on **2026-07-28** when its `2026-07-25` "healthy" report crossed the 3-day boundary: the staleness alert fired, so the test asserted `[]` but got one open alert. This blocks CI `test` on every PR opened today. (The two "stuck/stale" tests happen to pass now but are equally fragile — they'd fail if run before their hardcoded dates aged past the window.)

## Fix

Test-only. Add an `_ago(days/hours/minutes)` helper returning `now - offset` in the stored `%Y-%m-%d %H:%M:%S` format (so it sorts lexically = chronologically against the check's `since_iso`), and seed the three fragile tests relative to `now`:
- **hold/resolve test**: all three reports recent (within the window) → the healthy report resolves the alert and staleness never fires.
- **two stuck/stale tests**: a 10-day-old earliest + a recent still-unknown report → the staleness/stuck finding fires deterministically.

Wide margins (3h vs 72h; 10d vs 3d) hold for any real `now` — no time-of-day / month-boundary / DST dependence.

## Scope

Only `test_posture_and_jobs.py` needed this — the code-reviewer verified the other hardcoded-date files in the dir (`test_crud_and_config.py`, `test_recall_probe.py`) compare against fixed `since_iso`/injected `now` values, not `datetime.now`, so they don't drift.

## Verify

`pytest tests/test_memory_integrity/test_posture_and_jobs.py` → 8 passed (incl. the previously-failing test). Code-reviewer: DONE, no BLOCKER/SHOULD-FIX — traced each rewrite through the real posture logic, confirmed intent preserved, format match, no vacuous assertions.

### DIFF
--- a/tests/test_memory_integrity/test_posture_and_jobs.py
+++ b/tests/test_memory_integrity/test_posture_and_jobs.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+from datetime import UTC, datetime, timedelta
+
 import aiosqlite
 import pytest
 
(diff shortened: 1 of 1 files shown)


WingedGuardian (commented):

Test-only change. Import ordering correct (stdlib before third-party). `_ago()` defaults are immutable ints. Conventional commit prefix valid. No new public API without docs.

---

PR #1252, opened 2026-07-30
fix(hooks): verify pr-create branch against live remote (close stale-ref fail-open)

WingedGuardian:
## What

Makes the `gh pr create` gate in `git_push_guard.py` correct and quiet, on top of #1241's un-gate. Three related fixes to `_pr_create_would_publish` + the create-arm:

1. **Close a stale-ref fail-open.** #1241 decided "already pushed?" from the LOCAL remote-tracking ref, which goes stale the moment a remote branch is deleted (a squash-merge auto-delete). The check now hits the **live remote** via `git ls-remote` (accepting only the line whose ref path is exactly `refs/heads/<branch>`, so a bare pattern can't tail-match a namespaced ref). Verified live: on the exact stale ref #1241's own merge left behind, the old guard returns `would_publish=False` (fail-open); this returns `True` (gates).

2. **Gate only the form that can publish.** Per `gh pr create --help` ("Use `--head` to explicitly skip any forking or pushing behavior") and gh source (`if opts.HeadBranch != ""`), only an **implicit** create (no `--head`) can push — gh may push the current branch when it isn't fully on the remote. So any non-empty literal `--head` (local, unpushed, or `owner:fork`) is un-gated. Empty (`--head=`/`--head ""`), repeated-flag-last-empty (gh's last-value-wins), and shell-expansion heads (`$VAR`/`$(…)`/backtick, which could resolve empty at runtime) are **not** trusted and fall through to the live verification.

3. **Actually stop the prompt.** The un-gated path now emits `permissionDecision: "allow"` instead of a bare exit-0. A bare exit-0 left the command to Claude Code's own permission flow, which **still prompts** for a non-allow-listed `gh pr create` — the real reason a standalone create on already-pushed code kept prompting after #1241. A push in the same command still asks first (its `ask` is returned before the create `allow`), so `git push && gh pr create` prompts **once**, for the push.

## Net behaviour

- standalone `gh pr create` on a pushed branch → **allow** (no prompt)
- `gh pr create --head <literal>` (incl. `owner:fork`) → allow
- `git push && gh pr create` → one ask (for the push)
- `gh pr create` from an unpushed current branch, or an empty/dynamic `--head` → ask (interactive) / deny (dispatched)
- force-push, merge-into-main, `gh pr merge` w/o `--admin`, `commit --no-verify` → unchanged hard-blocks

Fail-safe throughout: network error, timeout, missing/absent branch, detached HEAD, or any uncertainty → gate. `ls-remote` bounded by 10s inside the hook's 60s budget.

## Review

Two independent adversarial passes (Codex + Claude), both hunting for fail-opens:
- Both caught the empty-`--head` under-gate; Claude additionally caught the dynamic (`--head "$BRANCH"`) and last-value-wins variants. All fixed + regression-tested.
- Both confirmed: non-empty literal `--head` un-gate is gh-source-supported; the implicit `ls-remote` path fails closed on every uncertainty; `allow` bypasses no other guard and never overrides a pending push `ask`.

125 hook tests pass (`test_merge_review_gate` + `test_push_create_override` + `test_hook_input_contract`); ruff clean; private-data scan clean. The exotic `pushurl != fetchurl` edge is tracked separately (tabled `26ff0b2f`).

## Deploy

CC-session hook — merge then `git pull` on main; reloads on next Bash call. No restart.

### DIFF
--- a/.claude/skills/genesis-development/references/worktrees.md
+++ b/.claude/skills/genesis-development/references/worktrees.md
@@ -102,14 +102,20 @@ over.
   tool). Force push (`--force` / `--force-with-lease` / `-f` / `+refspec` /
   `--mirror`) is **hard-blocked** in every session. Multiple pushes in one
   command are blocked — each push needs its own approval.
-- `gh pr create` — **un-gated when its branch is already on the remote** (then
-  it's just a review request on pushed code), so a `git push && gh pr create`
-  prompts once — for the push — and the PR-create rides along. The exception:
-  gh will *push* (and can fork) a branch whose HEAD is not yet on the remote, so
-  a create from an **unpushed branch** is gated like a push (interactive → ask,
-  dispatched → deny). Verified against the **actual remote** via `git ls-remote`
-  (not the local tracking ref, which goes stale when a merged branch is deleted),
-  fail-safe — any network error / timeout / uncertainty gates.
+- `gh pr create` — a create only publishes code in the **implicit** form (no
+  `--head`), where gh may *push* (and can fork) the current branch when it isn't
+  fully on the remote. So an **explicit `--head`** (local, unpushed, or
+  `owner:fork`) is **un-gated** — `gh pr create --help`: "Use `--head` to
+  explicitly skip any forking or pushing behavior." The implicit form is un-gated
+  **only when the current branch is already on the remote** (then it's just a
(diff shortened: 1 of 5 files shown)


WingedGuardian:

---

PR #1255, opened 2026-07-30
fix(hooks): resolve the commit review gate's worktree from the payload cwd

WingedGuardian:
## What

Stops the commit review gate from false-blocking a `git commit` made from a **second worktree** (one that isn't the session's launch dir).

The gate figured out which worktree's review marker to check by parsing a leading `cd X && git commit` from the command string only. A **bare** `git commit` (no `cd` prefix) from a secondary worktree therefore resolved the marker against the wrong worktree and blocked the commit even though the review had been recorded — a real papercut hit repeatedly this session (10+ times).

## How

Both hooks of the pair fall back to the **PreToolUse/PostToolUse payload `cwd`** — the Bash tool's actual working directory, which tracks its persisted `cd` across separate calls (verified live: it reports the drifted shell dir, not the static `CLAUDE_PROJECT_DIR`). Precedence: explicit `cd X && …` (the command targets that dir; payload `cwd` would be the stale *pre-cd* dir) → payload `cwd` → `None`.

**Both** the checker (`review_enforcement_commit.py`) and the invalidator (`review_invalidate_on_commit.py`) get the **identical** resolution, so the marker cleared after a commit is the same one that was checked. Patching only the checker would desync the pair and reopen a stale-marker bypass — caught by review, fixed here, and guarded by an end-to-end regression test that drives both hooks (mark → PreToolUse allow → PostToolUse invalidate → a new unreviewed `add && commit` must still block).

`_payload_cwd` requires an absolute, existing directory; anything else (non-str, empty, relative, non-dir) falls through to prior behavior — over-strict, never permissive.

## Not weakened

A test marks the **wrong** worktree and confirms the real target still blocks; Rule 0 (`--no-verify`) is checked before `cwd` is computed; Rule 1 (main) keys on the same resolved `cwd` as the commit's real location.

## Review

Codex + two Claude adversarial passes. The whole-repo Claude pass found a **BLOCKER** — the PostToolUse invalidator hadn't been updated, desyncing the pair — now fixed and regression-tested (verified RED without the invalidator fix). Optional `os.path.isabs()` hardening applied to both copies. 41 hook tests pass; ruff clean; private-data scan clean. A pre-existing decoy-`cd` / `git -C` evasion surface of the gate is tracked separately (tabled `d21db5ea`).

## Deploy

CC-session hooks — merge then `git pull` on main; reload on next Bash call. No restart.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,6 +119,15 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **The commit review gate no longer false-blocks commits made from a second
+  worktree.** When one Claude Code session worked in a git worktree other than
+  the one it launched from, the hook that requires a code review before a commit
+  looked for the review record in the wrong worktree and blocked the commit even
+  though the review had been done. It now reads the Bash tool's actual working
+  directory from the hook payload, so a plain `git commit` resolves its own
+  worktree's review record — no `cd`-prefix workaround needed. The gate is not
+  weakened: an unreviewed commit still blocks, verified by a test that marks the
+  wrong worktree and confirms the real target is still gated.
 - **Genesis's safety guardrails work again.** The hooks that block dangerous
   actions in Claude Code sessions — force-pushing, `rm -rf` on your data
   directories or database, committing to `main` without a review, writing to
--- a/scripts/review_enforcement_commit.py
+++ b/scripts/review_enforcement_commit.py
@@ -64,9 +64,26 @@ def _extract_working_dir(command: str) -> str | None:
     return path if os.path.isdir(path) else None
 
 
+def _payload_cwd(payload: dict) -> str | None:
+    """The Bash tool's actual working directory, from the PreToolUse payload.
+
+    CC reports ``cwd`` = the shell's current directory at hook time, and it
+    TRACKS the Bash tool's persisted ``cd`` across separate calls (verified live
+    2026-07-29). So a bare ``git commit`` with no leading ``cd`` still resolves
+    to the right worktree via this, where the command-string parse can't see it.
+    ``CLAUDE_PROJECT_DIR`` is NOT usable here — it is the static session root, so
+    it points at the wrong worktree when this session drives a secondary one.
+    """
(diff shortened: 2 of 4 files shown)


WingedGuardian (commented):

---

Checked all modified `.py` files against Domain 1 rules:

- No `nohup`, `task_states` direct inserts, `/tmp/` usage, `git add -A/. ` (the `git add -A` string in the e2e test is a simulated command payload that the hook is verifying it **blocks** — not an actual git invocation), force-push/reset-hard, mutable defaults, bare `except:`, `os.killpg`, or raw SQL.
- Import hygiene: `import os` promotion from inside `_extract_working_dir` to module level is an improvement; all other imports are stdlib-only for both hook scripts.
- `$CLAUDE_PROJECT_DIR` appears only in a docstring explicitly cautioning against its use — not used as a Python env var.
- Both new `_payload_cwd` functions carry docstrings; test helpers have brief docstrings; no coverage gaps (modified scripts are in `scripts/`, not `src/`).
- Conventional commit prefix: `fix(hooks):` ✓

PII/Secrets scan of added lines: no credentials, private IPs, SSH keys, personal emails, hardcoded `/home/<user>` paths, or private hostnames found.

PR title/body/CHANGELOG wording: describes a worktree resolution behavior fix; no language implying security incidents, credential leaks, or sensitive infrastructure exposure.

---

WingedGuardian:
Superseded by #1254 (merged ~30 min before this, by a concurrent session), which made `review_enforcement_commit.py` worktree-aware via a more comprehensive `_effective_diff_cwd` (payload cwd + decoy-cd/subshell/`git -C` + docs-only skip) — a superset of this PR's simpler payload-cwd fallback. Shipping this would revert that.

IMPORTANT carry-over: #1254 updated the CHECKER but NOT the paired invalidator (`review_invalidate_on_commit.py` still uses `_extract_working_dir` only). That is the exact checker/invalidator ASYMMETRY an adversarial review of this PR flagged as a BLOCKER: a bare `git commit` is now checked against the real worktree's marker (via payload cwd) but the marker is cleared under the wrong (process-cwd) key, so it survives and a later add+commit chain passes the existence-only gate on a stale review. Tracking a focused follow-up to make the invalidator symmetric with #1254's checker + the cross-hook regression test from this branch.

PR #1263, opened 2026-07-30
feat(resilience): CC network preflight + surplus outage-awareness (PR-3)

WingedGuardian:
## What & why

Third PR in the internet-outage resilience series (after PR-1 class fixes #1256 and PR-2 connectivity sentinel #1259). Adds a **network-OFFLINE gate** so Genesis fails fast instead of hanging when the internet is down.

During the 2026-07-28 blackout, CC dispatches on the dead network burned **45/55/23 minutes** each against a 7200s (`timeout_s`) ceiling. This turns that worst case into a **sub-second raise** (measured **0.002s** end-to-end).

## Design — two consumers, one shared gate

`network_config.parking_decision()` combines the `parking_mode` lever + the live `network_state.json` snapshot (from PR-2's sentinel) + a staleness guard, returning `off | normal | shadow | park`. Both consumers gate on it so they can't drift:

1. **CC-invoker preflight** (`invoker.run` / `run_streaming`, *before* subprocess spawn): fresh-OFFLINE + `live` + WAN endpoint → raises the new `CCNetworkOfflineError`. LAN CC peers (classified by the new stdlib-only `util/netclass.py`) stay reachable and are exempt.
2. **Surplus tier filter** (`dispatch._filter_tiers_for_network`): OFFLINE + `live` strips cloud compute tiers, so internet-dependent tasks stay `pending` (never `mark_failed`) instead of churning into a dead network.

## CAVEAT A — resume-continuity class-fix

`CCNetworkOfflineError` is added to the fast-re-raise tuples in `conversation.py` `_try_invoke`, `_try_invoke_streaming`, **and** `_run_failover_peer`, so a network blip on a **resume** turn is not misread as a stale resume (which would needlessly fail the live CC session and retry fresh).

## Fail-safe

Lever off, an absent/stale/NORMAL/DEGRADED snapshot, a LAN endpoint, or **any error** in the gate/classifier all fall through to a normal dispatch — identical to pre-sentinel behavior.

## Ships in shadow (default) — important

`parking_mode: shadow` is the shipped default. In shadow the preflight **never raises** — behavior is unchanged on deploy (pure observation). Flipping to `live` has a **hard prerequisite**: periodic CC consumers (inbox retry budget, autonomy/ego outcome signals, direct_session) must first be made outage-aware so a `CCNetworkOfflineError` is treated as "not attempted", not a failure — otherwise an outage would pollute calibration/retry state. Found in code review; tracked as a high-priority PR-3b follow-up and noted inline in `config/resilience.yaml`.

## Scope / compatibility
- Reuses PR-2's sentinel modules; **no schema changes**; new `CCNetworkOfflineError(CCError)` is contract-compatible with every existing terminal handler.
- Out of scope (documented): the remote-SSH `claude -p` path bypasses CCInvoker (a remote-module endpoint, orthogonal).

## Tests / verification
- **96 targeted** (netclass ranges/cache/last-known-good fallback, `parking_decision` gate, preflight incl. pre-spawn proof, surplus never-`failed`, CAVEAT A ×3) + **1227 regression** green; ruff clean.
- **Live E2E** against a real OFFLINE state file: `run()` raised in **0.002s** (vs 45-55min); stale→normal, shadow→observe, NORMAL→no-op, LAN-exempt all confirmed.

### DIFF
--- a/config/resilience.yaml
+++ b/config/resilience.yaml
@@ -34,6 +34,11 @@ network:
   enabled: true
   # Degraded-mode parking lever (PR-3): off | shadow | live. Default shadow —
   # observe one real outage before granting parking authority. Invalid → shadow.
+  # PREREQUISITE before flipping to `live`: harden the periodic CC-invoker
+  # consumers (inbox retry budget, autonomy/ego outcome signals, direct_session)
+  # to treat CCNetworkOfflineError as "not attempted" rather than a failure —
+  # otherwise an outage pollutes calibration/retry state. In `shadow` the preflight
+  # never raises, so that class is dormant. See the PR-3b follow-up.
   parking_mode: shadow
   # Restore push-retry lever (PR-4): off | live. Invalid → off.
   backup_push_retry: live
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -1010,9 +1010,18 @@ verified: b662f3e3 2026-07-17
   and opts out of the state machine's symmetric flap protection. The axis drives
   watchdog zombie-restart suppression (a restart can't fix an ISP outage), the
   `is_any_degraded` recovery gate (at OFFLINE only — don't re-dispatch into a
-  dead network), and the dashboard "Internet" light; it is DELIBERATELY excluded
-  from `to_legacy_degradation_level` (call-site shedding is deferred to PR-3
-  under `network.parking_mode`). Lever/kill-switch in `network_config.py`
+  dead network), and the dashboard "Internet" light. Two OFFLINE-parking
+  consumers (PR-3) gate on `network.parking_mode` (off/shadow/live, default
+  shadow) through the shared `network_config.parking_decision()` gate: the
+  **CC-invoker network preflight** (`cc/invoker.py::_network_preflight` — raises
+  `CCNetworkOfflineError` BEFORE the subprocess spawns for a WAN endpoint,
+  turning a `timeout_s`-bounded hang (7200s default) into a sub-second fail; a
+  LAN CC peer, classified via `util/netclass.py`, still runs), and the
+  **surplus tier filter** (`surplus/dispatch.py::_filter_tiers_for_network` —
+  strips cloud compute tiers so internet-dependent tasks stay `pending` instead
+  of churning `failed`). The axis stays DELIBERATELY excluded from
+  `to_legacy_degradation_level` (broader legacy-level call-site shedding remains
+  deferred/tabled). Lever/kill-switch in `network_config.py`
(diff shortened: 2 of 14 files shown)


WingedGuardian:
## Code Review — PR #1263

> ⚠️ Note: GitHub prevents requesting changes on your own PR — posting findings as a comment. This review contains **HARD BLOCK** findings that would normally block merge.

---

### Structural Review: CLEAN

No pattern violations (no nohup, no direct task_states inserts, no /tmp usage, no bare `except:`, no mutable defaults, no git add -A, no force push, no raw SQL against genesis.db). Import hygiene clean. All new `.py` files in `src/` have corresponding test files (`src/genesis/util/netclass.py` → `tests/test_util/test_netclass.py`). New public functions/classes have docstrings. Conventional commit prefix present.

Deliberate pattern (not a violation): function-scope imports in `_network_preflight` and `_filter_tiers_for_network` are intentional and self-documented for bootstrap cycle safety.

---

### PII/Secrets/Wording Scan: HARD BLOCK

**Private IPs hardcoded in test files — install-specific addresses embedded in the diff.**

**`tests/test_util/test_netclass.py` — `test_local_ip_literals` parametrize list**

Four addresses in the added parametrize block reveal this install's internal network topology:

- Address annotated `# RFC1918 /8 (this install's Ollama)` — comment explicitly identifies this as a real service endpoint. Replace with a generic RFC1918 /8 example (e.g. `10.0.0.1`).
- Address annotated `# RFC1918 /16 (this install's LM Studio)` — same issue. Replace with a generic RFC1918 /16 example (e.g. `192.168.1.1`).
- Address annotated `# CGNAT / Tailscale` in the 100.64–127.x.x block — falls in the Tailscale CGNAT range explicitly called out as a hard block. Replace with a generic CGNAT example such as `100.64.0.1`.
- IPv6 address annotated `# Tailscale mesh (under fc00::/7)` — in the IPv6 ULA `fd00::/8` range. Replace with a generic ULA example such as `fd00::1` and remove the install-specific prefix from the comment.

**`tests/test_cc/test_invoker_network_preflight.py` — `_LAN_URL` constant**

The `_LAN_URL` constant contains the same `192.168.x.x` address as the LM Studio annotation above — same real host. Replace with the generic `192.168.1.1` used above (or `192.168.1.100` to avoid looking like a gateway).

**Remediation**: drop the `# this install's X` comment suffixes entirely (they add no test value — the classification ranges are documented in the module docstring and `_LOCAL_NETS`), and replace all five install-specific addresses with generic RFC-standard examples that cover the same address families.

**Not flagged**: the CIDR network definitions in `_LOCAL_NETS` in `src/genesis/util/netclass.py` are classification rules, not instance addresses.

**PR title and description: CLEAN** — no language advertising a credential/PII fix, no infrastructure hostnames or identifiers.

---

PR #1274, opened 2026-08-01
fix(memory): preserve stored memory_class on re-embed + re-sync diverged payloads

WingedGuardian:
## Problem

`EmbeddingRecoveryWorker.drain_pending` rebuilt each re-embedded point's Qdrant
payload and **unconditionally recomputed `memory_class`** from a content
heuristic (`classify_memory`), discarding the authoritative store-time class in
`memory_metadata.memory_class`.

This silently downgraded explicit class overrides. The clearest case:
reference-extraction deliberately stores URL-bearing content as `fact` to avoid
the 0.7× activation penalty that `reference` carries — a re-embed recomputed it
back to `reference`, re-applying the penalty and lowering how often it surfaces
in recall. SQLite stayed correct; only the Qdrant payload (which recall reads)
drifted, so the divergence was invisible on the SQLite-backed dashboards.

Triggers: outage recovery, the d0008 mirror requeue, and — newly — the nightly
`memory_integrity` reconcile lane's mirror requeue.

**Measured on a live corpus:** of 1,103 `fact`-classed memories with URL
content, 319 (29%) already read a different, penalized class in Qdrant
(318 → `reference`, 1 → `rule`). Rows stored as rule/reference did not drift —
the corruption is specifically in the fact-override direction.

## Fix

1. **Class fix (going-forward).** `get_taxonomy` now also returns
   `memory_class`; the worker uses `stored_class or classify_memory(...)` —
   the authoritative stored class wins, and the heuristic recompute happens
   only when there is no metadata row (legacy path). Verified safe: across
   66,482 metadata rows, `memory_class` is never NULL/empty (schema
   `DEFAULT 'fact'`, every write resolves a concrete class first), so the
   `or` fallback fires only on a genuinely absent row.

2. **Data repair (existing corruption).** New `memory_class_backfill` core +
   data-migration `d0009` re-sync already-diverged Qdrant payloads to the
   SQLite-authoritative class. Routed through the data-migration framework
   (not a hand-run script) so it heals **every** install on the next
   pull+restart, and is a clean no-op where nothing drifted. Divergence is
   judged against the *effective* read value (`payload.get("memory_class",
   "fact")`) so a missing key that already reads as the correct default is not
   needlessly rewritten. Structurally identical to the shipped
   `d0001_origin_class_qdrant` (read-only SQLite, writes Qdrant only → no
   write-lock contention).

## Tests

- `test_embedding_recovery.py`: re-embed preserves an explicit `fact` override
  on reference-flavored content (asserts `classify_memory(content) ==
  "reference"` while the worker output stays `fact`); fallback recomputes when
  no metadata row.
- `test_memory_class_backfill.py`: true-divergence repair, read-default
  non-write, no-metadata skip, exact-match skip, dry-run, pagination, verify
  counter.
- `test_db/test_memory.py`: `get_taxonomy` returns `memory_class` (default +
  explicit).
- Data-migration framework tests confirm d0009 is discovered cleanly.

Code-reviewed inline (zero findings; full consumer enumeration of `memory_class`).

Ledger: fde7f389aaa84a70b1c52819fedc0652


WingedGuardian:
## First-pass review — PR #1274

### Structural review

**WARNING — Missing direct test coverage for `d0009` migration entry points**

`src/genesis/db/data_migrations/d0009_resync_memory_class_qdrant.py` is a new `src/` file. Its core logic (`resync_memory_class`, `count_diverged_memory_class`) is well-covered in `tests/test_memory/test_memory_class_backfill.py`, but the migration entry points `migrate()` and `verify()` themselves have no direct test. The PR description notes the data-migration framework tests confirm d0009 is discovered, but a unit test that calls `migrate()` and `verify()` against fakes would also confirm the try/finally teardown and return-value contract. Low risk given the thinness of the wrapper, but the rule is: new `src/` file → corresponding test file.

**NOTE — `migrate()` missing docstring**

`d0009_resync_memory_class_qdrant.py:36` — `verify()` has a docstring; `migrate()` does not. Since both are called by the migration runner, both should be documented:

```python
def migrate() -> dict:
    """Re-sync memory_class onto diverged Qdrant payloads (see module docstring)."""
    ...
```

---

**CLEAN** — No API keys, tokens, private IPs, IPv6 ULA addresses, SSH keys, or connection strings in added lines. PR title, description, and CHANGELOG additions use neutral bug-fix language with no PII-leak, credential-exposure, or private-infrastructure wording.

---

Everything else is clean: no `nohup`, no `/tmp/` usage, no bare `except:`, no mutable defaults, no `git add .`/force-push, no `task_states` direct inserts, no star imports, import grouping correct throughout. The SQL parametrisation in `memory_class_backfill.py` is safe (the `noqa: S608` is correctly motivated — `placeholders` is `?`-only, values are bound).

---

WingedGuardian:
Merging on green CI + the code-reviewer agent (zero findings, full `memory_class` consumer enumeration) + the structural review (both items — d0009 wrapper test and `migrate()` docstring — addressed in 7784dcda). The Codex bot did not post on this PR (it reviewed the later #1277 but not this one after ~50 min), so per the merge-gate quota-absent rule this merges without a Codex pass. The data-repair migration was additionally validated with a live dry-run against real data (343 diverged points identified, zero false positives on rule/reference rows).

PR #1275, opened 2026-08-01
fix(contribution): gh secret set reads value from stdin (no --body-file)

WingedGuardian:
## What

The secret-sync added in #1272 invoked `gh secret set … --body-file -`. That flag does not exist in the shipped `gh` — the secret value is read from **stdin** when `--body` is omitted. The call therefore failed at runtime (non-fatal: it logged and skipped), so the `GENESIS_PRIVATE_PATTERNS` Actions secret was never populated.

## Why the tests missed it

The unit tests mock `subprocess.run`, so they verified the code's control flow but not the real `gh` CLI contract — the value already flowed via `input=` (stdin), which is correct; only the extra bogus flag was wrong. Caught by an end-to-end run against real `gh` during post-merge verification.

## Fix

- Drop `--body-file -`; the value flows via `input=` (stdin), never argv.
- Add a contract-lock assertion to `test_sync_secret_success`: the command must not contain `--body-file` or any `--body*`, and the secret value must never appear on the command line.

Verified against real `gh`: the secret now sets (`GENESIS_PRIVATE_PATTERNS` present on the repo). Follow-on to #1272.

### DIFF
--- a/src/genesis/contribution/fingerprints.py
+++ b/src/genesis/contribution/fingerprints.py
@@ -459,7 +459,9 @@ def sync_secret(
     print(f"  fingerprints: syncing {len(patterns)} pattern(s) to the CI secret on {slug}")
     try:
         proc = subprocess.run(
-            ["gh", "secret", "set", SECRET_NAME, "--repo", slug, "--body-file", "-"],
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

PR #1276, opened 2026-08-01
fix(learning): skill-evolution is propose-only — stop autonomous auto-apply

WingedGuardian:
## Problem

The weekly autonomous **skill-evolution** pipeline (`CronTrigger` Sat 04:00) auto-applied MINOR skill edits directly to the tracked in-repo `SKILL.md` templates. It repeatedly landed **principle-violating edits** — e.g. a *"Fallback & Silent Recoveries"* section instructing a skill to swallow errors and hide failures from the user — that the shadow semantic **Critic flagged in real time but could not block** (it is shadow-by-construction, and no `enforce` mode was ever built). Because the applicator wrote the tracked template with **no commit**, each edit also left a dangling working-tree change that blocked `scripts/update.sh`.

Verified via the cognitive-modification ledger: 5 consecutive weekly `skill_evolution` auto-applies to one skill, two of them Critic-flagged, none blocked.

## Change — propose-only

Autonomous skill-evolution now **proposes, never applies**:

- **`applicator.apply()` never writes a skill file** (or a ledger row) — every change size is *staged* for human/CC review and application via the normal worktree/PR flow.
- The **Critic verdict and the full proposed content ride the staged proposal** (previously only a 500-char preview was kept, making a proposal un-applyable).
- **Dampening:** a skill with an unresolved proposal is not re-proposed (exact category match — skill names can contain SQL `LIKE` metacharacters), which also breaks the self-reinforcing "each bad edit tanks the success rate that triggers the next" spiral and saves the refiner LLM call.
- **TTL:** the `skill_proposal` review-queue observation moves 14d → 60d so the live `resolve_expired` sweep can't silently empty the safety queue.
- The WS1 shadow gates (Critic + replay) and the skill-autonomy-graduation GROUNDWORK are **preserved** as the future `enforce` re-enablement path — this deliberately does *not* build enforce, only stops auto-apply now.

## Deferred (follow-ups)

A dedicated `skill_proposal_resolve` apply/reject tool (DB-only state transition — never git from the server runtime), a per-run proposal digest, and re-sourcing the replay gate's OLD baseline (OLD=on-disk / NEW=proposal) since it previously read the now-absent ledger pre-image (documented with a loud comment, not silently inert).

## Verification

- 755 passed across the full `test_learning/` subsystem + observations; ruff clean.
- The no-write invariant is asserted directly (file unchanged + zero ledger rows) across MINOR/MODERATE/MAJOR × validation pass/fail × router present/absent × Critic clean/flagged/raise/None.
[...]

PR #1279, opened 2026-08-01
fix(db): rename colliding d0009 migration to d0010 + CI guard for prefix dupes

WingedGuardian:
Two concurrently-merged PRs (#1274, #1276) each took the `d0009` data-migration prefix. The runner's duplicate guard raises and `run_data_migrations` swallows the error to protect boot — net effect: **all data migrations silently skip on every install, every boot** until resolved.

**Fix:** rename the later one (`backfill_skill_proposal_dampening` → `d0010`; #1274's `resync_memory_class` keeps `d0009`). Safe in every ordering — the guard prevented either from running while both existed, so no ledger row exists under a wrong name.

**Class-fix:** `test_no_duplicate_migration_prefixes_in_tree` — a static scan of both real migration directories (schema + data), so a prefix collision fails CI on the offending PR's merge ref instead of no-op-ing deploys fleet-wide. (Verified: the guard detects the live `{'d0009': 2}` collision on the pre-fix tree.)

16 tests pass; ruff clean.

### DIFF
--- a/src/genesis/db/data_migrations/d0010_backfill_skill_proposal_dampening.py
+++ b/src/genesis/db/data_migrations/d0010_backfill_skill_proposal_dampening.py
@@ -1,4 +1,4 @@
-"""d0009 — align pre-existing skill_proposal observations to propose-only.
+"""d0010 — align pre-existing skill_proposal observations to propose-only.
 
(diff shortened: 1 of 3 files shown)


WingedGuardian:

---

PR #1295, opened 2026-08-05
fix(dashboard): CSRF-harden cookie-authed /api mutations (same-origin check)

WingedGuardian:
### Problem
The app-level `/api` mutation gate (`check_api_mutation_auth`) accepted a session
cookie as proof of authorization when `DASHBOARD_PASSWORD` is set. The session
cookie is `SameSite=Lax`, which still attaches on **same-site** requests — so a
sibling origin (another port/subdomain on the dashboard host) could ride a
logged-in session into any of ~81 state-changing routes, including
`approvals/approve-all`, the `secrets` writer (PUT), `config-files` (PUT/DELETE),
`restart/*`, and `provision/*`. `SameSite=Lax` does not distinguish same-site
siblings, so it is not a CSRF defense here.

### Fix
- The **internal bearer token is now checked first** — it is CSRF-immune (an
  attacker cannot read the `0600` token file), so it stays origin-independent.
- The **cookie path now additionally requires the request to be same-origin**:
  primary signal `Sec-Fetch-Site` (browser-set, unforgeable by page JS; safe iff
  `same-origin`/`none`), falling back to matching the `Origin` (then `Referer`)
  host against the request `Host`. When no same-origin signal is present at all,
  it **fails closed** (403) — a legitimate machine caller uses the bearer, not the
  cookie. This follows the OWASP CSRF Fetch-Metadata guidance (treat absent
  `Sec-Fetch-*` as unknown; do not fail open).
- The `GENESIS_DASHBOARD_API_AUTH=off` kill switch still disables the whole gate.
- Fixes the stale `api.js` comment that asserted `SameSite=Lax` alone covered CSRF.

Same-origin dashboard fetches already send `Sec-Fetch-Site: same-origin`, so the
web UI needs no changes. Read-only calls, the `/v1` voice API, and a co-hosting
framework's own `/api/*` routes are untouched.

### Tests
Extends `tests/test_dashboard/test_api_mutation_gate.py` with the CSRF matrix
(same-origin/`none` pass; same-site/cross-site/no-header fail-closed; Origin
fallback match vs mismatch; bearer immune to cross-origin; kill-switch off).
Written test-first (the blocked cases fail against the pre-change code, then pass).

### Deploy
Runtime dashboard code only — `git pull` + server restart. No host/guardian paths.

### Review
Self-review + one `genesis-security-reviewer` pass: no blocking or should-fix
findings. Confirmed `Sec-Fetch-Site` is a Forbidden header (unforgeable by page
JS), the fail-closed branches have no fail-open hole, the bearer-first reorder is
behavior-preserving, and the host TCP passthrough does not strip/rewrite the
headers the check relies on. One optional case-insensitivity hardening was applied;
one scheme-comparison suggestion was consciously declined (would risk false-closed
403s on TLS-terminating installs without forwarded-proto trust).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,18 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   breaks, set `GENESIS_DASHBOARD_API_AUTH=off` to disable just this gate without
   removing the password.
 
+### Security
+
+- **The dashboard API is now hardened against cross-site request forgery (CSRF).**
+  With a password set, a state-changing API call authenticated by your login
+  cookie must now originate from the dashboard itself (verified via the browser's
+  `Sec-Fetch-Site`/`Origin` headers). Previously another page — including a
+  separate service sharing your dashboard's host — could ride your logged-in
+  session to trigger dashboard actions; that path is now refused. Genesis's own
+  components (which use an internal token) and read-only calls are unaffected, and
+  the same `GENESIS_DASHBOARD_API_AUTH=off` switch disables this along with the
+  rest of the gate.
+
 ### Fixed
 
 - **Re-embedding a memory no longer downgrades its recall ranking.** When a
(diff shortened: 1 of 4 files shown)


WingedGuardian:
@codex review

PR #1290, opened 2026-08-05
feat(memory): instrument the proactive-recall 503 path (WS-1 PR-2)

WingedGuardian:
## What

Makes the proactive-recall 503 path observable. Instrument-first — **no structural change to recall behavior.**

The dashboard route enforces its 4.5s budget thread-side (`run_coroutine_threadsafe` + `future.result(timeout=4.5)`) and, on expiry, `future.cancel()`s the recall coroutine (which schedules a thread-safe task-cancel on the loop → raises `CancelledError` at the current await) and returns 503. Previously `proactive_context` only logged timings when the coroutine **completed** slow, so the ~39%/24h of calls that time out logged nothing — no phase timings, no attribution. That was the observability hole behind the recall-timeout hunt.

## Changes

1. **`memory/proactive.py`** — `proactive_context` runs its awaited phase sequence (embed → recall → enrich → procedure) inside a `try/except asyncio.CancelledError`, accumulating per-phase timings incrementally and tracking the in-flight `phase`. On cancellation it WARNs the phase + partial timings (the data the 503 discarded) and **re-raises**. `CancelledError` is a `BaseException`, so the inner best-effort `except Exception` blocks never swallow it.
2. **`hosting/standalone.py`** — a `_loop_lag_sampler` measures event-loop scheduling drift every 500ms and WARNs once per stall **episode** (peak drift reported when it clears — debounced so a sustained stall does not flood the journal), making a loop stall timestamp-correlatable with the awareness-tick / dispatch / bundle lines that starve recall. Started in `serve()` after loop capture, cancelled at shutdown.

## Levers (generalizable, zero-config defaults)

- `GENESIS_LOOP_LAG_WARN_MS` (default 250) — drift threshold to warn at.
- `GENESIS_LOOP_LAG_SAMPLER` (`0`/`false`/`off`) — disables the sampler at startup.

## Tests

- `test_proactive_context_logs_partial_timings_on_cancellation` — cancellation logs phase + partials and re-raises.
- `test_loop_lag_sampler.py` — warn-above-threshold, silent-below, env-raises-bar, bad-env-fallback, sustained-lag-warns-once-then-clears.
- **28 targeted tests green, ruff clean.**

## Verification

[...]

PR #1297, opened 2026-08-05
feat(memory): attribute cancelled recalls to sub-stage + executor pressure (WS-1 PR-2c)

WingedGuardian:
## What

Instrumentation-only follow-up to #1290. The 503 cancellation WARN named only the coarse phase (`phase=recall`), leaving the ~4s a cancelled recall burns unexplained — and the loop-lag sampler cannot distinguish the two competing mechanisms behind it: event-loop starvation vs default-executor (`asyncio.to_thread`) saturation. Live soak data showed every organic 503 dying in the recall phase (~3.2-4.2s) while the success path completes it in ~0.5s, with logged loop stalls too small to account for the gap.

## How

- **New `util/loop_diag.default_executor_pending()`** — best-effort gauge of the running loop's default ThreadPoolExecutor (`pending`/`workers`/`max_workers`). Blanket-guarded: returns `None` before the executor exists, outside a loop, or on any CPython attr-shape change; never raises into a caller. The default executor is the pool Qdrant recall `to_thread` calls share with other background `to_thread` work (verified: no `set_default_executor` anywhere in `src/`), so a deep pending queue during a stall names thread-pool saturation.
- **Cancellation WARN upgraded** (`proactive.py`): folds the recall sub-stage timers the engine already wrote incrementally into the `stats` sink before the cancel (an empty fold ⇒ died in the first stage, the vector search), plus `phase_wall_ms` (wall time inside the cancelled phase) and `executor=` gauge. The success path now shares the same `_fold_engine_substages` helper (pure refactor of the previous inline fold — behavior preserved), so both success and 503 logs carry one sub-stage vocabulary.
- **Loop-lag sampler WARN** (`standalone.py`) carries the `executor=` gauge per lag episode, so an episode with low lag but deep queue is attributable to executor saturation.

No changes to the recall engine (`retrieval.py`), recall results, or the wire contract.

## Test plan

- New `tests/test_util/test_loop_diag.py`: gauge contract — None before executor creation / outside a loop, well-formed dict after `to_thread`, queued-work case pinned to a 1-worker pool (hang-safe ordering: worker released before assertions).
- Extended cancellation test asserts `executor=`, `phase_wall_ms`, `cancelled_after_ms` in the WARN; new test proves completed sub-stages (vector/fts) fold into the WARN while never-reached stages stay absent.
- 33 targeted tests + 39 `tests/test_hosting/` green; ruff clean.

### DIFF
--- a/src/genesis/hosting/standalone.py
+++ b/src/genesis/hosting/standalone.py
@@ -176,6 +176,11 @@ class StandaloneAdapter:
         timestamp-correlatable with awareness-tick / dispatch / bundle log lines,
         naming what starves the recall coroutine behind the route 503s.
 
+        The WARN also carries ``executor=`` — the default-executor pending depth
+        (PR-2c). A lag episode with a deep executor queue means ``to_thread`` work
+        (Qdrant recall, awareness-tick git) is saturating the thread pool, which
+        loop drift alone cannot distinguish from pure loop starvation.
+
         Debounced per stall EPISODE: one WARNING when drift first crosses the
         threshold, one INFO with the peak drift when it clears. A sustained
         multi-minute stall (the worst case, and exactly when log noise competes
@@ -186,6 +191,8 @@ class StandaloneAdapter:
         ``GENESIS_LOOP_LAG_SAMPLER`` (0/false/off) — disables the sampler at
         startup. Best-effort: any error ends the sampler without touching serve.
         """
+        from genesis.util.loop_diag import default_executor_pending
+
         interval = 0.5
         try:
             threshold_ms = float(os.environ.get("GENESIS_LOOP_LAG_WARN_MS", "250"))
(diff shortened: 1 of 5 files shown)


WingedGuardian:

Checked across 5 files (`standalone.py`, `proactive.py`, `util/loop_diag.py`, `test_proactive_engine.py`, `test_util/test_loop_diag.py`):
- No bare `except:`, mutable defaults, raw SQL, `/tmp/` usage, or git safety violations.
- Deferred `loop_diag` import inside `_loop_lag_sampler` method body is appropriate (non-hook module, imports only stdlib `asyncio` — no heavy-dep concern).
- New `loop_diag.py` carries module docstring + function docstring; `_fold_engine_substages` private helper documented.
- New `util/loop_diag.py` covered by new `test_util/test_loop_diag.py`; `proactive_context` cancellation path covered by extended + new test in `test_proactive_engine.py`.
- PR title, body, and diff contain no PII, secrets, private IPs, or security-incident disclosure language.

---

PR #1299, opened 2026-08-05
feat(hooks): deterministic review-coverage manifest in the review reminder

WingedGuardian:
## What

Adds `scripts/review_scope.py`, a deterministic per-file **review-coverage manifest**, and injects it into the review-enforcement reminder (`scripts/review_enforcement_prompt.py`).

## Why

Today `/review` picks specialists by aggregate scope signals, but each specialist independently re-runs `git diff` and self-selects hunks. Nothing enumerates the changed files or guarantees each one is covered, so on a large changeset a changed file can go unreviewed by every specialist. This names every file that MUST be covered, model-independently.

## How

- **Self-contained**: no `genesis`-package import (hooks run on the main venv even from a worktree, and can't assume the package imports cleanly). The only reuse is the sibling github-aware `_is_docs_or_config`.
- **Strictly additive + fail-open**: the base "unreviewed changes" reminder always prints first; the manifest is appended behind its own guards. Any git error/timeout returns `None` and the reminder stands alone. Runs on every prompt, so every git call is timed.
- **Reviewable set** = code + tests + fixtures, minus binary/vendored. Docs/config are excluded to stay consistent with the commit gate's own `_is_docs_or_config` policy (`.github/` workflows remain in-scope as code).
- **scope_tag** is one tag per file, first-match-wins, mirroring `gstack-diff-scope`; scope-indicated specialists are surfaced as input (labelled "before the skill's adaptive gating"), never as a competing verdict.
- **git plumbing** uses `git diff -z` NUL parsing (verbatim paths, correct under tabs/renames/binary); `diff_lines` is the unfiltered whole-diff sum to match the skill's <50 specialist gate.

## Evidence

A-B validation vs `ocr review --preview` (a deterministic no-LLM file selector) over 8 real merged PRs: mean Jaccard 0.976. The only systematic divergence was config `.yaml`/`.toml`, which this manifest excludes to match the existing commit-gate policy. No external binary dependency ships.

Reviewed by Codex and a Claude reviewer (two sequential rounds, all findings fixed): numstat all-or-nothing fail-open, test/fixture classification precedence, `-z` NUL parsing, `UnicodeDecodeError` fail-open, and tab-in-path reconstruction. 53 targeted tests; `ruff` clean.

## Verification

- `pytest tests/test_scripts/test_review_scope.py` (53 passing)
- E2E: the real `UserPromptSubmit` hook emits the base reminder plus the manifest on a live diff, and fail-opens (base reminder survives) when the manifest can't be built.

Context: capability-build campaign ledger `09ab9c12a8fd49c686cc644201b448c7` (item #12). The campaign row stays open.

### DIFF
--- a/scripts/review_enforcement_prompt.py
+++ b/scripts/review_enforcement_prompt.py
@@ -58,11 +58,15 @@ def main() -> None:
         "result (e.g., if you wired a notification, confirm it actually sends; "
         "if you fixed a data path, confirm the data actually flows). "
         "Ask: 'If the system restarts now, will this actually work?' "
-        "If you cannot answer yes WITH EVIDENCE, you are not done."
+        "If you cannot answer yes WITH EVIDENCE, you are not done.",
+        flush=True,  # flush BEFORE the manifest's git calls: hook stdout is
+        # block-buffered (piped), so if the manifest git work overruns the 10s
+        # hook timeout and Python is killed, the base reminder must already be out.
     )
 
     # Additive: deterministic per-file review-scope manifest. Fully fail-open —
     # any import/build error is swallowed so the base reminder above stands alone.
+    # build_manifest self-bounds its total git time under the hook timeout.
     try:
         from review_scope import build_manifest, render_reminder_block
 
--- a/scripts/review_enforcement_prompt.py
+++ b/scripts/review_enforcement_prompt.py
@@ -37,7 +37,10 @@ def main() -> None:
     if is_review_current():
         sys.exit(0)
 
-    # Unreviewed changes exist — inject warning
+    # Unreviewed changes exist — inject the base reminder UNCONDITIONALLY first.
+    # The deterministic review-scope manifest below is strictly ADDITIVE: it is
+    # built and appended behind its own guards so a manifest error can never
+    # truncate or suppress this core reminder.
     print(
         "MANDATORY: Unreviewed code changes detected. "
         "You MUST run /review and dispatch the superpowers:code-reviewer agent "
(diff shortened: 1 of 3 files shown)


WingedGuardian:
## First-pass review: PR #1299

**Structural review: 2 WARNINGS**

---

### WARNING — Late import inside function body

**File**: `scripts/review_enforcement_prompt.py` (inside `main()`)

```python
try:
    from review_scope import build_manifest, render_reminder_block   # ← inside function
    ...
except Exception:
    pass
```

The `from review_scope import ...` is a late import inside `main()`'s `try` block rather than at module level. This violates import grouping convention. The design intent (fail-open: if `review_scope` can't be imported, the base reminder still runs) is sound, but the canonical way to achieve this in an enforcement hook script is a module-level `try/except ImportError` — the same pattern `review_scope.py` itself uses for its sibling import:

```python
# at module level (consistent with sibling pattern in review_scope.py itself)
try:
    from review_scope import build_manifest, render_reminder_block as _build, _render
    _SCOPE_AVAILABLE = True
except ImportError:
    _SCOPE_AVAILABLE = False

def main() -> None:
    ...
    if _SCOPE_AVAILABLE:
        block = _render(_build())
        if block:
            print("\n" + block)
```

Alternatively, the current `except Exception` is intentionally broader than `ImportError` (catches attribute errors, syntax errors in `review_scope`, etc.) — if that breadth is deliberate, a module-level `try/except Exception` with a flag would preserve it while restoring grouping.

Low severity; the fail-open contract is what matters here.

---

### WARNING — `git add -A` in test helpers (4 sites)

**File**: `tests/test_scripts/test_review_scope.py`

`_git(repo, "add", "-A")` appears in `test_build_manifest_enumerates_branch_changeset`, `test_build_manifest_includes_test_files_as_reviewable`, `test_build_manifest_excludes_binary`, and `test_build_manifest_excludes_vendored`. All run on isolated throwaway repos under `tmp_path` — zero real risk — but the pattern still violates the convention (same finding as PR #1300 where it appeared in `test_escalation_cap.py`).

Consistent fix: use specific file names (`_git(repo, "add", "src/engine.py")`, `_git(repo, "add", "tests/test_new.py")`, etc.) to match the convention even in test scaffolding.

---

Everything else looks solid. The module-level `try/except ImportError` fallback for `_is_docs_or_config` in `review_scope.py` is the right pattern. The `except (subprocess.TimeoutExpired, FileNotFoundError, OSError, UnicodeDecodeError)` in `_git()` is properly typed. Fail-open contracts, NUL-delimited parsing, and test coverage all check out cleanly. No heavy third-party imports in the hook path.

---

PR #1301, opened 2026-08-05
fix(observability): procedure_rebuild deferred item reads as recovery, not incident

WingedGuardian:
## What

The `procedure_rebuild` self-heal path (provider exhausted mid-extraction → watermark advances + a compensating rebuild is queued to `deferred_work_queue`) surfaced on the dashboard **Deferred review** panel with a reason string near-verbatim the *old data-loss bug it prevents* — reading as an active incident when it is in fact recovery working as designed (8-attempt retry, then an honest `procedure_extraction_lost` observation if providers stay down).

## Change

One string literal at the single source of truth (`memory/extraction_job.py`, `_enqueue_procedure_rebuild`). It is echoed verbatim through the queues snapshot and both the Deferred and Discarded review panels, so the one edit fixes every surface.

Before: `procedure builder provider-exhausted; watermark already advanced`
After:  `provider exhausted — auto-rebuild queued (self-healing; drains next extraction cycle)`

## Risk

None. Text only, no behavioral change, no test asserts the string. The one in-flight pending row keeps the old wording until it drains (future enqueues get the new string — not a data repair).


WingedGuardian (commented):

---

PR #1302, opened 2026-08-05
fix(db): bounded jittered retry on lock errors + per-process busy_timeout + MCP read pool (WS-1 PR-1)

WingedGuardian:
## What

Hardens the shared SQLite write path against "database is locked" errors observed under multi-process write bursts (N concurrent-session MCP children + the server, one WAL file). Root-cause analysis points at unfair busy-wait poll loss — SQLite's `busy_timeout` has no queue fairness, so under a burst one waiter can lose every re-poll for its full timeout while short writes convoy past it. The fix class is correct for any busy-loss mechanism (convoy or transient long-holder): bounded jittered application-level retry, fewer convoy participants, and longer per-child patience.

## How

1. **`SerializedConnection` retry** (`db/connection.py`): every SQL method routes through `_retry_locked` — up to 3 jittered backoff retries (0.25/0.5/1.0s ±20%) on lock errors, while HOLDING the serialization lock (releasing mid-episode would interleave another coroutine into the same implicit transaction). On exhaustion, the existing `_handle_lock_error` path runs unchanged, now counting exhausted *episodes* (reconnect-after-5 preserved). Idempotency (legacy implicit-BEGIN mode, no `isolation_level` anywhere): a locked execute never applied; a locked COMMIT leaves the txn open — or, in the WAL post-commit-autocheckpoint case, re-commit no-ops; ROLLBACK likewise, and gains lock handling it previously lacked entirely (a permanently-failed ROLLBACK wedges `in_transaction=True` until restart). `executescript` is deliberately NOT retried (partial-apply is not idempotent). `executemany` materializes generator args before attempt 1 (a consumed generator would make the retry "succeed" with zero rows). Lock predicate widened: both "database is locked" and "database table is locked", case-insensitive. Worst-case episode costs are documented in-code at the schedule constant (server ≈22s) and the MCP setdefault site (≈62s) with the why-acceptable argument and the quiet-period regression signal.
2. **Per-process `busy_timeout` knob** (`env.db_busy_timeout_ms`, `GENESIS_DB_BUSY_TIMEOUT_MS`, default 5000, floor 100): all consumers routed through the accessor — the 3 connection.py sites plus every direct importer (4 `mcp/health` writer sites, `eval/cli`, `guardian/watchdog`, `learning/procedural/session_inject`). The MCP child entrypoint `setdefault`s 15s (N child writers race ONE WAL writer slot; 5s loses to any server-side batch), added to the secrets allowlist so an operator override wins.
3. **Memory MCP child read pool** (`scripts/genesis_mcp_server.py`): `_bootstrap_memory` now builds the `ReadConnectionPool` (same size knob + kill switch as the server-side wiring, failure degrades to shared-connection reads) and passes it to `init()` — the child's recall reads stop contending for the writer slot. Pool closes before the db in the lifespan so in-flight reads keep their documented fallback.

## Review

Local adversarial review (Claude code-reviewer): DONE_WITH_CONCERNS — 1 SHOULD-FIX (9 remaining direct `BUSY_TIMEOUT_MS` sites bypassing the knob) fixed in-session; 1 design note (server-side episode blast radius needed an explicit in-code acceptance argument) addressed in-code. Retry counting, idempotency reasoning, predicate, contention-test determinism all verified by the reviewer.

## Test plan

- NEW `tests/test_db/test_write_retry.py` (17 tests): schedule + jitter bounds via a recording sleep seam; exhaustion after exactly 4 underlying attempts + WARN; non-lock errors not retried; "table is locked" variant retried; commit/rollback retried; executemany generator survives retry; executescript not retried; CancelledError during backoff propagates; env-knob defaults/override/floor/garbage + live PRAGMA application; and a REAL `BEGIN IMMEDIATE` cross-connection contention run with event-driven holder release (no wall-clock-dependent assertions).
- `test_connection_recovery.py` rewritten to episode semantics (the old tests locked in no-retry behavior by design).
- Hermeticized busy_timeout assertions (`test_connection.py`, `test_standalone_health.py`) against ambient env overrides.
- Sweeps: 1440 green (test_db + test_env + standalone_health), 1254 green (test_db + test_eval + test_learning + watchdog), 9 green (db_resilience + recall_read_pool). ruff clean.

## Deploy note

Runtime-only (no schema change; single-PR revert). Server pick-up on restart; MCP children pick up busy_timeout + read pool at next session start. Regression watch: exhaustion WARNs in quiet periods (would indicate a genuine long-holder, not convoy loss), write p95, MCP tool-call latency tails.

### DIFF
--- a/scripts/genesis_mcp_server.py
+++ b/scripts/genesis_mcp_server.py
@@ -226,7 +226,8 @@ def _bootstrap_memory(transport_kwargs: dict) -> None:
     async def _lifespan(server) -> AsyncIterator[None]:
         from qdrant_client import QdrantClient
 
-        from genesis.db.connection import get_db
+        from genesis.db.connection import ReadConnectionPool, get_db
+        from genesis.env import recall_read_pool_off, recall_read_pool_size
         from genesis.mcp.memory_mcp import init
         from genesis.memory.embeddings import EmbeddingProvider
         from genesis.memory.reranker import VoyageReranker
@@ -234,6 +235,25 @@ def _bootstrap_memory(transport_kwargs: dict) -> None:
 
         # Long-lived shared connection via SerializedConnection (see _bootstrap_health).
         db = await get_db(_DEFAULT_DB, foreign_keys=False)
+
+        # Read-only recall pool (WS-1 PR-1): reads leave the shared write
+        # connection, so this child stops contending for the WAL writer slot on
+        # every recall read — fewer convoy participants. Mirrors the server-side
+        # wiring in runtime/init/memory.py: same size knob, same kill switch,
+        # and failure degrades to read_pool=None (all reads on the shared
+        # connection — the pre-pool behavior), never a startup failure.
+        read_pool = None
+        if not recall_read_pool_off():
+            try:
+                pool = ReadConnectionPool(_DEFAULT_DB, size=recall_read_pool_size())
+                await pool.open()
+                read_pool = pool
+            except Exception:
+                logger.warning(
+                    "recall read pool unavailable — reads stay on the shared connection",
+                    exc_info=True,
+                )
+                read_pool = None
 
         try:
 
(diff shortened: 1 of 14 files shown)


WingedGuardian:
## First-pass review — PR #1302

### Structural review

**WARNING — Circular import: `src/genesis/env.py` ↔ `src/genesis/db/connection.py`**

`db_busy_timeout_ms()` in `env.py` contains a deferred `from genesis.db.connection import BUSY_TIMEOUT_MS` inside the function body, while `connection.py` already imports `from genesis.env import db_busy_timeout_ms` at module top-level. This works at runtime (Python's module cache prevents an `ImportError`), but it inverts the correct dependency direction — `env.py` should sit *below* `connection.py` in the import graph, not reach back into it.

The natural fix: move `BUSY_TIMEOUT_MS = 5000` into `env.py` itself (it's an env-tunable default, not a DB-layer implementation detail), then have `connection.py` import the constant from `env` rather than defining it locally. That eliminates the back-edge entirely and aligns with the new accessor's semantic: the value is now an environment knob, so `env.py` is where it belongs.

Everything else is clean:
- No `nohup`, `task_states` direct inserts, `/tmp/` usage, bare `except:`, mutable defaults, star imports, or raw SQL against `genesis.db`.
- `tests/test_db/test_write_retry.py` covers the new retry logic. `test_connection_recovery.py` rewritten for episode semantics.
- All new public functions (`db_busy_timeout_ms`, `_retry_locked`, `_is_lock_error`) carry docstrings.
- Conventional commit prefix (`fix(db):`) present.

---

### PII/Secrets/Wording scan

**CLEAN** — no secrets, keys, private IPs, private hostnames, or security-incident language found in added lines, PR title, or PR body.

---

WingedGuardian:
Both round-2 findings addressed in c99f0891: the two recall read-pool env vars joined `_MCP_VARS` (Codex P2), and `BUSY_TIMEOUT_MS` moved to `genesis.env` with a same-name re-export from `connection.py` for historical importers (structural warning — no back-edge remains).

PR #1304, opened 2026-08-05
feat(reflex): ingest job.failed + manual resolve/dismiss with taste-corpus verdict (PR-2b/2c)

WingedGuardian:
## Summary

Two cohesive steps for the reflex arc ("nerve" — Genesis's self-bug detection):
lets it hear the largest class of its own failures, and gives an ingested signal
a way to *leave* the `new` lane before the auto diagnose/fix lanes exist.

## PR-2b — ingest `job.failed`

PR-2a (#1226) funnels background-job exceptions onto the bus as `job.failed`, but
reflex intake was left dark — the ingestor dropped everything that wasn't
`task.failed`, so the single largest class of internal defects never reached the
self-bug store. This widens intake to `job.failed`.

Only **exception-driven** failures are ingested: a guard skips any event without
an `error_type`. `failure_details()` sets `error_type` iff a real exception caused
the failure; a semantic/reason-only failure (`error_reason`, no `error_type`)
belongs to the `job_health`/Sentinel lane, and ingesting it here would manufacture
a bogus `UnknownError` signal. Both current emitters fire exception-only, so the
guard is defense-in-depth honoring that contract.

## PR-2c — manual resolve/dismiss with a taste-corpus verdict

Until the auto lanes ship, a signal sits on the dashboard forever. New MCP tool
`reflex_signal_resolve` is the human exit — and writes the **first taste-corpus
verdict** (`reflex_verdicts`). Three dispositions, mapped to the schema's
CHECK-constrained vocabularies:

- `fixed` → status `resolved`, **no verdict row** — a real bug fixed out-of-band
  (e.g. a normal PR) is a lifecycle close, not a card judgment, and the verdict
  vocabulary has no "fixed_external" value; forcing `dismiss_wontfix` would poison
  the corpus (training data for the future judgment layer).
- `not_a_bug` → `dismissed_notbug` + verdict `dismiss_notbug`.
- `wont_fix` → `dismissed_wontfix` + verdict `dismiss_wontfix`.

Safety: the status transition is guarded on the observed status (a lost race is a
`conflict`, never a clobber) and runs **before** the verdict write (no orphan
verdicts); a verdict-write failure after the transition returns `partial`, never a
masked `ok`. Idempotency keys on a new public `TERMINAL_DISPOSED` set (clean
terminals only) — deliberately **not** the recurrence-reopen set — so failure/
expiry signals a human still needs to close stay resolvable.

Adds `reflex_verdicts` CRUD (first writer) + `reflex_signals.get_by_id`.

## Verification

- New tests RED-on-unfixed / GREEN-on-fixed (proven via stash for the intake
  guard). `tests/test_reflex/` full suite **83 pass** (incl. 10 resolve-tool +
  the new ingest cases); `tests/test_mcp/` health **318 pass**; the tool
  live-registers on the FastMCP instance. `ruff` clean; `subsystem-map` check
  CLEAN (`CURRENT.md` updated + re-stamped).

## Deploy path

Runtime code — `git pull` + server restart; MCP tool lands at the next CC session
start. Reflex ships **dark by default** (`config/reflex.yaml ingest_enabled:
false`); no schema migration (tables predate this from #1177), no config default
change, no host action.

## Notes

Behavior is scoped to the reflex arc, which is off by default on a standard
install — no user-facing effect there, hence no CHANGELOG entry (consistent with
the earlier dark reflex PRs).

### DIFF
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -673,8 +673,8 @@ verified: ca875c4b 2026-07-24
   their own domain `.failed` pass `emit_event=False` to avoid double-counting.
   Fire-and-forget (a broken emit never blocks the `job_health` write), and
   `job.failed` is in the ego's `_REFLEX_OWNED_EVENT_TYPES` so it never spins a
-  reactive ego cycle. Dark until reflex intake widens (`ingest.py` still drops
-  non-`task.failed`).
+  reactive ego cycle. Ingested by the reflex arc as of PR-2b (`ingest.py`
+  consumes `job.failed`, guarded on `error_type` presence).
 - **perception/**: the real-time reflection engine — MICRO (and LIGHT without
   a CC bridge) run in-process via the router; DEEP/STRATEGIC go to the CC
   reflection bridge. GROUNDWORK: user-model-synthesis, pre-execution-gate
@@ -1272,16 +1272,17 @@ verified: 9037d45b 2026-07-07
 
 ## 14. Reflex arc — self-bug detection & repair
 
-The afferent nerve for Genesis's own screaming bugs: detect `task.failed`
-exceptions, fingerprint/dedup them, and (later phases) diagnose → card →
-fix under a human-gated tier model. **PR1 (dark) + PR1.5 (observability)**
-— ingestion scaffolding and its watch surface; no cards/sessions/LLM yet.
-Spec: `docs/superpowers/specs/2026-07-21-reflex-arc-design.md`.
+The afferent nerve for Genesis's own screaming bugs: detect `task.failed` /
+`job.failed` exceptions, fingerprint/dedup them, and (later phases) diagnose →
+card → fix under a human-gated tier model. **PR1 (dark) + PR1.5 (observability)
++ PR-2b (job.failed intake) + PR-2c (manual resolve/dismiss with verdict)** —
+ingestion, its watch surface, and the human exit lane; no auto diagnose/fix
+sessions/LLM yet. Spec: `docs/superpowers/specs/2026-07-21-reflex-arc-design.md`.
 
 ```yaml subsystem-map
 entry: reflex-arc
 modules: [reflex]
-verified: 302066ad 2026-07-23
+verified: 4cc75d50 2026-08-05
 ```
 
 - **reflex/**: `fingerprint.py` (pure: normalize task name, line-number-free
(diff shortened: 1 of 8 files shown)


WingedGuardian:
## Structural Review

**NOTE — Missing docstring** (`src/genesis/db/crud/reflex_signals.py`): the new `get_by_id` function has no docstring. The adjacent `get_by_fingerprint` is also undocumented (consistent with the file's existing style), so this is in line with the module's conventions — flagging for awareness only, not blocking.

All pattern checks — nohup, `/tmp/` usage, `task_states` direct inserts, `git add .`/`git add -A`, bare `except:`, mutable default args, `os.killpg` without pgid guard, raw SQL outside crud modules, `IntervalTrigger`, `$CLAUDE_PROJECT_DIR` Python env var — **CLEAN**.

Raw SQL in `reflex_verdicts.py` and `reflex_signals.py` is in the approved `genesis.db.crud` location. Both `except Exception:` clauses carry an explicit type and documented rationale (bus-observer contract / taste-corpus partial-write). New files have corresponding test coverage (`TestVerdictsCrud`, `TestResolveTool`). Conventional commit prefix `feat(reflex):` present. ✓

---

## PII/Secrets/Wording Scan

**Code diff — CLEAN.** No API keys, tokens, private IPs, SSH keys, credentials, or personal email addresses in any added line. Test fixtures use generic placeholder values (`CCProcessError`, `"fp1"`, timestamp constant). No hardcoded `/home/<username>/…` paths.

**PR metadata — CLEAN.** Title and body describe a feature extension to the reflex arc (job.failed intake + manual resolve). No security-incident language, no "fixed credential/IP leak" wording, no private infrastructure names. The no-CHANGELOG rationale (reflex ships dark by default) is explicit and appropriate.

---

WingedGuardian:
Addressed the P2 in commit 49c4c989: a partial verdict-write is now recoverable through the same API. The terminal-status branch detects a signal whose status matches the requested disposition but whose expected verdict is absent (the partial case) and REPAIRS it — writes the missing taste-corpus verdict and returns `repaired` — instead of no-op'ing forever. A genuinely already-recorded signal still no-ops. Test `test_partial_then_retry_repairs_the_missing_verdict` covers fail→partial→repaired→noop.

WingedGuardian:

---

WingedGuardian:
This P2 is a re-post of the same finding already addressed in commit 49c4c989 (the reviewer's line-142 anchor is the terminal-status branch, but the repair intercept sits immediately inside it at lines 143-161, *before* the noop return at 162). A partial-state signal retried with the same disposition detects the missing verdict and REPAIRS it (returns `repaired`), so the corpus row is recoverable via this API — it is not permanently lost. Covered by `test_partial_then_retry_repairs_the_missing_verdict` (fail→partial→repaired→noop). No further change needed.

PR #1315, opened 2026-08-06
fix(outreach): email replies register as engagement (reply→ledger bridge)

WingedGuardian:
## Problem

When someone replies to an outreach email, the reply poller matched it to its
`email_threads` row and updated thread state — but never wrote the engagement
back to `outreach_history`. Consequences:

- Real replies read as **zero engagement** (`engagement_outcome` stayed NULL,
  then the 24h timeout job stamped them `'ignored'/'timeout'`).
- The WS-2 ledger's `reply_received` prediction lane resolves off
  `outreach_history.engagement_signal`, so every real email reply **graded as
  silence (0)** at deadline — silently miscalibrating the ledger.

## Fix

Bridge the matched-reply event into the outreach ledger, injected at the
composition root so `genesis.mail` stays outreach-agnostic:

- **`EngagementTracker.record_reply_if_pending`** — guarded write: promotes
  NULL **or a machine-stamped state** to `'useful'/'user_reply'`. The guard is
  on the outcome+signal **pairing** (`('ignored','timeout')`,
  `('ambivalent','implicit_activity')`, `('ambivalent','auto_digest')`):
  the dashboard `/engage` route rewrites the outcome without touching the
  signal, so a human verdict can sit on a stale mechanical signal — any
  outcome outside `{ignored, ambivalent}` is definitionally human and is
  never overwritten. Late replies landing after the 24h timeout stamp (the
  poller is 4-hourly; 24–72h replies are common) now register correctly.
- **`MECHANICAL_ENGAGEMENT_SIGNALS`** contract in `outreach/types.py`
  (mirrors the `POSITIVE_ENGAGEMENT` pattern).
- **`make_reply_engagement_bridge`** — closure resolving `outreach_id` from
  the thread `context` stashed by `pipeline._deliver` at send time.
- **`ReplyPoller.set_engagement_bridge`** + `poll()` hook — fires after the
  `email_threads` reply record and **before** the reply handler, in its own
  try/except: engagement registers even when the handler declines to act
  (e.g. sanitizer-blocked), and a bridge failure never breaks reply tracking.
- Wiring in `runtime/init/outreach.py` (mirrors the existing thread-tracker
  wiring; mail init runs first so the poller exists).
- `CLAUDE.md`: adds the timezone rule (store/compute UTC; schedule/display via
  `user_timezone()`) — user-directed this session, intentional ride-along.

## Tests (all verified RED-first)

`tests/test_mail/test_reply_poller_engagement.py` (9 tests, real code paths —
no mocking of the tracker/engagement/grader chain):
baseline documenting the bug · bridge happy path · human-verdict no-clobber
(incl. the stale-signal dashboard case) · late-reply timeout override ·
mechanical-ambivalent override · missing-`outreach_id` degradation · layering
guard (`genesis.mail` imports no `genesis.outreach`) · E2E through
`ledger.writers.on_outreach_delivered` → `grader.grade_due_predictions`
proving a bridged reply grades `reply_received=1` while a silent send grades 0.

## Review record

3 reviewer rounds: R1 found the timeout-race BLOCKER (fixed as the
mechanical-signals class), R2 found the stale-signal pairing BLOCKER (fixed +
regression test), R3 faithfulness pass: PASS, cleared for merge. Local commit
gate raised a false-positive despite a current marker (identical scripts,
synthetic run allows) — landed with a logged review-override; follow-up filed
to root-cause the gate divergence.

## Deploy path

Runtime code only — `git pull` + server restart. No schema change, no config.
Fresh-install/empty-state safe: without registered threads or the bridge wired
(no IMAP creds), every path degrades to a logged no-op.

Ledger: 919884d9329047d7863a58f071aab2ba

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,9 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   as zero and the prediction ledger graded every real reply as silence. The
   reply poller now writes the engagement back to the outreach record (without
   overwriting a richer outcome you set manually), so reply metrics and ledger
-  calibration reflect reality.
+  calibration reflect reality. Automated messages (out-of-office responders,
+  bounces, list mail) and replies from an address other than the one contacted
+  are filtered out, so they can't inflate the reply rate.
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ## [Unreleased]
 
+### Fixed
+
+- **Email replies now count as engagement.** When someone replies to an email
+  Genesis sent (outreach pitch, follow-up), the reply was recorded for thread
+  tracking but never registered as an engagement outcome — so reply rates read
+  as zero and the prediction ledger graded every real reply as silence. The
+  reply poller now writes the engagement back to the outreach record (without
+  overwriting a richer outcome you set manually), so reply metrics and ledger
+  calibration reflect reality.
+
 ### Added
 
 - **Real spreadsheets and polished decks from the deliverable builder.** When the
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -403,6 +403,8 @@ When a user shares a file path or URL in conversation:
   repairs ONE instance on ONE install. Label it "data repair", and fix the
   mechanism in the same session — or get the user's explicit deferral. Never
   report a data repair as "fixed".
+- **Timezones**: store/compute in UTC; schedule/display via
+  `genesis.env.user_timezone()` (every CronTrigger + user-facing timestamp).
 - **Telegram reminders**: use `outreach_send` with `preferred_timing`,
   NOT the `/schedule` skill (that's Claude Code's remote scheduler).
 - **Cognitive co-pilot, not order taker.** On every task, ask: "what else
--- a/src/genesis/mail/reply_poller.py
+++ b/src/genesis/mail/reply_poller.py
@@ -36,7 +36,7 @@ class ParsedReply:
 
     __slots__ = (
         "message_id", "in_reply_to", "references", "sender",
-        "subject", "body_preview", "imap_uid",
+        "subject", "body_preview", "imap_uid", "is_auto",
     )
 
     def __init__(
@@ -49,6 +49,7 @@ class ParsedReply:
         subject: str,
         body_preview: str,
         imap_uid: int,
+        is_auto: bool = False,
     ) -> None:
         self.message_id = message_id
         self.in_reply_to = in_reply_to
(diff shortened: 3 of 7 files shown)


WingedGuardian:
## First-pass review — PR #1315

### Structural review

**⚠️ WARNING — Raw SQL in `engagement.py`**

`EngagementTracker.record_reply_if_pending` (new method) uses `self._db.execute("UPDATE outreach_history …")` directly. Per the pattern rule, non-migration code should go through `genesis.db.crud` modules.

Flagging as WARNING rather than ERROR: the rest of `EngagementTracker` (`record_reply`, `record_implicit_engagement`, etc.) already uses this pattern and CI suppresses it via `# noqa: S608`. The new method follows the established class precedent. The real fix is a full crud migration for the class — flagging a single new method as a hard block while the class body does the same thing is not useful first-pass signal. But this is worth tracking.

---

**📝 NOTE — Forward-reference annotation on `make_reply_engagement_bridge`**

```python
def make_reply_engagement_bridge(tracker: EngagementTracker):
```

`make_reply_engagement_bridge` is placed **before** `class EngagementTracker:` in `engagement.py`. Without `from __future__ import annotations` at the top of the file, Python evaluates function annotations at definition time and this raises `NameError: name 'EngagementTracker' is not defined` on module load.

The tests passed (PR says RED-first verified), so `from __future__ import annotations` is likely present — but confirm it's there. If the file doesn't have it, this is a silent import-time crash waiting for a production restart.

---

**📝 NOTE — Deferred local import in method body**

```python
async def record_reply_if_pending(self, outreach_id: str, reply_text: str) -> bool:
    ...
    from genesis.outreach.types import MECHANICAL_ENGAGEMENT_SIGNALS_SQL_IN
```

`outreach.types` is a same-package constants module. A deferred import inside the method body is the pattern for avoiding circular imports, but `types.py` shouldn't import `engagement.py`. If there's no circular dependency, this can be a top-level import (makes the intent clearer and avoids re-importing on every call). If the deferral IS for circular-import avoidance, a brief comment explaining why would help the next reader.

---

### PII/Secrets/Wording scan

**CLEAN** — no API keys, private IPs, credentials, SSH keys, or connection strings in added lines. `@example.com` addresses in tests are on the ignore list. PR title, body, and CHANGELOG additions contain no incident/exposure/credential language.

---

WingedGuardian:
## Review dispositions

Thanks — three good catches. Dispositions:

**P2 (reply_poller.py:196) — foreign/automatic messages inflating engagement → FIXED in this PR.**
`match_reply` matches on `Message-ID` alone, so an OOO responder / bounce / list echo could match a thread and get stamped `useful`. Added a gate on the *bridge only* (thread recording is unchanged): the engagement write is skipped when the message is automated (RFC-3834 `Auto-Submitted != no`, `Precedence: bulk/junk/list`, common vendor `X-Auto*` headers) or when the reply's sender address (parsed, bare) doesn't match the address we mailed. Four new tests cover auto-responder, `Precedence: bulk`, foreign-sender, and a genuine reply with a display name still counting.

**P2 (reply_poller.py:196) — durable retry of failed bridge writes → deferred (tracked).**
Correct that a transient write failure is currently logged + counted but not retried. The proper fix is a reconciliation sweep over `email_threads` in `'replied'` state whose linked outreach row is still unset, run on the existing hourly engagement job — which self-heals transient failures *and* retroactively bridges pre-bridge replies. That needs record-time auto-detection stored on the thread (so the sweep applies the same genuine-reply filter as the live gate), which is a schema touch best done as its own change. Split out to keep this PR tight; tracked as a follow-up.

**P2 (engagement.py:125) — late reply vs. prediction deadline → pre-existing, tracked separately.**
This is a real property of the WS-2 ledger grader (terminal one-shot grading + the twice-daily cadence), and it predates this PR — the same ordering dependence exists on the Telegram engagement path. The fix (persist reply arrival time and have `_resolve_reply_received` compare it against `deadline_at`) belongs in the ledger lane, not the mail bridge. Tracked as a follow-up.

**Structural warning (raw SQL in engagement.py) — consciously accepted.**
The `IN (...)` list is a trusted module constant (`MECHANICAL_ENGAGEMENT_SIGNALS_SQL_IN`, sorted literals, no user input), the exact pattern already used by `POSITIVE_ENGAGEMENT_SQL_IN` in `db/crud/outreach.py`. Bound params are used for all values. `S608` is `noqa`'d with that justification.

WingedGuardian:
Re-review dispositions (Codex re-emitted the three P2s on the gate commit as line numbers shifted):

**Foreign/automatic messages (reply_poller.py:225) — now resolved by this commit.** The finding reads the `poll()` call site as an unconditional bridge, but the gate lives inside the injected bridge closure (`engagement.py` `make_reply_engagement_bridge`): it skips the engagement write when the message is automated (RFC-3834 `Auto-Submitted`/`Precedence` + vendor `X-Auto*`) or when the sender's bare address doesn't match the recipient. Thread recording is intentionally left unconditional; only the engagement write is gated. Covered by the four new tests (auto-responder / `Precedence: bulk` / foreign-sender skip; display-name genuine reply still counts).

**Retry durability + late-reply timestamp** — unchanged from the prior disposition: both tracked as follow-ups (reconciliation sweep for durable retry + backfill; deadline-aware grading in the WS-2 ledger lane). Neither blocks this PR.

PR #1323, opened 2026-08-06
perf(health): serve top_tags stale-while-revalidate + offload /proc scans off the loop (MW-0)

WingedGuardian:
## What

Removes the dominant chronic event-loop blocker in genesis-server: `memory/health.py::distribution_stats` scanned the entire `memory_fts` corpus (~69k rows) and tag-counted it in pure Python **on the event loop**, fired on every health snapshot by internal consumers (ego context builders, sentinel, morning report). Measured on real data: ~596ms per call. Its only output, `top_tags`, has **zero production consumers**.

## The fix

**A1 — `top_tags` served stale-while-revalidate** (`memory/health.py`). The cheap `by_collection`/`total` aggregates stay fresh on the caller connection. `top_tags` is served from a module cache that a single-flight background `tracked_task` refreshes off-loop, via a **separate read-only** sqlite connection (`mode=ro`, WAL-visible) so it never touches the shared write connection. Staleness = never-built OR corpus-count drift >10% OR age >1h. Mirrors the existing `memory/intent.py::TagCooccurrenceIndex` SWR pattern. Env kill switch `GENESIS_TOP_TAGS_DISABLED` neutralizes the background rebuild.

**A2 — off-loop offloads.** The ~1s synchronous `/proc` scan `enumerate_cc_slots()` and the session-observer's two filesystem walks are wrapped in `asyncio.to_thread`, at every call site on the loop (awareness tick, infrastructure snapshot, CC-sessions detail route, pending-observations processing).

## Impact (measured, real corpus)

| | before | after |
|---|---|---|
| `distribution_stats` hot-path cost | ~620ms (on-loop) | **29ms** (cheap queries only) |
| top-tags scan | every snapshot, on-loop | hourly, off-loop, single-flight |

Return shape is backward-compatible (`by_collection`, `total`, `top_tags` unchanged; adds `top_tags_warming` bool — no consumer reads `top_tags`).

## Tests

- RED witnessed first: `distribution_stats` scanning `memory_fts` on the caller connection.
- `distribution_stats_never_scans_fts_on_caller_connection` regression guard (file-backed, proves the scan runs on a separate connection); `_scan_top_tags` correctness + RO-write-rejected; rebuild-populates-cache; single-flight + staleness-delta; kill switch.
- A2 offload guards (thread-dispatch spies) for all four sites.
- 86 affected-suite tests green + 176-test collateral ring (loop, health-snapshot consumers, ego context) green; ruff clean.

## Rollback

`GENESIS_TOP_TAGS_DISABLED=1` + restart neutralizes A1's background machinery (serves empty top_tags); A2 offloads are pure `to_thread` wrappers with no switch needed.

## Follow-up

If the post-deploy soak still shows lag-WARN episodes clustering with snapshot activity, add a ~20–30s serve-stale cache inside `HealthDataService.snapshot()` (internal-snapshot TTL — pre-scoped fast-follow).

### DIFF
--- a/src/genesis/awareness/loop.py
+++ b/src/genesis/awareness/loop.py
@@ -1458,7 +1458,8 @@ async def _check_cc_slot_memory(db, slots: list[dict] | None = None) -> None:
             enumerate_cc_slots,
         )
         if slots is None:
-            slots = enumerate_cc_slots()
+            # /proc scan is ~1s of sync syscalls — keep it off the event loop.
+            slots = await asyncio.to_thread(enumerate_cc_slots)
     except Exception:
         logger.debug("cc_slot memory check: enumeration failed", exc_info=True)
         return
--- a/src/genesis/dashboard/routes/cc_sessions.py
+++ b/src/genesis/dashboard/routes/cc_sessions.py
@@ -16,6 +16,7 @@ waypoint spine (this is its first reader), and repo-pulse annotations.
 
 from __future__ import annotations
 
+import asyncio
 import json
 import logging
 import sqlite3
@@ -261,7 +262,8 @@ async def cc_sessions_detail():
     if not rt.is_bootstrapped or rt.db is None:
         return jsonify({"error": "not bootstrapped"}), 503
 
-    slots = enumerate_cc_slots()
+    # /proc scan is ~1s of sync syscalls — keep it off the event loop.
+    slots = await asyncio.to_thread(enumerate_cc_slots)
     deploy = await last_successful_update(rt.db)  # (completed_at, new_commit) | None
     return jsonify(await _collect_detail(rt.db, slots, deploy=deploy))
 
--- a/src/genesis/memory/health.py
+++ b/src/genesis/memory/health.py
@@ -5,11 +5,17 @@ Pure SQL aggregates and Qdrant vector queries — NO LLM judgment.
 
 from __future__ import annotations
 
+import asyncio
 import logging
+import os
 import random
+import sqlite3
+import time
 from datetime import UTC, datetime, timedelta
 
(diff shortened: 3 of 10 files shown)


WingedGuardian:

**Structural:** 2 WARNINGs — no ERRORs

---

### WARNING — Missing `@pytest.mark.asyncio` on new test

**File:** `tests/test_dashboard/test_cc_sessions_routes.py`
**Function:** `test_detail_route_offloads_slot_enumeration`

```python
async def test_detail_route_offloads_slot_enumeration(app, db, monkeypatch):
```

The function is declared `async def` but has no `@pytest.mark.asyncio` decorator. Every other async test added in this PR carries the decorator. Without it (absent an `asyncio_mode = auto` config), pytest-asyncio silently collects the test but never runs the coroutine body — the guard passes vacuously and provides zero regression protection.

Fix: add `@pytest.mark.asyncio` above the function definition, consistent with the rest of the suite.

---

### WARNING — Raw `sqlite3` against genesis.db outside a migration module

**File:** `src/genesis/memory/health.py`
**Function:** `_scan_top_tags`

```python
conn = sqlite3.connect(f"file:{db_path}?mode=ro", uri=True)
```

This opens a synchronous `sqlite3` connection directly to the genesis.db path. The project convention is to access genesis.db via the `genesis.db.crud` modules, except in migrations. The architectural justification here is solid — the function must run in a thread (off the event loop) and needs a separate read-only connection that aiosqlite's async API cannot provide — but the deviation is worth a conscious sign-off. A short inline note or a module-level exemption comment (`# not using crud: needs a sync RO connection in asyncio.to_thread`) would document the intent for the next reader.

---

PR #1326, opened 2026-08-07
feat(campaigns): github_activity_pending pre-check to cost-gate the GitHub activity digest

WingedGuardian:
## What

Adds a campaign pre-check, `github_activity_pending`, and plumbs the runner's
`db` handle into the pre-check `ctx`. The pre-check gates a 6h "GitHub activity
digest" campaign (the LLM tier that consumes the account-activity monitor's
observations): it passes — allowing an LLM session to dispatch — only when the
monitor has recorded at least one **unresolved** `github_account_activity`
observation (source `recon`) since the campaign's `last_run_at`. A quiet window
therefore spawns no session and costs nothing.

This resolves the standing gap where `github_account_activity` observations had
no consumer: the digest campaign (shipped separately as install-local user data,
per the campaigns public/private contract) is that consumer, and this pre-check
makes its dispatch cost-gated.

## How

- `check_github_activity_pending(campaign, *, ctx)` counts unresolved
  `github_account_activity`/`recon` observations created after `last_run_at` via
  the existing `count_recent_unresolved_by_type_and_source` CRUD. `>0` → pass;
  `0` → skip with a reason.
- **Fail-open** on a missing `db` handle or first run (no `last_run_at`): a
  spurious pass is cheap (the digest no-ops on an empty read), whereas a wrong
  gate would silently swallow a contributor signal.
- Runner `ctx` gains `"db": self._db` — inert for the three existing pre-checks,
  which read only their own keys.

### Timestamp-compare safety

`last_run_at` is `datetime.now(UTC).isoformat()` (`+00:00` suffix); the monitor's
obs `created_at` is `_now_z()` (`Z` suffix). The count is a lexical
`created_at > since` compare. Both share the fixed-width `YYYY-MM-DDTHH:MM:SS`
prefix, so lexical order equals chronological order except within a single
second — where the differing suffix biases toward counting a same-second obs as
new (harmless false-positive, made idempotent by the digest resolving rows it
consumes). It **never** produces a false-negative, so a genuinely-new obs is
never dropped. Boundary tests pin both the same-second bias and the tightest
worst-case sub-second case.

## Tests

RED-first (`tests/test_campaigns/test_prechecks.py::TestGithubActivityPending`):
gating on no / predating / resolved / foreign-type-or-source activity; passing
on new activity; same-second and worst-case sub-second boundaries; both
fail-open branches; and registry + `db`-in-ctx wiring end-to-end through
`run_prechecks`. Full `test_campaigns` suite: 69/69. Ruff clean.

## Review

Reviewed by a local code-reviewer agent (verdict: clean — no functional bug;
its one finding, a test-coverage gap on the same-second boundary claim, is fixed
in this PR by the two boundary tests). The external cloud reviewer was
unavailable at authoring time (usage quota resets Aug 8); this PR can either
merge on green CI or hold for that pass, at maintainer discretion.

## Notes

- No behavior change on any existing campaign — the new check only runs for a
  campaign that explicitly lists `github_activity_pending` in its `pre_checks`.
- No private data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,10 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   be delivered, it's retried until it lands rather than lost. Off by default on a
   fresh install (it baselines quietly first); enable pings by setting
   `mode: live` in a `config/github_steward.local.yaml` overlay once it has run.
-  Kill switch: `GENESIS_GITHUB_STEWARD_DISABLED=1`.
+  Kill switch: `GENESIS_GITHUB_STEWARD_DISABLED=1`. A companion campaign
+  pre-check (`github_activity_pending`) lets an opt-in digest campaign batch the
+  non-urgent activity into a periodic Telegram summary while only spending on
+  ticks that actually have new activity — quiet windows spawn nothing.
 
 ### Fixed
 
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -318,7 +318,10 @@ verified: 9037d45b 2026-07-07
   campaign names/prompts/targets are USER DATA (DB + private backups), never
   tracked source; zero shipped defaults. `CampaignRunner` cron-ticks
   programmatic prechecks then dispatches DirectSessions; a 120s reaper
-  reconciles finished sessions.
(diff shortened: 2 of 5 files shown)


WingedGuardian:

Reviewed `prechecks.py`, `runner.py`, `test_prechecks.py`, `CHANGELOG.md`, `CURRENT.md` against all Domain 1 and Domain 2 rules. Nothing to flag.

---

Issue #1310, opened 2026-08-06
A "|" or newline in an inbox item corrupts the digest table


WingedGuardian:
# A `|` or newline in an inbox item corrupts the whole digest table

> 👋 **This one is reserved for a first-time contributor.** If you've contributed
> to open source before, please leave it for someone who hasn't. If this is your
> first PR ever, it's yours — I'll help you through it and respond within 24 hours.

## The bug

`_format_digest()` builds Markdown tables by dropping each item's text straight
into a table row like `| {priority} | {content} | {strategy} |`. Item text is
free-form, so if a `content`, `strategy`, `resolution_notes`, or `source` value
contains a `|` or a newline, it breaks the table:

- a `|` inside the text injects an extra column, shifting that row
- a newline (`\n`) splits the row in half and breaks **every** row after it

Example: a follow-up whose content is `"Compare X | Y"` renders as a broken row
instead of showing the text `Compare X | Y` in one cell.

## Where

One file: **`src/genesis/mcp/health/inbox_digest.py`**, the `_format_digest()`
function (around lines 100–172). The affected rows are built at lines 121, 133,
147 (and the values come from `content`, `strategy`, `resolution_notes`,
`source`).

## What to change

1. Add a tiny helper near the top of the function (or just above it) that makes a
   value safe for a single Markdown table cell — escape `|` and flatten newlines:
   ```python
   def _cell(value: str) -> str:
       """Escape a value so it stays inside one Markdown table cell."""
       return value.replace("\n", " ").replace("|", "\\|")
   ```
2. Wrap the **free-text** fields in `_cell(...)` where they're interpolated into
   the rows — the `content` values (pending, ideas, resolved), `strategy`,
   `resolution_notes` (the `notes` variable), and `source`. The priority / date /
   age fields are fixed-format and don't need it (wrapping them anyway is
   harmless if you prefer consistency).

Roughly a 3–6 line change. Don't restructure the tables — just sanitize the cells.

## How to test it

Add a test to the existing class **`TestFormatDigest`** in
**`tests/test_inbox/test_inbox_digest.py`** (next to `test_format_with_data`):

```python
def test_format_escapes_pipe_and_newline(self):
    from genesis.mcp.health.inbox_digest import _format_digest

    pending = [
        {"priority": "high", "content": "Compare X | Y\nsecond line",
         "strategy": "ego_judgment"},
    ]
    output = _format_digest(pending, [], [], 7)
[...]

PR #1328, opened 2026-08-07
Escape markdown table cells in inbox digest

AyushkhatiDev:
## Summary

Escape Markdown table cell values in `_format_digest()` to prevent table
corruption when free-text fields contain pipe characters (`|`) or newlines.

## Changes

- Add `_cell()` helper to sanitize Markdown table cell values.
- Escape `|` as `\|`.
- Replace embedded newlines with spaces.
- Apply sanitization to free-text fields used in digest tables.
- Add characterization test covering pipes and newlines.

## Testing

```bash
pytest tests/test_inbox/test_inbox_digest.py

### DIFF
--- a/src/genesis/mcp/health/inbox_digest.py
+++ b/src/genesis/mcp/health/inbox_digest.py
@@ -97,6 +97,11 @@ async def _impl_inbox_digest(
         return {"error": f"Failed to generate inbox digest: {exc}"}
 
 
+def _cell(value: str) -> str:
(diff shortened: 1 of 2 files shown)


WingedGuardian:
Merged — thanks @AyushkhatiDev! This nicely fixes the table-cell corruption from #1310: a `|` or newline in an item no longer breaks (or can inject a row into) the digest table. Clean, well-scoped, and tested. 🙌

PR #1330, opened 2026-08-07
fix(memory): incremental byte-offset transcript read + off-loop extraction (WS-1)

WingedGuardian:
## Problem
The memory-extraction cycle re-read every extractable transcript from byte 0 **synchronously on the asyncio event loop** each pass, blocking it 6–7s and starving memory-recall (HTTP 503s). Confirmed by overnight py-spy: 6/6 mega-spike event-loop-lag WARNs caught MainThread on-CPU in `read_transcript_messages`, timestamp-matched to the WARNs.

## Fix
- **`read_transcript_delta`** (`util/jsonl.py`): binary-mode byte-offset resume — `seek()` to a stored offset and read only the delta; stat-gate (skip without opening) when file size == stored offset; reset to a full scan on truncation/rotation.
- **`run_extraction_cycle`**: offloads `_find_transcript` + the read via `asyncio.to_thread`, and persists a new `cc_sessions.last_extracted_byte` watermark **in lock-step** with `last_extracted_line`.
- **Invariant** (test-gated): `last_extracted_byte`, when set, is the byte start of line `last_extracted_line`. NULL → legacy full scan, then self-populates. A truncation reset also resets the line base so the invariant holds on the shrunk file.
- **Schema**: nullable `last_extracted_byte` via migration `0080` + base-path `_try_alter` (dual path); no backfill, idempotent, `down()` present.

`read_transcript_messages` left unchanged (parity-tested). Chronic sub-second lag from infra-snapshot/SSL work on the loop is tracked separately.

## Tests (RED-first)
- `test_jsonl_incremental.py` (10): byte-accurate delta, the invariant, stat-gate, truncation reset, partial-trailing-line, multibyte, legacy fallback, parity.
- `test_extraction_offload.py` (5): off-loop execution (`get_ident`), dual-write invariant on both the messages and empty paths (real-DB persistence), stat-gate on the second cycle, **truncation-with-new-content regression** (seen RED first), **cap-break** lock-step.
- `test_migration_0080_*.py` (4): apply/idempotent/NULL/down.

Local review: 2 rounds (1 blocker — truncation watermark staleness — found, fixed RED→GREEN, confirmed clean). Full collateral ring + ruff green.

Follow-up: 470cec53

### DIFF
--- a/src/genesis/db/crud/cc_sessions.py
+++ b/src/genesis/db/crud/cc_sessions.py
@@ -507,7 +507,7 @@ async def get_extractable(
     status_ph = ",".join("?" for _ in statuses)
     cursor = await db.execute(
         f"SELECT id, cc_session_id, source_tag, last_extracted_at, "
-        f"       last_extracted_line, started_at "
+        f"       last_extracted_line, last_extracted_byte, started_at "
         f"FROM cc_sessions "
         f"WHERE source_tag IN ({tag_ph}) "
         f"  AND status IN ({status_ph}) "
@@ -524,12 +524,26 @@ async def update_extraction_watermark(
     *,
     last_extracted_line: int,
     last_extracted_at: str,
+    last_extracted_byte: int | None = None,
 ) -> bool:
-    """Update the extraction watermark for a session."""
-    cursor = await db.execute(
-        "UPDATE cc_sessions SET last_extracted_at = ?, last_extracted_line = ? WHERE id = ?",
-        (last_extracted_at, last_extracted_line, id),
-    )
+    """Update the extraction watermark for a session.
+
+    ``last_extracted_byte`` is the incremental-resume hint (byte offset of the
+    START of line ``last_extracted_line``). When None the column is left
+    untouched, preserving any prior value — callers that computed a byte offset
+    pass it; legacy/reference paths omit it.
+    """
+    if last_extracted_byte is None:
+        cursor = await db.execute(
+            "UPDATE cc_sessions SET last_extracted_at = ?, last_extracted_line = ? WHERE id = ?",
+            (last_extracted_at, last_extracted_line, id),
+        )
+    else:
+        cursor = await db.execute(
+            "UPDATE cc_sessions SET last_extracted_at = ?, last_extracted_line = ?, "
+            "last_extracted_byte = ? WHERE id = ?",
+            (last_extracted_at, last_extracted_line, last_extracted_byte, id),
+        )
     await db.commit()
     return cursor.rowcount > 0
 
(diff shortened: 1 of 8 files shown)


WingedGuardian:

Checked all 8 changed files across both passes:

**Structural (Domain 1)**
- No `nohup`, no `task_states` direct inserts, no `/tmp/` paths, no destructive git commands.
- No bare `except:`, no mutable default arguments (all new `None` defaults are safe).
- `import asyncio` correctly promoted from two deferred inline positions to top-level — clean import hygiene.
- No star imports, no circular imports introduced.
- Migration `_add_column` uses f-strings for `PRAGMA` and `ALTER TABLE`, but both are fed only hardcoded strings from `up()`/`down()` — no user input path. Acceptable in a migration file.
- All new `src/` files (`0080_cc_sessions_extracted_byte.py`) have corresponding test coverage (`test_migration_0080_cc_sessions_extracted_byte.py`). Modified paths covered by `test_extraction_offload.py` and `test_jsonl_incremental.py`.
- All new public APIs (`TranscriptDelta`, `read_transcript_delta`, `_parse_transcript_line`) carry docstrings.
- Commit prefix `fix(memory):` ✓. No `IntervalTrigger`, no `$CLAUDE_PROJECT_DIR`, no hook scripts with heavy imports.

**PII/Secrets/Wording (Domain 2)**
- No API keys, tokens, passwords, private IPs, IPv6 ULA addresses, SSH keys, or DB connection strings in any added line.
- No personal email addresses, hardcoded `/home/<user>` paths, hardware identifiers, or private hostnames.
- PR title and body describe a performance/correctness fix; no language suggesting a security incident, credential cleanup, or private infrastructure detail.

---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
@codex review — addressed in 58e5fae2: TranscriptDelta.failed now set on both stat()/read OSError paths and the empty-branch watermark write is gated on it, so an I/O error never advances or initializes the watermark. Unit + integration tests added.

WingedGuardian:
@codex review — findings 2 & 3 addressed in 482a39fe: (2) read_transcript_delta now validates the stored offset is a line boundary before seeking and falls back to a full scan on a reshaped/replaced file; (3) run_extraction_cycle hoists `if delta.failed: continue` above the messages check so a partial-read failure is discarded (not extracted/advanced). RED→GREEN tests added; 20 passed.

WingedGuardian:
@codex review — finding 4 addressed in 2e807840: the max_lines cap now counts from the effective scan base (emit_from), so it stays bounded after a truncation/boundary reset instead of subtracting the stale start_line. Seeking behavior unchanged. RED→GREEN test added.

PR #1334, opened 2026-08-07
fix(telegram): nudge instead of confusing on topic-root approval replies

WingedGuardian:
## Problem
Replying a bare "yes"/"approve" to a forum **topic root** (the topic header) instead of the specific approval/proposal/content message matched no resolver in `handle_text` and fell through to a fresh, context-free CC conversation session. The topic-root service message has no `.text`, so the session saw only a naked "yes" and replied "I don't have anything to confirm — what are you saying yes to?".

## Fix (scope: minimal-safe — nudge only)
Detect the topic-root reply (`reply_to_message.message_id == message_thread_id`) in a decision topic (`approvals`/`content_review`/`ego_proposals`) that parses as a bare decision but matched nothing, and reply with a nudge to reply to the specific message (or tap ✅). It **deliberately does not auto-resolve** — resolving most-recent-pending from an ambiguous topic-root reply risks acting on the *wrong* item.

The fuller fix (a structured content-approval resolution with a terminal-state column + publish-idempotency, so a topic reply can safely resolve a specific content item) is deferred to its own design follow-up — content approvals currently have no resolver at all and rely on emergent CC re-derivation.

## Review & verification
1 review round, **clean**. Verified: correct topic-root signal; no false-positive interception (specific-message replies and substantive messages fall through); ordering masks no real resolver; nudge-only / no writes; fail-closed on runtime unavailability. Ruff clean; 6 nudge tests + 74 telegram-handler regression tests pass.

Partially addresses follow-up 8785140f (the confusing-fresh-session half). Bug A (CLI-approval topic-root fallback) and the structured content-approval resolution remain in that follow-up.

Follow-up: 8785140f

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Replying "yes" to an approval topic no longer starts a confused new chat.**
+  If you replied "yes"/"approve" to the *topic itself* (the forum topic header)
+  instead of the specific approval/proposal/content message, Genesis got no
+  context and spun up a fresh conversation that answered "I don't have anything
+  to confirm — what are you saying yes to?". It now recognizes that case and
+  asks you to reply to the specific message (or tap its ✅ button) rather than
+  guessing — it deliberately won't act on an ambiguous topic-level reply.
+
 - **Email replies now count as engagement.** When someone replies to an email
   Genesis sent (outreach pitch, follow-up), the reply was recorded for thread
   tracking but never registered as an engagement outcome — so reply rates read
--- a/src/genesis/channels/telegram/_handler_messages.py
+++ b/src/genesis/channels/telegram/_handler_messages.py
@@ -1229,6 +1229,63 @@ async def _try_ego_correction_store(ctx: HandlerContext, msg) -> bool:
     return False
 
 
+def _is_topic_root_reply(msg) -> bool:
+    """True if ``msg`` is a reply to a forum topic's ROOT service message rather
+    than to a specific message. Telegram sets ``reply_to_message.message_id`` to
+    the topic id (== ``message_thread_id``) for such replies; the root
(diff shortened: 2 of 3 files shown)


WingedGuardian:

Checked:
- No bare `except:` (both catches are `except Exception:` with proper handling)
- Lazy `from genesis.runtime import GenesisRuntime` inside function body is the correct deferred-import pattern to break a circular import — not flagged
- New private functions `_is_topic_root_reply` and `_try_topic_root_decision_nudge` both carry docstrings
- Test file `tests/test_channels/test_topic_root_nudge.py` present and covers the detection logic, the nudge path, and all relevant skip conditions
- Commit prefix `fix(telegram):` correct
- CHANGELOG addition and PR body have no security/PII language

---

WingedGuardian:
@codex review

WingedGuardian:
@codex review — a bare approval ("yes"/"approve") that quote-replies a forum TOPIC ROOT (reply_to_message.message_id == message_thread_id) matched no resolver and spawned a confused context-free CC session. This is a MINIMAL-SAFE fix: it only sends a nudge, resolves NOTHING, publishes NOTHING, and changes no approval guard. Please adversarially check: (1) `_is_topic_root_reply` cannot false-positive on a specific-message quote-reply or a non-reply; (2) the nudge fires ONLY in decision topics (approvals/content_review/ego_proposals) on a BARE decision, and a substantive topic-root message still falls through to conversation; (3) no path here can auto-resolve or double-publish an item; (4) `forum_topic_created` service messages (no `.text`) are handled without raising.

WingedGuardian:
@codex review — addressed your P2 in 36e6938f: the nudge now uses the whole-message `phrase_decision` matcher, so decision-prefixed substantive replies ("yes, but shorten the title" / "no, what alternatives?") fall through to conversation instead of being swallowed. Left `_bare_decision` untouched (the separate `_try_bare_approval_resolution` path intentionally uses leading-token semantics — and I filed a follow-up noting that sibling path has the same swallow but auto-resolves, which belongs with the deferred content-approval work). Added a parametrized regression test. Please re-review.

PR #1337, opened 2026-08-07
fix(cc): steer dispatch durable-learning to Genesis memory, not MEMORY.md

WingedGuardian:
## What
Adds one guidance sentence to the `interact`, `research`, and `campaign` dispatch-session profile addenda (`_PROFILE_ADDENDA` in `src/genesis/cc/direct_session.py`): persist durable learnings to Genesis memory (`procedure_store`/`observation_write`), never by hand-editing a `MEMORY.md` index. Plus a regression test pinning the guidance in all three.

## Why (root cause — corrected)
Ego-dispatch sessions run with cwd `~/.genesis/background-sessions`, so CC scopes their file-memory to a **separate** project dir from the user's foreground index. Three 2026-08-06 dispatches truncated *that shared-among-dispatches* index via full-file Write (dispatch profiles grant Write but not Edit), losing consolidated context until the next (irregular) consolidation.

A prior follow-up (`0361719f`) framed this as corrupting the user's *shared/live* memory index — that was a **mislabel** (the dirs are cwd-isolated, inode-distinct; no dispatch code references the shared path). No data-integrity risk to user memory, no data repair needed; the follow-up was corrected HIGH→medium.

## Why guidance-only (not a hook)
A PreToolUse block on Write→MEMORY.md would also block the **legitimate** consolidation job that regenerates the index. Steering the three investigation profiles (the only ones with Write + memory tools) is the clean fix; the tools named are reachable for exactly those profiles (they lack `_NO_MEMORY_WRITES`).

## Test
`test_write_profiles_steer_persistence_to_genesis_memory` — RED on main (base addenda contain neither `procedure_store` nor `MEMORY.md`), green here. Full profiles suite: 88 passed. Reviewed clean (feature-dev:code-reviewer, DONE).

### DIFF
--- a/src/genesis/cc/direct_session.py
+++ b/src/genesis/cc/direct_session.py
@@ -286,7 +286,7 @@ _PROFILE_ADDENDA: dict[str, str] = {
 
 You have: Write, browser MCP tools, memory MCP tools, outreach send.
 You do NOT have: Edit, Bash, NotebookEdit.
-Your final message IS your deliverable. Write files to `~/.genesis/output/`.
+Your final message IS your deliverable. Write files to `~/.genesis/output/`. Persist durable learnings to Genesis memory (`procedure_store` / `observation_write`) — NEVER by writing or editing a `MEMORY.md` index (a dedicated consolidation job owns that file; a hand-edit truncates it and loses the other sessions' context).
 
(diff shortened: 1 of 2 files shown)


WingedGuardian:

---

PR #1339, opened 2026-08-07
fix(routing): surface the legacy degradation fallback instead of silent

WingedGuardian:
## What
`CircuitBreakerRegistry.compute_degradation_level()` has a legacy provider-count fallback that runs only when **no** essential-site map was injected. It's inert in a correctly-configured install (the map is always built via `build_essential_provider_map` at `runtime/init/router.py`), but if it's ever reached in production it means a misconfig/wiring bug — and degradation silently switches to a count-based heuristic that can false-alarm "all paid providers down ⇒ ESSENTIAL".

## Fix
Emit a **once-per-instance** `logger.warning` when the legacy fallback is entered, so the condition is surfaced instead of silent (provision-or-surface). No behavior change beyond the log line; bare unit-test registries legitimately hit this path and warn once (harmless).

## Why it's not redundant
`essential.py` only errors when `build_essential_provider_map` returns an *empty* dict (config-content bug). This warning also covers a **distinct** failure mode: a caller constructing the registry without going through that builder at all (a wiring bug).

## Test
Two caplog tests: warns-once on the legacy path (no re-warn on the same instance); no-warn when an essential map is present. Existing `test_degradation_l0/l1/l2/l3/l5` unaffected (13 degradation tests green). Reviewed clean (feature-dev:code-reviewer, DONE).

Honest note: low-value/low-risk defensive hardening of an inert path — shipped for completeness of the ego-reliability sweep.

### DIFF
--- a/src/genesis/routing/circuit_breaker.py
+++ b/src/genesis/routing/circuit_breaker.py
@@ -358,7 +358,23 @@ class CircuitBreakerRegistry:
                 return DegradationLevel.ESSENTIAL
             return DegradationLevel.NORMAL
 
-        # Legacy provider-count fallback (no essential map injected).
+        # Legacy provider-count fallback (no essential map injected). In a
+        # correctly-configured install the essential map is always present
+        # (build_essential_provider_map), so reaching here in production means it
+        # came back empty (a misconfig that essential.py already logs at build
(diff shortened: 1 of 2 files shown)


WingedGuardian (commented):

---

PR #1342, opened 2026-08-07
feat(ego): surface completed dispatch findings in the dashboard

WingedGuardian:
## What
Makes completed ego-dispatch findings **discoverable** in the dashboard. After a dispatch finishes, the proposal detail showed only the 1000-char `user_response` summary; the full investigation output lived unseen in `cc_sessions.metadata`. This joins them back.

- **CRUD** `cc_sessions.dispatch_info_for_proposals(ids)`: one bounded query (`json_extract(metadata,'$.caller_context') IN (...)`) mapping each proposal → its dispatch session's `output_excerpt` (capped 4000) + `transcript_path` + `session_id`. Newest session wins per proposal.
- **Route** `ego_proposals_all`: attaches a `dispatch` field per proposal.
- **Template**: a collapsible `<details>` "Dispatch output" block (Alpine `x-text`, XSS-safe), shown only when `p.dispatch` exists.

Discoverability-first: **read-only surfacing, no LLM extraction, no auto-filed follow-ups** (the prior `record_follow_ups` harvest ran 84% noise and stays disabled).

## Robustness (review SHOULD-FIX addressed)
The join is best-effort so it can never regress the core proposals view: `contextlib.suppress` at the call site **and** a `json_valid(metadata)` guard in the query (SQLite short-circuits `AND`) so a legacy non-JSON metadata row can neither 500 the endpoint nor wipe out discoverability for good rows. `json_extract` raises on malformed JSON (verified) — the guard is load-bearing.

## Test
`tests/test_db/test_cc_sessions_dispatch.py` (5): maps output, newest-wins, empty-ids, excerpt-capped, malformed-metadata-skipped (RED without the `json_valid` guard). Verified **live against prod data** — real dispatch output surfaces for historical proposals (e.g. the voice identity-slice dispatch). Reviewed (feature-dev:code-reviewer): 1 SHOULD-FIX, fixed.

### DIFF
--- a/src/genesis/dashboard/routes/ego.py
+++ b/src/genesis/dashboard/routes/ego.py
@@ -338,6 +338,8 @@ async def ego_cadence():
 @_async_route
 async def ego_proposals_all():
     """Return all proposals with optional status filter and limit."""
+    import contextlib
+
     from genesis.db.crud import ego as ego_crud
     from genesis.runtime import GenesisRuntime
 
@@ -349,6 +351,18 @@ async def ego_proposals_all():
     limit = min(request.args.get("limit", 50, type=int), 200)
 
     proposals = await ego_crud.list_proposals(rt._db, status=status, limit=limit)
+    # Surface completed ego-dispatch findings: attach the dispatch session's
+    # output excerpt + transcript pointer so the debrief is discoverable in the
+    # dashboard, not just the 1000-char user_response summary.
+    from genesis.db.crud import cc_sessions as cc_sessions_crud
+
+    # Best-effort: a join failure (e.g. a legacy row with non-JSON metadata) must
+    # never 500 the core proposals list — degrade to no dispatch info.
+    dispatch_map: dict = {}
+    with contextlib.suppress(Exception):
+        dispatch_map = await cc_sessions_crud.dispatch_info_for_proposals(
+            rt._db, [p["id"] for p in proposals]
+        )
     return jsonify(
         [
             {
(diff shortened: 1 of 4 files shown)


WingedGuardian:
## First-pass review — PR #1342

### Structural review

**No ERRORs or WARNINGs found.**

---

**NOTE — `import contextlib` inside function body** (`src/genesis/dashboard/routes/ego.py`)

`contextlib` is stdlib with zero initialization overhead; there's no reason to import it lazily inside `ego_proposals_all`. It should move to the module-level import block alongside the other stdlib imports. The existing function-level imports (`genesis.db.crud`, `genesis.runtime`) are understandable for avoiding circular-import issues at module load time, but `contextlib` has no such constraint.

---

**NOTE — No route-level test for modified `ego_proposals_all`** (`src/genesis/dashboard/routes/ego.py`)

The new `dispatch_info_for_proposals` CRUD function is well-tested (5 targeted tests). The route-level integration — particularly the `contextlib.suppress(Exception)` degradation path and the `dispatch` field appearing in the JSON response — is not exercised by the test suite. Since the change is additive this won't break existing tests, but a route test verifying (a) `dispatch` is present on a matched proposal and (b) the endpoint doesn't 500 when the CRUD call raises would close the gap.

---

**CLEAN** — no secrets, private IPs, credentials, personal emails, or private hostnames in added lines. Test fixture paths use a synthetic `/x/` prefix. PR title and body contain no PII-leak language or security-incident references.

---

WingedGuardian:
@codex review

PR #1343, opened 2026-08-08
fix(reflection): read-only tool lockdown + reserve follow-up board for foreground

WingedGuardian:
## Why

Autonomous background **reflection** sessions (deep/strategic) ran with
`--dangerously-skip-permissions` and **no tool restriction** — full built-ins
(`Bash`/`Write`/`Edit`/`Task`) plus the entire `genesis-health` + `genesis-memory` MCP
surface, and (because `--mcp-config` is additive) any user/project-scoped MCP servers too.
A reflection used this to write an **ungrounded follow-up straight onto the accountability
board**. Reflection's contract is read-only investigation whose only output is observations
(parsed from its structured result), so unrestricted tool access was an unintended leak — a
read-only config existed but was never wired into the reflection dispatch path.

## What

- **Reflection is locked to read-only + `observation_write`.** A derived `disallowed_tools`
  denylist is wired into the reflection `CCInvocation`, with `strict_mcp_config=True` so only
  reflection's own MCP servers load (never user/project-scoped ones). The denylist is derived
  as `(live registry − read-allowlist − observation_write)` + the write/action built-ins, so a
  future write tool is auto-denied. `--allowedTools` is **not** a strict allowlist under
  `--dangerously-skip-permissions` (verified via the CC init-event tool list), so the scope is a
  denylist; `Task`/`Workflow`/`Skill` are denied so a subagent can't escape the lockdown.
- **The hot follow-up board is foreground-only.** Any autonomous/dispatched session's
  `follow_up_create` is routed to the cold `tabled` lane, and `follow_up_update` cannot promote
  an item onto the board (a foreground session can). Re-affirming an existing board item stays a
  no-op (never a demotion).
- **`REFLECTION_DEEP.md`:** read-only framing; the "missed insights" anchor becomes an
  observation rather than a follow-up; `observation_write` retained.
- **Guardrails:** a sentinel test (known write tools stay denied), a static AST scan (rejects any
  write-shaped tool in the read-allowlist — it caught a recall tool doing a vector-store write), a
  server-list drift test, and a fail-closed default-deny check.

## Verification

- **Live (CC init-event tool list):** with strict config, user-scoped MCP servers are absent;
  built-in and MCP write tools are stripped; `--disallowedTools` honors the `mcp__server__*`
  wildcard; the read set + `observation_write` remain.
- **1703 tests pass**; ruff clean.

## Reviews

Three review rounds (security → architecture → final security); all findings resolved. Codex was
unavailable (a model-manager parse error), so Claude reviewers were used per the documented fallback.

Ledger: 3358cf1b6fa347569776f8548127efa4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,17 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Changed
 
+- **Reflection sessions are now strictly read-only.** Genesis's autonomous
+  background reflections (deep/strategic) can read freely to investigate, but can
+  no longer call any write or action tool — their only output is observations (and
+  the structured reflection result the system parses). Previously reflections ran
+  with unrestricted tool access and could, in rare cases, mint a follow-up or other
+  write from ungrounded reasoning. Relatedly, any follow-up created by an
+  autonomous/dispatched session now lands in the recoverable *tabled* lane for
+  review rather than directly on the actionable follow-up board — the board stays
+  reserved for your (foreground) work. A foreground session can promote a tabled
+  item to the board.
+
 - **Free-tier model refresh.** Groq is retiring Llama 3.3 70B (the model behind
   several of Genesis's free reasoning/extraction/tagging steps) on 2026-08-16, so
   those steps now use Groq's recommended replacement, gpt-oss-120b. Structured-output
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -194,8 +194,26 @@ any task bigger than an LLM call.
 ```yaml subsystem-map
 entry: execution-cc
 modules: [cc]
-verified: 0a3fc19d 2026-07-23
+verified: 5dcd9fd4 2026-08-07
 ```
+
+- **Reflection tool lockdown — read-only + observations only**
+  (`session_config.build_reflection_disallowed`, wired at `reflection_bridge/_bridge.py`
+  into the reflection `CCInvocation.disallowed_tools`). Deep/strategic reflections run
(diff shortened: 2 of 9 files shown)


WingedGuardian:
## First-pass review — structural + PII/wording

### Structural

**WARNING — import inside conditional block** (`src/genesis/mcp/health/follow_up_tools.py`)

In `_impl_follow_up_update`, `import os` is placed inside the `if resolved_kind == "follow_up" and existing.get("kind") != "follow_up":` conditional rather than at the module top-level. `os` is a stdlib module and belongs in the import block at the top of the file. If it's already imported there (likely), this is a redundant inline import; if not, it's an import-grouping violation. Either way: move it to the module top.

**NOTE — late imports inside `except` block** (`src/genesis/cc/reflection_bridge/_bridge.py`)

`from genesis.cc.session_config import _REFLECTION_DENY_BUILTINS, _REFLECTION_MCP_SERVERS` and `from genesis.env import repo_root` live inside the `except Exception:` handler. The comment documents the intent (avoiding dependency on the config builder in the failure path), which is a legitimate reason. Still unconventional grouping — at minimum `from genesis.env import repo_root` could be a module-level import without adding a config-builder dependency. Worth tidying on a follow-up pass; not blocking.

No other structural violations found. No bare `except:`, no mutable default args, no `/tmp/` usage, no direct `task_states` inserts, no raw SQL, no `nohup`, no `git add .` / `--force`. New public functions carry docstrings. New test file added for the new source code. Conventional commit prefix present.

---

### PII / Secrets / Wording

**CLEAN** — no API keys, tokens, private IPs, SSH keys, database connection strings, or secret-valued env vars in any added line. PR title, body, and CHANGELOG additions describe the fix without advertising exploitable specifics or naming private infrastructure.

---

PR #1344, opened 2026-08-08
feat(recon): github steward — account notifications lane (mentions + outbound-contribution responses)

WingedGuardian:
## GitHub steward — account-level notifications lane

The account-activity monitor deep-polls the owner's OWN repos for external contributors, but it was blind to two things: **@mentions of the owner on other repos**, and **responses to the owner's outbound contributions** (issues/PRs they filed on third-party projects). This adds an account-level notifications lane that surfaces both.

### How
- One `gh api notifications` call per tick (repo-independent, shares the watermark), filtered to a configured `reason` allowlist — default `mention` / `team_mention` / `author`.
- `author`/`subscribed` are dropped on repos the owner **owns** (the deep-poll already records those); `mention`/`team_mention` pass on any repo. Window is `since < updated_at <= wm`.
- Notifications carry no inline actor, so the human is resolved via the notification's `latest_comment_url` (fallback: `subject.url`); the owner's own activity and bots are skipped.
- Reuses the existing `github_account_activity` observation type (a new `kind` inside the content JSON) — **no new table, no schema-guard tax**. Deduped by thread id + `updated_at`, so a *new* update on a thread re-pings but a repeat is a no-op.
- Immediate Telegram ping in `live` (both mentions and author-responses); `observe` records only; a failed ping degrades to the existing 6h digest (the observation is always recorded).
- Own sidecar cursor (sentinel key, can't collide with an `owner/name` repo); advances only on a clean sweep, holds on any gh/classification error so a missed window re-polls. First run baselines silently (no ping storm).

### Safety / scope
- Ships `observe` by default; the ping-capable `live` mode is an opt-in lever. Same `off`/`observe`/`live` gate as the rest of the steward.
- No new egress door — reuses the existing `submit_raw` notification path.

### Tests / verification
- 15 new unit tests: reason filter, owner-repo `author` filter vs mention-any-repo, the `updated_at` window, actor resolution + null-`latest_comment_url` fallback, hold-on-error (API + classification), owner/bot skip, dedup + new-update re-ping, observe-vs-live, baseline first run, api→html url (issue + PR), config damage-tolerance. 41 pass in the file; ruff clean.
- **Live-probed against a real account** (read-only): correctly surfaces a real outbound author-response on a third-party repo plus cross-repo mentions, each actor-resolved, with `advance_ok=True`.

### Follow-ups
- No per-tick cap on the notifications feed (the `reason` allowlist keeps the in-window set tiny); a cap can be added if volume ever warrants.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ## [Unreleased]
 
+### Added
+
+- **GitHub steward now surfaces responses to your upstream contributions.** Beyond
+  the flagship-repo deep-poll, the account-activity monitor gained an account-level
+  notifications lane: it pings you when someone @mentions you on any repo, or
+  responds on an issue/PR you filed on another project's repo (your outbound
+  contributions — the flagship deep-poll can't see those). Tunable via the
+  `notifications` reason-allowlist in `github_steward.yaml`; respects the same
+  `off`/`observe`/`live` lever and pings immediately in `live`.
+
 ### Changed
 
 - **Reflection sessions are now strictly read-only.** Genesis's autonomous
--- a/config/github_steward.yaml
+++ b/config/github_steward.yaml
@@ -35,3 +35,14 @@ automation_denylist: []
 
 # Loud-truncation cap: activity events processed per tick.
 max_events_per_tick: 100
+
+# Account-level notifications lane — surfaces activity BEYOND the flagship repos:
+# @mentions of the owner anywhere, and responses on issues/PRs the owner authored
+# on OTHER people's repos (their outbound contributions). The flagship deep-poll
+# already covers the owner's own repos, so `author`/`subscribed` reasons are
+# filtered to NON-owned repos. `reasons` is the GitHub notification-`reason`
+# allowlist; everything else (ci_activity, subscribed-to-own, ...) is dropped.
+# Both mentions and author-responses ping immediately in `live` mode.
+notifications:
+  enabled: true
+  reasons: ["mention", "team_mention", "author"]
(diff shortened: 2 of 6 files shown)


WingedGuardian:

---

WingedGuardian:
## Codex adversarial review — dispositions

Ran the owed Codex pass (after a first `feature-dev:code-reviewer` round). Codex raised **2 BLOCKER + 4 SHOULD-FIX**; each verified against the code and the real ones fixed with a design simplification: **per-item failures never hold the account-level cursor** (one bad item must not freeze the whole lane), and unresolvable/unsupported items are recorded **digest-only** so nothing is silently lost.

### Fixed
- **BLOCKER — truncation tie-safety.** The cap boundary now advances to the newest processed timestamp *strictly before* the first deferred item (mirrors the repo lane), so a same-second twin at the cap edge isn't stranded by the exclusive `since`.
- **BLOCKER — a persistently-unresolvable item froze the cursor forever** (e.g. a deleted comment 404ing on `latest_comment_url`). Per-item resolution/classification failures now record digest-only and the cursor **advances**; only a top-level feed-poll failure holds. `_resolve_notification_actor` is best-effort — it tries `subject.url` even when `latest_comment_url` errors, and never signals a hold.
- **Discussion data-loss** → recorded digest-only (GraphQL actor resolution is a follow-up) instead of dropped.
- **`latest_comment_url` mis-attribution** → the owner as the *latest* commenter on a mention is kept digest-only (the mention is real even if the owner replied last), not silently dropped; a genuine author-self-reply is skipped.
- **Malformed-JSON silent skip** → folded into the unresolved → digest-only path.

### Accepted (with rationale)
- **Failed live ping is not durably retried.** The observation is always recorded, so the 6h digest is a genuine fallback surface; the dominant "failure" is quiet-hours, where the digest *is* the correct behavior; a `WARNING` is logged. A `github_ping_pending`-style durable retry for notifications is a follow-up if wanted.

### Clean (Codex-confirmed)
- Reason/owner filtering; reuse of the `github_account_activity` obs type across all consumers.

Added 6 tests for the new paths (tie-safe boundary, unresolved→digest-only, classification-unresolved-surfaces, discussion-digest-only, owner-mention-digest-only, author-self-skip). **46 tests pass**; ruff clean. Live-probed against the real account pre- and post-remediation.

PR #1346, opened 2026-08-08
fix(awareness): make user_model_delta staleness monitor flap-resistant

WingedGuardian:
## Problem

The WS-2 M7 `user_model_delta` staleness monitor flapped. A fixed **14-day** threshold, a 3-day `infrastructure_alert` TTL, and an **in-memory monotonic cooldown** (reset on every server restart) combined so that during any quiet stretch the monitor re-minted the *same* `high`-priority alert every ~3 days — and each alert could trigger a **paid ego investigation**.

The 2026-08-06 diagnosis established the stream is **sparse by design**: reflection emits a `user_model_delta` only when its model of the user genuinely *changes* (the anti-confabulation prompt returns "no change" otherwise). A 288-delta burst in one intense March session, then ~3 deltas across four months of heavy interaction — all correct. So **silence alone is not a defect**, and the alarm must not treat it as one.

## Fix

Recalibrate to a false-positive-resistant "worth a look" nudge (all four parts):

- **45-day threshold** — well clear of observed honest gaps (~19d under heavy use).
- **Interaction gate** — alert only when ≥ 5 foreground sessions (interactive CC + voice) ran in the window; silence during a genuine user absence stays quiet.
- **Episode-scoped DB dedup** — one alert per silence episode, immune to the TTL re-mint and restart-proof (replaces the in-memory cooldown, now removed). Silence-since-last-delta anchors on the delta; the *never-any-delta* condition is one standing episode (any prior alert suppresses), so it doesn't slide and re-fire.
- **Medium priority** + rewritten content that states the stream is sparse by design and warns **against** lowering `MIN_DELTA_CONFIDENCE` or loosening the prompt to force emissions (which would confabulate deltas and corrupt the measured signal).

The alert still auto-resolves when a fresh delta lands, and a genuinely new episode after recovery re-alerts once.

## Verification

- 11 monitor tests pass; the flap-fix behaviors — episode dedup surviving a TTL resolution, both interaction-gate paths, and never-delta-fires-once — are each **verify-RED'd** (neutering the logic makes the test fail).
- Full `tests/test_awareness/` suite: 278 pass.
- **Replay over real history** (787 deltas, 1287 foreground sessions): the current 7-day gap produces **0 alerts** (the old 14d-gap flap is gone), and a synthetic 45d+ drought produces **exactly 1** alert that does **not** re-mint after a TTL resolution.

### DIFF
--- a/src/genesis/awareness/loop.py
+++ b/src/genesis/awareness/loop.py
@@ -206,9 +206,13 @@ _last_embed_backlog_band: str = ""
 #     few days (each re-mint re-triggered a paid ego investigation).
 #   - 'medium' priority: a slow user-model stream is never urgent.
 _USER_MODEL_STALE_DAYS = 45
-# Minimum foreground (interactive CC + voice) sessions since the last delta for
-# a silence to count as a possible defect rather than an expected quiet stretch.
-_USER_MODEL_MIN_INTERACTION_SESSIONS = 5
+# Minimum foreground (interactive CC + voice) sessions ACTIVE since the last
+# delta for a silence to count as a possible defect rather than an expected quiet
+# stretch. Counted by last_activity_at (reused long-lived sessions advance it), so
+# 1 is the meaningful floor: a single persistently-reused topic session is enough
+# to mean "the user was present." The 45d threshold + episode dedup keep this from
+# over-firing (at most one medium alert per drought episode).
+_USER_MODEL_MIN_INTERACTION_SESSIONS = 1
 
 
 def _embed_backlog_band(failed: int) -> str:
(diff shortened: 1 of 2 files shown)


WingedGuardian:
## Structural review — first-pass

> ⚠️ `REQUEST_CHANGES` could not be submitted (GitHub blocks self-review). Findings below are ERROR-level and should be treated as blocking.

---

### ERROR: Raw SQL against genesis.db (2 new blocks)

`src/genesis/awareness/loop.py` — `_check_user_model_staleness`

Two SQL queries are **new to this PR** (the pre-existing `SELECT max(created_at) FROM observations WHERE type='user_model_delta'` predates it and is out of scope for this review):

**1. Episode-dedup query**

```python
dedup_sql = (
    "SELECT 1 FROM observations WHERE source = 'user_model_staleness_monitor' "
    "AND type = 'infrastructure_alert' AND created_at > ? LIMIT 1"
)
# and the never-delta variant with no created_at anchor
```

`genesis.db.crud.observations` has `exists_recent_by_type` and `count_recent_unresolved_by_type_and_source`, but neither fits: both filter `resolved = 0`, and neither accepts an arbitrary ISO timestamp anchor. The dedup here intentionally covers **resolved** alerts too — that's the flap fix. Suggested wrapper: `has_observation_after(db, *, source, type, after_iso: str | None) -> bool` in `crud/observations.py`, covering both the `created_at > ?` branch and the no-anchor (never-delta) branch.

**2. Interaction-count query**

```python
"SELECT COUNT(*) FROM cc_sessions "
"WHERE session_type = 'foreground' AND started_at > ?"
```

`genesis.db.crud.cc_sessions` has `get_status_counts` (hour-window, all session types) but no foreground-since-ISO helper. Suggested wrapper: `count_foreground_sessions_since(db, *, since_iso: str) -> int`.

---

All added lines scanned — no secrets, tokens, private IPs, absolute paths with usernames, or private hostnames. PR title, body, and metadata contain no language that advertises a data-exposure fix or names private infrastructure.

---

WingedGuardian:
@codex review

PR #1341, opened 2026-08-07
feat(contribution): Contributor Work-Log — sanitized public-issue supply for contributors

WingedGuardian:
## Contributor Work-Log — public-issue supply for contributors

Turns curated backlog/codebase items into public GitHub issues for contributors, gated behind per-item owner approval. This PR is the committed **infrastructure** (propose + post path); the curator campaigns that feed it are local, uncommitted user data.

### Flow
1. A local curator drafts an issue and calls the `contributor_issue_propose` MCP tool (server-side, trusted boundary).
2. The tool sanitizes title + body + labels with the fail-closed `scan_prose` (blocks private IPs/paths/emails/secrets); if clean, it **holds** the draft in `pending_issue_posts` + a linked `approval_requests` row (approval row FIRST, crash-safe).
3. Owner approves **per item** on the dashboard — batch "approve all" excludes these, exactly like the WS-8 email gate.
4. The `contributor_issue_watcher` drain (every 5 min, `max_instances=1`) resolves approved holds under the `contributor_worklog` mode lever:
   - `propose_only` (default) → shadow-observe the egress once + mark `dry_run` (terminal; a later flip to `live` never retro-posts it).
   - `live` → `gh issue create`, deduped against open issues. `mark_posted` runs **before** `mark_consumed`, and a pre-post `gh issue list` dedup adopts an already-open same-title issue, so a crash can't double-post.
   - `off` → leave held. rejected/cancelled/expired/orphan → close out.

### Safety
- Fail-closed sanitizer at the **trusted server boundary** — a misbehaving curator subprocess can't bypass it; it covers every string that egresses (title, body, **and** labels).
- The new GitHub issue-create egress door is **shadow-gated** (`observe_github_issue_create`, observe-only, best-effort).
- Ships `propose_only` — **nothing posts** until `mode: live` is set in a `config/contributor_worklog.local.yaml` overlay. Env kill switch: `GENESIS_CONTRIBUTOR_WORKLOG_DISABLED=1`.
- Retention prune of terminal rows (>30d); **held rows are never pruned**.

### Tests / verification
- Unit + integration (~90 tests): sanitizer prose fail-closed, hold-store transitions (`dry_run` terminal + single-flip), propose tool (block / dedup / backpressure / mode / approve-all exclusion), drain state machine (live / propose_only / off / dedup-adopt / list+create failure / orphan / pending), retention (held never pruned), cell-risk-class validity.
[...]

PR #1349, opened 2026-08-09
fix(db): renumber MW-1 migration 0079 -> 0081 (duplicate prefix on main)

WingedGuardian:
## Problem
`#1341` landed `0079_pending_issue_posts.py` on main; then `#1325` merged `0079_mw1_extraction_judgment.py`, creating a **duplicate 0079 prefix**. The migration runner's pre-flight raises `RuntimeError` on duplicate prefixes (they collide on `schema_migrations.id`), leaving **main un-deployable** and `#1330`'s migration-check failing. GitHub's mergeable check is textual-only, so the two differently-named 0079 files merged without conflict and migration-check never ran against the merged tree.

## Fix
`0080` is owned by the incoming `#1330`, so this migration takes the next free prefix, **0081**. Rename only — migration logic is byte-identical. Updates the test's `import_module` string and the `_migrations.py` comment. No install ever recorded `0079_mw1` as applied, so no stale `schema_migrations` id results.


WingedGuardian:

---

PR #1355, opened 2026-08-09
feat(onboarding): readiness enrichment fields on setup-status (PR-B2a)

WingedGuardian:
Adds four **non-gating enrichment** fields to the dashboard `GET /api/genesis/setup-status`
payload, above the existing cumulative tier model. All additive and backward-compatible —
no existing field renamed or removed. Backend only; the persistent readiness panel that
renders the tier + these chips is a follow-up (PR-B2b).

## New fields
- `web_search_keyed_providers` (list) — which premium web-search providers
  (brave/tavily/exa/tinyfish) have a key configured. Web search is available regardless
  (the keyless SearXNG primary is the bootstrap default), so this reports *which paid
  providers augment* the baseline — never a floor/tier gate.
- `voice_configured` (bool) — S2S voice **deliberately** set up: an explicit
  `VOICE_S2S_PROVIDER` (openai/gemini) plus that provider's key. `s2s_provider()` defaults
  to "openai", so a bare `OPENAI_API_KEY` (an LLM key) must NOT read as voice — this
  requires the explicit opt-in.
- `ego_cadence_minutes` (int) — the ego think-cadence.
- `autonomy_level` (int) — the shipped autonomy default level (config posture, NOT the
  live earned per-install level, which is DB state).

## Design
- `readiness.EnrichmentStatus` + `compute_enrichment` mirror `compute_readiness`: pure over
  persisted secrets plus two **injected** config ints, so the module stays free of
  ego/autonomy imports on the `setup-status` hot path. Injected ints are coerced via a
  defensive `_as_int`, so a malformed cadence/level can never raise.
- `autonomy.config_read.read_autonomy_default_level`: fail-safe `yaml` read of the shipped
  default; returns L1 on **any** failure (missing/malformed/non-mapping root/absent key/
  non-int), `isinstance`-guarded so a non-dict YAML root can't `AttributeError`.
- The route reads the ego cadence **off** the `ego_enabled` critical path, so a bad cadence
  value can never disturb the tier.

## Invariants preserved
- `setup_status` never 500s into first-run (every new read is fail-safe).
- The floor is computed exactly once per request.
- Enrichment is non-gating: nothing here changes the floor or the tier.

## Tests
- `tests/test_autonomy/test_config_read.py` (new): every fail-safe path → L1.
- `tests/test_onboarding/test_readiness.py`: web-search keyed-provider table, voice
  truth-table (incl. the bare-OpenAI-key false-positive guard), `compute_enrichment`
  packaging + non-coercible-int fallback.
- `tests/test_dashboard/test_setup_wizard_api.py`: exact-dict fresh-install extended,
  populated enrichment case, cadence fail-safe, backward-compat.

Self-reviewed + independent reviewer pass (two findings fixed: cadence/tier coupling;
non-dict YAML root). Verified E2E through the real route handler (HTTP 200; output equals
independent recomputation for every field; tier stays consistent).

### DIFF
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -1133,7 +1133,14 @@ verified: b662f3e3 2026-07-17
   bool (no runtime import on the hot path); web-search/autonomy-posture/surplus are
   enrichment, never tier gates (autonomy has no on/off switch — it's always
   initialised and gated per-action by the mandatory approval gate). `setup-status`
-  emits `tier`/`tier_name`/`telegram_configured`/`ego_enabled` additively.
+  emits `tier`/`tier_name`/`telegram_configured`/`ego_enabled` additively, plus
+  non-gating **enrichment** from `readiness.compute_enrichment` +
+  `autonomy.config_read.read_autonomy_default_level` (fail-safe, presence-based):
+  `web_search_keyed_providers` (premium providers augmenting the keyless SearXNG
+  baseline), `voice_configured` (explicit `VOICE_S2S_PROVIDER` opt-in — a bare
+  OpenAI LLM key does NOT count), `ego_cadence_minutes`, and `autonomy_level`
+  (shipped config default, not the live earned level). The persistent panel that
+  renders the tier + these chips is PR-B2b (frontend, not yet built).
 - **db/**: aiosqlite WAL behind `SerializedConnection` (an asyncio.Lock —
   without it interleaved commits pin `in_transaction` until restart). Two
   schema paths coexist: base DDL (`schema/_tables.py`, ~113 CREATE TABLE; docs
--- a/src/genesis/autonomy/config_read.py
+++ b/src/genesis/autonomy/config_read.py
@@ -39,11 +39,11 @@ def read_autonomy_default_level(
     """The shipped *default* autonomy level for ``category`` (fail-safe → 1).
 
     Reads ``config/autonomy.yaml``'s ``defaults`` block with ``yaml.safe_load`` and
-    returns ``int(defaults[category])``. Returns ``1`` on ANY failure — missing file,
-    unreadable/non-UTF-8 file, malformed YAML, absent ``defaults``/``category`` key, or
-    a non-int value — mirroring ``AutonomyStateMachine._load_config``'s graceful
-    fallback, so a surface that reads this (the readiness enrichment) can never be
-    500'd or misled by a bad config.
+    returns ``int(defaults[category])`` floored at ``1``. Returns ``1`` on ANY failure —
+    missing file, unreadable/non-UTF-8 file, malformed YAML, absent ``defaults``/
+    ``category`` key, a non-int value, or a sub-``1`` (0/negative) value — mirroring
+    ``AutonomyStateMachine._load_config``'s graceful fallback, so a surface that reads
+    this (the readiness enrichment) can never be 500'd or misled by a bad config.
 
     ``category`` defaults to ``direct_session`` — the autonomy category governing
     autonomous CC sessions, the posture most relevant to the "Autonomous" readiness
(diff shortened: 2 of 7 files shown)


WingedGuardian:
**Structural review:** nearly CLEAN — one NOTE below.
**PII/Secrets/Wording scan:** CLEAN — no hard blocks, soft warnings, or wording issues.

---

### NOTE: Missing docstring on `EnrichmentStatus.as_dict()`

**File:** `src/genesis/onboarding/readiness.py`

`as_dict()` is a new public method on a new public class. It has an inline `#` comment explaining the key-namespace design decision, but no formal docstring. Convention: new public functions/classes get docstrings.

Suggested addition:
```python
def as_dict(self) -> dict[str, object]:
    """Serialise enrichment signals as a flat dict for the setup-status JSON payload.

    Key names are deliberately distinct from the tier/floor fields — the route merges
    all three dicts into one flat payload, so these must not collide with existing keys.
    """
```

No other structural, PII, or wording issues found. The fail-safe exception chains, test coverage across all new paths, import grouping, and no-nohup/no-task_states/no-/tmp violations are all correct.

---

WingedGuardian:
Fixed the P2 in 78221a9a: `_web_search_keyed_providers` now checks each provider's canonical `API_KEY_<TYPE>` (what the Brave/Tavily/Exa/TinyFish adapters actually read) instead of routing's 3-pattern resolver, so alias-only keys no longer advertise an unusable provider. Regression test added.

@codex review

WingedGuardian:
Closed the class in 7e89769a (robust-by-construction): `compute_enrichment` is now wrapped so any internal raise degrades to empty enrichment; `_as_int` and `read_autonomy_default_level` both catch `OverflowError` (the `int(float('inf'))` case). Added inf/nan/None coercion + route-level `.inf`-cadence→200 tests. A fresh adversarial audit (authoritative-source, enumerate-the-class) came back clean.

@codex review

WingedGuardian:
Addressed both in 952c2217:

- **Voice parity (readiness.py:247)** — fixed. Dropped the `.strip()` so `_voice_configured` matches the runtime exactly: `s2s_provider()` reads the raw env value and `s2s_enabled()` compares `== "openai"`/`"gemini"` with no normalization, so `" openai "` now reads as *not* configured here (as it would be off at runtime). Regression test added.

- **Autonomy level (config_read.py:61)** — floored at L1 (a 0/negative default clamps up). I did **not** cap above L4, though: your premise that levels max at L4 doesn't hold for this category — `autonomy.yaml` sets `ceilings.direct_session: 7` ("effectively uncapped"), so a legitimately-high default must pass through. The per-category ceiling is enforced by `AutonomyStateMachine`, not this generic display read.

@codex review

WingedGuardian:
You're right and I was wrong — thanks for holding the line. I pushed back citing `ceilings.direct_session: 7`, but I hadn't read the enum: `AutonomyLevel` defines only L1–L4 (`autonomy/types.py`, "L5–L7 deferred to V5"), and `load_or_create_defaults` does `AutonomyLevel(level)` which would `ValueError` on 5–7. The ceiling caps *actions*, not the level enum.

Fixed in bd230945 by validating against the constructible enum: `int(AutonomyLevel(int(...)))`, so any non-member (0, negative, 5–7) hits the existing L1 fail-safe and the field can't advertise a level the runtime couldn't construct. Auto-tracks the enum if V5 widens it.

@codex review

WingedGuardian:
Verified and fixed in 89b4caeb. Confirmed against source that fractional cadence is real end-to-end — `validate_ego_config` accepts `int|float`, `load_ego_config` doesn't int-coerce the scalar, and `EgoCadenceManager` hands it to `IntervalTrigger(minutes=)` (float-capable). Replaced the int coercion with `_as_finite_number`: preserves finite fractional values (`45.5` stays `45.5`; whole numbers normalise to int for clean JSON), still falling back to 60 on inf/nan/non-numeric. Route+jsonify test proves `45.5` survives.

@codex review

PR #1356, opened 2026-08-09
fix(cc): secure-by-default MCP scoping for autonomous CC sessions

WingedGuardian:
## Why

Claude Code's `--mcp-config` flag is **additive**: unless `--strict-mcp-config` is also
passed, a session loads the servers in the config file *plus* the user-scoped servers from
`~/.claude.json` (and, at repo cwd, project-scoped ones). Genesis dispatches many autonomous
background CC sessions with a Genesis-only `--mcp-config` but **without** `--strict-mcp-config`,
so those sessions additively inherited the operator's user-scoped MCP servers (code-editing and
graph tools) — whose write operations (`replace_symbol_body`, coordinated `rename`,
`delete_project`, …) **bypass the `Edit`/`Write` PreToolUse hooks** entirely, since those hooks
match only the built-in `Edit`/`Write`/`NotebookEdit` tools.

`CCInvocation.strict_mcp_config` defaulted to `False`, so every autonomous site had to remember
to opt in — and most didn't. This is the class fix behind the reflection tool lockdown (#1343),
which addressed one path; ~15 autonomous `CCInvocation` sites had the same leak.

Empirically verified on Claude Code 2.1.x via the live `init`-event tool list:
- genesis config, no strict → user-scoped servers **load** (the leak).
- genesis config, `--strict-mcp-config` → **only** the config's servers load (the fix), exit 0.
- `--bare` + `--strict-mcp-config` → CC exits non-zero (hence the invoker suppresses strict under bare).

## What

- **`CCInvocation.strict_mcp_config` now defaults to `True`** (secure-by-default). `--mcp-config`
  becomes authoritative for every dispatched session; a site that forgets to scope itself fails
  **closed** (zero MCP), never open.
- **Invoker** suppresses `--strict-mcp-config` under `--bare` (they're mutually incompatible).
- **Foreground/interactive opt-outs**: the owner-attended conversation paths (`conversation.py` ×3)
  and the checkpoint resume (`checkpoint.py`) set `strict_mcp_config=False` — a human is driving,
  so the full toolset is legitimate; the flip must not silently strip it.
- **Reflection**: extracted a shared `_reflection_lockdown_kwargs` helper and applied it to **all
  three** reflection entry points. The weekly self-assessment and quality-calibration arms
  previously ran fully unrestricted (a gap left by #1343).
- **Autonomy executor**: CODE/VERIFICATION steps are pinned to the Genesis-only reflection MCP
  profile **and** paired with a denylist that drops the write/admin Genesis MCP tools (since
  `build_mcp_config` loads servers wholesale) while keeping Genesis read tools and the built-ins
  (`Bash`/`Write`/`Edit`) that code steps need. Code edits still flow through the audited
  `Edit`/`Write` hooks.
- **Defense-in-depth**: the user-scoped servers are also denied by name (a shared
  `_USER_SCOPED_MCP_WILDCARDS` constant) in the DirectSession universal denylist and the reflection
  denylist — belt-and-suspenders behind the categorical strict guarantee.

## Verification

- **Live init-event probe** (CC 2.1.x): strict-scoped sessions load only `genesis-health` +
  `genesis-memory` (zero user-scoped) even at repo cwd; the foreground opt-out preserves the full
  toolset. Real code paths emit the correct flags per session type (reflection / DirectSession /
  foreground / code-step), including code steps keeping their built-ins.
- **Tests**: `2117 passed / 2 skipped` across `test_cc` + `test_autonomy` + `test_reflection`;
  `126` targeted (new `test_mcp_strict_default.py` guards: strict-default, bare-guard, reflection
  arms all use the helper (AST), step-CODE denylist recipe, foreground opt-out (AST); updated
  `test_invoker.py`). ruff clean. Verify-RED confirmed on the bare-guard.

## Review

One review round (security lens). Codex was unavailable (models-manager parse error), so a Claude
reviewer was used per the documented fallback. It caught one BLOCKER — the code-step MCP config was
initially granted without a denylist, exposing the full `genesis-health` admin surface (incl.
`settings_update`); fixed by scoping it to the reflection denylist's MCP portion. All findings
resolved; one pre-existing sentinel fail-silent path tracked as a separate follow-up.

Follow-up: a8f15c94

### DIFF
--- a/.claude/docs/background-sessions.md
+++ b/.claude/docs/background-sessions.md
@@ -71,6 +71,17 @@ it also reaches the `genesis-recon` discovery tools (GitHub/model-intel/skill
 scanning, findings storage) — the only profile that does.
 Use `observe` for read-only investigation.
 
+**MCP scoping is secure-by-default.** `CCInvocation.strict_mcp_config` defaults to
+True, so every background session gets `--strict-mcp-config`: it loads ONLY the
+servers in its generated `--mcp-config` (its `mcp_profile`) and never additively
+inherits the operator's user-scoped `~/.claude.json` MCP servers (Claude Code's
+`--mcp-config` is additive without strict — probe-verified). A profile that maps to
+no genesis servers therefore runs with zero MCP tools (fail-closed), not the
+operator's full set. Only human-driven foreground/interactive sessions
+(`cc/conversation.py`, `cc/checkpoint.py`) opt out (`strict_mcp_config=False`) to
+keep the full user-scoped toolset. As defense-in-depth, `_UNIVERSAL_DISALLOW` also
+denies the user-scoped servers by name (`_USER_SCOPED_MCP_WILDCARDS`).
+
 **`steward` is the one built-in Bash-enabled profile** — its Bash is restricted
 to the `gh` CLI only, enforced by `scripts/bash_safety_hook.sh` via the
 `GENESIS_BASH_ALLOWLIST` env var set from `CCInvocation.bash_allowlist`. It
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,16 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   reserved for your (foreground) work. A foreground session can promote a tabled
   item to the board.
 
+- **All autonomous background sessions are now MCP-scoped by default.** Genesis
+  runs many kinds of background Claude Code sessions (reflection, sentinel,
+  inbox/mail triage, autonomy executor, ego gates, research). These are now secure
+  by default: each session loads only the Genesis MCP tools it is explicitly given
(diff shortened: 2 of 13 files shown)


WingedGuardian:

---

WingedGuardian:
Addressed the Codex P2 (`mcp_profile="full"` under strict) in 953ec3ab: `_build_invocation` now sets `strict_mcp_config = (mcp_profile != "full")`, so an explicit trusted `full` profile opts out of strict (full means full) while every other profile stays strict — a `None` from an unknown/failed profile still fails **closed** to zero servers rather than silently leaking the full default. Guarded by a behavioral test (real `_build_invocation`, remaps a profile to `full`) plus an invariant that no built-in profile maps to `full`. Also hardened the round-1 tests (step-dispatcher denylist recipe + source-wiring; AST-based foreground opt-out check).

WingedGuardian:
@codex review

PR #1364, opened 2026-08-10
fix(install): stop /etc/environment write leaking 'Permission denied' on non-root installs

WingedGuardian:
## Problem

On a non-root install, `scripts/install.sh` prints a stray error on stderr near the start of setup:

```
scripts/install.sh: line 80: /etc/environment: Permission denied
```

## Root cause

The HOME guard tried to persist `HOME` with:

```sh
echo "HOME=$HOME" >> /etc/environment 2>/dev/null || true
```

`/etc/environment` is root-owned (0644), so a non-root user cannot append to it. The intent was to fail silently, but the shell sets up redirections left-to-right: it opens the `>> /etc/environment` target **before** applying `2>/dev/null`, so the "Permission denied" is emitted to the real stderr during redirection setup. Neither `2>/dev/null` nor `|| true` can suppress an error that happens before the command runs. Two consequences:

1. Every non-root install prints a spurious error line.
2. The write silently no-op'd, so `HOME` was never actually persisted to `/etc/environment` — the guard's stated purpose ("so all future sessions inherit it") was never achieved.

## Fix

Persist `HOME` best-effort using the sudo idiom already used throughout this script:

- write directly when `/etc/environment` is writable (root installs),
- otherwise via passwordless `sudo -n tee -a`,
- otherwise skip silently.

`HOME` is already exported for the current session just above, so persistence here is optional — but when privileges allow, it now actually happens instead of failing.

## Verification

Ran the full install on a fresh Ubuntu 24.04 cloud VM (arm64), before and after the change:

- Before: the `Permission denied` line appears once in install output.
- After: the line is gone (grep count 1 → 0); no error patterns remain in output.
- Install completes with `9 passed, 0 failed, 2 skipped`; `genesis-server` active; health endpoint returns HTTP 200 once the server finishes binding.

`bash -n scripts/install.sh` passes.


WingedGuardian:
Structural review: CLEAN — shell-script-only change; no Python pattern violations apply. Conventional commit prefix (`fix:`) present. No `nohup`, `/tmp/`, force-push, or other flagged patterns.

PII/Secrets/Wording scan: CLEAN — no secrets, private IPs, hardcoded paths, or hardware identifiers in added lines. PR title uses "leaking" in the sense of a stderr message escaping suppression, not a credential or data exposure; no wording concern.

---

WingedGuardian:
Structural review: CLEAN — shell-script only change; Python structural rules not applicable.

---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

PR #1367, opened 2026-08-10
fix(hooks): heredoc-aware guard parser — stop false commit/rm/push blocks

WingedGuardian:
## What

Make the shared git-guard command parser (`scripts/hooks/shell_parse.py`
`split_segments`) **heredoc-aware**: a *quoted* heredoc (`<<'EOF'`) body is inert
stdin data (no shell expansion, never executed), so it is no longer split into
command segments.

## Why (diagnose-first)

The recurring "worktree commit false-block" was root-caused by live
instrumentation + transcript archaeology to a **heredoc parsing bug**, not the
worktree-key divergence long assumed (a correctly-marked worktree commit already
passes on current code). Because the parser split heredoc *bodies* into segments:

- a commit-message line like `cd into the module` was parsed as a real `cd` →
  effective cwd went UNKNOWN → false **"Direct commits to main"** block;
- prose mentioning `rm`/`push` in a heredoc body **falsely tripped** the rm/push
  guards.

The parser backs all four guards (commit gate, push guard, protected-paths/rm
guard, invalidator), so the fix hardens every one.

## The fix

- `split_segments` skips **quoted-delimiter** heredoc bodies. Unquoted `<<EOF`
  (whose bodies expand) and `<<<` here-strings stay conservatively parsed;
  **fail-open** on an unterminated heredoc.
- **SECURITY:** a heredoc operator inside a `#` comment or inside a
  `` `…` ``/`$(…)` command substitution does **not** earn a body-skip — both
  would hide the real commands that follow. Gated by `in_comment` +
  backtick/`$()` context tracking.

## Adversarial review (genesis-architect)

An independent adversarial audit found **two real guard-bypasses**, both fixed
and regression-tested before this PR:
- **CRITICAL** — commented-out heredoc hid a following `rm`/`push`/`--no-verify`
  from every guard.
- **HIGH** — a backtick-substitution-scoped heredoc hid the following real
  commands.

**Safety invariant:** only quoted heredoc bodies (which never execute) are
skipped; everything that can execute is parsed. The change can only turn a
false-block into an allow or remove a false-positive — it cannot hide an
executable command.

## Verification

- **Outcome-verified (MEASURED):** both bypass cases now expose the dangerous
  command to the guards; a legit quoted-heredoc commit message with a `cd`-line
  body resolves to the worktree (was UNKNOWN → false Rule 1); a genuine
  commit-to-main still blocks.
- 292 guard tests pass (shell_parse, protected-paths, push guard, review gates);
  ruff clean. New `TestHeredoc` class (incl. both bypass regressions) + a
  hook-level integration cell.

**Hook change → effective at next CC session start** (this session keeps the old
parser until restart — which is why writing this very PR tripped the old bug).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,17 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **The git-guard command parser no longer misreads heredoc bodies.** The shared
+  parser behind the commit/push/rm safety guards split a *quoted* heredoc
+  (`<<'EOF'`) body into command segments — so a commit-message line like "cd into
+  the module" was parsed as a real `cd` and blocked the commit as if it targeted
+  `main`, and prose mentioning `rm`/`push` inside a heredoc body could falsely
+  trip a guard. The parser now skips quoted-heredoc bodies (inert stdin data that
+  never executes), fixing the false blocks without weakening any guard: a real
+  dangerous command *outside* a heredoc is still detected, and an unterminated
+  heredoc fails open. Unquoted heredocs (whose bodies do expand) keep the prior
+  conservative parse.
+
 - **The autonomous-CLI approval gate now ships ON by default.** Every background
   Claude Code session Genesis dispatches must be rooted in an explicit user
   approval — but the committed policy config shipped the gate *off*, so a fresh
--- a/scripts/hooks/shell_parse.py
+++ b/scripts/hooks/shell_parse.py
@@ -109,6 +109,69 @@ class Segment:
     depth: int = 0  # 0 = top level, >0 = inside a sh -c script
 
 
+def _quoted_heredoc_delim(command: str, i: int) -> tuple[str, bool, int] | None:
+    """If ``command[i:]`` starts a QUOTED heredoc operator, return
+    ``(delimiter, dash_strip, index_after_operator)``; otherwise ``None``.
+
+    Only QUOTED delimiters (``<<'W'`` / ``<<"W"``, optional ``<<-``) are matched:
+    a quoted heredoc body undergoes NO shell expansion (no ``$var`` / ``$(...)`` /
+    backtick), so it is inert data that never executes — excluding that body from
+    segment parsing is provably safe. UNQUOTED heredocs (whose bodies DO expand)
+    are intentionally NOT matched here, so they keep the conservative
+    (over-matching, fail-safe) parse. ``<<<`` (here-string) is not a heredoc.
+    """
+    n = len(command)
+    if command[i : i + 2] != "<<":
+        return None
+    j = i + 2
+    if j < n and command[j] == "<":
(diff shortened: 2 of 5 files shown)


WingedGuardian:
## First-pass review: structural + PII/wording

### Structural: 1 WARNING, 2 NOTEs

---

**WARNING — `git add -A` in test fixture code**

`tests/test_hooks/test_review_enforcement_commit_worktree_repro.py` — `repo` fixture and `_second_repo` helper both call `_git(r, "add", "-A")`. The pattern rule flags `git add -A`; specific files should be staged instead. Risk is near-zero here (isolated ephemeral `tmp_path` repos), but the convention applies to all code in the repo. Should be `_git(r, "add", "f.py")` / `_git(r, "add", "g.py")` respectively.

---

**NOTE — File's own docstring says "Do not merge as-is"**

`tests/test_hooks/test_review_enforcement_commit_worktree_repro.py` opens with:

> `NOT a fix, not a permanent regression suite — a throwaway evidence-gathering harness for fix/commit-gate-worktree-falseblock. Do not merge as-is; the findings belong in the design doc, this file is the raw proof.`

If this file is intentionally being kept (e.g. as a permanent regression suite), the docstring needs updating to reflect that intent. If it genuinely was meant to be temporary, it should be removed. As-is it contradicts its own presence in a merge PR.

---

**NOTE — Import ordering in the new test file**

`tests/test_hooks/test_review_enforcement_commit_worktree_repro.py`:

```python
sys.path.insert(0, str(_REPO_ROOT / "scripts"))
from review_state import _worktree_key  # noqa: E402
```

The `sys.path.insert` + `from review_state import` are mixed into module-level setup between constant declarations rather than kept in a distinct local-imports group after a `# noqa: E402` block separator. The `noqa` suppresses the ruff violation, so this is style-only — but a `# Local (path-hacked)` separator and moving the import to the end would be cleaner.

---

- No API keys, tokens, passwords, or credentials in added lines.
- No private IPs, ULA IPv6, SSH keys, or DB connection strings.
- `"t@e.st"` in test fixtures is a non-real test domain — not flagged.
- PR title, body, and CHANGELOG addition contain no PII-leak/security-incident language.

---

WingedGuardian:
Converting to draft: Codex found 4 P1 guard-bypasses (verified real — e.g. `bash <<'EOF'` executes its body, so skipping quoted-heredoc bodies can hide executable commands). The heredoc-skip approach trades a fail-closed false-positive for fail-open bypasses — wrong direction for a safety guard. Stepping back to research proper shell-parsing approaches before rearchitecting or withdrawing.

WingedGuardian:
Closing: the heredoc-skip approach was abandoned after adversarial review falsified its premise (a quoted heredoc body is NOT inert — the receiving interpreter executes stdin), and subsequent empirical reproduction showed heredocs were never the false-block trigger at all: the real cause was an unresolvable variable cwd hitting Rule 1's misleading message. Superseded by the teach-only fix on branch fix/guard-fail-closed.

PR #1374, opened 2026-08-12
fix(recon): career-outreach auto-run honors the external engine's accuracy gate

WingedGuardian:
## What & why

The opt-in career-outreach monitor (ships `off`) drives an external career-agent
module over an SSH `claude -p` bridge to stage first-touch outreach **drafts** into
the owner's mail Drafts (never sends — owner clicks Send). The **auto-run** (write)
path failed intermittently while the **read** path stayed clean.

**Root cause (diagnosed from live evidence, root-cause theory retracted twice on the way):**
the auto-run prompt said "reply with ONLY compact JSON, no prose", but the external
engine has its **own** UserPromptSubmit hooks that fire on outreach/draft wording and
inject a "run your verification gate + lead your reply with a verdict line + confirm
interactively" flow. That contradicts JSON-only, requires a headless-impossible
confirm, and the full gated flow (research → draft → verify → stage) blew the old
240s timeout and the hardcoded `--max-turns 25`. The read prompt has no trigger words,
so it never collided — matching the observed read-clean/write-flaky split.

## The fix (honor the gate, never bypass it)

- **`_AUTORUN_PROMPT` rework** — drive the engine's COMPLETE flow **including** its own
  verification gate under a headless contract: reword an unverifiable claim rather than
  fabricate a token or strip substance; report `verify_failed` (a soft outcome, **not**
  a job-health failure) when a draft can't honestly verify; lead with the engine's
  verdict line then compact JSON (`_parse_json` already tolerates it).
- **`_live`** — `verify_failed` handling + a same-company repeat guard (the self-selecting
  engine may lack durable "attempted" state); a no-progress tick emits a WARNING event so
  a persistently self-refusing engine is not invisible to job-health.
- **`ipc.py`** — a per-call `max_turns` override (default 25 **unchanged** for every other
  caller; bool-guarded).
- **Config** — `dispatch_timeout_s` 240→900 (cap 1800; there is **no** fixed SSH cap —
  `_send_cc` honors `timeout_s` with no ceiling, correcting a false comment), new
  `dispatch_max_turns` (80), new `autorun_note` overlay knob (auto-run prompt only, so
  install-specific gate vocabulary can't trip the engine's hooks on the clean read path).

## Verification

- **Design proven live E2E _before_ building** (the intended prompt): converged 328s /
  42 turns, ran the real verification gate, and staged a verified draft (confirmed in the
  mailbox). 42 turns > the old 25-cap is why `max_turns` is required.
- 80 targeted tests + 908-test recon/modules/surplus sweep GREEN; ruff clean.
- Adversarial `genesis-architect` review (round 1/3): 1 SHOULD-FIX + 1 NOTE fixed
  (no-progress visibility; blank-`verify_failed` presence-check), rest dispositioned.
- Ships `off`; overlay is gitignored; no install literals/PII in shipped files.

Ledger: ffceaa361d5a4f71a575f0f65cf863f5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,9 +15,12 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   once enabled, drives a configured external career-agent module to stage
   first-touch outreach drafts into your mail Drafts, then sends you a single
   Telegram nudge to review and send them — it never sends mail itself (you click
-  Send). Ships `off`; flip `off → observe → live` in `career_outreach.yaml`
-  (env kill switch: `GENESIS_CAREER_OUTREACH_DISABLED=1`). No-ops cleanly on
-  installs that don't have a career-agent module configured.
+  Send). Each auto-run drives the career-agent's OWN accuracy/verification gate
+  end-to-end, so a draft is staged only if it passes that gate; a draft that
+  can't be verified is skipped, not staged. Ships `off`; flip
+  `off → observe → live` in `career_outreach.yaml` (env kill switch:
+  `GENESIS_CAREER_OUTREACH_DISABLED=1`). No-ops cleanly on installs that don't
+  have a career-agent module configured.
 
 - **cc-tmp blast-radius isolation now converges on its own.** The dedicated,
   size-capped volume that stops a runaway temp write from filling the container
--- a/config/career_outreach.yaml
+++ b/config/career_outreach.yaml
@@ -21,4 +21,13 @@ enabled: true
 mode: off
 reasoning_module: ""   # set your career-agent module's name in career_outreach.local.yaml
 max_auto_runs_per_tick: 3
-dispatch_timeout_s: 240
+# Per-dispatch bridge timeout (seconds). No fixed SSH cap — the auto-run drives the
+# engine's COMPLETE gated flow (research -> draft -> verify -> stage), measured ~5.5
+# min live. Default 900 covers it; capped at 1800.
+dispatch_timeout_s: 900
+# Per-dispatch --max-turns for the agentic auto-run (the read dispatch keeps the ipc
+# default of 25). The gated flow needs >25 turns (measured 42). Capped at 200.
+dispatch_max_turns: 80
+# autorun_note: appended to the AUTO-RUN prompt only (never the read prompt). Ships
+# empty; put any install-specific gate guidance in career_outreach.local.yaml.
+autorun_note: ""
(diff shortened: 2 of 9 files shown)


WingedGuardian:
@codex review

WingedGuardian:
## Automated First-Pass Review — PR #1374

---

WingedGuardian:

---

WingedGuardian:
@codex review

WingedGuardian:
**Update — robust-by-construction redesign (escalation-cap switch-approach).**

Three review rounds each surfaced a new edge in the auto-run reply handling (the churn the cap flagged). Rather than patch instance #4, the reply handling was **redesigned**: `classify_autorun_reply` maps the reply into a **closed frozen-dataclass outcome union** (NoneLeft/Staged/VerifyFailed/ModuleError/ProtocolError); `_live` matches on outcome *type*. Every malformed shape collapses to `ProtocolError` in one exhaustively-table-tested place. A fresh adversarial review verified the classifier is a **complete, correct partition** (the class is closed) and the match is exhaustive with a fail-loud guard. It also caught + fixed a `_parse_json` trailing-content regression (verdict-line + fenced JSON now parses via an outermost-span fallback). Codex P2s from the prior pass are all addressed (or reasoned-pushback, see thread). Behavior-equivalent on valid input; the earlier live E2E (staged a verified draft) still holds.

WingedGuardian:
**Codex P2s (redesign pass) — both FIXED with rules, not patches (`1a156d3f`):**
- `career_outreach.py:186` (parser combines two objects on verdict+fence): replaced the consumes-to-end scan + outermost-span fallback with ONE rule — `_parse_json` returns the decoded JSON value with the **largest end-index** (the rightmost-reaching complete object = the trailing payload). Handles leading verdict-is-JSON, fenced payload, trailing prose, AND nested objects in one rule (no fallback). Locked incl. the compound verdict+fence case.
- `career_outreach.py:282` (company alongside none_left): a terminal signal (`none_left`/`error`) carrying a `company` is now a `ProtocolError`, not a silent `NoneLeft` success (exactly-one-outcome; `verify_failed` still requires its company). Locked.

WingedGuardian:
@codex review

WingedGuardian:
**Codex confirmation-pass findings — all fixed (`1eaf9052`):**
- `career_outreach.py:276` (company alongside none_left): now tests **key presence** (`"company" in payload`), not the normalized value — so non-string/blank/null company variants are rejected too. This is the class fix; the prior value-check was the incomplete instance version. Locked with 3 new table cases.
- `runners.py:161` (no-progress warning on failed ticks): gated on `not result.errors` — a mixed tick records the hard failure without the misleading "verifier refused every attempt" warning. Locked.
- P3 docs: corrected the four "no ceiling" spots to reflect the 3600s SSH-adapter ceiling (config caps at 1800).

The classifier partition was proven complete in the redesign review; these tighten it. 126 targeted green, CI green. Merging on green per plan.

PR #1375, opened 2026-08-12
feat(resilience): distinguish down vs event-loop-starved genesis-server

WingedGuardian:
## What

The in-container watchdog restarts genesis-server on 900s `status.json` staleness. But `status.json` is written **on the event loop**, so a *starved* (not dead) loop makes it go stale — and the watchdog would blind-restart a starved-but-alive server, re-triggering the boot-extraction backlog that caused the starvation (a mutual false-positive storm). This lets the watchdog tell the two apart.

## How

- **`util/loop_health.py`** (new) — off-loop publish/read seam for the loop-lag sampler's reading. A GIL-atomic frozen-dataclass reference swap (writer = loop thread, readers = Flask worker threads / the cross-process watchdog via HTTP); `age_s()` computed at read time from `time.monotonic()`.
- **`hosting/standalone.py`** — the loop-lag sampler now `publish()`es every iteration. If starvation is severe enough that `asyncio.sleep` stops returning, publishes cease and the sample's age grows — that growing age is the WEDGED signal.
- **`dashboard/routes/health.py`** — NEW **sync** `GET /api/genesis/liveness`. No `@_async_route`, so Flask's threaded server answers it **off the event loop** even during starvation (when `/api/genesis/heartbeat` and every other async route hangs). Returns `{alive, loop: {…sample_age_s} | null, awareness: {…} | null}`, fail-closed to null. Uses `peek()` (never lazily constructs a runtime).
- **`autonomy/watchdog.py`** — tri-state gate at the top of `_restart_if_allowed` (genesis-server, staleness/zombie reasons only): `down/unknown/wedged/responsive → restart` (unchanged path); `starved` (fresh, high-lag sample) → **bounded** suppress (`max_starved_skips=6`, ~30 min) + one deduped alert. A dead process can never produce a fresh 200 sample, so it always restarts. `starved_skips` resets on any non-starved verdict and on `_record_failure`.
- **`config/autonomy.yaml`** — `liveness_url` / `liveness_lag_suppress_ms=1000` / `liveness_sample_stale_s=120` / `liveness_max_starved_skips=6`.
- **`scripts/proactive_memory_hook.py`** — the degraded banner appends live loop-lag / executor-backlog on the reachable-but-loaded branches (never on connect-refused).

## Safety

The load-bearing property — **a genuinely dead server is never left unrestarted** — was adversarially audited (genesis-architect): every `_probe_liveness` failure mode (connection refused, timeout, HTTPError, JSON garbage, null/partial loop, non-numeric fields, stale sample) terminates in a restart-allowed path. Suppression is the only non-restart branch and requires a positive fresh sample. Ships **live** (observability; no lever) — it can only make restart behavior *less* aggressive, bounded at ~30 min, and every ambiguous verdict restarts.

Note: on this install the blind-restart has never actually fired (`restart_history: []`); this is preventive plus honest hook labeling and the first readable loop-lag metric. The **guardian-side** probe consult (its probes are also loop-bound) is a tracked **phase-2 follow-up** to keep this PR reviewable.

## Testing

- New: liveness-route (5, incl. a RED guard that a configured-but-stopped loop still 200s — proving the route is sync); sampler-publish (2); watchdog probe-classification (8) + restart-gate tri-state (9, full down/unknown/wedged/responsive/starved matrix incl. bound + bridge-inactive/relay never-probe); hook liveness-detail suffix (4). 111 tests green; `test_scripts` + hook subprocess green.
- Verify-RED confirmed on the suppression gate (disabled → starved cases wrongly RESTART) and the `starved_skips` reset.

Subsystem map reviewed: no entry's judgment-layer prose is made stale (the change is a watchdog-internal decision refinement, not a subsystem-capability change), so no map edit — noted here for transparency.

Addresses follow-up b4421dea (down-vs-starved distinguisher).

### DIFF
--- a/config/autonomy.yaml
+++ b/config/autonomy.yaml
@@ -65,6 +65,17 @@ watchdog:
   flap_window_seconds: 21600             # 6h rolling per-reason window
   flap_threshold: 3                      # same-reason restarts before damping kicks in
   flap_backoff_max_seconds: 7200         # cap escalating flap backoff at 2h
+  # Down-vs-starved distinguisher: before restarting genesis-server for a
+  # staleness/zombie reason, the watchdog queries the off-loop liveness probe
+  # (a SYNC Flask route that answers even when the event loop is starved). A
+  # fresh, high-lag reading = UP-but-starved, not dead — a restart there just
+  # re-triggers the boot backlog that caused the starvation. Suppress + alert,
+  # bounded by max_starved_skips so a genuinely-stuck loop still recovers. Port
+  # matches the shipped server default (5000); change both together if relocated.
+  liveness_url: "http://127.0.0.1:5000/api/genesis/liveness"
+  liveness_lag_suppress_ms: 1000         # suppress restart only above this loop lag (well over the 250ms warn floor)
+  liveness_sample_stale_s: 120           # sample older than this = loop fully wedged → restart, don't suppress
+  liveness_max_starved_skips: 6          # max consecutive suppressed restarts (~30 min at 300s cadence) before restarting anyway
 
 # CC rate limit detection patterns
 # These are parsed from CC CLI stderr/stdout when rate limits are hit.
--- a/scripts/proactive_memory_hook.py
+++ b/scripts/proactive_memory_hook.py
@@ -1006,6 +1006,12 @@ def _liveness_detail() -> str:
         loop = resp.json().get("loop")
         if not isinstance(loop, dict):
             return ""
+        # A wedged loop (or a dead sampler) keeps returning the LAST sample with a
+        # growing sample_age_s — don't present stale drift/executor depth as current
+        # load. Past a freshness window, report the staleness instead of the number.
+        age = loop.get("sample_age_s")
+        if isinstance(age, (int, float)) and age > 30:
+            return f" (loop sample stale {age:.0f}s — possibly wedged)"
         parts: list[str] = []
         lag = loop.get("lag_ms")
         if isinstance(lag, (int, float)):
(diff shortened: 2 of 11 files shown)


WingedGuardian:
## First-pass review — PR #1375

### Structural review

**WARNING — Missing coverage: `src/genesis/util/loop_health.py`**

New source file with no dedicated test file. Per convention, a new `.py` file in `src/` should have a corresponding file in `tests/`. The module's three public functions (`publish`, `read`, `age_s`) are covered by integration tests across two other test files:
- `tests/test_dashboard/test_liveness_route.py` — exercises `read()` and `age_s()` via the route
- `tests/test_hosting/test_loop_lag_sampler.py` — exercises `publish()` and `read()` via the sampler

The edge-case behavior of `age_s` (the `max(0.0, ...)` monotonic-reference clamp) has no direct test. Given the tight scope of the module (three thin functions + one frozen dataclass), a `tests/test_util/test_loop_health.py` with a handful of unit tests covering the clamp and the `now=` injectable override would close the gap.

No other structural findings: no pattern violations (no bare `except:`, no `/tmp/`, no raw SQL, no `git add .`), import grouping clean, all new public functions and classes have docstrings, conventional commit prefix present.

---

**CLEAN**

- No API keys, tokens, passwords, or secrets in added lines.
- `127.0.0.1:5000` in `config/autonomy.yaml` is loopback — not in any blocked private IP range.
- No private hostnames, personal email addresses, or absolute paths containing usernames in added lines.
- PR title, body, and CHANGELOG additions contain no security incident language, credential leak mentions, or exploitable vulnerability details.

---

WingedGuardian:
@codex review

Since your first pass: all 3 P2s fixed in 480176da (proxy-free opener for the loopback probe; starved iff lag >= liveness_lag_suppress_ms — dropped the lagging-flag OR; stale-sample rejection in the degraded-banner detail), plus c21cd9c0 (test-only: autouse fixture insulating the watchdog tests from the real alert queue) and 5abf6498 (dedicated loop_health unit tests per the first-pass structural warning).

PR #1377, opened 2026-08-12
perf(observability): offload infra-snapshot /proc & cgroup reads off the loop

WingedGuardian:
## What

The infrastructure health snapshot did synchronous `/proc` & `/sys/fs/cgroup` reads **inline on the event loop** — CPU (`/proc/stat` + PSI), disk, `cc_tmp`, and container cgroup memory. Usually sub-millisecond, but they stall exactly under memory/CPU pressure (kernel-side contention on PSI / meminfo / memory.stat), producing a chronic class of sub-second event-loop lag WARNs — distinct from the extraction spikes fixed in #1330.

## How

- **`_collect_host_metrics()`** (cpu + disk + cc_tmp) and **`_collect_container_memory()`** — the sync reads, each keeping its original per-section `try/except` and error shapes.
- **`infrastructure()`** offloads both via `asyncio.to_thread` — two hops placed at the original call-site positions so the snapshot's key ordering is unchanged. Mirrors the existing `enumerate_cc_slots` offload (#1323).
- **`_collect_cpu_usage()`** guards its module-global delta baseline (`_last_cpu_reading`) with a `threading.Lock` now that it runs on worker threads. Defensive: `infrastructure()` snapshots are single-flighted (`health_data.py` coalesces), so the lock never actually contends.
- **`awareness/signals.py`** `ContainerMemoryCollector.collect` offloads the same cgroup read (runs on every 5-min awareness tick).

## Safety

Snapshot output is **byte-identical** to before — the adversarial review (genesis-architect) diffed every moved field (disk/cc_tmp character-identical; container_memory branching, ternaries, guards, caps, and unavailable/error fallbacks all preserved) and confirmed the CPU-baseline lock produces identical `used_pct` on every path with no deadlock/race. Empty-state (fresh install, cgroup paths absent) degrades to the same `unavailable`/`error` shapes.

## Testing

- New offload tests: both collectors dispatched through `to_thread`; a **thread-identity RED guard** (reverting to inline runs on the main thread → fails); CPU delta-baseline persistence across calls. 23 offload/health tests + 27 awareness signal tests green; existing `test_health_psi_cpu.py` / `test_health_data.py` shape tests unchanged.

Follow-on: the SSL-context sharing half of follow-up 6bbd2aa7 (httpx per-client context construction on the loop) is evidence-gated and ships separately.

### DIFF
--- a/src/genesis/awareness/signals.py
+++ b/src/genesis/awareness/signals.py
@@ -290,7 +290,10 @@ class ContainerMemoryCollector:
     async def collect(self) -> SignalReading:
         from genesis.autonomy.watchdog import get_container_anon_memory
 
-        mem = get_container_anon_memory()
+        # Offload the synchronous cgroup reads off the event loop — this runs on
+        # every 5-min awareness tick and the /sys/fs/cgroup reads can stall under
+        # memory pressure (same chronic loop-lag class as the snapshot path).
+        mem = await asyncio.to_thread(get_container_anon_memory)
         if mem is None or mem[1] == 0:
             return _bootstrap_placeholder_reading(self.signal_name, "cgroup")
 
--- a/src/genesis/observability/snapshots/infrastructure.py
+++ b/src/genesis/observability/snapshots/infrastructure.py
@@ -6,6 +6,7 @@ import asyncio
 import logging
 import os
 import shutil
+import threading
 import time
 from datetime import UTC, datetime
 from typing import TYPE_CHECKING
@@ -161,6 +162,11 @@ def _cpu_status(used_pct: float | None) -> str:
 
 # Module-level state for delta-based CPU reading (no blocking sleep)
 _last_cpu_reading: tuple[int, int, float] | None = None  # (idle, total, monotonic_time)
+# Guards the read-modify-write of _last_cpu_reading. _collect_cpu_usage now runs on
+# a worker thread (offloaded off the event loop via _collect_host_metrics), so the
+# delta baseline's compare-and-swap must be atomic even though snapshot computes are
+# single-flighted in practice (health_data.py coalesces concurrent callers).
+_cpu_lock = threading.Lock()
 
 
 def _collect_cpu_usage() -> dict:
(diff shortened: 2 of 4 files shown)


WingedGuardian:
Structural review: one WARNING, no ERRORs.

**WARNING — Missing coverage: `ContainerMemoryCollector.collect` offload (`src/genesis/awareness/signals.py:293`)**

The `awareness/signals.py` change adds `await asyncio.to_thread(get_container_anon_memory)` to `ContainerMemoryCollector.collect`. The equivalent change in `infrastructure.py` ships with an explicit thread-identity RED guard (`test_host_metrics_run_off_the_main_thread` asserts the collector runs off the main thread, and it fails if you revert the offload). No equivalent dispatch test exists for `ContainerMemoryCollector.collect`. The PR body says 27 awareness signal tests pass, but those are pre-existing shape/logic tests — they won't catch a future regression that reverts the offload back to inline. Consider adding a similar thread-identity guard for this path.

---

PII/Secrets/Wording scan: CLEAN — no hard blocks, soft warnings, or wording issues.

---

WingedGuardian:
@codex review

Since the first-pass structural review: added the missing thread-identity RED guard for the ContainerMemoryCollector.collect offload (test_container_memory_reads_run_off_the_main_thread in tests/test_awareness/test_signals.py, commit 30ec0a3f), mirroring test_host_metrics_run_off_the_main_thread. Verify-RED confirmed (reverting the offload fails the main-thread assertion).

WingedGuardian:
@codex review

Merged origin/main to resolve a conflict in infrastructure.py: PR #1353 (PID-budget monitoring) added an inline `infra["pids"] = _collect_pid_budget()` to the same function this PR offloads. Resolution folds the pid-budget read into the offloaded `_collect_host_metrics()` helper (preserving main's cpu→disk→pids→cc_tmp order) so it rides the thread-offload instead of blocking the loop; second region resolves to the offloaded `_collect_container_memory()`. A fresh-context adversarial audit confirmed the merged output is the union of both parents (14 keys, nothing dropped) and that _collect_pid_budget is thread-safe to offload. Offload test now asserts `pids` is folded in.

PR #1379, opened 2026-08-12
feat(commands): add /list-skills on-demand skill roster

WingedGuardian:
## What

A new Claude Code slash command **`/list-skills`** that lists every available skill (Tier-1 always-indexed + Tier-2 specialized) with a one-line description each, read from the prebuilt `~/.genesis/skill_catalog.json`. Optional keyword filter: `/list-skills browser`.

## Why

Genesis has 65 skills across three locations, but the only discovery path today is the automatic skill-injection nudge, which surfaces just **one** matching skill per prompt. There's no way to browse the full set on demand. This is the cheap on-demand slice of the larger session-start capability-surface work — independently useful, not a substitute for it.

## Design

- **Reuses the existing catalog pipeline** — `scripts/generate_skill_catalog.py` already scans all three tiers into `~/.genesis/skill_catalog.json` (self-heals hourly via the injection hook). The command just reads it; no new enumeration code.
- **Fresh-install correctness** — regenerates the catalog only when missing (via `git rev-parse --show-toplevel`, no hardcoded paths), so it never races the hourly refresh; guards a missing/empty/malformed catalog instead of rendering an empty roster.
- **Injection-safe** — the keyword filter is applied model-side and is never interpolated into a shell command.
- **Convention-matched** — frontmatter + `## Arguments` + `## Steps`, mirroring `.claude/commands/shelve.md`.
- Named `/list-skills` (not `/skills`) to avoid shadowing any bundled command.

## Verification

- E2E-tested the mechanism: frontmatter parses; fresh-install fallback resolves the generator via git-root; catalog reads (18 Tier-1 + 47 Tier-2); roster formats with first-sentence truncation; keyword filter returns correct matches (e.g. 'voice' → 4 skills).
- Adversarial genesis-architect audit: **no BLOCKER/SHOULD-FIX**; verified injection-safe, schema-accurate, atomic-write, generalizable, convention-conformant. NOTE-level empty-state hardening applied; remaining NOTEs consciously accepted (documented in the review).
- Final interactive check (`/list-skills` live) is user-facing and best done in a CC session.

## Deploy path

Hooks/commands land at next CC session start (no restart needed). Empty-state and any-install correct.


WingedGuardian:
@codex review

WingedGuardian:
**Automated first-pass review — PR #1379**

Markdown-only change (`.claude/commands/list-skills.md`): structural review skipped per protocol.

- No secrets, API keys, tokens, or credentials in added lines.
- No private IP addresses (10.x, 192.168.x, Tailscale CGNAT) or IPv6 ULA.
- No personal email addresses, hardcoded usernames, hardware identifiers, or private hostnames.
- `~/.genesis/` paths use `~` — no username leakage.
- PR title, description, and any changelog additions contain no wording referencing security incidents, credential leaks, or private infrastructure.

---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
## Round-3 finding dispositions

All three verified against source — each is pre-existing behavior of the catalog *subsystem* the command reads from, not a defect introduced by this command (which has converged after the redesign):

1. **Match filters against complete descriptions** — real bug, **fixed in #1380**: the catalog generator parsed frontmatter with a truncating regex; #1380 switches to `yaml.safe_load`, so `genesis-voice`/`gitnexus-*`/`code-intelligence` descriptions parse in full. Once #1380 lands and the catalog regenerates, the filter matches complete text.

2. **Scope the catalog to the invoking worktree** — pre-existing property of the single global `~/.genesis/skill_catalog.json` (the skill-injection hook has the same behavior). Tracked as a follow-up (`694c58be`, architectural, low priority); out of scope for a read-only list command.

3. **Regenerate existing empty catalogs before retrying** — same class as the hook's mtime-only freshness check (it never regenerates a fresh-but-empty/corrupt catalog). Tracked in the hook self-heal follow-up (`17a8e95b`, broadened to cover empty tiers). The command already tells the user to delete the file for the corrupt case; the durable fix belongs in the hook, not the command.

Per the 3-round escalation cap, not patching the command further around subsystem issues that don't live in it. Ready to merge once #1380 lands (pending owner approval).

WingedGuardian:
Dispositioning the P2s (prose command-definition file; the command only READS ~/.genesis/skill_catalog.json and delegates regeneration to the skill-injection hook):
- Resolve main venv / Regenerate empty catalogs before retry — concern the generator/hook, not this command; the command intentionally never regenerates (reads only; instructs retry or delete-and-retry). Out of scope for the command file.
- Scope the catalog to the invoking worktree — the catalog is install-global (home-anchored) by design; worktree-scoping would be incorrect.
- Match filters against complete descriptions — already the case: step 3 filters on full name/description/keywords; only the display (step 2) truncates to the first sentence. Working as intended.
Accepted as-is; branch was refreshed to current main (pure additive), command file unchanged.

WingedGuardian:
@codex review

PR #1386, opened 2026-08-13
fix(security): don't log a non-numeric TELEGRAM_ALLOWED_USERS entry verbatim (CodeQL #160)

WingedGuardian:
Fixes CodeQL alert #160 — `py/clear-text-logging-sensitive-data` (HIGH).

## Problem

`build_bridge_config` logged a non-numeric `TELEGRAM_ALLOWED_USERS` entry
verbatim (`log.warning("Invalid UID …: %r", uid)`). That value is taint-tracked
from a secret, and the realistic failure is a **bot token mis-pasted into the
allowed-users field** — a bot token has the shape `<id>:<secret>`, exactly what
the existing test fixture (`123456:ABC-token`) mimics — which would then be
echoed in clear text to the logs (journald / dashboard).

## Fix

Log the offending entry's **1-based position**, not its content:

```
TELEGRAM_ALLOWED_USERS entry #<n> is not a numeric user ID — skipping it
```

The tainted value no longer reaches any log sink (breaking the CodeQL taint
path), while the diagnostic stays useful — the operator can locate the bad entry
by position. No other behaviour change: valid numeric UIDs still parse, empty
entries are still skipped, and the return-None contract is unchanged.

## Test

`test_invalid_uid_is_not_logged_verbatim` passes a bot-token-shaped value and
asserts it is absent from the captured logs while a position-based warning is
still emitted. Teeth-verified against the old behaviour.

### DIFF
--- a/src/genesis/channels/bridge_config.py
+++ b/src/genesis/channels/bridge_config.py
@@ -59,12 +59,19 @@ def build_bridge_config(
     allowed_users: set[int] = set()
     allowed_raw = secrets.get("TELEGRAM_ALLOWED_USERS", "")
     if allowed_raw:
-        for uid in allowed_raw.split(","):
+        for idx, uid in enumerate(allowed_raw.split(","), start=1):
             uid = uid.strip()
             if uid.isdigit():
                 allowed_users.add(int(uid))
             elif uid and log:
-                log.warning("Invalid UID in TELEGRAM_ALLOWED_USERS: %r", uid)
+                # Never echo the raw entry: a value mis-pasted into this field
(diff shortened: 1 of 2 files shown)


WingedGuardian:
PII/Secrets/Wording scan: CLEAN — test fixture `"123456:AA_secret_bot_token_value"` is an intentionally fake bot-token-shaped value (not a real credential); PR title and body describe a log-sanitization behavior fix without advertising an incident or naming private infrastructure.

---

WingedGuardian:
@codex review

PR #1391, opened 2026-08-13
fix(cc): deny subagent-spawn (Agent) across restricted sessions

WingedGuardian:
## Problem

Restricted, locked-down Claude Code sessions (deep/strategic reflection, surplus
brainstorm, the direct-session profiles, the inbox/mail judges, the experimentation
single-turn completion) run with a denylist to keep them read-only and non-acting.
Part of that denylist's job is to block the **subagent-spawn tool** — a session that
can spawn a subagent escapes its own lockdown, because the subagent runs with a
fresh, unrestricted toolset (Bash/Write/Edit).

The live spawn tool is named **`Agent`**; **`Task`** is the obsolete name. The
denylists had **drifted**: `sentinel`-degraded blocked both names, reflection blocked
only `Task` (so the live `Agent` slipped through), and the direct-session profiles +
surplus blocked **neither**. Net effect: several restricted session types could, in
principle, spawn a subagent and act outside their sandbox.

## Fix

- Introduce one source of truth: `cc/types.SPAWN_TOOL_NAMES = ("Agent", "Task")`
  (`Agent` current, `Task` legacy). A stdlib-only leaf module already imported by
  every consuming site — no new import edge, no cycle.
- Reference it from every restricted-session denylist: reflection
  (`_REFLECTION_DENY_BUILTINS`), surplus (`_READONLY_DISALLOWED`), every
  direct-session profile (`_UNIVERSAL_DISALLOW`), the inbox/mail judges, and the
  experimentation completion string.
- Two extra hardening points in `direct_session`:
  - the per-request `tool_exceptions` path now discards the spawn tools, so a caller
    cannot re-enable spawn (mirrors the existing recursive-spawn discard);
  - `add_profile` injects the spawn tools unconditionally, so an install-local
    overlay profile cannot opt out.
- `sentinel/dispatcher._DEGRADED_DISALLOWED_TOOLS` keeps its own literal (a
  deliberate import-cycle avoidance) and already blocked both names.

## Guardrail

`tests/test_cc/test_spawn_lockdown.py` asserts every restricted-session denylist is a
superset of `SPAWN_TOOL_NAMES` (iterating profiles exhaustively), plus behavioral
tests for the `tool_exceptions` and overlay paths — so a new restricted-session
denylist that forgets spawn-blocking fails CI instead of silently reopening the gap.

## Scope / non-goals

`autonomy/executor/research.py` and the code-step executor
(`step_dispatcher`) legitimately hold broader tools and are **deliberately left out**
of this change (tracked separately). This is a strictly additive hardening: every
denylist gains entries, none loses any.

## Verification

- verify-RED first, then green: reflection/surplus/direct-session tests fail without
  the wiring, pass with it.
- E2E data-flow: the actual invocation configs that reach `claude -p` (reflection
  deep/strategic, surplus, inbox/mail judges, the experimentation `--disallowedTools`
  string) all deny `Agent` (+`Task`).
- Adversarial security review: no findings; strictly additive, no gate weakened.
- ~400 tests across affected suites pass; `ruff` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,17 +42,16 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
-- **Read-only reasoning sessions can no longer escape their sandbox by spawning.**
-  Genesis's strictly read-only Claude Code sessions (deep/strategic reflection, the
-  inbox/mail judges, and the experimentation completion) run with a locked-down
-  toolset. That lockdown blocked the obsolete subagent-spawn tool name
-  but not the current one — nor the `Workflow`/`Skill` spawn paths — so such a
-  session could in principle spawn a child that ran with full, unrestricted tools.
-  These denylists now deny the whole spawn class (subagent, workflow, and skill
-  spawns) from a single shared definition, with a guardrail test so a new read-only
-  session can't silently reopen the gap. (Working background sessions that
-  legitimately orchestrate — e.g. the deep-research `Workflow` path — are
-  intentionally out of scope and tracked separately.)
+- **Restricted reasoning sessions can no longer escape their tool restrictions by
+  spawning.** Several of Genesis's restricted Claude Code sessions (deep/strategic
+  reflection, the inbox/mail judges, and the experimentation completion) could spawn a
+  subagent that ran with full, unrestricted tools — escaping the restrictions placed on
+  the parent. Their denylists blocked the obsolete subagent-spawn tool name but not the
+  current one, nor the `Workflow`/`Skill` spawn paths. These denylists now deny the whole
+  spawn class (subagent, workflow, and skill spawns) from a single shared definition,
+  with a guardrail test so a new session can't silently reopen the gap. (Working
+  background sessions that legitimately orchestrate — e.g. the deep-research `Workflow`
+  path — are intentionally out of scope and tracked separately.)
 
 - **The autonomous-CLI approval gate now ships ON by default.** Every background
   Claude Code session Genesis dispatches must be rooted in an explicit user
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,15 +42,17 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
-- **Autonomous background sessions can no longer escape their tool sandbox by
-  spawning a subagent.** Genesis's restricted sessions (reflection, the
-  direct-session profiles, the inbox/mail judges, and others) run with a
-  locked-down toolset — no shell, no file writes. That lockdown blocked the
-  obsolete subagent-spawn tool name but not the current one, so a restricted
-  session could in principle spawn a subagent that ran with full, unrestricted
-  tools. Every restricted-session denylist now denies the subagent-spawn tool
-  from a single shared definition, with a guardrail test so a new one can't
-  silently reopen the gap.
+- **Read-only reasoning sessions can no longer escape their sandbox by spawning.**
+  Genesis's strictly read-only sessions (deep/strategic reflection, surplus
+  brainstorm, the inbox/mail judges, and the experimentation completion) run with a
+  locked-down toolset. That lockdown blocked the obsolete subagent-spawn tool name
+  but not the current one — nor the `Workflow`/`Skill` spawn paths — so such a
+  session could in principle spawn a child that ran with full, unrestricted tools.
+  These denylists now deny the whole spawn class (subagent, workflow, and skill
+  spawns) from a single shared definition, with a guardrail test so a new read-only
+  session can't silently reopen the gap. (Working background sessions that
+  legitimately orchestrate — e.g. the deep-research `Workflow` path — are
+  intentionally out of scope and tracked separately.)
 
 - **The autonomous-CLI approval gate now ships ON by default.** Every background
   Claude Code session Genesis dispatches must be rooted in an explicit user
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,16 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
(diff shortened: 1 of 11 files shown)


WingedGuardian:
@codex review

WingedGuardian:

---

WingedGuardian:
## Disposition of the two P1 findings

**P1 (types.py) — include every spawning builtin: FIXED.** Verified the gap — only the
reflection denylist blocked `Workflow`/`Skill`; surplus, direct-session, inbox/mail, and
the experimentation completion did not, and `Workflow` is confirmed reachable inside
background sessions. `SPAWN_TOOL_NAMES` is now the full spawn/escape class
`("Agent", "Task", "Workflow", "Skill")`, and reflection references it directly (dropping
the redundant literals). All six restricted-session denylists now deny the whole class;
the guardrail test pins the tuple. (commit 54e7561d)

**P1 (test — extend to every restricted denylist): dispositioned per-site.**
- `autonomy/executor/research.py` and `step_dispatcher.py` — **deliberately out of scope**
  per an explicit maintainer decision: these hold broader tools by design and may
  legitimately spawn a decomposition subagent. Your Bash-bypass point is valid and is
  captured in a tracked follow-up for per-site evaluation, rather than a blind block.
- `ego/session.py` — its `ego_*` authority tools are gated **server-side** by
  `is_dispatched_session_env()`, and a spawned child inherits that env, so the authority
  denial holds even via a subagent — defense-in-depth, not a live hole.
- `eval/bench/arms.py` — runs full builtins **by design** (benchmark fairness contract),
  so spawn-denial there would change the measured premise. Out of scope.

The test docstring now states the covered scope precisely (read-only/restricted lockdown
sessions) rather than claiming every session.

WingedGuardian:
@codex review

WingedGuardian:
## Round-2 disposition (commit 998b7793)

**P2 — "Preserve the documented background Workflow path": FIXED by narrowing scope.**
Confirmed via `.claude/docs/background-sessions.md:33-50` — channel-dispatched deep
research routes to `direct_session_run(profile="research")` precisely so a long
`Workflow` (100+-agent fan-out) finishes under the extended background wait ceiling (a
real inline-kill incident on 2026-07-20). Blocking the spawn class in `_UNIVERSAL_DISALLOW`
broke that path, so **direct_session.py (and its profile test) is reverted to
`origin/main`** — it blocked neither `Agent` nor `Workflow` before, so no regression. The
full spawn class now applies **only to the strictly read-only sessions** where spawning is
always an escape (reflection, surplus, inbox/mail, experimentation). direct-session's spawn
story — which needs a sandbox-preserving `Workflow` (children inherit the parent denylist)
or an accepted-porousness decision, exactly as you suggested — is tracked as committed
follow-up work alongside the autonomy-executor sessions.

**P3 — types.py doc path: FIXED** (`docs/.claude/…` → `.claude/docs/background-sessions.md`).

**P3 — CHANGELOG accuracy: FIXED.** Dropped the inaccurate "no shell, no file writes"
claim (interact/research/campaign retain `Write`; steward retains allowlisted `Bash`) and
reframed it as a read-only spawn-lockdown note; removed direct-session from the entry.

WingedGuardian:
@codex review

WingedGuardian:
## Round-3 disposition (commit d8411a81)

**P3 — "Exercise the live surplus path instead of an orphaned list": RESOLVED by dropping surplus from scope.** You were right, and it's more than a test nit — I verified against the code that **surplus is not a spawn-capable CC session at all**: the live executor `SurplusLLMExecutor.execute` calls `self._router.route_call(...)` (a tool-less Router LLM call, `surplus/executor.py:518`), not a `claude -p` session, and `build_surplus_config()`'s `disallowed_tools` is consumed only for its `system_prompt` (`direct_session.py:1288`). So `_READONLY_DISALLOWED` has no live spawn surface, and claiming surplus is "spawn-locked" was inaccurate.

Fix: reverted `_READONLY_DISALLOWED` to `origin/main` (with a NOTE explaining why it's out of scope), and removed surplus from the guardrail test, CHANGELOG, `CURRENT.md`, and the `SPAWN_TOOL_NAMES` docstring.

The scope is now **genuine spawn-capable CC sessions only** — reflection, the inbox/mail judges, the experimentation completion, and sentinel-degraded — and each assertion targets the **live `--disallowedTools` value** (reflection via `build_reflection_disallowed`; the inbox/mail/experimentation constants are exactly what their `CCInvocation` passes), not an orphaned constant.

(The prior round-3 review errored out mid-run — re-requesting a full pass.)

WingedGuardian:
@codex review

WingedGuardian:
## Round-4 disposition (commit 6c5069d0)

Thanks — all three verified. This PR stays scoped to the **spawn class**; the two larger items are tracked as deliberate separate work.

**P1 (inbox/mail judges leave Bash + MCP writes open on external input): CONFIRMED, tracked as a separate HIGH follow-up.** Verified: both judges run `skip_permissions=True` with `Bash` NOT in their denylist, and the inbox judge loads `build_mcp_config("reflection")` without the derived write-denylist (so `memory_store`/`settings_update` are exposed). This is a real pre-existing injection→shell/mutation vector, and it means "read-only boundary" overclaimed for the judges. Per an explicit maintainer decision this is handled as its own focused fix (deny `Bash` in both judges + apply the reflection MCP write-denylist to inbox), not bolted onto this spawn-lockdown PR. I've **softened every "read-only" claim here to "spawn-hardened / restricted"** and added NOTEs pointing at the follow-up, so nothing overclaims in the interim.

**P2 (guardrail should auto-discover new read-only denylists): valid, tracked follow-up.** The suite hard-codes the known constants, so it wouldn't catch a *future* read-only session that forgets spawn-blocking. Filed for an AST-enumeration or shared-constructor approach.

**P3 (doc-sync): FIXED.** Corrected the drift-history (reflection previously blocked all *but* `Agent` — Task/Workflow/Skill) and the stale inbox/mail comments that still described `SPAWN_TOOL_NAMES` as only `Agent`+`Task` (now the full class).

Merging the spawn-class fix now (the mechanism is sound and CI green); the judge Bash/MCP hardening ships as its own change.

PR #1393, opened 2026-08-13
fix(skills): full SKILL.md descriptions via a bounded YAML parser

WingedGuardian:
# PR: fix(skills): full SKILL.md descriptions via a bounded YAML parser

## Problem
`generate_skill_catalog.py::_parse_frontmatter` extracted descriptions with a
regex that stopped at the first apostrophe/quote/newline, dropping searchable
text (genesis-voice cut at "user's" ≈121 vs 210 chars; gitnexus/aws cut at the
first `\"`; multi-line plain scalars cut to the first line). Descriptions feed
the injection-hook nudge (`desc[:80]`), the autonomy executor's skill-awareness
list (`resources.py`, `desc[:120]`), and `/list-skills`.

## Approach — bounded, alias-forbidding `yaml.safe_load`
Parse frontmatter with `_FrontmatterLoader` (a `SafeLoader` subclass that
**refuses YAML aliases at the composer** and loads scalars as strings), behind a
64 KB size cap that routes pathologically large blocks to a linear legacy regex.
This is the same loader genesis-architect approved on a prior PR, and the same
choice `export_agents_md.py` already makes for AGENTS.md.

**Why yaml, not a hand parser:** a regex/string parser cannot match YAML's scalar
semantics (blank-line folding, inline-comment stripping, the full escape set) —
chasing that is unwinnable. Using yaml makes the parser correct *by construction*
and consolidates on one frontmatter parser. The DoS surface that motivated
avoiding yaml is closed two ways: **aliases refused** (billion-laughs / `*a`
multiplication) + **size cap** (mapping amplification) → legacy fallback.

## Validation
- 32 tests (truncation, every scalar form, blank-line fold, inline comment,
  `\u`/`\x` escape decode, **alias-bomb refused+bounded**, **oversized→legacy**).
- E2E over the real 65-skill catalog via `generate_catalog()`: 0 keyword diffs,
  0 description regressions, 15 improved. (One pre-existing malformed-YAML skill
  degrades gracefully to a non-empty description via the legacy fallback.)
- Measured DoS: alias-bomb → legacy ~3 ms; oversized(70 KB) → legacy ~0.7 ms.
- Consumer smoke: injection-hook `_score_skill` + `resources.py` consume the new
  catalog cleanly.

## Accepted residual (documented)
The only residual is **aggregate CPU** if an attacker plants ~100 crafted ≤64 KB
files in a scan root — but the scan roots are repo-versioned (PR-reviewed) or the
**user-curated** `~/.genesis/skill-library` (no autonomous ingestion; verified),
so exploiting it needs local FS write to the user's own dirs, and the impact is a
slow *background* refresh. The real amplifier — the hook re-spawning generators
with no concurrency guard — is a pre-existing hook bug filed separately
(follow-up `7ea8e6314c014b04ae421fc6c447732e`), not this parser's concern.

## Deploy path
Runtime script consumed by a hook + the autonomy executor: lands on `git pull`;
the catalog self-heals within 1 h — no migration, no restart-critical path.

Supersedes the closed hand-parser approach (which spiraled on YAML fidelity).

### DIFF
--- a/scripts/export_agents_md.py
+++ b/scripts/export_agents_md.py
@@ -140,11 +140,13 @@ def parse_mcp_tools(mcp_root: Path) -> list[dict]:
 def _frontmatter_description(skill_dir: Path) -> str | None:
     """Full, whitespace-normalized description from a skill's frontmatter.
 
-    The shared catalog scanner uses a line-based regex that truncates on
-    apostrophes/quotes and unfolded multi-line values; a proper YAML load of
-    the frontmatter block recovers the complete description. Returns None (so
-    the caller keeps the scanner's value) if yaml is unavailable or the block
-    doesn't parse to a usable string.
+    The shared catalog scanner now returns full descriptions via its own
+    string parser, but a YAML load here stays authoritative for AGENTS.md: it
+    resolves any scalar form (exotic escapes, multi-line quoted) exactly, so a
+    shipped artifact never depends on the scanner's documented edge-case limits.
+    Runs at build time over repo-tracked tiers only. Returns None (so the caller
+    keeps the scanner's value) if yaml is unavailable or the block doesn't parse
+    to a usable string.
     """
     if _yaml is None:
         return None
@@ -176,7 +178,7 @@ def collect_skills(repo_root: Path) -> list[dict]:
     install-specific ``~/.genesis/skill-library`` is intentionally NOT
     scanned (a shipped AGENTS.md must stay generalizable and must not leak
     a user's private skill names). Tier-1 wins on a name collision.
-    Descriptions are re-parsed with YAML so they aren't truncated.
+    Descriptions are re-parsed with YAML for authoritative full-fidelity values.
     """
     tier1 = _gsc._scan_tier(repo_root / ".claude" / "skills", 1, repo_root)
     seen = {s["name"].lower() for s in tier1}
--- a/scripts/export_agents_md.py
+++ b/scripts/export_agents_md.py
@@ -140,13 +140,11 @@ def parse_mcp_tools(mcp_root: Path) -> list[dict]:
 def _frontmatter_description(skill_dir: Path) -> str | None:
     """Full, whitespace-normalized description from a skill's frontmatter.
 
-    The shared catalog scanner now returns full descriptions via its own
-    string parser, but a YAML load here stays authoritative for AGENTS.md: it
-    resolves any scalar form (exotic escapes, multi-line quoted) exactly, so a
-    shipped artifact never depends on the scanner's documented edge-case limits.
-    Runs at build time over repo-tracked tiers only. Returns None (so the caller
-    keeps the scanner's value) if yaml is unavailable or the block doesn't parse
-    to a usable string.
+    The shared catalog scanner now parses frontmatter with a bounded YAML loader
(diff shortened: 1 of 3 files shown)


WingedGuardian:
@codex review

WingedGuardian:

---

WingedGuardian:
Addressed the P2 in 7a600e6f: `_collect_indented` now folds only lines indented strictly deeper than the `description:` key (the YAML continuation rule), so a uniformly-indented mapping no longer absorbs the sibling `keywords:` key, and plain-scalar comment lines are excluded (block scalars still keep `#` as content). Verified against `yaml.safe_load`: the extractor now matches YAML ground truth on all 66 real SKILL.md files (0 diffs). @codex review

WingedGuardian:
**Reworked to a bounded YAML parser (c4782eea).** The three R4 divergences were all "the hand parser doesn't match `yaml.safe_load`" — an unwinnable chase, so per the escalation cap I switched to the robust-by-construction fix: parse with a bounded, alias-forbidding `yaml.safe_load` (the loader used on a prior PR, and what `export_agents_md.py` already uses). All three are now correct **by construction** (verified: blank-line fold, inline-comment strip, and `\u2014`→— all match yaml exactly).

DoS is closed two ways: aliases refused at the composer (billion-laughs / `*a` multiplication → measured 1.9ms, flat RSS) and a 64KB size cap → linear legacy fallback (mapping amplification). genesis-architect verified the alias closure against 4 payload forms + oversized + recursion.

**Accepted residual (documented):** aggregate CPU (~1.4s/file) if ~100 crafted ≤64KB files are planted in a scan root. Reachability is low — scan roots are repo-versioned or the user-curated `~/.genesis/skill-library` (no autonomous ingestion), so it needs local FS write to the user's own dirs, and the impact is a slow *background* refresh, memory bounded. The real amplifier (the hook re-spawning generators with no concurrency guard) is a pre-existing hook bug filed as a separate follow-up. 34 tests, ruff clean, 0 corpus regressions. @codex review

WingedGuardian:
## First-pass review — PR #1393

**PII/Secrets/Wording scan:** CLEAN on code diff. No credentials, private IPs, personal emails, private hostnames, or hardware identifiers in any added line.

---

### NOTE — PR body describes a different implementation than what was committed

The PR title says "linear frontmatter parser" and the body describes an approach built around functions `_extract_description`, `_collect_indented`, `_unescape_double`, and `_norm_ws`. None of those functions appear in the diff. The PR body also explicitly states *"Builds no YAML object graph"*.

The actual implementation uses `yaml.load` with a custom `_FrontmatterLoader` SafeLoader subclass. It **does** build a YAML object graph — the DoS surface is bounded by the 64 KB size cap (`_MAX_FRONTMATTER_CHARS`) and alias refusal at the composer, not by structural absence of a graph.

The implementation itself is sound (bounded graph + alias refusal + scalar-only type guards is a defensible approach), but the PR description is a verbatim account of a different design. Anyone doing a future security audit of the `yaml.load` call will read the PR body, find "no YAML object graph" and "linear parser", and be confused — the `# noqa: S506` justification is in the code but the PR-level rationale contradicts it.

Suggest updating the PR description to reflect what was actually committed before merge.

---

WingedGuardian:
@codex review

PR #1394, opened 2026-08-13
fix(hooks): per-worktree review evidence + never-shared worktree key (complete #1244)

WingedGuardian:
## What & why

Completes #1244. That PR made the review **marker** and **round counter** per-worktree so concurrent CC sessions don't clobber each other's review state — but two residual races remained, and both bit live this session:

1. **`_worktree_key` collapsed to a shared `"default"`** whenever `git rev-parse --show-toplevel` failed (timeout/error) — *likeliest under the concurrent load this keying exists to survive*. On that fallback every session shared one key, reopening the exact cross-session clobber #1244 fixed. Observed live as a review marker vanishing and the escalation round counter resetting `4→1` mid-workflow.
2. **The review-evidence file was global** (`--agent-output` defaulted to `~/.genesis/last_code_review.txt`). Concurrent sessions overwrote each other's audit, and the depth gate's `_evidence_is_adversarial` check validated a marker against whatever content happened to be in that shared file — a correctness hole, not just noise.

## Changes (all in `scripts/review_state.py` + instruction text)

- **`_worktree_root(cwd)`**: git primary; on failure, walk up to the nearest `.git` (a FILE in a linked worktree, a dir in the main tree), else the resolved cwd — **never a shared constant**. `_worktree_key` re-based on it; the git-success path is `realpath`'d so it matches the fallback for the same worktree.
- **Per-worktree evidence**: `_evidence_file(cwd)` under a new `_EVIDENCE_DIR`, keyed like the marker. `mark`'s default `--agent-output` now points there. New **`evidence-path`** CLI prints it (and creates the dir, so a `> $(evidence-path)` redirect can't fail on a fresh host).
- The two commit-hook instruction messages (`review_enforcement_commit.py`) and `SKILL.md` repointed to the per-worktree path.

## Tests

`tests/test_scripts/test_review_evidence_isolation.py` (6, verify-RED against the old `"default"` fallback / missing `_evidence_file`), including the core **git-success ↔ fallback key-agreement** property (a wrong fix here would block *every* commit), a subdir-cwd test, a real `mark_reviewed`-wiring test, and the evidence-dir-creation test. 229 existing gate/review tests (`test_hooks` + `test_scripts`) green; ruff clean.

genesis-architect audited; its SHOULD-FIX (evidence dir never created → gate could wedge on a fresh host) and the realpath asymmetry were both fixed in-branch. Both changes are fail-closed (never false-allow).

### DIFF
--- a/.claude/skills/genesis-development/SKILL.md
+++ b/.claude/skills/genesis-development/SKILL.md
@@ -519,7 +519,7 @@ above (full definitions in `.claude/agents/genesis-architect.md`):
   conscious, logged act (like `# review-override`); adding it — or falsely passing
   `--clean` — WITHOUT the honest review result is the same violation as ignoring the
   prose above (the `--clean` flag mirrors the review record you write to
-  `~/.genesis/last_code_review.txt` at the same moment: falsifying one falsifies the
+  the per-worktree evidence path (`review_state.py evidence-path`) at the same moment: falsifying one falsifies the
   other).
   Caveats: **multiple findings in a single pass = one round** (not an
   escalation); the same defect reappearing (an incomplete prior fix) is a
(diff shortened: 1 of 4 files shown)


WingedGuardian:

---

WingedGuardian:
@codex review

PR #1398, opened 2026-08-14
fix(awareness): suppress false critical_failure under event-loop starvation + stall sampler

WingedGuardian:
## What

Two changes: stop a false `critical_failure` alarm, and instrument its root cause.

**B — `critical_failure` no longer fires on loop-starvation-induced probe timeouts.**
The `critical_failure` signal probes local infra (DB, Qdrant, + Ollama if enabled) with a
3s aiohttp timeout. When background work stalls the event loop, those probes time out and
report DOWN → `critical_failure=1.0` even though the infra is healthy. Because
`critical_failure` is a `_MICRO_CRITICAL_SIGNALS` member, each false spike woke a Micro
reflection + Telegram alert.

The collector now reclassifies a DOWN to `0.0` **only** when every DOWN probe timed out
**and** the `loop_health` sample (the off-loop seam from #1375) shows the loop is not
scheduling normally — a fresh + lagging sample, or a stale sample inside a wedged window
(1s–30s). Fail-closed: a hard-error (connection-refused) DOWN, a mixed batch, a healthy
loop, a stale sample past the 30s ceiling (i.e. the sampler itself died), or absent
evidence all keep `1.0`; and the suppression is wrapped so any internal exception keeps the
un-suppressed value (the `_safe_collect` wrapper would otherwise turn an escaping exception
into `0.0`). Probes now carry `ProbeResult.timed_out` to distinguish a timeout from a hard
error.

**Instrumentation — off-loop stall-stack sampler (`util/loop_stall.py`).**
The on-loop lag sampler can only measure a stall *after* it clears, so it never captures the
synchronous frame that blocked the loop. A new off-loop daemon thread reads the `loop_health`
heartbeat and, when it goes stale (loop wedged now), snapshots the loop thread's stack via
`sys._current_frames()` and logs it — one dump per episode. Diagnostic-only; env-gated
(`GENESIS_LOOP_STALL_SAMPLER` / `GENESIS_LOOP_STALL_MS`), no-op if the heartbeat publisher is
disabled.

## Why

The false alarm was firing repeatedly under normal background load. The underlying
loop-starvation source is not yet identified (a trace disproved the obvious hypothesis), so
the stall sampler is the instrument to catch it live; B stops the false alarm at the signal
boundary regardless of the source.

## Testing

Test-first (verify-RED → GREEN): probe timeout-flag (timeout vs hard error); suppression for
all-timeout + wedged / fresh-lagging; fail-closed for hard-error / mixed batch / healthy-loop
/ dead-publisher-stale / absent-evidence / suppression-exception; stall sampler
dedup/reset/none-sample/missing-frame/shutdown. New tests + health-probe and
standalone-shutdown-ordering regressions green; `ruff` clean. Pre-implementation review +
two adversarial diff reviews (genesis-architect); a SHOULD-FIX on dead-publisher staleness
was closed with a regression test.

## Deploy

Runtime code — `git pull` + server restart. No migration. The stall sampler starts with the
server; the suppression takes effect on the next steady-state awareness tick.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,17 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **No more false "critical failure" alarms when the system is briefly busy.** The
+  health signal that watches your local infrastructure (database, vector store, and
+  Ollama if enabled) probes those services with a short timeout. When background work
+  momentarily stalls Genesis's event loop, those probes could time out even though the
+  services were perfectly healthy — firing a spurious "critical failure" that triggered
+  a reflection and a Telegram alert. Genesis now recognizes when a probe timed out
+  because the loop was starved (rather than because a service is actually down) and
+  suppresses the false alarm, while still firing on a genuine outage. A new diagnostic
+  also captures what code was blocking the loop during such a stall, to help track down
+  the underlying cause.
+
 - **Restricted reasoning sessions can no longer escape their tool restrictions by
   spawning.** Several of Genesis's restricted Claude Code sessions (deep/strategic
   reflection, the inbox/mail judges, and the experimentation completion) could spawn a
--- a/src/genesis/hosting/standalone.py
+++ b/src/genesis/hosting/standalone.py
@@ -120,41 +120,62 @@ class StandaloneAdapter:
             "false",
             "off",
         ):
-            if lag_sampler_task is None:
-                logger.warning(
-                    "loop-stall sampler enabled but its heartbeat publisher "
-                    "(the loop-lag sampler, GENESIS_LOOP_LAG_SAMPLER) is disabled "
-                    "— the stall sampler will no-op until the publisher runs",
-                )
-            from genesis.util.loop_stall import run_loop_stall_sampler
-
+            # The sampler is diagnostic-only: any failure to start it (import
+            # error, thread-limit RuntimeError from Thread.start under resource
+            # pressure) must disable only the sampler, never abort serve() before
+            # Flask/Telegram come up.
             try:
-                stall_ms = float(os.environ.get("GENESIS_LOOP_STALL_MS", "1000"))
-            except ValueError:
-                stall_ms = 1000.0
-            # float() accepts nan/inf/negatives — reject them here so a misconfig
-            # is surfaced (the sampler's __init__ also clamps as a backstop).
-            # `not (x > 0)` catches NaN (NaN>0 is False), 0, and negatives.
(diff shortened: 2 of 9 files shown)


WingedGuardian:
@codex review

WingedGuardian:

---

WingedGuardian:
Triaged the 4 P2s (verified each against source):

- **P2-3 (preserve DEGRADED)** — fixed. Suppression now recomputes the residual worst-status over the non-timeout-suppressed probes, so a genuine `DEGRADED` (503) coincident with a starvation timeout yields 0.5, not 0.0. Test added.
- **P2-2 (metadata not persisted)** — fixed. The tick serializer + `signal_format` drop `SignalReading.metadata`, so the suppression explanation now lives in `baseline_note` (which both retain), making a suppressed reading distinct from a healthy 0.0. Test added.
- **P2-4 (non-finite/non-positive stall threshold)** — fixed. `LoopStallSampler.__init__` now clamps NaN/inf/zero/negative to the default (the env `float()` guard only catches `ValueError`). Test added.
- **P2-1 (mask a genuine timeout during a coincident wedge)** — documented-accept. This is unresolvable from a single reading (`loop_health` proves the loop is starved, never that the dependency is reachable); it self-heals on the next tick with a healthy loop, and `critical_failure` is a reflection signal, not the hard-failure alarm (status.json watchdog owns that). Comment added at the suppression site.

@codex review

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
Final disposition (merging with a documented stale-review-override — Codex reviewed the substance across four rounds; the remaining delta is diagnostic-only hardening):

- **"Do not infer a wedge from a recently stopped publisher"** (critical_failure.py:113) — verified real but **documented-accept**. It requires the lag sampler to *crash* (it has almost nothing that can raise) AND a coincident timeout-type outage within the 1–30s window AND a tick landing there; it self-heals past 30s (fail-closed) and `critical_failure` is a reflection signal, not the hard-failure alarm (status.json watchdog owns that). Tracked as a hardening follow-up: invalidate loop_health on sampler exit so a dead publisher reads as `None` → fail-closed.
- **"Keep diagnostic thread failures from aborting startup"** + **"Keep the stall threshold above the heartbeat cadence"** — FIXED (16fe788b): sampler startup wrapped in try/except; `GENESIS_LOOP_STALL_MS` floored at 2× the ~500ms cadence.
- The rest are re-raises of already-fixed/accepted items (preserve-DEGRADED, persist-metadata→baseline_note, validate-stall_ms, retry-transient-dump, coincident-wedge P2-1).

None are BLOCKER/P1. All CI green.

PR #1400, opened 2026-08-14
perf(memory): offload entity-registry difflib fuzzy-match off the event loop

WingedGuardian:
## What

Offload the entity-registry fuzzy-match (`_closest`) off the event loop.

## Why

The loop-stall sampler added in #1398 captured the live event-loop starvation source on
its first wedge: `entity_registry._closest` runs synchronous O(n·m) `difflib` fuzzy-matching
over every same-cluster candidate **on the event loop** during the memory-extraction cycle.
As the entity registry grew past the module's "small enough to scan in-process" assumption,
this blocked the loop ~1s+ per resolution (measured: WEDGED stack dumps at ~1050ms each,
`resolve_entity → _closest → difflib.ratio`). That loop starvation is what times out the
health probes into false `critical_failure` alarms and drives the recall 503s.

## Change

`resolve_entity` now runs `_closest` via `asyncio.to_thread`, so the loop keeps scheduling
while the match runs on a worker thread. `_closest` is a pure function over an
already-materialized candidate list (no DB, no shared state, no `self`), so it is safe
off-loop; behavior is unchanged.

## Testing

The 33 existing entity-layer + write-wiring tests pass unchanged (behavior-equivalence lock —
the offload does not alter `resolve_entity`'s output or the fuzzy AMBIGUOUS tier). `ruff` clean.
Concurrency safety (purity, await-ordering, executor/GIL) adversarially reviewed.

## Deploy

Runtime code — `git pull` + server restart. E2E: after restart, the `difflib` frames should no
longer appear in the loop-stall sampler's WEDGED dumps.

Addresses the difflib starvation source tracked under follow-up fb3cf7be (other candidate
sources — inbox scanner sync-I/O, the ~30-min timer thundering-herd — remain to confirm once
the sampler runs at steady state).

### DIFF
--- a/src/genesis/memory/entity_registry.py
+++ b/src/genesis/memory/entity_registry.py
@@ -21,6 +21,7 @@ rewriting here.
 
 from __future__ import annotations
 
+import asyncio
 import difflib
 
 import aiosqlite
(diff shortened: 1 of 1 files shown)


WingedGuardian:
@codex review

WingedGuardian:

---

PR #1403, opened 2026-08-15
fix(memory): OR-fallback for FTS5 recall to end implicit-AND starvation

WingedGuardian:
FTS5 treats a bare space-separated `MATCH` string as an implicit **AND**, so
multi-word natural-language queries (`reference_lookup` / `knowledge_recall`, and
memory recall on its non-expanded fallback path) required **every** token to be
present or returned nothing. Root cause re-verified on the live `knowledge_fts`:
an 11-term AND query returned 0 rows while the same terms OR-joined returned the
target ranked #1.

## Fix
New shared `db/crud/_fts.py::fetch_fts` runs the precise AND query first and —
**only when it returns zero rows and the query is not an already-structured
boolean expression** — retries the terms OR-joined (on a copy of `params`). It
only ever *adds* results when AND found none; it never changes a query that
already hit.

Wired into the recall surfaces:
- `knowledge.search_fts` (`knowledge_fts`; always applies — no boolean mode)
- `memory.search_ranked` (`memory_fts`, `boolean=False` path — the fallback the
  hot recall path uses when `expand_query` cannot expand, e.g. Qdrant down)

`memory.search` is left **strict-AND on purpose**: its only callers are the
entity-name resolver (`linker._find_entity_by_name`), which picks `results[0]`
and must not be widened to single-term matches (it would bypass its difflib
quality gate → spurious graph links). Audited-clean and unchanged:
`extraction_job` dedup already OR-joins; `voice/hygiene` matches a fixed constant.

## Tests / verification
- `_fts` helper units + integration on both recall paths (partial-term → fallback
  surfaces; all-present → AND unchanged; `boolean=True` → no fallback); a lock
  test pinning `memory.search` AND-only; one existing test re-probed off an
  incidental AND-strictness assumption.
- **82 passed** (incl. `test_linker` suites); an earlier full sweep of
  **2789 passed** across `test_db` + `test_memory` + memory MCP.
- FTS5 operator/numeric edge cases probed live (OR-joined lowercased tokens never
  form an operator); `params` copy is non-destructive; `match_index=0` correct in
  both callers.

Adversarially reviewed (genesis-architect): the single SHOULD-FIX — a linker
resolution gate that `memory.search` feeds — is resolved by **not** widening
`memory.search`, verified firsthand and by the linker test suites.

No migration, no schema change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,24 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ## [Unreleased]
 
+### Fixed
+
+- **FTS5 recall no longer starves on multi-word queries.** `_prepare_fts5` builds a
+  bare space-separated FTS5 MATCH, which SQLite treats as an implicit AND — so a
+  verbose query (`reference_lookup` / `knowledge_recall` natural-language text, and
+  memory recall on its non-expanded fallback path) required *every* token to be
+  present and otherwise returned nothing. A shared `db/crud/_fts.py::fetch_fts` now
+  runs the precise AND query first and, only when it returns zero rows and the query
+  isn't an already-structured boolean expression, retries the terms OR-joined —
+  adding partial matches where there were none while leaving every already-matching
+  query unchanged. Applied to the recall surfaces `knowledge.search_fts` and
+  `memory.search_ranked` (the latter's `boolean=False` path, which the hot recall
+  path falls back to when `expand_query` can't expand, e.g. Qdrant unavailable).
+  `memory.search` is left strict-AND on purpose — its only caller resolves entity
+  names by `results[0]` and must not be widened to single-term matches.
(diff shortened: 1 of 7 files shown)


WingedGuardian:

---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
Disposition of the 2 non-blocking Codex P2s (consciously accepted):

1. **Filter stopwords out of OR retries** — already implemented in this PR (`or_fallback` drops ~48 English stopwords from the OR join; commit `dde4a789`). This is a non-deterministic re-review of an already-fixed finding.
2. **Skip fallback when every term is a stopword** — deliberate: when *every* token is a stopword we OR them all as a last resort, since an OR still recalls something vs. returning nothing. This fires only on the zero-AND fallback for an all-stopword query, and BM25 rank + LIMIT bound it. Accepted as a judgment call.

WingedGuardian:
@codex review

PR #1408, opened 2026-08-19
fix(inbox): end the approval-request storm — idempotent approvals, real DM scroll-up, faithful park resume

WingedGuardian:
## What

Seven commits fixing five measured defects plus four hardening features, all root-caused from a live incident (48 phantom approval requests/day for four days) and adversarially reviewed per-commit + branch-wide.

### Approval storm (the headline fix)
- `get_all_known()` rebuilt as ONE recency-ordered scan — a permanently-failed row can no longer clobber a newer completed row's hash (the storm seed: phantom "modified" every 30-min tick).
- New/modified inbox content while an approval is pending now **parks onto the same request** (per-file supersession, delta recomputed) — the cancel-and-recreate refresh that sent a fresh Telegram message per tick is gone. The orphan guard is the only remaining cancel path.
- Per-policy re-ask control: `reask_overrides: {inbox_evaluation: 0}` ships (never re-ask a delivered inbox request); all other policies keep the 24h re-ask. Delivery-failure recovery (no successful send on record) always retries, for every policy — recovery, not a reminder.
- Tri-state approval consume (consumed / already-consumed / failed-with-retry+ERROR); approved-then-edited content is invalidated AND its approval consumed at dispatch time, so post-approval edits require a fresh approval.

### DM continuity ("real scroll-up")
- `conversation_history` gains optional `chat_id` (scoped) + `before` (cursor pagination, full-length messages); unscoped cross-chat default preserved.
- Session-recovery recap is byte-budgeted and tail-biased (the END of long replies survives; the old per-message 300-char head chop dropped exactly the conclusions).
- Every fresh telegram session gets a conversation-identity block naming its real chat id (correct for negative group ids).

### Rate-limit park integrity
- Park payload/queue/consumer now round-trip the full execution shape (`system_prompt`, `source_tag`, `skills`, `tool_exceptions`, `planning_instruction`) — locked by a required-set test; a parked campaign no longer resumes without its strategy doc.
- Campaign bookkeeping follows parks (structured `park_id` metadata stamp): open park = still running (bounded 7 days, fail-closed on corrupt rows), resumed = capture the delivering session's real result, terminal = real failure.

### Hardening / features
- Firecrawl as an explicit **paid** escalation backend on `web_fetch`/`web_search` (never in the auto chains — locked by exhausted-chain purity tests); stale Firecrawl-is-MCP labels fixed.
- CC login health: hourly expiry warning for the fixed-lifetime OAuth refresh token (`GENESIS_CC_LOGIN_EXPIRY_WARN_DAYS`), plus fallback to a stored 1-year setup-token ONLY on hard-expired + probe-confirmed-dead login (never on ambiguity; never for peer-routed/bare/cleanroom invocations).
- Settings writes provenance-stamped (`# set-by:` headers, whole hand-written comment block preserved unconditionally); disabling `manual_approval_required` requires a strict-typed confirmation flag on both surfaces and is loud on both (Telegram alert / critical observation). Fixed a pre-existing silent data-loss bug in the surplus PUT (merge result discarded).
- cc-update process: new-capability detection now produces a ready follow-up naming the exact instruction change (adoption path, not just an informational note).

## Review
Six per-commit adversarial audits + a branch-wide fresh-context deep review + a sequential security review (no CRITICAL; both SHOULD-FIXes applied in the final commit). Every fix landed test-first; every new test was verified RED via the bug or deliberate sabotage.

## Deploy notes
- Config default change: `config/autonomous_cli_policy.yaml` gains `reask_overrides` (deep-merge means existing overlays inherit it; restoring inbox reminders requires an explicit `inbox_evaluation: 24`).
- Runtime code + MCP-server changes: server restart + next CC session start respectively; no schema migrations.
- First post-deploy inbox tick is a quiet no-op on this install (verified against live rows).

Ledger: ebf7fd58951f4750a5d0a9e99eb23e11
Ledger: 26f151bec33a46d69561e21fcb37e241
Follow-up: af265c4c1ce845248893ed1e61c498c3
Follow-up: 7f18b3ab8d0a43cb89639bb9de250635
Follow-up: 6fd24166616f48deaa8e5e34592ab45b
Follow-up: ebf2acaa586f4088a9a5e0d3391471fe

### DIFF
--- a/.claude/agents/genesis-researcher.md
+++ b/.claude/agents/genesis-researcher.md
@@ -8,8 +8,9 @@ You are a research agent for Genesis with access to powerful web and code intell
 
 ## Web Tools (MCP — use these, NOT CC WebFetch/WebSearch)
 
-- **`web_fetch(url)`** — Fetch any URL with smart anti-bot bypass. Scrapling TLS impersonation → Crawl4AI JS rendering → httpx. Returns structured content.
+- **`web_fetch(url)`** — Fetch any URL with smart anti-bot bypass. Auto chain: TinyFish (anti-bot, JS) → Scrapling TLS impersonation → Ladder → Crawl4AI JS rendering → httpx. Returns structured content.
 - **`web_search(query)`** — Search the web via TinyFish (primary) → SearXNG (self-hosted fallback) → Brave. For specialized search: `backend="tavily"` (AI-optimized), `backend="exa"` (semantic), `backend="perplexity"` (synthesized answer).
+- **Escalation when the free chain dead-ends** (hard anti-bot, paywalled, heavy JS): `web_fetch(url, backend="firecrawl")` / `web_search(query, backend="firecrawl")` — a PAID cloud scraping API (burns credits; never used automatically). Reach for it only after the auto chain has actually failed, and say so in your findings.
 - **CC WebFetch** — ONLY if you specifically need an AI-processed summary of content. Otherwise use `web_fetch`.
 - **CC WebSearch** — ONLY for trivial general lookups. Otherwise use `web_search`.
 
@@ -40,6 +41,7 @@ When searching for repos, libraries, or implementation patterns on GitHub:
 | Need | Tool |
 |------|------|
 | Fetch a webpage | `web_fetch(url)` |
+| Fetch failed on anti-bot/paywall/JS | `web_fetch(url, backend="firecrawl")` (paid, last resort) |
 | Search the internet | `web_search(query)` |
 | Search GitHub repos | `gh search repos "query"` via Bash |
 | Search GitHub code | `gh search code "query"` or grep.app via `web_fetch` |
--- a/.claude/skills/cc-update/SKILL.md
+++ b/.claude/skills/cc-update/SKILL.md
@@ -69,3 +69,24 @@ the delta touches:
 
 Capture any such change as a follow-up + a doc entry — that is how the next update stays
 execute-not-rediscover.
+
+## Detection → behavior (new-capability adoption path)
+
+The impact eval treats a new feature as *informational* by design ("available"
+is not "required") — correct for alerting, but it left no path from "detected"
+to "Genesis actually starts using it". So, when the changelog review flags a
+NEW skill/command/flag that Genesis would plausibly WANT (it overlaps an
+existing Genesis workflow, replaces a hand-rolled mechanism, or covers a known
+gap), do BOTH:
+
+1. File the informational KB entry as usual (no alert — calibration unchanged).
+2. Create a `follow_up_create` row (`work_state="ready"`, low priority) naming
+   the SPECIFIC instruction change, e.g. "once CC >= vX.Y.Z is pinned: prefer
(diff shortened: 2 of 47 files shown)


WingedGuardian:
## First-pass review — structural + PII/secrets/wording

Verdict: **REQUEST_CHANGES** (one ERROR blocks merge; posting as comment since self-review is not permitted via the review API)

---

### 🔴 ERROR (must fix before merge)

**`src/genesis/campaigns/runner.py` — `_follow_park` — raw SQL against genesis.db**

Two `db.execute(...)` calls issue SQL directly against `cc_rate_limit_parks` and `cc_sessions` in a non-crud, non-migration module. This violates the raw-SQL pattern rule:

- The `cc_sessions` query should go through `genesis.db.crud.cc_sessions` (that CRUD module exists).
- The `cc_rate_limit_parks` query should go through the service layer in `src/genesis/cc/rate_limit_park.py` (or a new helper there if this access pattern is new).

Inline `db.execute` in business logic is the pattern we're enforcing against — the CRUD modules exist precisely so query logic isn't scattered across the codebase.

---

### 🟡 WARNING

**`src/genesis/providers/firecrawl_client.py` — no dedicated test file**

This is a new file in `src/` and no `tests/test_providers/test_firecrawl_client.py` exists. The `test_web_tools.py` additions mock the client at the call site, which tests the integration wiring, but the client itself (singleton init, `_get_key()`/`_headers()` helpers, HTTP error paths for 401/402/429, the `data.get("web", [])` defensive branch in `search()`) has no direct unit coverage. Worth adding a dedicated test file.

**`src/genesis/cc/conversation.py` — `import os` inside method body**

`import os` appears inside the `_build_recovery_context` try block rather than at the module top level. Functional but inconsistent with project convention; move it to the module-level import block.

---

### 🔵 NOTE

**`src/genesis/cc/login_health.py` — `_get_probe_lock` creates `asyncio.Lock()` in a sync function**

Safe in Python 3.10+ (no event-loop binding at construction time), but worth a one-liner comment noting the version assumption since this is non-obvious to future readers.

**`src/genesis/db/crud/inbox_items.py` — `supersede_parked_rows` docstring**

The case where supersession leaves zero live rows (orphan guard picks them up later) is described only in an inline comment, not the docstring. That's the most operationally surprising path for a caller — worth surfacing in the docstring.

---

### ✅ PII / Secrets / Wording — CLEAN

Scanned all added lines (`+`): no API keys, tokens, private IPs, SSH material, or credentials found. `secrets.env.example` additions are all empty placeholders. The `<tailscale-ip>` comment is a placeholder, not a real address.

PR title and body describe functional changes only — no security-incident language, PII-cleanup disclosure, or exploitable vulnerability details.

---

**Summary:** 1 ERROR · 2 WARNING · 2 NOTE · 0 hard blocks · 0 PII findings. The raw-SQL issue in `_follow_park` is the only blocking item.

---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
@codex review

PR #1412, opened 2026-08-19
fix(hooks): resolve gh pr merge PR from the positional, not a value-flag's value (E3)

WingedGuardian:
## What

`_extract_pr_number` (the `gh pr merge` PR-number resolver behind every merge gate) read the **first bare-digit token** in the command and did **not** skip a value-flag's argument. So an unquoted numeric flag value was misread as the target PR:

- `gh pr merge --subject 123 5` → returned `123` (the `--subject` value), while gh actually merges the trailing positional **PR 5**. Every downstream gate then evaluated the wrong PR's facts.

gh's own semantics (docs): `--subject` / `--body` / `--body-file` / `--author-email` / `--repo` / `--match-head-commit` are value-taking flags; the lone positional is the PR (number / URL / branch).

## Fix

Walk the tokens with gh-equivalent consumption, reusing the value-flag specs the sibling parsers already use (`_GH_MERGE_VALUE_FLAGS`, `_short_cluster_consumes_next` — the same set `_merge_match_head` / `_comment_target` rely on):

1. A value flag (long, short-single, or a short cluster whose trailing value-letter has no glued remainder like `-db`) consumes its next token — skip both. Never swallow a shell separator as a value (a dangling `--subject ; gh pr merge 999` must not leak the chained command's `999`).
2. A positional PR ref never starts with `-`. Guarding the number/`#N`/URL matchers on that also closes a **glued-value sibling** of the same class found in review: `--body=…/pull/999` and `-b…/pull/999` are single `-`-prefixed tokens that don't consume a next token, and the URL matcher's `\S*` prefix would otherwise read `999` out of the flag's value.

The two sibling parsers in this file already skipped value flags; this resolver was the one that didn't.

## Reachability / severity

Contained on the normal merge path (the shadow-flag belt refuses `--subject`/`--body`/… once the head-binding engages; the only non-shadow value flags — `--repo`, `--match-head-commit` — hit repo-unresolvable / match-head-mismatch blocks, and GitHub enforces `--match-head-commit` server-side). The reachable observable is under `# stale-review-override` (freshness waived → shadow belt + binding skipped), where the finding scanners then run on the mis-resolved PR. Correctness / defense-in-depth fix; the resolver is the contract, not the belts. Sole caller is fail-closed (unresolved → block or numberless re-resolve); a wrong skip can only over-block.

## Tests (RED-first)

- Unit (`TestResolvePrNumber`): unquoted value flags (`--subject/-t/--body/--match-head-commit/--repo/-db 123 5` → `5`), glued-value URL siblings (`--body=…/pull/999 5` → `5`), chained-command guard (`--subject ; gh pr merge 999` → `None`), `--` end-of-options (`-- 5` → `5`), and positional/quoted/URL/separator regression locks. Each new case verify-RED'd (returned the wrong PR before the fix).
- End-to-end (`test_e3_stale_override_gates_positional_pr_not_flag_value`): drives `main()` with `--subject 123 5 # stale-review-override`, seeds an inline P1 on PR 5, and asserts the scan targets PR 5 not 123 (verify-RED at the main() level: reverted source → gate scanned PR 123 and allowed).

Full `tests/test_hooks/` + `tests/test_scripts/` green; ruff clean.

Ledger: de9f128c

### DIFF
--- a/scripts/hooks/git_push_guard.py
+++ b/scripts/hooks/git_push_guard.py
@@ -436,20 +436,50 @@ def _extract_pr_number(cmd: str) -> str | None:
         tokens = shlex.split(spaced)
     except ValueError:
         tokens = spaced.split()
-    for tok in tokens:
+    seps = {";", "&", "|", "&&", "||"}
+    i = 0
+    while i < len(tokens):
+        tok = tokens[i]
         # Stop at the end of THIS command — tokens after a separator
         # belong to a chained command, and their digits must not be read
         # as this merge's target (`gh pr merge 123; echo 456` merges 123
         # but this loop would otherwise return 456). 2026-07-10 review.
-        if tok in {";", "&", "|", "&&", "||"}:
+        if tok in seps:
             break
-        if tok.isdigit():
-            return tok
-        if tok.startswith("#") and tok[1:].isdigit():
-            return tok[1:]
-        url = re.match(r"\S*/pull/(\d+)\b", tok)
-        if url:
-            return url.group(1)
+        # A value-taking flag consumes the NEXT token as its value, so an
+        # UNQUOTED numeric value is never misread as the PR: gh parses
+        # `gh pr merge --subject 123 5` as subject="123" + PR 5, and the old
+        # loop returned 123 (the flag value) → the gates checked the WRONG PR
+        # (E3, 2026-08-19). Covers long/short-single value flags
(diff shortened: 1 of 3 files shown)


WingedGuardian:
@codex review

WingedGuardian:

---

PR #1421, opened 2026-08-21
fix(install): apply memory-resilience provisioning on fresh container installs

WingedGuardian:
## What

`scripts/install.sh` (the fresh-container installer) never applied the
memory-resilience provisioning that `bootstrap.sh` and `update.sh` already do —
systemd-oomd pressure-kill, the swap invariant, and the raised per-user-slice
`TasksMax`. Because nothing auto-runs `update.sh` after a fresh install, a newly
installed box sat unprotected against the OOM-thrash / `Cannot fork` wedge until
its first manual update.

This inserts a guarded block into install.sh Step 11 that mirrors
`bootstrap.sh`, using install.sh's **own** `_install_pkg` idiom — bootstrap's
bare `install_pkg` is undefined in install.sh, so a verbatim copy would
hard-abort a fresh install under `set -euo pipefail`.

## Why it's safe

- The lib entrypoints (`memory_resilience_apply`, `pid_budget_apply`) are
  contractually always-return-0; `_install_pkg … || echo` neutralizes a failed
  package install. The block cannot abort the installer.
- Verified by an isolated dry-run: oomd-installs, oomd-install-fails, and
  re-run cases all exit 0 with the lib taking its clean skip path.
- 3 new locking tests in `test_memory_resilience.py` (idiom, install→apply
  ordering, guard) — verified RED before the fix, green after (24 passed).
- Adversarial review (genesis-architect, fresh context): no BLOCKER/SHOULD-FIX;
  two NOTE-level doc-staleness gaps it surfaced are fixed in this PR.

## Scope

Closes gap **(a)** of follow-up `d7ffc73b`. The **(b)** consolidation of the
13+ other `install.sh` ↔ `bootstrap.sh` provisioning gaps remains open as
separate work.

## Docs

Corrected `docs/reference/memory-resilience.md`, `docs/architecture/CURRENT.md`,
and the lib header to list `install.sh` as a third caller. CHANGELOG entry added.

Ledger: ccc9b74c788c49f89064777f902551ca

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,14 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Fresh container installs now get OOM/fork-wedge protection.**
+  `scripts/install.sh` (the fresh-container path) never applied the
+  memory-resilience provisioning (systemd-oomd pressure-kill, swap invariant,
+  raised per-user-slice `TasksMax`) that `bootstrap.sh` and `update.sh` already
+  did — so a freshly installed box sat unprotected against the OOM-thrash /
+  `Cannot fork` wedge until its first `update.sh` run. It now applies the same
+  idempotent, adaptive provisioning at install time.
+
 - **The core Claude Code spawner now uses the shared hardened group-kill.**
   `cc/invoker.py` — the launcher behind every CC session Genesis runs — carried
   the patterns the repo-wide sweep retired everywhere else: `preexec_fn` spawns,
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -1460,7 +1460,7 @@ verified: 3c514f3e 2026-08-10
   don't merge them. Memory-resilience invariants are first-class facts:
   container `cgroup_memory_swap_max` (tri-state — "0" IS the 2026-07 wedge
   state) + `oomd_user_slice_kill` (config-plane scan of user.slice.d drop-ins,
-  laid down by `scripts/lib/memory_resilience.sh` from bootstrap/update) and
+  laid down by `scripts/lib/memory_resilience.sh` from install/bootstrap/update) and
   host-plane `swap_total_kb`, so the annotation layer flags unprotected
   installs (see docs/reference/memory-resilience.md). Network-resilience
   invariants are first-class too: container `networkd_keep_configuration` +
--- a/docs/reference/memory-resilience.md
+++ b/docs/reference/memory-resilience.md
@@ -23,9 +23,9 @@ The fingerprint, if you're diagnosing it live:
 
 ## What Genesis sets up (and what it only warns about)
 
-`scripts/lib/memory_resilience.sh` runs from `bootstrap.sh` on fresh installs
-and from every `update.sh` (which re-runs bootstrap), so existing installs
-retrofit automatically. It is idempotent and **adaptive — every threshold is a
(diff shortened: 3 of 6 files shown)


WingedGuardian:

Notes (not blocking):
- Verified `_install_pkg systemd-oomd systemd-oomd-defaults` matches the existing `_install_pkg <pkg> [<pkg-for-dnf/yum>]` helper signature already defined in `install.sh` (no bare `install_pkg` call, which would be undefined there).
- Confirmed the new block is a single top-level insertion (no duplicate sourcing of `lib/memory_resilience.sh` elsewhere in `install.sh`) and mirrors the guarded pattern already used in `bootstrap.sh`.
- New test coverage (`test_install_wires_the_lib`, `test_install_provisions_the_oomd_package_with_its_own_idiom`, `test_install_guards_the_lib_source`) follows the file's existing comment-over-docstring convention — consistent, no NOTE needed.
- Diff is scoped correctly: no secrets, private IPs, credentials, or PII in added lines; PR title/body/CHANGELOG entry read as ordinary infra-resilience language with no leak/incident wording.


---

---

WingedGuardian:
**Structural review (first-pass, automated)**

Scope: `scripts/install.sh`, `scripts/lib/memory_resilience.sh`, `tests/test_scripts/test_memory_resilience.py`, plus doc/CHANGELOG updates. No Python source under `src/` is touched, so most of the Python-specific ERROR checks (task_states inserts, bare `except`, mutable defaults, `os.killpg`, raw SQL, import hygiene) don't apply here.

**Verified claims** (read the actual code, not just the PR description):
- `_install_pkg` in `install.sh` (line ~323) takes `(pkg_apt, pkg_dnf_optional)` — the new call `_install_pkg systemd-oomd systemd-oomd-defaults` (line 1212) matches the real signature. Confirmed `install_pkg` (bootstrap's bare name) is genuinely absent from `install.sh`, so the idiom fix is real, not cosmetic.
- `memory_resilience_apply` and `pid_budget_apply` in `scripts/lib/memory_resilience.sh` both terminate every branch with `return 0` — the "contractually always-return-0" claim in the PR body checks out by inspection.
- `bash -n` passes clean on both `scripts/install.sh` and `scripts/lib/memory_resilience.sh`.
- New tests are well-isolated (read-only text assertions against the script, no side effects) and correctly grouped under existing `# ──` section banners.
- Commit message and PR title both carry a conventional prefix (`fix(install): …`).

**NOTE — provisioning runs after genesis-server is already started**

In `install.sh`, the new memory-resilience block (line 1198) is placed *after* the `genesis-server` enable/start block (lines 1187–1196), and nothing later in the script restarts the service. In `bootstrap.sh` the equivalent block runs earlier in the flow, before any service is (re)started there.

Functionally this is likely harmless — TasksMax, the swap invariant, and systemd-oomd's PSI-based pressure-kill are cgroup/kernel-level settings that apply going forward without needing the already-running process to restart — so this is not a correctness bug, just a minor ordering inconsistency versus `bootstrap.sh`'s placement. Worth a second look only if there's a reason (not visible in this diff) that `genesis-server`'s first few seconds of life on a fresh install need the ceiling already in place.

No ERROR-level pattern violations found (nohup, `/tmp/` usage, `git add -A`/force-push, bare `except`, mutable defaults, raw SQL, etc. — none present in this diff). No missing test coverage — the diff adds tests alongside the behavior change. Doc updates (`CURRENT.md`, `memory-resilience.md`, lib header, CHANGELOG) are consistent with the code change.




---

WingedGuardian:
@codex review

PR #1422, opened 2026-08-21
fix(ego): deadlock when approval gate disabled + truthful cycle-liveness on the dashboard

WingedGuardian:
## Problem

Both egos silently stopped cycling for 3 days, while the dashboard reported them **healthy / "ego active"** the entire time. Two independent bugs:

1. **Deadlock (liveness).** The user set `manual_approval_required=false` (a deliberate config choice). The gate (`ensure_approval`) auto-approves when off and creates no new row — but a *separate* pre-flight (`_approval_pending`, and the shared `find_site_pending`) blocked whenever a pending `autonomous_cli_fallback` row existed, **without consulting the flag**. A leftover row raised earlier (while the gate was on) parked both egos forever (rows created with no timeout, never expired). Multiple restarts did not clear it.

2. **The system lied about it (observability).** `egoSemantic()` and the cadence card derived "healthy" from liveness *proxies* — `is_running` (the consumer-loop task, alive while deadlocked), `next_fire_at` (APScheduler reschedules on every gated no-op), and `consecutive_failures` (a gated cycle isn't a failure). None read the honest signal (`job_health.last_success`, advanced only on a real completed cycle).

## Fix

**Part A — deadlock** (`ego/cadence.py`, `autonomy/approval_gate.py`)
- `_approval_pending` / `find_site_pending`: report not-gated when `manual_approval_required` is off (fail-safe: any policy-read error keeps the gate-on scan). The gate-ON path is byte-for-byte unchanged; the default stays on; nothing auto-approves when the gate is on.
- `ensure_approval` gate-off branch cancels leftover pending rows for *that exact site* (`action_type` + `subsystem` + `policy_id`) so a stale "action required" clears on the next dispatch. TOCTOU-safe (guarded `WHERE status='pending'`); never touches email-gate / contributor-issue rows.

**Part B — truthful liveness** (`ego/liveness.py` new, `dashboard/routes/ego.py`, `dashboard/webui/js/dashboard.js`, `awareness/loop.py`, `db/crud/observations.py`)
- New pure helper computes "stalled" from the `job_health.last_success` gap, **conservatively**: only past `max(4x the current interval, 12h floor)`, so a legitimate adaptive backoff (up to 72h) or a quiet-hours lull never false-flags. `gated`/`paused`/fresh-install are never stalls.
- Dashboard status gains `last_success_at` / `stalled` / `stall_reason`; `gated` is now policy-aware. `egoSemantic()` renders `stalled → error`, `gated → needs-action`, before the healthy fall-through (forward-compatible with an old server).
- New hourly `_check_ego_liveness` awareness posture check: one self-superseding `high` `ego_alert` per ego, auto-resolving when a cycle lands (mirrors `_check_infra_protection_posture`).

## Review
- `genesis-security-reviewer` on Part A: gate-ON path provably unchanged, cancel scoped + TOCTOU-safe, no dispatch-without-approval path. Ready-to-merge: Yes.
- `genesis-architect` on Part B: false-red safety verified (attacked restart-interval-reset, degenerate intervals, clock skew, quiet-hours — all fail toward not-stalled). One SHOULD-FIX (register `ego_alert` TTL) + logging NOTEs applied.

## Testing
- RED-first for every fix; pure helper (7) + alert state machine (6) + gate tests (3) + `_approval_pending` policy-awareness (2); full `test_autonomous_dispatch`, `test_cadence`, `test_awareness/`, `test_ego_routes`, `test_observations` green.
- **Live E2E:** the stale rows were data-repaired mid-session; both egos immediately resumed and completed real cycles (`total_runs` 90→92 / 88→89, `last_success` advanced from Aug-18 to now). Confirms the root cause and that the mechanism fix prevents recurrence.

## Deploy
Runtime code — `git pull` + server restart (the awareness check + endpoint land on restart; `dashboard.js` at next page load). Additive-only; no migration. The restart's boot-first-fire re-verifies egos cycle.

Ledger: af76e90dd0fe4c6bb9f81f396141f55b

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,20 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Ego cycles no longer deadlock when the approval gate is disabled, and the
+  dashboard stops reporting a stalled ego as healthy.** With
+  `manual_approval_required` set to false, a leftover pending approval row
+  (raised earlier while the gate was on) kept blocking both egos' pre-flight
+  check forever — cycles silently stopped while every status surface still
+  showed "ego active". The pre-flight now honors the gate-off setting (and the
+  gate clears the stale row on the next dispatch), so cycles resume
+  immediately. Separately, the dashboard and a new hourly liveness check now
+  read the ego's last *completed* cycle (not the loop-alive flag), so a stalled
+  ego reads "stalled" / "waiting on approval" instead of green — with a
+  conservative threshold that never false-flags a legitimate slow cadence or
+  quiet-hours lull. The mandatory approval gate itself is unchanged (default
+  stays on; nothing auto-approves when it is on).
+
 - **Fresh container installs now get OOM/fork-wedge protection.**
   `scripts/install.sh` (the fresh-container path) never applied the
   memory-resilience provisioning (systemd-oomd pressure-kill, swap invariant,
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -878,6 +878,12 @@ verified: ca875c4b 2026-07-24
   deployed_commit via `~/.genesis/host_gateway_state.json`; collectors in
   `observability/snapshots/deploy_health.py`), `high` on any drift, `critical`
   only sustained (≥7d AND ≥20 commits, or a missing unit alerted >24h).
+  Also (hourly) ego cycle liveness (`_check_ego_liveness`, `ego/liveness.py`): an
+  ego with no COMPLETED cycle past a conservative multiple of its current
+  interval (the `job_health.last_success` gap — never the `is_running`/heartbeat/
+  `next_fire_at` proxies that stay green through a deadlock) raises one
+  self-superseding `high` `ego_alert` per ego, auto-resolving when a cycle lands;
+  `gated`/`paused`/fresh-install are never stalls.
   Also per-tick (WS-2 M10) the SINGLE designated `alert_events` writer:
   `_persist_health_alerts` recomputes the firing set via the pure
   `mcp/health/errors.py::_compute_alerts()` and reconciles a durable open-set
--- a/src/genesis/autonomy/approval_gate.py
+++ b/src/genesis/autonomy/approval_gate.py
@@ -635,7 +635,18 @@ class AutonomousCliApprovalGate:
         Does NOT weaken the gate: the default stays True; with the gate on the
         pending-row scan below is unchanged.
         """
-        if not self._policy().manual_approval_required:
+        try:
+            gate_off = not self._policy().manual_approval_required
+        except Exception:
+            # Fail-safe toward the mandatory gate: a policy-read error keeps the
+            # gate-on pending-row scan (never silently skips a real block). Log
+            # it so a persistent policy misconfiguration is not invisible.
+            logger.debug(
+                "find_site_pending: policy read failed; keeping gate-on scan",
+                exc_info=True,
+            )
+            gate_off = False
+        if gate_off:
             return None
         pending = await self._approval_manager.get_pending()
         for row in pending:
--- a/src/genesis/autonomy/approval_gate.py
+++ b/src/genesis/autonomy/approval_gate.py
@@ -216,6 +216,18 @@ class AutonomousCliApprovalGate:
             # and must stay True — never flip the default, and never add a code
             # path that reaches this branch implicitly. See AutonomousCliPolicy
             # (cli_policy.py) and CC memory `cli_gate_not_negotiable`.
+            #
+            # Clear any LEFTOVER pending row for THIS exact site (raised while
+            # the gate was on): with the gate off it will never be delivered to
+            # or resolved by the user, and the pre-flights (`_approval_pending`,
+            # `find_site_pending`) now ignore it — so without this it would sit
+            # pending forever, showing a stale "action required" on the
+            # dashboard. Scoped strictly to this action_type + (subsystem,
+            # policy_id); never touches email-gate / contributor-issue rows.
+            # Best-effort — a cleanup failure must never block dispatch.
+            await self._cancel_stale_site_approvals(
+                subsystem=subsystem, policy_id=policy_id, action_type=action_type,
+            )
             return ("approved", None, "manual approval disabled by config")
 
         # Serialize per (subsystem, policy_id) — prevents duplicate rows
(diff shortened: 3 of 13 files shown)


WingedGuardian:
@codex review

WingedGuardian:
**Structural first-pass review** (patterns/structure only — not a logic review; PR already notes `genesis-security-reviewer` + `genesis-architect` passes).

## ERROR
None found. No `nohup`, no direct `task_states` writes, no `/tmp/` usage, no `git add -A`/force-push patterns, no bare `except:`, no mutable default args, no raw SQL against `genesis.db` outside test fixtures (all new prod-path reads/writes go through `genesis.db.crud.{ego,job_health,observations}` as required), no `os.killpg` usage.

## WARNING
None. Import grouping is stdlib/third-party/local throughout; no star imports. The repeated pattern of importing `genesis.*` modules inside function bodies (`approval_gate.find_site_pending`, `awareness/loop._check_ego_liveness`, `dashboard/routes/ego._cadence_snapshot`, `ego/cadence._approval_pending`) is consistent with this file's existing lazy-import convention (used to sidestep import-time circularity between `autonomy`/`ego`/`awareness`/`dashboard`), not a new circularity introduced by this diff.

Coverage: the new `src/genesis/ego/liveness.py` module has a corresponding `tests/ego/test_liveness.py` (7 cases), and the new `_check_ego_liveness` alert path has `tests/test_awareness/test_ego_liveness_check.py` (6 cases, including idempotency + recovery). `_approval_pending`/`find_site_pending`/`ensure_approval` policy-awareness changes are covered in `tests/ego/test_cadence.py` and `tests/test_autonomy/test_autonomous_dispatch.py`. Verified all referenced CRUD functions (`job_health_crud.get_job_last_success`, `observations.resolve_by_source_and_type`, `observations.supersede_except_hash`, `ego_crud.has_pending_cli_approval`) and manager attributes (`current_interval_minutes`, `source_tag`, `is_paused`) exist with matching signatures — no dangling references.

## NOTE
- `src/genesis/dashboard/routes/ego.py`: the two new `except Exception:` blocks each do a local `import logging` — matches this file's existing deferred-import style, but could be hoisted to one shared import if this function grows further. Not worth a separate change here.
- New public functions/classes (`compute_ego_liveness`, `EgoLiveness`, `stall_threshold_minutes`, `_check_ego_liveness`) all have docstrings; `docs/architecture/CURRENT.md` and `CHANGELOG.md` updated to match behavior change.
- Commit/PR title carries a proper `fix(ego):` conventional prefix.

---

---

WingedGuardian:
## Structural review

No ERROR-level pattern violations. Code is well-integrated: all cross-module calls checked against actual signatures (`job_health_crud.get_job_last_success`, `ego_crud.has_pending_cli_approval`, `observations.create/resolve_by_source_and_type/supersede_except_hash`) and all match. No bare `except:`, no mutable defaults, no raw SQL, no `nohup`/`/tmp/`/`git add -A` patterns. Import hygiene is clean (stdlib/third-party/local grouping correct in new files; `hashlib`/`uuid`/`logging` already imported at the top of `awareness/loop.py`, no new unguarded imports). New public functions (`compute_ego_liveness`, `stall_threshold_minutes`, `EgoLiveness`) and the new `ego/liveness.py` module are docstring'd. Commit/PR prefix convention followed (`fix(ego): ...`).

**WARNING — inconsistent fix coverage on a sibling endpoint.** `src/genesis/dashboard/routes/ego.py`'s `_cadence_snapshot()` (used by `/api/genesis/ego/status`) is correctly made policy-aware in this PR — `gated` now checks `manual_approval_required` before trusting `has_pending_cli_approval`. But the separate `ego_cadence()` route (`/api/genesis/ego/cadence`, ~line 296) was **not** updated and still sets `gated = await ego_crud.has_pending_cli_approval(...)` unconditionally, with no gate-off check. This endpoint backs `egoModal.cadence`, which `dashboard.js`'s `egoSemantic()` falls back to when per-ego data isn't yet available (`!ue && cadence?.available`) — so the same "phantom gated on a leftover row" symptom this PR fixes for the primary status card can still surface through the modal's fallback path. Given the PR's stated goal is exactly "the dashboard stops reporting a stalled/gated ego wrongly," this looks like a missed call site rather than an intentional scope cut. Confirmed via `tests/test_dashboard/test_ego_routes.py::TestEgoCadence` — those tests still assert the old, non-policy-aware `gated` semantics for that endpoint and were not touched by this PR.

**NOTE — no route-level test for the new `_cadence_snapshot` fields.** The pure liveness logic is thoroughly tested (`tests/ego/test_liveness.py`, `tests/test_awareness/test_ego_liveness_check.py`), but no test in `tests/test_dashboard/test_ego_routes.py` asserts that `/api/genesis/ego/status` now returns `stalled` / `last_success_at` / `stall_reason` per ego. Worth a quick route-level test given this is the field `dashboard.js` reads to render the "stalled" health state.

**NOTE — minor duplication.** In `dashboard/routes/ego.py::_cadence_snapshot`, both new `except Exception:` blocks do `import logging; logging.getLogger(__name__).debug(...)` separately rather than reusing a single module-level or function-level logger. Harmless (matches this file's existing per-function-local-import style), just a small dedup opportunity.

## PII/Secrets/Wording scan

CLEAN. No secrets, private IPs, credentials, or connection strings in the diff. PR title/body/CHANGELOG describe the deadlock and observability bug in general terms (leftover pending-approval row, liveness proxies vs. `job_health.last_success`) with no private infrastructure names, no exploit-level detail, and no "leak"/"incident" framing — this is a legitimate bug-fix narrative, not a disclosure.



---

WingedGuardian:
Thanks — dispositioned all 6 P2s (commit 9eec4170):

**Fixed (false-red/correctness):**
- **Global pause** (loop.py:892): stall detection now suppresses on a global `GenesisRuntime.paused`, not just the ego's own pause. Both dashboard + awareness.
- **Quiet-hours suppress window** (liveness.py:31): new `quiet_suppress_floor_minutes` raises the threshold above a configured `suppress` window >12h; 'floor' mode still ticks so needs no floor.
- **Resolve on disabled ego** (loop.py:878): a standing `ego_alert` now clears when the ego is disabled (manager gone) instead of leaking open.
- **Resolution note** (loop.py:909): now accurate (cycle-completed OR now gated/paused).

**Documented-accept:** *never-first-cycle* (liveness.py:83) — `None` last_success → not-stalled is the deliberate fresh-install-safety choice; distinguishing wedged-before-first-cycle from a genuinely-fresh box needs an uptime reference and reintroduces the false-red risk this PR exists to avoid. The reported deadlock always had a prior last_success.

**Deferred (same class, distinct subsystem):** *unpark Sentinel on gate disable* (approval_gate.py:230) — real sibling; `sentinel/dispatcher.py` reads `_require_approval` once at init and never converges a parked sentinel to approved when the gate is off. Tracked as a follow-up with its own sentinel tests + security review, rather than expanding this PR's blast radius into the sentinel state machine.

@codex review

WingedGuardian:
**Converged the liveness signal (commit bade974d) — the whack-a-mole is resolved architecturally.**

Rather than keep enumerating suppression conditions (round 1: global-pause/quiet/disabled; round 2: idle/quiet-enablement), the signal now excludes them **by construction**:

`stalled = last INTENT to cycle (_last_proactive_fire_at) leads last COMPLETED cycle (job_health.last_success) by > max(4×interval, 3h), and not gated.`

`_last_proactive_fire_at` is written only *after* `_on_tick` clears every cadence gate (setup/floor/paused/global-pause/circuit/**user-active**/quiet-hours) and pushes a signal — so every legitimate non-running reason prevents an intent from being recorded, the lag never opens, and there is no per-condition list to keep in sync. `gated` is excluded explicitly. Net −95 lines; the prior round-1/2 P2s are moot (that code is gone).

The deadlock is still detected: `_on_tick` advances intent independently of the gated consumer (self-expiring signals keep the queue from saturating).

Reviewed by genesis-architect: key-match verified (no false-green), false-red safety verified by construction. **Documented coverage boundary:** this deliberately does NOT cover *total* scheduler cessation (both timestamps freeze) — that's a distinct failure mode (scheduler death; main also shows healthy there today, so no regression) tracked as a follow-up for the complementary `ego_heartbeat`-staleness monitor.

@codex review

PR #1423, opened 2026-08-21
fix(security): origin-police every observations-table content consumer (read-side injection close-out)

WingedGuardian:
## What & why

Follow-on to the user-model poisoning gate (#1407). That PR gated the two
privileged-**write** consumers of external-origin observations; a security
review found the injection surface was broader — external-origin sessions (the
inbox judge etc.) write observations stamped `external_untrusted`, and many
**read** consumers surfaced that content unfiltered into LLM context. This
closes the read side.

Every consumer of the `observations` table's content is now classified and
enforced:

- **Hard exclusion** (external dropped in SQL, NULL kept) on the surfaces where
  wrapping isn't enough: the reflection pipeline (evidence, recent-obs funnel,
  calibration, perception's reflection context) and the always-loaded L1
  `essential_knowledge.md`. This severs the laundering chain where an external
  digest reached the deep-reflection prompt and re-emerged as a trusted-origin
  `user_model_delta` (deltas are stamped by session-window origin, not content
  lineage).
- **Wrapping** in `<external-content>` markers on the ego / sentinel / guardian
  / surplus context builders and the `observation_query` MCP tool.
- **Write-side authz**: `observation_write` refuses `user_model_delta` from
  external sessions (denylist), and constrains the metadata columns that render
  *outside* the wrapper (`source`/`type`/`category`/`priority`) to an
  identifier charset — so injection text can't ride in a sibling column.
- **Suppression guard**: `observation_resolve` refuses an external session
  resolving a non-external row (hiding internal alerts is the dual of forgery).
- **Channel-aware `task_detected` stamping**: terminal + Telegram (owner-
  attended) stamp `owner`; gateway/ambient channels stay unstamped (visible,
  never dispatch-authorized).
- **Discovery-based guardrail** (`test_observation_surface_coverage.py`):
  AST-enumerates every raw-SQL and `observations.query` consumer over
  git-tracked sources and fails CI on any lacking a classified verdict, with
  policy tokens asserted inside each function's own AST.

## Design notes

- New crud `exclude_origin_class` filter drops only the named class **before**
  the LIMIT while **keeping NULL** — deliberately the opposite NULL semantics
  from the privileged-read `origin_class_in` (all pre-provenance history is
  NULL; measured 214/214 in-window rows, so excluding NULL would wipe the L1
  feed). Passing both raises.
- Marker strip is case/whitespace-insensitive and normalizes zero-width +
  confusable-hyphen codepoints, so a forged closing tag built from look-alikes
  can't escape the wrapper.
- No schema change (`origin_class` exists since migration 0057); no index added
  (the OR-NULL negative predicate isn't index-servable; reads are recency+LIMIT
  scans).

## Honest residuals (not closed here)

- Reflection can still recall external-stamped memory-store content — it arrives
  **wrapped**, and a deep-reflection prompt rule forbids deriving
  `user_model_updates` from `<external-content>` material. Full closure is the
  provisional-provenance-tier design (tracked separately).
- The remaining source/type-pinned `SAFE_INTERNAL` reads (recon markers,
  version signals, retrospective calibration) are numeric/count parses — content
  is never rendered as prose into a prompt.

## Testing

- New/changed targeted suites green (crud filter, reflection/L1 exclusion
  verify-RED-first, wrap spot-checks, write/resolve authz integration against a
  temp MCP DB, channel-stamp table, guardrail).
- Full suite: 18895 passed. The only failures are 5 pre-existing full-suite
  ordering flakes (all pass standalone; none touch changed files) + the
  install-local `credential_bridge` env-leak tests — both tracked in a separate
  follow-up, neither introduced here.

## Review

Adversarial review by `genesis-security-reviewer` + `genesis-architect`
(multiple rounds incl. a fresh-context exhaustive audit that surfaced — and
this PR fixes — a sibling-column injection class, four mis-classified prompt
surfaces, and the ungated resolve path). Independent-model (Codex) review to
follow on this PR.

Follow-up: 0a32f6b2549b48efb4a2f7eb5f79a23c

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,44 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ## [Unreleased]
 
+### Security
+
+- **External-origin observation content is now origin-policed at every
+  surfacing point (read-side memory-provenance work).** Follow-on to the
+  user-model poisoning gate: the observations table's content consumers are
+  now all classified and enforced. The reflection pipeline (deep-reflection
+  evidence, recent-observations funnel, calibration, perception's reflection
+  context) and the always-loaded L1 knowledge file HARD-EXCLUDE
+  `external_untrusted` rows in SQL (new crud `exclude_origin_class` filter +
+  shared `EXCLUDE_EXTERNAL_ORIGIN_SQL` fragment; NULL rows kept — unstamped
+  internal/legacy writers keep flowing, deliberately the opposite NULL
+  semantics from the privileged-read `origin_class_in`). Ego, sentinel,
+  guardian-briefing, and surplus context builders plus the `observation_query`
+  MCP tool WRAP external-origin content in `<external-content>` markers
+  (`provenance.wrap_if_external`; truncate-then-wrap; embedded forged closing
+  tags are stripped so the wrapper can't be escaped). This closes the
+  laundering path where externally-written digests reached the deep-reflection
+  prompt unfiltered and re-emerged as trusted-origin user-model deltas
+  (delta stamping is session-window-based); a deep-reflection prompt rule
+  additionally forbids deriving user-model updates from `<external-content>`
+  material. A discovery-based guardrail
+  (`tests/test_security/test_observation_surface_coverage.py`) AST-enumerates
+  both raw SQL and crud-query consumers (docstrings skipped, regex evasion
+  net, git-tracked sources only) and fails CI on any consumer without a
+  classified verdict, with policy tokens asserted inside each function's AST.
+- **`observation_write` refuses privileged observation types from
+  external-origin sessions.** DENYLIST (`user_model_delta`): external
+  sessions keep their legitimate digest/finding types; the delta type's only
+  legitimate writers are server-side (reflection output, perception writer —
+  no session env), so the refusal closes the injection→user-model write path
+  at zero functional cost. Refusals return a `refused:` string and log a
+  warning — never raise.
+- **`task_detected` conversation producers stamp channel-aware origin.**
+  Terminal + Telegram (owner-attended) stamp `owner`, so legitimate task
+  requests will dispatch once the autonomy pickup path is wired;
+  gateway/ambient channels (web gateway, WhatsApp, voice) deliberately stay
+  unstamped — visible but never dispatch-authorized.
+
 ### Changed
 
 - **Executor Gate 2 (`17_executor_review`) leads with paid DeepSeek V4-pro.**
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -220,7 +220,7 @@ any task bigger than an LLM call.
 ```yaml subsystem-map
 entry: execution-cc
 modules: [cc]
-verified: d71d1d39 2026-08-18
+verified: 6830dd78 2026-08-21
 ```
 
 - **Subagent-spawn lockdown — one source of truth across the restricted sessions**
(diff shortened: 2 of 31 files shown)


WingedGuardian:
@codex review

WingedGuardian:

**Structural review: 2 WARNING, 1 NOTE** — no ERRORs.

### WARNING — Missing coverage

1. **`src/genesis/ego/context.py::EgoContextBuilder._observations_section`** — this PR adds `wrap_if_external` here (wrapping external-origin content in `<external-content>` markers), and the mechanism is enforced at the AST-token level by `test_observation_surface_coverage.py::test_verdict_enforcement_tokens_present`. But unlike the sibling surfaces this PR touches — `GenesisEgoContextBuilder` (`tests/test_security/test_observation_surface_wrap.py::test_genesis_ego_observations_section_wraps_external`), sentinel, guardian briefing, and world snapshot — there's no behavioral test that plants an `external_untrusted` row and asserts the *combined* `EgoContextBuilder`'s rendered output actually contains the wrapper markers. `tests/ego/test_context.py::test_observations_section` only covers the pre-existing unwrapped-content path.

2. **`src/genesis/ego/user_context.py::UserEgoContextBuilder._recurring_patterns_section`** — the new `any_external` GROUP-BY taint (`MAX(CASE WHEN origin_class = ? THEN 1 ELSE 0 END)`, wrapping the sample when any row in a repeating-pattern group is external) has no test exercising it. `tests/ego/test_user_context.py`'s existing `_recurring_patterns_section` tests (threshold, resolved-exclusion) don't plant an external-origin row inside a 3+ pattern group, so the new taint/wrap path is unverified beyond the AST-token check.

### NOTE — Documentation

3. **`src/genesis/mcp/memory/observations.py::observation_write`** — docstring still reads `"Write processed reflection/observation. Returns observation_id."`, but the function now also returns a `"refused: ..."` string on the two new authz-refusal paths (denied type, metadata-charset violation). Worth a one-line docstring update so callers reading just the docstring know to check for the `refused:` prefix.

Everything else structural looks solid: import ordering is clean throughout, no mutable defaults, no bare `except:`, all 6 commits carry proper conventional prefixes, and the new discovery-based coverage guardrail (`test_observation_surface_coverage.py`) is a good forcing function. The raw-SQL-outside-crud pattern touched in `ego/*`, `sentinel/context.py`, `guardian/briefing.py`, `world_snapshot.py`, and `essential_knowledge.py` predates this PR (only a column/filter is added to existing queries), so not flagged as new.

---

**PII/Secrets/Wording scan: CLEAN** — no secrets, private IPs, credentials, or personal paths in the added lines; no PII-leak/incident language in the PR title, body, or CHANGELOG entry. The PR description is a normal, appropriately-scoped account of defensive security work (prompt-injection/provenance hardening), not a leak disclosure.



---

WingedGuardian:
**First-pass structural review** (patterns/structure only — not a logic review; ruff/CI-caught issues skipped)

Scope: 31 changed files (~1749/-54), all under `src/genesis/{autonomy,cc,db,ego,guardian,mcp,memory,perception,reflection,security,sentinel,surplus}` and `tests/`. Read the full diff.

## ERROR
None. No `nohup`, no direct `task_states` inserts, no `/tmp/` usage, no `git add .`/`-A`, no force-push/reset patterns, no bare `except:`, no mutable default args, no `os.killpg` misuse.

## WARNING

1. **Raw SQL vs. crud-layer inconsistency for the new origin filter.** This PR adds `exclude_origin_class` to the crud choke point (`db/crud/observations.py::query`), and `perception/context.py`, `reflection/context_gatherer.py`, and `cc/reflection_bridge/_prompts.py` correctly route through it. But `ego/genesis_context.py::_execution_outcomes_section`, `ego/user_context.py::_execution_outcomes_section`/`_genesis_escalations_section`, and `essential_knowledge.py` instead hand-roll raw `db.execute()` with the imported `EXCLUDE_EXTERNAL_ORIGIN_SQL`/`_PARAMS` fragments (pre-existing raw-SQL sites, not newly introduced here, but the fix extends rather than consolidates them). Worth a note on why these can't route through `observations.query()` now that it supports the needed filter — some (GROUP BY + CASE ordering in `_recurring_patterns_section`) clearly can't, but the two `_execution_outcomes_section` copies and `_genesis_escalations_section` look like plain filtered selects that could.

2. **Missing behavioral test for two WRAPPED sites.** `tests/test_security/test_observation_surface_wrap.py` adds live-DB spot checks for `genesis_context.py::_observations_section`, `sentinel/context.py`, `guardian/briefing.py`, `world_snapshot.py`, and `user_context.py::_genesis_escalations_section` — but not for `ego/context.py::_observations_section` (combined-ego) or `ego/user_context.py::_recurring_patterns_section` (the new `any_external` `MAX(CASE WHEN origin_class = ?...)` GROUP-BY-taint logic). Both are only checked by the AST-level `test_verdict_enforcement_tokens_present` (confirms `wrap_if_external` appears somewhere in the function body, not that it fires correctly against real external-origin rows). `_recurring_patterns_section` in particular has non-trivial new aggregation logic that a live-data test would be well-suited to catch a param-ordering or off-by-one bug in.

3. **Repeated function-local imports with no apparent circular-import need.** `from genesis.memory.provenance import wrap_if_external` and `from genesis.db.crud.observations import EXCLUDE_EXTERNAL_ORIGIN_PARAMS/SQL` are imported inside function bodies across `ego/context.py`, `ego/genesis_context.py` (×2), `ego/user_context.py` (×2), `ego/world_snapshot.py`, `guardian/briefing.py`, `sentinel/context.py`, `surplus/executor.py`, and `mcp/memory/observations.py`. Checked `genesis.memory.provenance`'s own imports (only `genesis.security.sanitizer`) and none of the calling modules already import it — no evidence of a real cycle. If there isn't one, hoisting to module-level imports would match the rest of each file's style and avoid repeated per-call import overhead.

## NOTE

- `cc/conversation.py::_task_origin_for_channel` locally imports `ORIGIN_OWNER` (a single string constant) inside the function — minor, could be module-level for the same reason as above.
- Docstrings are consistently present and useful on all new/changed public and private surfaces (`wrap_if_external`, `_observation_type_permitted`, `_external_metadata_violation`, `query()`'s updated origin-filter contract, MCP tool docstrings).
- `docs/architecture/CURRENT.md` and `CHANGELOG.md` are both updated to reflect the closure of the previously-tracked PARTIAL/OPEN item — good hygiene.
- All 6 commits carry conventional prefixes (`feat`/`fix`/`test`).

## Notable strengths (not required by the checklist, but worth calling out)
The `test_observation_surface_coverage.py` AST-based discovery guardrail (raw-SQL + crud-query-caller enumeration over git-tracked sources, with a regex evasion net and per-function token enforcement) is a strong forcing function against silent regressions on new consumers. The deliberate, well-documented asymmetry between `origin_class_in` (NULL-excluded, privileged reads) and `exclude_origin_class` (NULL-kept, surfacing reads) plus the marker-forgery hardening (case/whitespace-insensitive, zero-width + confusable-hyphen normalization) reflect real adversarial thinking, not just pattern-matching.



---

WingedGuardian:
Superseded by #1431 (merged as ce25d324, deployed 2026-08-23). The v1 per-read classification approach was replaced by write-boundary provenance (origin_class stamped at the crud chokepoint, fail-closed NULL) + laundering-critical read exclusions. All 10 findings from this PR's review are structurally addressed or documented residuals in the v2 design; see #1431's PR body.

PR #1425, opened 2026-08-22
feat(hooks): stamp Genesis-Session provenance trailer on commits

WingedGuardian:
## What

`prepare-commit-msg` now appends a second opaque trailer, `Genesis-Session: <id8>`, alongside the existing `Install:` trailer. This makes a merged PR's commits traceable back to the working session that built them — turning provenance into a lookup instead of forensics.

## Why

Commits currently carry only an `Install:` fingerprint; nothing links a commit/PR to the session that produced it. Reconstructing that after the fact is effectively impossible (time-correlation is defeated by long-lived sessions). A single opaque trailer on the natural chokepoint (every commit) closes the gap going forward.

## How it stays safe on a public repo

- The value is the **first 8 hex** of the session id, resolved from env at commit time (`GENESIS_SESSION_ID` for dispatched sessions → `CLAUDE_CODE_SESSION_ID` for foreground → legacy `CLAUDE_SESSION_ID`). Git hooks get no stdin, so env is the only route.
- It is **hard-clamped to `^[0-9a-fA-F]{8}`** via `sed`, so a hostile/garbled env value cannot inject commit-message content (adversarially fuzzed: shell metachars, embedded newlines, `--trailer`-looking values → all reduced to a bare token or skipped).
- A bare hex token matches **none** of the leak gates (`commit-msg` IP/ULA classes, CI `leak-detector`, contribution sanitizer) — no path/IP/email/hostname/name shape.
- The trailer maps back to `cc_sessions.topic` **only privately, on the authoring install**. The human-readable session name is never published — only the opaque id.

## Behavior

- Each trailer is independent and best-effort: a missing id skips only its own trailer and **never blocks a commit**; merge commits stay unstamped; idempotent on `--amend`.
- **Non-retroactive by design** — existing commits keep only `Install:`.

## Deploy path

Hooks land via `scripts/hooks/sync-hooks.sh` (already wired into bootstrap/update) — takes effect on the next commit after pull. `.genesis-hook-versions` records the new hook hash so `sync-hooks.sh` auto-updates stale installs.

## Tests

`tests/test_scripts/test_prepare_commit_msg.py` — 8 new cases (foreground stamp, env precedence, idempotency, absent-env omission, non-hex skip, both-trailers coexist, merge unstamped, bare-hex shape). 14/14 pass. Adversarial review: 1 BLOCKER (a real session id used as a test fixture) found and fixed to a synthetic id; injection/idempotency/non-blocking/hook-versions verified.


WingedGuardian:

Notes (informational, no action needed):
- `scripts/hooks/prepare-commit-msg`: session-id resolution (`GENESIS_SESSION_ID` → `CLAUDE_CODE_SESSION_ID` → `CLAUDE_SESSION_ID`), 8-hex clamp via `sed`, and merge-commit skip all match the PR description and are covered by the new tests in `tests/test_scripts/test_prepare_commit_msg.py`.
- The in-code comment referencing `sqlite3 ~/genesis/data/genesis.db ...` is a documentation path pattern, not a live credential/path leak — consistent with known false positives for this repo.
- The synthetic session id used in test fixtures (`5e551011-0000-...`) is clearly non-real; PR body's mention of catching a real-id fixture during adversarial review pre-merge is normal process narration, not a leak disclosure.
- No new `.py` source files added, so no missing-test-file warning applies; existing test file was extended in-place with matching coverage for the new trailer.



---

WingedGuardian:

Scope: `scripts/hooks/prepare-commit-msg`, `tests/test_scripts/test_prepare_commit_msg.py`, `.genesis-hook-versions` (bash hook + test coverage + hook-version bump, 3 files).

Checked and clear:
- **Pattern violations (ERROR):** none — no `nohup`, no direct `task_states` writes, no `/tmp/` usage, no `git add .`/`-A`, no force-push/`reset --hard`, no `pip install -e` worktree path, no bare `except:`, no mutable default args (`env_extra: dict[str, str] | None = None` in the test helper is correctly `None`-defaulted), no `os.killpg`, no raw SQL against `genesis.db` (the `sqlite3 ...` line in the header comment is documentation for a manual lookup, not application code).
- **Import hygiene:** test file imports (`json`, `subprocess`, `pathlib.Path`) are stdlib-only and correctly ordered; no star imports, no circularity.
- **Coverage:** the two resolution functions (`_resolve_install_id` rename, new `_resolve_session_id`) are covered by the 8 new cases — env precedence, idempotency on rerun, non-hex rejection, merge-commit exclusion, and value-shape assertion. Good.
- **Docs:** both trailer resolution orders are documented inline above their functions; no undocumented behavior change.
- **Genesis conventions:** no heavy imports in the hook (stdlib bash only), no `$CLAUDE_PROJECT_DIR` misuse, no `IntervalTrigger`, commit carries the `feat(hooks):` prefix.

One non-blocking NOTE: `import re` is done inside `test_session_trailer_value_is_bare_hex` rather than at module scope alongside the other stdlib imports — harmless (test-local, stdlib), just inconsistent with the rest of the file's import style.




---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
**Disposition — Codex P2 (empty-message abort):** consciously accepted, not fixed.

The stripspace guard that would preserve git's empty-message abort was tried and reverted: `prepare-commit-msg` runs on the comment-only template *before* the editor opens, so guarding on emptiness there strips trailers from every interactive editor commit (a worse regression). The hook cannot distinguish "comment-only template that will get content" from "genuinely empty" at pre-editor time.

Accepted because: (1) it matches the long-standing `Install:` trailer behavior on this same hook; (2) Genesis authors all commits via `-m`/`-F` with real content, so an empty/aborted commit is never produced in practice; (3) the trailer on a rare hand-made empty commit is harmless. The multiline-clamp and prose-safe idempotency P2s were fixed.

WingedGuardian:
@codex review

WingedGuardian:
**Disposition — Codex R4 P2 (multiple pre-existing Genesis-Session trailers):** consciously accepted, not fixed.

`--if-exists replace` normalizes the single-trailer case (the actual bug that was fixed) but only replaces the last of *multiple* pre-existing `Genesis-Session` trailers. That state is unreachable in Genesis's flow: the hook runs `replace` on every commit, so a Genesis-authored commit chain never accumulates two — duplicates could only arise from a manual rebase-squash that concatenates messages, or an externally-imported message, neither of which the provenance feature targets (it identifies the CC session behind Genesis's automated -m/-F commits). A 'strip-all-then-add' normalization would add materially more shell logic (and finding surface) to a hand-rolled hook to cover a case that doesn't occur in practice.

This is the residual edge accepted under the deliberate round-4-then-merge decision; the escalation cap (3 defect-bearing rounds) is well past. The core provenance path is adversarially verified across four rounds.

PR #1401, opened 2026-08-15
feat(ssh): one-click "lobby" door to reattach the whole CC fleet

WingedGuardian:
Adds a single-connection "lobby" SSH door so a client (e.g. after a laptop
reboot) can reattach ALL persistent Claude Code tmux sessions at once, instead of
reopening one connection per slot. The sessions persist server-side in tmux; only
the client dies on reboot — this is a client-entry convenience, no persistence
change.

## Mechanism
`generate-ssh-config.sh` now emits a dedicated `Host <host>-lobby` block **ahead
of** the numeric-slot wildcard `Host <host>-*`. ssh uses the first matching
`RemoteCommand`, so:
- `<host>-lobby` → `tmux -u new-session -A -s lobby` (attach-or-create a stable
  landing session on the **same** tmux socket as the `cc-*` slots). From there,
  `Ctrl-b s` lists and jumps to any live slot.
- `<host>-N` (numeric) → still `cc-slot.sh <host>-N`, unchanged.

`lobby` is not a `cc-N` name, so it consumes no slot and never trips cc-slot.sh’s
session cap (which counts only `^cc-[0-9]+$`). Placing the specific block before
the wildcard is load-bearing: otherwise `<host>-lobby` would fall through to
cc-slot.sh and be rejected as a non-numeric slot.

## Verification
- **Real `ssh -G` resolution** (not just text order), asserted in tests and run
  live: `<host>-lobby` resolves to the tmux RemoteCommand; `<host>-2` still
  resolves to cc-slot.sh.
- Live tmux: a `lobby` session on the shared default socket sees all `cc-*` and
  consumes no slot.
- `tests/test_scripts/test_generate_ssh_config.py` — emission + `ssh -G` resolution
  tests (skipif ssh absent). 7 passed in-file; 21 across `tests/test_scripts`.
  ruff / shellcheck / `bash -n` clean.

## Notes
- No new network exposure: reached only via the existing SSH surface; `lobby` is a
  plain shell with the same trust as any SSH login.
- Adversarially reviewed (genesis-architect): scope clean, ready-to-merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1426, opened 2026-08-22
feat(ssh): one-click "lobby" door — IP-keyed, opens the session picker

WingedGuardian:
One-click reattach to all persistent Claude Code tmux sessions after a client
reboot, over one SSH connection. The sessions persist server-side in tmux; only
the client dies on reboot — this is a client-entry convenience.

## Mechanism
`generate-ssh-config.sh` emits a `Host <host>-lobby` block **ahead of** the
numeric-slot wildcard (ssh uses the first matching `RemoteCommand`) that attaches
a stable `lobby` tmux session on the **same socket** as the `cc-*` slots and opens
straight into the session picker (`choose-tree`). `lobby` is not a `cc-N` name, so
it consumes no slot and never trips cc-slot.sh’s cap; numeric-slot routing is
unchanged.

- **HostName is keyed on the stable Tailscale IP**, not the MagicDNS name, so the
  config keeps resolving even when the client’s DNS/MagicDNS is disrupted (a real
  failure mode on Windows when the NRPT rule drops).
- The lobby `RemoteCommand` sets PATH (an ssh RemoteCommand doesn’t source
  `.bashrc`), mirroring `cc-slot.sh`, so `tmux` resolves even when user-local.

## Verification
- `tests/test_scripts/test_generate_ssh_config.py`: emission + real `ssh -G`
  first-match resolution (ordering, IP-not-MagicDNS, PATH-prefix + choose-tree,
  numeric → cc-slot.sh). 8 pass; `bash -n` + ruff clean.
- Adversarially reviewed (genesis-architect, three passes) + Codex clean on the
  prior iteration.

Supersedes #1401 (this is a clean single-commit re-creation to drop a
documentation-placeholder value from that PR’s intermediate commit history; the
code is byte-identical to #1401’s reviewed head).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,6 +121,20 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **One-click "lobby" terminal door — reattach the whole CC fleet after a client
+  reboot.** `generate-ssh-config.sh` now emits a dedicated `Host <host>-lobby`
+  block (placed ahead of the numeric-slot wildcard, since ssh takes the first
+  matching `RemoteCommand`) that attaches a stable `lobby` tmux session on the
+  same socket as the `cc-*` slots and opens straight into the session picker
+  (`choose-tree`). One reconnect brings every live session back — they persist in
+  tmux on the box, only the client died — pick any slot and jump in (`Ctrl-b s`
+  reopens the picker). All emitted blocks are keyed on the stable Tailscale IP
+  (not the MagicDNS name), so they keep resolving even when the client's DNS is
+  disrupted. `lobby` is not a `cc-N` name, so it never consumes a slot or trips
+  the slot cap, and numeric-slot routing through `cc-slot.sh` is unchanged. Pair
+  with one Windows Terminal shortcut (`wt.exe ssh <host>-lobby`) for a
+  one-double-click return to the whole fleet.
+
 - **Claude Code session exits are recorded.** When a CC session's process exits
   — a clean quit, a crash, or an OS/kill signal — its exit status (with a
   signal-decoded hint) and a tail of the terminal are written to
--- a/scripts/generate-ssh-config.sh
+++ b/scripts/generate-ssh-config.sh
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 # generate-ssh-config.sh — Generate client SSH config for tmux slot access.
 #
-# Auto-detects this machine's Tailscale hostname and outputs an SSH config
-# snippet to paste into ~/.ssh/config on client devices.
+# Auto-detects this machine's Tailscale identity and outputs an SSH config
+# snippet (keyed on the stable Tailscale IP, DNS-resolver-independent) to paste
+# into ~/.ssh/config on client devices.
 #
 # Usage: ./scripts/generate-ssh-config.sh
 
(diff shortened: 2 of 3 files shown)


WingedGuardian:
@codex review

WingedGuardian:
Structural review (first-pass, patterns/structure only — not deep logic review).

**Scope:** `CHANGELOG.md`, `scripts/generate-ssh-config.sh`, `tests/test_scripts/test_generate_ssh_config.py` (new).

## WARNING

- **Dead variable left over from the IP-keying refactor** — `scripts/generate-ssh-config.sh:23,32`. `TS_DNSNAME` is still computed and still gates the script's success (`-z "$TS_DNSNAME"` in the error check), but nothing in the emitted heredoc references it anymore: both prior usages (`Host ${TS_HOSTNAME}-*` and the commented-out `Direct access` block) were switched to `${TS_IP}` in this diff. Functionally harmless right now — its emptiness tracks `TS_HOSTNAME`'s since both derive from the same `DNSName` field — but it's dead code from an incomplete refactor that could mislead a future reader into thinking it still drives an emitted value. Worth removing (or repurposing) in a follow-up.

## Notes (non-blocking)

- No ERRORs found — no `nohup`, no raw SQL, no `/tmp/` usage, no bare `except`, no mutable defaults, no `git add -A`/force-push patterns, nothing autonomy-gate-adjacent.
- Import hygiene in the new test file is clean (stdlib, then `pytest`, correctly grouped).
- New script logic ships with its own test coverage (`tests/test_scripts/test_generate_ssh_config.py`), including real `ssh -G` first-match resolution checks rather than just text-order assertions — good practice for a "which `RemoteCommand` wins" invariant like this one.
- Verified the lobby `RemoteCommand`'s `PATH` construction (`${REMOTE_HOME}/.n/bin:...`) does genuinely mirror `cc-slot.sh`'s `export PATH=...` line, as the inline comment claims.
- Commit/PR title carries the `feat(ssh):` conventional prefix.



---

WingedGuardian:

- No ERROR-level findings (nohup/systemd, `/tmp`, `task_states`, `git add .`/`-A`, force-push/reset, bare `except:`, mutable defaults, raw SQL — none apply; this diff touches only a bash script, `CHANGELOG.md`, and a new test file).
- Import hygiene in `tests/test_scripts/test_generate_ssh_config.py` is correctly grouped (stdlib: `shutil`, `stat`, `subprocess`, `pathlib` → third-party: `pytest`).
- The changed script (`scripts/generate-ssh-config.sh`) has new, non-trivial test coverage (emission tests + real `ssh -G` resolution tests), so no coverage gap.
- No Genesis-convention violations (no hooks, no `$CLAUDE_PROJECT_DIR` misuse, no `IntervalTrigger`); PR title carries a valid conventional prefix (`feat(ssh): ...`).

**PII/Secrets/Wording scan: CLEAN** (one soft note below)
- Code diff: no secrets/tokens, no private IPs in the blocked ranges. The only IP literals in the diff (`192.0.2.5`, `2001:db8::1`) are RFC5737/RFC3849 documentation-range addresses in the test fixture, explicitly commented in the test file as placeholders — not real Tailscale/private infra, and outside the `10.x`/`192.168.x`/`100.64–127.x`/`fd00::/8` blocklist.
- No hardcoded usernamed paths, personal emails, or hardware identifiers introduced.
- **Soft note (wording, not a block):** the PR body says this PR is "a clean single-commit re-creation to drop a documentation-placeholder value from that PR's [#1401] intermediate commit history." This doesn't name a specific secret/leak and reads as ordinary history hygiene, so it doesn't meet the hard-block bar — but if that placeholder value was ever a real credential or IP rather than a doc placeholder, worth confirming #1401's branch/commits aren't still reachable anywhere with that value intact.

No hard blocks or structural errors — no changes requested by this pass.



---

PR #1430, opened 2026-08-22
fix(hooks): block full-suite/whole-directory local pytest runs; fix concurrent-guard quoted-pipe FP

WingedGuardian:
## Problem

`full_suite_guard.py` (a CC PreToolUse hook) exists to discourage running the full test suite locally — CI runs it authoritatively on every push, and a local full run duplicates that work and starves the shared box (it OOM-killed the run mid-suite). But the guard only warned on path-LESS `pytest` and treated any arg containing `/` as "targeted", so a whole-directory run like `pytest tests/test_scripts/` (1285 tests) sailed through silently.

Separately, `concurrent_test_guard.py` matched `pytest` via a raw regex that read a `|` inside a quoted grep pattern as a shell pipe, false-blocking commands like `grep -E 'a|pytest|b' file`.

## What this does

- **full_suite_guard**: warn → **BLOCK** (exit 2) on a pytest run that targets no specific `.py` file and no `-k`/`-m`/`--pyargs` selector (bare `pytest`, `pytest -v`, or a bare directory). Explicit `# full-suite-ok` override for the rare local full run. Detection routes through the quote-aware `shell_parse.analyze`.
- **concurrent_test_guard**: `_command_runs_pytest` delegates to `shell_parse.command_runs_pytest`, fixing the quoted-`|pytest` false-block.
- **shell_parse**: two additive helpers (`is_pytest_invocation`, `command_runs_pytest`); no existing symbol changed.
- Adds the missing `test_full_suite_guard.py`.

## Testing

- 101 hook tests green (new `test_full_suite_guard.py` covers block/allow/override + the quoted-grep case); 330 `shell_parse`-consumer tests green (`git_push_guard` etc. — the additive change is non-regressive); ruff clean.

## Review

Pre-push `genesis-architect` found and I fixed a BLOCKER (a sibling test still calling the removed `_is_full_suite` → 12 CI failures), a `.py`-substring directory bypass, and a `--pyargs` false-block.

### DIFF
--- a/scripts/hooks/concurrent_test_guard.py
+++ b/scripts/hooks/concurrent_test_guard.py
@@ -21,22 +21,22 @@ invoked with `-m pytest`.
 from __future__ import annotations
 
 import os
-import re
 import sys
 
 # Self-locate so hook_input resolves whether run as a script or imported (tests).
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 from hook_input import field, read_payload  # noqa: E402
+from shell_parse import command_runs_pytest  # noqa: E402
 
 
 def _command_runs_pytest(cmd: str) -> bool:
-    """Check if a shell command will invoke pytest (any variant)."""
-    # Match pytest as a standalone command or as a module invocation.
-    # Covers: pytest, python -m pytest, python3 -m pytest, chained commands,
-    # and env-var prefixed invocations (PYTHONPATH=src pytest ...).
-    # Does NOT match: "grep pytest", "cat pytest.ini", etc. — requires word boundary.
-    # The (?:\w+=\S*\s+)* handles env var assignments before the command.
-    return bool(re.search(r"(?:^|&&|;|\|)\s*(?:\w+=\S*\s+)*(?:python3?\s+-m\s+)?pytest\b", cmd))
+    """Whether a shell command will invoke pytest (any variant), quote-aware.
+
+    Delegates to the canonical shell parser so a ``pytest`` mentioned only inside a
+    quoted argument — e.g. ``grep 'a|pytest' f`` — is NOT treated as a run (the old
+    raw-regex scan matched the ``|pytest`` inside such a grep pattern and false-blocked).
+    """
+    return command_runs_pytest(cmd)
 
 
 def _argv_is_pytest(argv: list[str]) -> bool:
--- a/scripts/hooks/full_suite_guard.py
+++ b/scripts/hooks/full_suite_guard.py
@@ -1,105 +1,99 @@
-"""PreToolUse hook: warn when running full pytest suite without a specific path.
+"""PreToolUse hook: BLOCK a full-suite / whole-directory local pytest run.
 
-Running the entire test suite (5900+ tests, ~10 min) during iterative
-development wastes time.  Targeted tests (specific file or directory)
-run in seconds.  This hook detects bare `pytest` or `pytest -v` (no
-test path) and warns the user.
+Running a whole test directory locally on this shared box (let alone ``tests/``)
+duplicates CI's authoritative full run and starves the live Genesis services —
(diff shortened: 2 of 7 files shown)


WingedGuardian:
@codex review

WingedGuardian:

Checked: `scripts/hooks/concurrent_test_guard.py`, `scripts/hooks/full_suite_guard.py`, `scripts/hooks/shell_parse.py`, and the corresponding test files (`tests/test_hooks/test_concurrent_test_guard.py`, `tests/test_hooks/test_full_suite_guard.py` (new), `tests/test_hooks/test_shell_parse.py`, `tests/test_hooks/test_deliverable_and_fullsuite.py`).

- Import hygiene: stdlib-only in all three hook/lib files (`os`, `sys`, `re`, `shlex`, `dataclasses`); local imports (`hook_input`, `shell_parse`) follow the existing self-locate + `# noqa: E402` convention already used elsewhere in `scripts/hooks/`. No circular imports, no star imports.
- No bare `except:` — both new `except Exception:` blocks (in `full_suite_guard.main` and `shell_parse.command_runs_pytest`) name the exception type and document the fail-open rationale.
- No nohup/task_states/tmp/git-add-all/force-push/reset-hard/killpg/raw-SQL patterns present.
- Coverage: all three changed source files have matching updated or new test files; no new untested `.py`.
- Docstrings: module docstring and public functions (`is_pytest_invocation`, `command_runs_pytest`) are documented and reflect the new blocking behavior; `full_suite_guard.py`'s module docstring was updated to match its new allow/block semantics.
- Conventional commit prefix present in the PR title (`fix(hooks): ...`).

- No secrets, tokens, private IPs, SSH keys, connection strings, or private hostnames in the added lines — the diff is pure control-flow/parsing logic and tests.
- No personal emails, username-bearing paths, or location-revealing identifiers.
- PR title/body describes an internal shared-box OOM incident without naming specific infrastructure, and doesn't fall under the leak/security-incident wording pattern.



---

WingedGuardian:
Structural first-pass review of PR #1430 (7 files, hooks + tests).

## Summary

No ERROR-level pattern violations found. No nohup/raw-SQL/bare-except/mutable-defaults/git-hygiene issues, no heavy imports in hook scripts (stdlib + local `hook_input`/`shell_parse` only), import grouping is correct throughout, both commits use conventional prefixes (`fix(hooks):`, `test(hooks):`), and the new/changed public functions (`is_pytest_invocation`, `command_runs_pytest` in `shell_parse.py`) have docstrings. Coverage is solid — `test_full_suite_guard.py` is new and covers the rewritten hook, `shell_parse` additions are tested in `test_shell_parse.py`, and the `concurrent_test_guard` fix has a dedicated regression test for the quoted-pipe false-positive this PR fixes.

## WARNING — behavior change dropped without matching test coverage

`scripts/hooks/full_suite_guard.py:72-97` (`_targets_specific_test`): for an arg that starts with `-`, the very first check —

```python
if arg in ("-k", "-m") or arg.startswith(("-k", "-m", "--keyword")):
    return True
```

— returns `True` (i.e. "targeted, allow") for a **dangling** `-k`/`-m` with no following value, e.g. `pytest -k` or `pytest -m`. The prior implementation explicitly treated a valueless `-k` as the full-suite case (`test_k_without_value_is_full_suite: assert _fsg._is_full_suite("pytest -k") is True`, in the now-deleted `TestFullSuiteSelector`). That test was dropped in this PR rather than migrated to `test_full_suite_guard.py`, so the new behavior for this edge case is untested in either direction.

Real-world impact is low — pytest itself will error out on a valueless `-k`/`-m` (`argument -k/--keyword: expected one argument`) before any full-suite run could happen — but since this hook's whole job is precise blocking logic, worth a one-line test in `TestTargetsSpecificTest` (e.g. `test_k_without_value` / `test_m_without_value`) asserting the intended behavior explicitly, rather than leaving it as a silent side effect of the rewrite.

No other structural issues found.



---

PR #1431, opened 2026-08-22
feat(security): observation write-provenance + laundering-critical read exclusions (WS-3 PR-1)

WingedGuardian:
## What

Closes the observation-content injection surface (WS-3 PR-1): observation rows now carry a definite ORIGIN stamped at the write boundary, and the laundering-critical READ surfaces exclude external/unknown-origin content — severing the path by which external observation content could reach the deep-reflection prompt and re-enter the user model as a trusted delta (the gap #1407's write-gate could not close alone).

## Why

`observation_write` lets a caller set `source`/`type` freely, and `observations.create(origin_class=None)` was the default — so provenance was not on the data, and read-side classification could not be reliable. The robust fix puts provenance at the WRITE chokepoint and makes it fail-closed, so the (uncurated, ~200-value) `source` space is non-load-bearing for security: a missed/novel external writer degrades to excluded, never trusted.

## How

**Write boundary (definite, fail-closed origin):**
- The CRUD chokepoint (`create()`/`upsert()` → `derive_observation_origin`) classifies every caller: explicit origin → dispatching-session origin → source classification → NULL (fail-closed). Unknown source → NULL → excluded at read.
- Channel-attended `task_detected` writes stamp `owner`; gateway ones `external_untrusted` (`cc.types.task_detected_origin`).
- The two stdlib-only hooks that bypass CRUD stamp origin inline: the conversation-pivot writer (`owner` — only reachable from interactive owner sessions) and the post-commit audit writer (`first_party`).
- A one-time migration backfills historical rows (session-JOIN + source classification; unknown → NULL).

**Read boundary (exclude external/unknown from laundering-critical surfaces):**
- The always-loaded L1 file (`essential_knowledge` — recent decisions + active session pivots).
- The deep-reflection / perception prompt pipeline (`context_gatherer` eval-context ×4 + recent-observations + calibration; `perception.context` memory-hits + light-chain; the prior-light-summary fetch).
- A deep-reflection prompt-rule belt (user-model updates never from external/quoted/third-party material).
- Filter is fail-closed (`origin_class IN ('owner','first_party')`, NULL excluded). Metric-only reads (backlog counts, assessment volume) stay unfiltered so a flood cannot suppress a reflection.

**Guardrail + hardening:**
- A coverage guardrail fails CI on any raw `INSERT/REPLACE INTO observations` outside the chokepoint + the two stamped hooks, and pins that `create()`/`upsert()` still resolve origin.
- The pushed-surfaces supervision exemption for interactive sessions is now restricted to owner-attended channels (terminal/Telegram); gateway conversations no longer receive it.

## Scope / follow-ups

- **PR-2 (tracked follow-up):** read-side WRAP breadth — the ego/sentinel/guardian/surplus/dashboard/MCP reads that should WRAP (not exclude) external content via a shared `render_safe` renderer, a read-side discovery guardrail, and boundary-marker zero-width normalization on the wrap path.
- **Watch post-deploy:** external findings still count toward the (unfiltered) reflection-trigger backlog but are excluded from what reflection gathers, so the deep-reflection job may re-fire more often until those rows hit their TTL. Not breakage; a separate external-row resolution path is the fix if cadence climbs.

## Testing

- New/updated unit tests for every write path (pivot/audit origin, channel predicate), every read exclusion (L1, reflection, perception, light-summary), the constant pinning, and the coverage guardrail. Verify-RED confirmed on the L1 and light-summary filters.
- Pre-existing tests that seeded NULL-origin rows expecting them surfaced were updated to stamp first-party (matching production) or to seed a genuine legacy NULL row via raw insert (the fail-closed gate test).
- Post-merge: deploy + a live E2E functionality check (backfill classification, write chokepoint, both read-exclusion paths, pivot writer) against the live DB.

Follow-up: 0a32f6b2549b48efb4a2f7eb5f79a23c


WingedGuardian:

**Notes (review scope):** 27 files changed (+1388/-47). Reviewed the full diff directly via git against the merge-base with `main` (not the API diff dump). Checked: nohup/systemd, direct `task_states` inserts, `/tmp/` usage, `git add -A`/`.`, force-push/reset --hard, `pip install -e` on a worktree path, bare `except:`, mutable default args, `os.killpg`, raw SQL against `genesis.db` outside `crud`/migrations, circular/star imports, import grouping, new-file test coverage, docstrings, hook stdlib-only hygiene, `$CLAUDE_PROJECT_DIR` misuse, `IntervalTrigger` on >1h intervals, commit prefixes, secrets/keys/tokens, private IPs (10.x/192.168.x/100.64-127.x/fd00::/8), SSH keys, personal emails, username-bearing paths, and PR title/body/CHANGELOG wording for incident-revealing language. None found.

Highlights: the new `observations.origin_class` write chokepoint (`db/crud/observations.py::_resolve_origin` → `memory/provenance.py::derive_observation_origin`) is fail-closed by design (unknown source → `None` → excluded, never trusted), the two stdlib-only raw-INSERT hooks (`scripts/proactive_memory_hook.py`, `scripts/hooks/emit_bugfix_audit.py`) both stamp `origin_class` inline, and `tests/test_security/test_observation_surface_coverage.py` is a solid AST+regex guardrail pinning both the chokepoint wiring and the raw-writer allowlist. New migration file (`0085_backfill_observation_origin.py`) has a matching test file with good edge-case coverage (idempotence, missing tables, absent/pre-stamped session rows). Every read-side call site that now filters on `SAFE_SURFACING_ORIGINS`/`SAFE_ORIGIN_SQL` has a corresponding origin-exclusion test.

Non-blocking observation: `db/crud/observations.py::count_external_by_ids` has a self-documented caveat that its `= 'external_untrusted'` predicate treats NULL as first-party — the opposite of the fail-closed read contract used everywhere else in this PR. The docstring already flags this as safe-for-now (feeds a SHADOW-only gate) with a tracked follow-up to route it through `effective_origin_class` before that gate enforces. No action needed here — just confirming it's the one deliberately-carved exception, not an oversight.



---

WingedGuardian:
## Structural review (first-pass, automated)

Reviewed the full diff (27 files, +1388/-47) against head `6482e10`.

### WARNING — gate-4 channel wiring in `ConversationLoop` is untested at the call-site level

`src/genesis/cc/conversation.py` now computes `supervised=is_owner_attended_channel(channel)` (the `handle_message`/`handle_message_streaming` invocation sites) and `origin_class=task_detected_origin(channel)` (the two `task_detected` observation writes) instead of the previous hardcoded `supervised=True`. This is the PR's stated "gate-4" fix — restricting the pushed-surfaces supervision exemption to owner-attended channels.

The new tests (`tests/test_cc/test_types.py::test_task_detected_origin_owner_vs_gateway`, `::test_is_owner_attended_channel`) only exercise the two pure helper functions in isolation. `tests/test_cc/test_conversation.py` — which already has the exact capture pattern this needs (`inv = mock_invoker.run.call_args[0][0]`, used by `test_handle_message_omits_routing_for_web` around line 631-643 for a `ChannelType.WEB` call) — wasn't extended to assert `inv.supervised is False` for that WEB-channel invocation (vs. `True` for the existing TERMINAL/TELEGRAM tests), and no test asserts the `origin_class` actually passed to `observation_write` for a `task_detected` row on a gateway channel. The security-relevant behavior change (gateway channels lose the supervision exemption) is therefore verified only at the standalone-predicate level, not at the production call site that consumes it.

### NOTE — pre-existing caveat comment, already tracked

`src/genesis/db/crud/observations.py::count_external_by_ids` gained a documentation-only addition (no logic change) flagging that its bare `= 'external_untrusted'` predicate treats NULL as first-party — the opposite of the fail-closed contract this PR establishes on every other read surface. The comment says it's safe today (feeds a SHADOW-mode gate only) and defers the fix to the PR-2 follow-up. No action needed here — flagging only so the follow-up row is confirmed to exist before that gate flips to ENFORCE.

### Clean

No ERROR-level pattern violations found: no `nohup`, no raw `task_states` inserts, no `/tmp/` usage, no `git add -A`/force-push/`reset --hard`, no bare `except:`, no mutable default args, no unguarded `os.killpg`, no bare-except, no new raw-SQL-against-`genesis.db` outside the migration exception (the two stdlib-only hook raw `INSERT`s are pre-existing and are now covered by a new CI coverage guardrail, `tests/test_security/test_observation_surface_coverage.py`). Import grouping/ordering, docstrings on new public functions, and commit prefixes (`feat(security):`, `test(security):`) all look clean. The one new `src/` file (migration 0085) has a matching test file.




---

WingedGuardian:
Thanks — all five addressed in `68464a58`.

- **P1 gateway origin** — `cc.types.session_origin_for_channel` now stamps `origin=external_untrusted` on all three conversation `CCInvocation` sites for non-owner-attended channels, so a gateway session's own `observation_write` calls are stamped untrusted rather than coalescing to `first_party`.
- **P1 classify internal writers** — completed `_FIRST_PARTY_OBS_SOURCES` from the full writer census (foreground_reaper, cognitive_ledger, stability_monitor, the git/wal/pid/embedding/dead_letter/cc_* monitors, task_executor, …). Added a new AST-based source-classification coverage test that forces every static observation source to classify — it immediately caught two more the census had missed.
- **P2 reflection cutoff/control queries** — gated the three `cc_reflection_deep` window markers, the `detect_pending_work` backlog trigger, and the `reflection_quality` cohort on origin, so a forged external `cc_reflection_deep` row can't move the window, inflate the trigger, or skew the score.
- **P2 migration 0085** — replaced the live-classifier import with a frozen, self-contained in-file snapshot (verified identical classification to the live path over all sources), so the one-time backfill is deterministic across deployment history.
- **P2 dedup by provenance** — `skip_if_duplicate` now includes `origin_class` (NULL-safe `IS`), so a gateway/external duplicate can't suppress a later owner request.

The read-side WRAP breadth (ego/sentinel/guardian/dashboard) remains PR-2 as noted in the description.

WingedGuardian:
@codex review

WingedGuardian:
Round-2 findings addressed in `127a98e5` (+ a clean `origin/main` merge). Dispositions:

**A (P1) — durable gateway session origin: FIXED, and extended to the whole gateway class.**
Confirmed real. `get_or_create_foreground` stamps gateway foreground sessions; red-teaming it surfaced that **voice** sessions bypass that path entirely (`register_voice_session`), so they'd stay NULL → read first-party by `reflection_window_origin`. Both are now stamped `external_untrusted`; migration 0085 gains a step-0 that grandfathers pre-deploy gateway/voice `cc_sessions` rows (keyed on the actual stored values — `channel IN ('web','whatsapp')`, `'voice_s2s'`, `source_tag='voice'`) *before* the observations session-JOIN, so historical session-attributed rows inherit the right origin. NULL-channel adoptees (owner CLI) and telegram/terminal untouched.

**C (P2) — indirect-writer provenance: FIXED via origin-aware derivation (NOT allowlisting).**
Confirmed real, with a sharper root cause: the triage pipeline runs in-server for inbox/mail (external email) too, so blanket-allowlisting `retrospective`/`cc_debrief` would have *laundered* external content into L1/reflection as first-party — the opposite of a fix. Instead they're stamped with the analyzed conversation's channel origin (new `observation_origin_for_channel`), threaded through `ObservationWriter.write` and `route_learning_signals` to **both** the observations row and its MemoryStore copy (this also closes a pre-existing memory-side first-party hole). `follow_up_watchdog` + an `ego_domain_redirect:` prefix are safe first-party additions; `retrospective`/`cc_debrief` live in a new `_CHANNEL_STAMPED_OBS_SOURCES` set (stamped at the write site, never source-derived → an un-stamped caller fails closed to NULL). The coverage guardrail now discovers the indirect `ObservationWriter.write` path and resolves module-constant source Names (45→53 sources; verify-RED confirmed it bites).

**D (P2) — Telegram quote-reply intent forge: FIXED as a class.**
Confirmed real, and it's broader than `/task`: the unanchored parser also let quoted bot text flip `/model`/`/effort`/`/resume`. `handle_message`/`handle_message_streaming` gain an `intent_text` param — slash intent is parsed from the owner's raw reply only; `task_detected` content is owner-only; the composite stays the LLM prompt (quoted context preserved). Every other channel/path is unchanged (`intent_text=None`). A real-`IntentParser` test proves a `/task` in quoted text is NOT detected while an owner-reply `/task` is.

**B (P1) — backfill trusts historical source labels: KEPT source-based, bounded residual documented.**
This is a deliberate, bounded acceptance (documented in 0085's docstring): it affects only pre-0057 rows (no provenance was *read* before this PR, so there was no gate to evade for gain), those rows are TTL-expired out of the 7–14d surfacing windows, and every forward write is stamped at the chokepoint from the live session env, which a forged `source` cannot override. `retrospective`/`cc_debrief` are deliberately not snapshot-classified (channel-stamped live) → historical rows stay NULL → fail-closed.

Adversarially reviewed (genesis-architect: SAFE-TO-COMMIT, all focus areas clean). 141 touched-file tests green on the final snapshot; broader run 3642 + 1615 passed, 0 failures.

@codex review

PR #1433, opened 2026-08-23
test(guardian): isolate host credential sources in combined-bridge tests

WingedGuardian:
The combined-bridge tests only isolated `HOME` + `secrets_path`, but the composite `propagate_guardian_credentials` fans out to five legs, two of which read dedicated host-credential files that neither isolation covers:

- `propagate_cc_oauth_token` reads an import-frozen module constant (`_CC_TOKEN_SOURCE`), which a `HOME` monkeypatch cannot retro-redirect.
- `propagate_internal_api_token` reads `GENESIS_HOME/internal_api_token` (fallback `~/.genesis`); an absolute `GENESIS_HOME` bypasses the `HOME` patch entirely.

On any install where those files exist, the legs escrowed the real tokens, adding stray `cc_oauth_token.env` / `internal_api_token.env` entries that broke the four exact-name assertions — green in CI, red on a live install (a suite that reddens locally trains people to ignore red).

Neutralize both sources so the legs no-op: extend the `TestCombinedBridge` autouse fixture (covers its 3 tests + any future one) and inline-isolate `test_combined_bridge_includes_passphrase`, mirroring the existing `TestCCOAuthTokenSync` exemplar.

Both leaks reproduced RED on a live install and are now GREEN. Full file: 42 passed. The "includes cc-token" contract stays covered by `test_combined_bridge_includes_cc_token`.

Follow-up: bf401452e30e4b069804c3d1c6d8b491

### DIFF
--- a/tests/test_guardian/test_credential_bridge.py
+++ b/tests/test_guardian/test_credential_bridge.py
@@ -15,7 +15,7 @@ from genesis.guardian.credential_bridge import (
 
 
 @pytest.fixture(autouse=True)
-def _isolate_credential_sources(tmp_path: Path):
+def _isolate_credential_sources(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
     """Fail-closed default for EVERY test in this module: point all three host
     credential sources at absent tmp paths so NO combined-bridge invocation can
     escrow a real host token into pytest's output dirs. (Reproduced on a live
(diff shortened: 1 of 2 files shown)


WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
@codex review

PR #1435, opened 2026-08-23
fix(dashboard): navigate the file browser to the upload destination

WingedGuardian:
After a Files-tab upload the browser stayed put, so a folder upload landed but the user couldn't see where it went. Navigate to the destination instead: a folder upload jumps INTO the new top-level folder, a single loose file to its directory, multiple loose files to the uploads root (falling back to a refresh in place).

The folder segment is derived from the SERVER-sanitized relpath (`uploaded[0]` = `data.filename`), not the raw client path: the backend rewrites unsafe filename chars (e.g. `Q3 (final)` → `Q3 _final_`), so navigating to the raw name would 404 and silently no-op — reproducing the very bug this fixes for any folder name with special characters. Reuses the existing `fetchFiles(path)` method.

Verified: uploading a folder named `Q3 (final)` stores it on disk as `Q3 _final_`; the computed navTarget matches and resolves against the live files API (HTTP 200).

Follow-up: 846de35181d3409dbb5425b50c06333a

### DIFF
--- a/src/genesis/dashboard/webui/js/dashboard.js
+++ b/src/genesis/dashboard/webui/js/dashboard.js
@@ -996,7 +996,11 @@
               const resp = await fetchApi("/api/genesis/files/upload", { method: "POST", body: form });
               if (resp && resp.ok) {
                 const data = await resp.json();
-                uploaded.push(data.filename);
+                // Only count a real string relpath: keeps uploaded.length honest for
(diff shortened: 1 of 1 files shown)


WingedGuardian:
@codex review

WingedGuardian:
@codex review

PR #1436, opened 2026-08-23
fix(dashboard): empty-state loading gates for 5 low-traffic panels

WingedGuardian:
Extend the shipped P4 empty-state pattern (reference: the observations panel) to the remaining panels that flashed a false "No X" while their fetch was still in flight: campaigns, tracked repositories (watchlist), recent knowledge units, modules, and providers.

- `dashboard.js`: register `fetchState` entries for campaigns/watchlist/knowledgeRecent and wire `startFetch`/`finishFetch`/`failFetch` into their fetch functions (mirroring `fetchObservations`). modules/providersDetail were already wired — template-only there.
- templates: add `Loading…` and error rows and gate each `length === 0` empty state on `!['loading','error'].includes(panelState('<name>'))` so the empty state shows only after a successful fetch returns zero items.

watchlist and recent-knowledge are fetched once on tab open (no poll), so their error copy says "reload to retry" rather than the "will keep retrying" used by the polled panels — an accurate promise per panel.

Follow-up: 143f54a1f2a34a5b9b10b89ad34ef71e

### DIFF
--- a/src/genesis/dashboard/templates/partials/tabs/campaigns.html
+++ b/src/genesis/dashboard/templates/partials/tabs/campaigns.html
@@ -14,7 +14,13 @@
           Pause/resume, run one now, or edit cadence, model, effort, daily cap and jitter.
         </div>
 
-        <template x-if="$store.genesisDashboard.campaignsList.length === 0">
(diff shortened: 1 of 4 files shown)


WingedGuardian:
@codex review

PR #1437, opened 2026-08-23
feat(dashboard): page-top banner when backups are unconfigured or failing

WingedGuardian:
Backup health had no page-level visibility — a barren or failing backup state was only apparent on the Backup tab. Add a persistent header banner (reusing the existing system-banner pattern) that surfaces on every tab when backups are unconfigured, have never run, or the last run failed, and stays hidden when backups are healthy.

- `dashboard.js`: `backupBanner()` computed returns a styled `{text,color,bg,border}` for a problem state or null when healthy (null-safe before the first fetch, so no flash). Fetched fire-and-forget on load (NOT awaited — the `/backup/status` route shells out to git + systemctl, and must not gate initial paint) plus a slow 180s refresh so an unattended scheduled-backup failure surfaces without a reload.
- `header.html`: the banner template, warn (unconfigured / never-run) vs crit (last run failed) styling.

Data comes from the existing `/api/genesis/backup/status` route (no backend change). Message text is rendered via `x-text` (no HTML injection).

Follow-up: c50d288254244ddfaf9a8165e44aaec9

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1441, opened 2026-08-24
test(campaigns): defuse wall-clock time-bomb in park-status tests

WingedGuardian:
## Problem

`tests/test_campaigns/test_runner.py::TestParkAwareSessionStatus` seeded a
`cc_rate_limit_parks` row with a **fixed** literal `created_at="2026-08-17T10:00:00Z"`.
The production path under test ages that timestamp against the **real wall clock** —
`runner.py:430` `datetime.now(UTC) - created > _PARK_WAIT_BOUND` (bound = 7 days), with no
injectable clock. So the "reads as still running" tests passed only for **7 days** after
the seed date, then flipped to failing on **2026-08-24** — reddening the `test` CI job for
every open PR, not just this one.

This is a **wall-clock time-bomb**: an absolute seed compared against `now()` with a
threshold works when written, then detonates once the calendar crosses the threshold.

## Fix

Seed the park (and the session's `started_at`/`last_activity_at`) **now-relative**, so it
always sits inside its 7-day wait window on any run date, with a comment explaining why (to
prevent re-hardcoding). The over-bound tests, which deliberately backdate `created_at`
afterward, are unaffected — a fixed *old* date stays always-over-bound and never flips.

## Whole-class sweep

Rather than fix only the site on fire, the entire class was swept (four parallel triage
agents traced every absolute datetime literal seeded into test state across the tree to its
production consumer, plus an adversarial cross-check of production staleness windows longer
than a week). **This was the only wall-clock time-bomb found** — the codebase otherwise
injects clocks (`now=`/`clock=`) or seeds now-relative.

## Verification

- Reproduced RED (both tests fail today with the literal seed) → GREEN after the fix.
- `tests/test_campaigns/test_runner.py` = 30 passed; verify-RED done (reverted → fail →
  restored → pass); `ruff` clean.
- Unblocks the `test` CI job for #1434 and #1437, whose only failure was this bomb.

Follow-up: 202367ef817345f6ab692da401a81b73

### DIFF
--- a/tests/test_campaigns/test_runner.py
+++ b/tests/test_campaigns/test_runner.py
@@ -362,11 +362,20 @@ class TestParkAwareSessionStatus:
     async def _seed_failed_parked_session(self, db, park_status: str):
         from genesis.db.crud import cc_sessions
 
+        # Seed the park NOW-RELATIVE, not with an absolute literal. _check_session_status
+        # ages the park's created_at against the REAL wall clock (runner.py:430,
+        # `datetime.now(UTC) - created > _PARK_WAIT_BOUND`, bound = 7 days) and there is no
+        # injectable now, so a fixed seed like "2026-08-17" silently works only until the
+        # calendar crosses the bound, then flips these "still running" tests to failing (a
+        # wall-clock time-bomb). A recent seed keeps the park within its wait window on any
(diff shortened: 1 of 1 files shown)


WingedGuardian:
@codex review

PR #1442, opened 2026-08-24
fix(cc): daily session reset fires at local midnight, not UTC midnight

WingedGuardian:
## Problem
The CC-session daily reset computed its day boundary in UTC:
`datetime.now(UTC).replace(hour=day_boundary_hour)`. On any install whose
timezone is offset from UTC, the reset fired at UTC-midnight instead of local
midnight — e.g. ~20:00 for a US-Eastern install with the default
`day_boundary_hour=0`. Sessions were being retired mid-evening rather than
overnight.

Two call sites shared the bug: `ConversationLoop._should_reset` (the live path,
hit on every message) and `SessionManager.check_morning_reset`.

## Fix
Extract one helper, `genesis.util.tz.local_day_boundary(hour, now=None)`, that
computes the most-recent `hour:00` boundary **in the user's timezone** and
returns it **UTC-aware**. Both call sites delegate.

Returning UTC is load-bearing for `check_morning_reset`: its boundary is
serialized and string-compared (`<`) against stored `+00:00` timestamps inside
`cc_sessions.query_stale`. A local-offset ISO string (`...-04:00`) would sort
lexicographically wrong against `+00:00` values and silently break the query —
so the helper normalizes to UTC, matching every `last_activity_at` write site
(all `datetime.now(UTC).isoformat()`).

`now` is an injectable seam (defaults to the current instant) so the boundary
logic is deterministically testable without a time-freezing dependency.

## Tests
- `TestLocalDayBoundary`: local-vs-UTC boundary, DST correctness (EST vs EDT),
  UTC-install no-op (byte-identical behavior), and the `+00:00` serialization
  that closes the string-comparison trap.
- Wiring tests exercise both real call sites via the injected clock.
- **Verified-RED**: reverting the helper to the old UTC-midnight computation
  fails exactly the 4 distinguishing tests; the fix makes them pass.

Generalizable: a UTC install is unaffected; the boundary follows
`user_timezone()`, no hardcoded offset.

### DIFF
--- a/src/genesis/cc/conversation.py
+++ b/src/genesis/cc/conversation.py
@@ -7,7 +7,7 @@ import json
 import logging
 from collections.abc import Awaitable, Callable, Coroutine
 from dataclasses import replace
-from datetime import UTC, datetime, timedelta
+from datetime import UTC, datetime
 from typing import TYPE_CHECKING, Any
 
 from genesis.cc import rate_limit_park, roster
@@ -37,6 +37,7 @@ from genesis.cc.types import (
 )
 from genesis.db.crud import cc_sessions
 from genesis.observability.call_site_recorder import record_last_run
+from genesis.util import tz
 
 if TYPE_CHECKING:
     from genesis.cc.contingency import CCContingencyDispatcher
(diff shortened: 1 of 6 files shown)


WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
Superseded by #1449 — recreated on a clean branch because this branch's history contained a private timezone identifier that force-push (blocked on this repo) can't rewrite in place.

PR #1445, opened 2026-08-24
fix(contribution): scan_prose no longer over-blocks dash-leading Markdown lines

WingedGuardian:
## Summary

`scan_prose` — the secret-scan floor in the contribution sanitizer — ran
`detect-secrets scan --string <line>` per line. argparse then misread any line
whose content starts with `-` (a Markdown `---` horizontal rule, a `--flag`
example — both common in issue/PR prose) as an unknown option (exit 2), and the
fail-closed nonzero-exit branch turned that into a spurious BLOCK. Legitimate
contributor prose was over-blocked.

## Fix

Use the `--string=<value>` form so argparse binds the value literally even when
it starts with a dash. Scan semantics are unchanged for every other input.

Same-function hardening (from review):

- Broaden the crash catch to `OSError`, so an oversized argv (E2BIG) still fails
  **closed** rather than propagating.
- Document that the blank-line guard is load-bearing (an empty `--string=` makes
  detect-secrets scan the working tree instead of the string).

## Tests

Real-binary regression tests (detect-secrets is a core dependency, so they run
in CI):

- a `---` / `--flag` body scans clean;
- a secret on a dash-leading token is still genuinely detected — asserting the
  detection *message*, not merely `ok is False`, so a crash-into-fail-closed
  can't masquerade as a real scan.

Both were verify-RED confirmed on the pre-fix code. Full `test_sanitize.py`: 57 passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved privacy scanning for contributor submissions containing Markdown horizontal rules, examples, or other dash-prefixed lines.
  * Prevented valid Markdown content from being incorrectly blocked during scanning.
  * Preserved detection of genuine secret-like values, including those beginning with a dash.
  * Scanning failures now produce a blocking result instead of being silently ignored.

* **Documentation**
  * Updated the unreleased changelog with these privacy-scanning improvements.


WingedGuardian:

- Checked all added/modified `.py` files (`src/genesis/contribution/sanitize.py`, `tests/test_contribution/test_sanitize.py`) against Domain 1 rules: no `nohup`, no direct `task_states` writes, no `/tmp/` usage, no `git add -A`/force-push patterns, no bare `except:` (the widened `except (subprocess.TimeoutExpired, OSError)` is still explicitly typed — `FileNotFoundError` is a subclass of `OSError`, so this is a legitimate broadening, not a bare catch), no mutable default args, no raw SQL.
- Import hygiene: stdlib/third-party/local grouping intact, no star imports, no new circular-import risk.
- Coverage: the changed function (`_run_detect_secrets`) has corresponding tests updated in the same PR, including two new regression tests that exercise the real `detect-secrets` binary (necessary since the bug lives in argparse's CLI parsing, invisible to a mocked subprocess).
- Docs: no public API surface changed; the added comments correctly document the load-bearing blank-line-skip and the `--string=` argv-parsing rationale.
- Commit/PR title uses a conventional prefix (`fix(contribution): ...`).

- No secrets, credentials, private IPs, or connection strings introduced in the diff.
- `AKIAIOSFODNN7EXAMPLE` in the new test is AWS's own published example/placeholder access-key ID (used intentionally as a test fixture to confirm the scanner still detects real secret shapes) — not a live credential, and this pattern is a standard secret-scanner test fixture.
- The `CHANGELOG.md` entry and PR description describe a false-positive **over-blocking** bug (legitimate Markdown prose was being rejected) and its fix — not a leak, security incident, or exploitable-detail disclosure. No wording concerns.



---

WingedGuardian:

Scope: 3 files (`CHANGELOG.md`, `src/genesis/contribution/sanitize.py`, `tests/test_contribution/test_sanitize.py`), single commit with a conventional `fix(contribution):` prefix.

Notes from a first-pass structural review (not a logic review — I additionally ran the actual fix against the real `detect-secrets` binary to confirm the mechanism, which goes beyond the usual first-pass scope, but is worth recording since it's easy to verify):

- No pattern-violation checklist items apply here (no `nohup`, no raw SQL, no `git add -A`/`--force`, no bare `except:`, no mutable defaults, no `/tmp/` usage — tests use pytest's `tmp_path` fixture correctly).
- Import hygiene: no new imports, no star imports, no reordering issues.
- Test coverage: the changed function (`_run_detect_secrets`) gets two new targeted regression tests exercising the real CLI (the autouse mock is deliberately bypassed via a captured pre-patch `shutil.which`/`subprocess.run`, which is the only way to catch an argparse-level bug like this).
- Verified independently: installed `detect-secrets` and reproduced the exact bug locally — `detect-secrets scan --string ---` and `--string --verbose` both exit 2 (argparse misreads the dash-leading value as an unknown option) under the pre-fix code, and both scan clean under `--string=<value>`. Also confirmed a real secret on a dash-leading token (`-AKIAIOSFODNN7EXAMPLE`) still fires `AWSKeyDetector` under the fix — no fail-open regression.
- The broadened `except (subprocess.TimeoutExpired, OSError)` is a strict superset of the prior `FileNotFoundError` (subclass of `OSError`), consistent with the fail-closed contract described in the function's docstring.
- No new `.py` file added, so the "new file needs a test file" check doesn't apply; the changed function's docstring didn't need updating (still accurately describes `--string` mode, just the `=` binding form).



---

WingedGuardian:
@coderabbitai review

WingedGuardian:
Rebased on current main (was 91 behind). The core `--string=<value>` dash fix is still needed — the bug is live on main (`detect-secrets scan --string ---` → exit 2 vs `--string=---` → exit 0). Resolved the 2 conflicts by keeping BOTH improvements: `[ds_bin, "scan", f"--string={text}"]` (main's PATH-independent resolve + this PR's attached-form dash fix), and main's broader `except Exception` floor.

CodeRabbit's ask (catch ValueError + NUL regression test) is resolved by the merge — main's `except Exception` already catches the embedded-NUL ValueError and ships `test_detect_secrets_nul_byte_line_fails_closed`. No new code needed. Full sanitize suite: 59 passed (both dash tests ran).

@codex review

PR #1447, opened 2026-08-24
fix(hooks): offline allowlist so a transient ls-remote failure doesn't re-prompt an approved branch push

WingedGuardian:
## Problem

The push-approval hook prompts only on a branch's **first** push; a re-push of
fixes to the same, already-published branch must be silent. But the "already on
the remote?" check is a live `git ls-remote` that fail-closes to a **prompt** on
any network hiccup, so a flaky network re-prompted every re-push — the
first-push-only guarantee was network-fragile.

## What this does

- **`scripts/hooks/push_allowlist.py`** (new, stdlib-only): a local allowlist
  that caches the confirmed-on-remote fact in `~/.genesis/pushed_branches.json`,
  so re-pushes are decided **offline**. `is_recorded` / `record`, keyed on
  `(branch, push-URL set)` — never the remote name, so the same branch name on a
  different repo is never conflated. 90-day prune on write; atomic
  `tmp` + `os.replace`; fail-open everywhere (a corrupt/absent file can only
  cause a redundant prompt).
- **`scripts/hooks/git_push_guard.py`**: the non-force re-push arm now checks the
  allowlist offline first, falls back to the live `ls-remote` leg, and records on
  a **HIT** (the sole writer). A guarded import degrades to the pure `ls-remote`
  path if the module is broken/absent — mirroring the existing `review_state`
  soft-import (a raising hook import exits 1 → CC treats non-2 as non-blocking →
  every fail-closed gate would silently vanish).

## Security invariant

The allowlist can **only** relax a re-push of a branch a live `ls-remote` just
proved is on the remote (i.e. already approved on its first push) — it can never
authorize a genuine first push. Force pushes, dispatched sessions, and pushes to
`main`/`master` are hard-blocked **before** this arm; the allow is deferred so a
compound `git push … && git commit --no-verify` still hard-blocks. A security
review of the diff confirmed the invariant holds (no CRITICAL/HIGH).

Consumer-side recording only — a pre-implementation architecture review found
that a PostToolUse recorder keyed on the current branch could mis-seed a
never-pushed branch into a silent first push; it was dropped, removing that
class entirely.

Conscious tradeoff (documented): a recorded branch stays trusted for the 90-day
window even if its remote copy is later deleted — a liveness check would
reintroduce the network dependency this change removes.

## Tests

16 unit (`test_push_allowlist.py`) + 5 integration (`test_push_create_override.py`,
including offline-allow when `ls-remote` fails, verify-RED'd RED→GREEN) + a
home-isolation guard. 93 push-gate + 593 sibling guard/`test_scripts` tests
green; ruff clean; no writes to the real `~/.genesis`.

Ledger: fb8b780245cc4e1b9e8a9160fbfaee54

https://claude.ai/code/session_012NPVb1ckaMnde48hzvumyt

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,8 +49,20 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
-- **A scheduled job that has run repeatedly but never once succeeded now raises a
-  health alert.** Such a job was invisible to every alarm: the "silently failing"
+- **A transient `git ls-remote` failure no longer re-prompts an already-approved
+  branch push.** The push-approval hook prompts only on a branch's FIRST push; a
+  re-push of fixes to the same, already-published branch should be silent. But the
+  "already on the remote?" check was a live `git ls-remote` that fail-closes to a
+  prompt on any network hiccup, so a flaky network re-prompted every re-push. A new
+  stdlib allowlist (`scripts/hooks/push_allowlist.py`, state in
+  `~/.genesis/pushed_branches.json`) caches the confirmed-on-remote fact locally so
+  re-pushes are decided OFFLINE. It is keyed on (branch, remote push-URL set) — never
+  the remote name — so the same branch name on a different repo is never conflated,
+  and it is written ONLY on a live ls-remote HIT (which proves the branch is already
+  on the remote), so it can never authorize a genuine first push. Corrupt/absent
+  state and any error fail OPEN to the existing prompt path; entries expire after 90
+  days (a recorded branch stays trusted for that window even if its remote copy is
+  later deleted). Such a job was invisible to every alarm: the "silently failing"
   check needs a prior success to measure a gap against, and the consecutive-failure
   counter resets on every restart. So a job that failed from its very first run —
   e.g. a daily actuator whose external login expired on day one — could fail silently
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,7 +62,9 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   on the remote), so it can never authorize a genuine first push. Corrupt/absent
   state and any error fail OPEN to the existing prompt path; entries expire after 90
   days (a recorded branch stays trusted for that window even if its remote copy is
-  later deleted). Such a job was invisible to every alarm: the "silently failing"
+  later deleted).
+- **A scheduled job that has run repeatedly but never once succeeded now raises a
+  health alert.** Such a job was invisible to every alarm: the "silently failing"
   check needs a prior success to measure a gap against, and the consecutive-failure
   counter resets on every restart. So a job that failed from its very first run —
   e.g. a daily actuator whose external login expired on day one — could fail silently
--- a/scripts/hooks/git_push_guard.py
+++ b/scripts/hooks/git_push_guard.py
@@ -123,6 +123,19 @@ from shell_parse import (  # noqa: E402
     split_segments,
 )
 
+# Local push allowlist (offline re-push cache). SOFT dependency, guarded exactly
+# like the review_state import above: a module-LOAD exception must degrade to
+# None (→ the pure live-ls-remote republish path, i.e. today's behavior), NEVER
+# propagate — a raising import exits 1, which CC treats as non-blocking, silently
+# disabling EVERY fail-closed gate in this file. push_allowlist itself is
+# fail-open by contract (its calls never raise); None here only covers an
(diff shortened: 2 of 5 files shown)


WingedGuardian:
@codex review

WingedGuardian:

Notes (non-blocking):
- `push_allowlist.py` is stdlib-only (json/os/tempfile/datetime/pathlib), matches hook conventions.
- The mid-file guarded `import push_allowlist` in `git_push_guard.py` mirrors the existing `review_state` soft-import pattern already in the file, and is documented as such — not flagging as import-hygiene.
- New module has a corresponding test file (`tests/test_hooks/test_push_allowlist.py`); `git_push_guard.py`'s changed re-push logic has updated coverage in `test_push_create_override.py`.
- No bare `except:` (all typed `except Exception:`), no mutable defaults, no `/tmp/` literals (uses `tmp_path`/`GENESIS_HOME`), no raw SQL, no `nohup`/`task_states` inserts.
- No secrets, private IPs, credentials, or personal paths in the diff. CHANGELOG/PR wording describes a hook-behavior bug fix, not a security incident.

---
Generated by an automated first-pass review (structural + PII/secrets/wording scan).




---

WingedGuardian:
## Structural review (first-pass)

Scope: pattern violations, import hygiene, coverage, docs, Genesis conventions. Not a logic review.

**No ERROR-level pattern violations found.** `push_allowlist.py` is stdlib-only (json/os/tempfile/datetime/pathlib), uses `except Exception:` (not bare `except:`), no mutable default args, atomic tmp+`os.replace` write (not under `/tmp/`), and the soft-import in `git_push_guard.py` correctly mirrors the existing `review_state` guard pattern (degrades to `None` on load failure, never propagates). Commit and PR title both carry the `fix(hooks):` conventional prefix. New module has full test coverage (`tests/test_hooks/test_push_allowlist.py`, 16 unit tests) plus 5 new integration tests in `test_push_create_override.py` covering the offline-allow / fail-closed / cross-repo-URL / isolation cases. Control-flow check: the restructured `if/elif/else` in `git_push_guard.py:3139-3181` preserves the original fail-safe semantics — `ask_reason` is still checked before `push_allow_reason` at the end of `main()` (lines 3570/3576), so a hard-block in a compound command still wins.

### WARNING — CHANGELOG.md entry corrupted by the edit

`CHANGELOG.md`'s `### Fixed` section: the new bullet ("A transient `git ls-remote` failure...") was spliced into the *middle* of the pre-existing "A scheduled job that has run repeatedly but never once succeeded now raises a health alert." bullet, rather than being inserted as its own bullet. The result is one bullet whose bold title is about the push-allowlist fix, but whose text silently runs on (with no paragraph break) into the unrelated health-alert bullet's body ("...later deleted). Such a job was invisible to every alarm: the 'silently failing' check needs a prior success..."). The original bullet's own bold title is gone entirely — that fix is no longer identifiable as its own entry in the changelog. Recommend re-splitting into two separate bullets before merge.

### NOTE (non-blocking)

- `git_push_guard.py:3172-3176` and `:3177-3181` — the new inner `else` (reached when neither the allowlist nor a live `ls-remote` confirms the branch) and the outer `else` (reached when the branch isn't a plain update of `cur`) build the identical `ask_reason` string. Not a correctness issue (both are effectively "no relaxation applies → default prompt"), just duplicated text — could be hoisted if it comes up again.




---

WingedGuardian:
@codex review

WingedGuardian:
Disposition of the 4 findings on the redesign head:

- **#1 (Preserve SSH user identity)** — **fixing**. Correct catch: an scp path resolves relative to the SSH login's home, so `alice@host:repo` ≠ `bob@host:repo`. Stripping the scp `user@` (from the earlier credential revision) conflated them. Keeping the full scp form; scp has no password field so nothing there is a credential (scheme-URL userinfo is still stripped via urlsplit). Added a distinctness test.
- **#2 (Resolve raw URL destinations)** — accepted (fail-safe). A direct-URL push resolves to an empty push-URL set and simply falls back to the ls-remote prompt — never a wrong allow, only an extra prompt for an unusual form.
- **#3 (fetch vs push URL) / #4 (`git -c pushurl` override)** — accepted as out of scope per this gate's declared threat model (see the module docstring: it does not model "a malicious operator crafting argv" nor "a multi-actor repo … retarget"; "reply as such rather than adding another patch"). Both are triangular-multi-remote / operator-crafted-argv cases where the operator already owns `--admin`/override.

Root: keying on push-URLs mirrors git's full push-destination resolution — an unbounded semantic tail. The redesign closed the string-parsing tail; the residual edges fall under the declared single-author, non-adversarial model.

WingedGuardian:
@codex review

PR #1449, opened 2026-08-24
fix(cc): daily session reset fires at local midnight, not UTC midnight

WingedGuardian:
Replaces #1442 (recreated on a clean branch: its history contained a private timezone identifier that could not be rewritten in place, since force-push is blocked).

## Problem
The CC-session daily reset computed its day boundary in UTC
(`datetime.now(UTC).replace(hour=day_boundary_hour)`). On any install whose
timezone is offset from UTC, the reset fired at the UTC-midnight instant
instead of local midnight — with the default `day_boundary_hour=0`, sessions
were retired at UTC-midnight local-time rather than overnight. Two call sites
shared it: `ConversationLoop._should_reset` (live, every message) and
`SessionManager.check_morning_reset`.

## Fix
Extract one helper `genesis.util.tz.local_day_boundary(hour, now=None)` that
computes the most-recent `hour:00` boundary **in the user's timezone** and
returns it **UTC-aware**. Both call sites delegate.

Returning UTC is load-bearing for `check_morning_reset`: its boundary is
string-compared (`<`) against stored `+00:00` timestamps in
`cc_sessions.query_stale`. A local-offset ISO string would sort wrong against
`+00:00` values and silently break the query, so the helper normalizes to UTC,
matching every `last_activity_at` write site. A naive `now` is treated as UTC
(matching `parse_utc_iso`). `now` is an injectable seam for deterministic tests.

## Tests
- `TestLocalDayBoundary`: local-vs-UTC boundary, a second fixed offset (proves
  the boundary follows the zone), UTC-install no-op, naive-`now`, and the
  `+00:00` serialization that closes the string-comparison trap.
- Wiring tests exercise both real call sites via the injected clock.
- **Verified-RED** against a reverted UTC-midnight helper.
- Boundary tests use fixed `Etc/GMT+N` offsets (no real timezone). The display
  tests (`fmt`/`fmt_short`) need a DST-observing named zone, so they use
  `Europe/London` (install-generic) instead of a real install timezone.

Generalizable: a UTC install is byte-identical; the boundary follows
`user_timezone()`, no hardcoded offset.

### DIFF
--- a/src/genesis/cc/conversation.py
+++ b/src/genesis/cc/conversation.py
@@ -7,7 +7,7 @@ import json
 import logging
 from collections.abc import Awaitable, Callable, Coroutine
 from dataclasses import replace
-from datetime import UTC, datetime, timedelta
+from datetime import UTC, datetime
 from typing import TYPE_CHECKING, Any
 
 from genesis.cc import rate_limit_park, roster
@@ -37,6 +37,7 @@ from genesis.cc.types import (
 )
 from genesis.db.crud import cc_sessions
 from genesis.observability.call_site_recorder import record_last_run
+from genesis.util import tz
 
 if TYPE_CHECKING:
     from genesis.cc.contingency import CCContingencyDispatcher
@@ -1620,11 +1621,19 @@ class ConversationLoop:
         except Exception:
             logger.debug("User correction scan failed", exc_info=True)
 
-    def _should_reset(self, session: dict) -> bool:
+    def _should_reset(self, session: dict, *, now: datetime | None = None) -> bool:
         """Check if session is from a previous day boundary.
+
+        The boundary is local-midnight (``day_boundary_hour`` in the user's
+        timezone), not UTC-midnight — otherwise the daily reset fires at the
+        UTC-midnight instant instead of local midnight on any UTC-offset install
+        (see :func:`genesis.util.tz.local_day_boundary`).
 
         Supergroup topic sessions (thread_id set) are persistent — they
         only compact when CC context limits are hit, never by day boundary.
+
+        *now* is injectable for deterministic tests; defaults to the current
+        instant.
         """
         # Supergroup topic sessions are persistent — no daily reset
         if session.get("thread_id"):
@@ -1635,10 +1644,5 @@ class ConversationLoop:
         started_dt = datetime.fromisoformat(started)
         if started_dt.tzinfo is None:
             started_dt = started_dt.replace(tzinfo=UTC)
-        now = datetime.now(UTC)
-        boundary = now.replace(
-            hour=self._day_boundary_hour, minute=0, second=0, microsecond=0,
-        )
-        if now < boundary:
-            boundary -= timedelta(days=1)
+        boundary = tz.local_day_boundary(self._day_boundary_hour, now=now)
         return started_dt < boundary
--- a/src/genesis/cc/session_manager.py
+++ b/src/genesis/cc/session_manager.py
@@ -17,6 +17,7 @@ from genesis.cc.types import (
     session_origin_for_channel,
 )
 from genesis.db.crud import cc_sessions
+from genesis.util import tz
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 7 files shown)


WingedGuardian:
@codex review

WingedGuardian:

Scope: `src/genesis/cc/conversation.py`, `src/genesis/cc/session_manager.py`, `src/genesis/util/tz.py`, plus tests in `tests/test_cc/` and `tests/test_util/`. Not documentation/config-only, so a full pass was done.

Checked and clear:
- No nohup/raw-SQL/bare-except/mutable-default/`/tmp/`/force-push/`git add -A` patterns.
- Import hygiene: new `from genesis.util import tz` import in both `conversation.py` and `session_manager.py` is correctly grouped with local imports and alphabetized (after `genesis.observability` / `genesis.db.crud` respectively). No circular imports, no star imports.
- New public function `genesis.util.tz.local_day_boundary` has a complete docstring covering behavior, the UTC-return rationale (the `query_stale` string-comparison trap), and the `now` injection seam. Docstrings on `_should_reset` and `check_morning_reset` were updated to match the new behavior.
- Test coverage: `TestLocalDayBoundary` in `tests/test_util/test_tz.py` covers local-vs-UTC boundary, a second fixed offset, UTC-install no-op, naive-`now` handling, and `+00:00` serialization. Both call sites (`ConversationLoop._should_reset`, `SessionManager.check_morning_reset`) get dedicated regression tests via the injected clock.
- Commit message uses the conventional prefix (`fix(cc): ...`).

Nit (not blocking): `_should_reset`'s docstring says "Supergroup topic sessions ... only compact when CC context limits are hit, never by day boundary" — this is existing prose, unrelated to this diff, no action needed.



---

WingedGuardian:
Second-pass review covering both structural patterns and PII/secrets/wording (the earlier comment on this PR only ran the structural pass). Note: GitHub blocks a request-changes review on a PR authored by this same account, so posting findings as a comment instead.

**Structural review: CLEAN**
No pattern violations. Confirms the prior structural comment: no nohup/raw-SQL/bare-except/mutable-default/`/tmp/`/force-push/`git add -A` patterns; import grouping is correct in both `conversation.py` and `session_manager.py`; the new `local_day_boundary` helper and the two modified call sites have complete docstrings and dedicated regression tests; commit uses a conventional `fix(cc):` prefix.

**PII/Secrets scan (code diff): CLEAN**
No secrets, private IPs, credentials, or other Domain-2 code-level issues in the added lines. Note the test suite change is a *positive* signal here: it deliberately swaps `America/New_York` fixtures for install-generic `Europe/London` / `Etc/GMT+N` zones specifically to avoid leaking a real install timezone into the public repo.

**Wording scan: CRITICAL — flagged for human attention**

PR description, opening paragraph:

> Replaces #1442 (recreated on a clean branch: its history contained a private timezone identifier that could not be rewritten in place, since force-push is blocked).

This discloses that PR #1442's git history contains a leaked private timezone identifier, and explicitly says it could not be scrubbed (force-push blocked). That's exactly the semantic-layer case this review is meant to catch: it doesn't repeat the IANA/offset value itself, but it does two things that matter more than a regex hit —
1. Confirms a real data exposure occurred (a private identifier is permanently in `#1442`'s commit history).
2. Points anyone reading this PR directly at where to go find it (`#1442`, "history … could not be rewritten in place").

Recommend rewriting the "Replaces #1442" note to state only the mechanical reason for the new PR (e.g. "recreated on a clean branch to drop an identifier that shouldn't have been committed") without confirming that the leak is unremediated and locatable. Whether `#1442` itself should be closed/scrubbed further is a separate, human decision — this review only covers this PR's own text/diff.

Since a request-changes review isn't possible on a self-authored PR, please treat the wording finding above as blocking until addressed.

---

WingedGuardian:
---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
@codex review

Round-2 (fresh cycle): replaced the one-directional walk-back with windowed-max to close the 4th DST edge (>1h fall-back regressing the local date across the boundary hour — Antarctica/Casey hour=0 returned a 24h-stale, non-monotonic boundary). local_day_boundary now generates fold=0 hour:00 candidates over a window one local date ahead of now through the lookback window and returns max candidate <= now (UTC). Added a bounded exhaustive brute-force test vs an independent wider-window oracle (never-future + monotonic) across pathological zones x hours x a 2-year transition grid. Focus on window sufficiency and any residual DST edge.

PR #1453, opened 2026-08-25
feat(dashboard): server-authoritative backup-health banner (supersedes #1437)

WingedGuardian:
## What

Moves the dashboard backup-health banner from a client-side precedence ladder to a
single server-authoritative pure function. `routes/backup.py::_backup_health` computes
`{state, code, reason}` and adds it to `/api/genesis/backup/status`; the banner
(`header.html` + `dashboard.js::backupBanner()`) just renders it.

**Supersedes #1437** (the client-side version). This resolves that PR's full set of
review findings by construction + design rather than porting the client ladder:

- **Client fetch-coordination findings** (status/config race, form re-hydration
  clobber) vanish — the banner no longer depends on the backup-config fetch at all
  (dropped the on-open config fetch; the Backup tab still hydrates its own form).
- **Semantic findings**: `configured` keys on a real `.git` clone (not bare `is_dir()`);
  the off-site-incomplete check gates on the **resolved** Tier-2 backend, not a stale
  status record; a malformed `{}` status record reads as *unreadable*, not healthy; a
  valid custom schedule with a future next-run is not false-flagged overdue (while a
  known-interval active timer that silently skips runs still flags).

## Security

`failure_reason` is served on the **unauthenticated** status route and can embed a
home-directory path (username). It's now sanitized (`/home/<user>` → `~`) at the
allowlist-projection seam, so the raw sibling field and the banner copy both carry the
scrubbed value — not just the banner. (The field was already served unauthenticated
before this change; this closes the path leak.)

## Review

- genesis-security-reviewer + genesis-architect both reviewed; both independently
  flagged the raw-`failure_reason` leak (now fixed) and otherwise approved. Architect
  NOTEs applied: preset-map drift guard added to the consistency test; precedence
  reorder for a more actionable message.
- `x-text` (textContent) render + fixed-map `:style` — no XSS/CSS-injection.

## Test

- 18 new state-table cases in `tests/test_dashboard/test_backup_config_api.py`
  (one per health code + malformed record + resolved-backend gating + real-clone
  `configured` + whole-body scrub + preset drift guard). **53 passed**, ruff clean.
  Verify-RED confirmed on the newly-designed branches.
- Real-data E2E: `GET /api/genesis/backup/status` → `backup_health: {code: healthy}`
  on a recently-backed-up host, with no home path anywhere in the response body.
- Full Alpine served-visual check is post-deploy (the live dashboard runs the prior JS
  until this ships).

Follow-up: 62c598ed416e4e4eb9a86195fe90245b

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -185,6 +185,19 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **Page-top backup-health banner (server-authoritative).** The dashboard now
+  surfaces a page-top banner when backups need attention — unconfigured, never run,
+  last run failed, timer stopped, overdue, replication incomplete, or an unreadable
+  status record — and stays hidden when healthy. The health verdict
+  (`{state, code, reason}`) is computed once on the server in
+  `routes/backup.py::_backup_health` and added to `/api/genesis/backup/status`; the
+  client only renders it. Computing it server-side removes the client-side
+  fetch-coordination and clock-skew failure modes entirely, keys "configured" and
+  the off-site-incomplete check on current signals (a real `.git` clone, the
+  resolved Tier-2 backend) rather than a stale status record, treats a malformed
+  status record as unreadable instead of healthy, and does not false-flag a valid
+  custom schedule as overdue. The failure reason rendered on the (unauthenticated)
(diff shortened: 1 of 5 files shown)


WingedGuardian:
@codex review

WingedGuardian:
@codex review

PR #1456, opened 2026-08-25
feat(hooks): configurable required scheduled-review kinds (leaks irreducible)

WingedGuardian:
## What

Makes the merge gate's required **scheduled-review kinds** configurable via local config instead of a hardcoded tuple, so an install can relax the structural **code-review** routine to *advisory* — without weakening the public default or the secret scanner.

## Why

The scheduled `code-review` and `leaks` routines are quota-capped; when `code-review` is rate-limited it blocks merges even though an equivalent adversarial pass (genesis-architect / the local review-depth gate) already ran. Operators should be able to make code-review advisory locally while keeping the secret scanner mandatory — a secret reaching a public repo is irreversible.

## Changes

- **`scripts/hooks/git_push_guard.py`** — new `_required_scheduled_review_kinds()`:
  - reads `merge_gate.required_scheduled_reviews` (a list) from `~/.genesis/config/genesis.yaml`;
  - **DEFAULT unchanged** (`code-review` + `leaks`) — no install that doesn't set the key is affected;
  - **`leaks` is irreducible** — always unioned in, never removable by config;
  - **fail-closed** to the full default on any missing / unreadable / malformed / wrong-type config; kinds lowercased to the marker grammar.
  - The check function consumes it; report path and merge-block path share the one decision.
- **`SKILL.md`** — documents the lever generically.
- **Tests** — `TestRequiredScheduledReviewKinds` + `TestScheduledGateUsesRequiredKinds` (default=both, relax→leaks-only, leaks-irreducible-even-if-omitted, empty/malformed→fail-closed, case-insensitive, gate integration); pinned the pre-existing characterization corpus's autouse fixture to the default policy so it stays install-agnostic.

## Review

Adversarial security review (genesis-security-reviewer): **fail-open enumeration all-clear** — `leaks` cannot be dropped on any branch. Found + fixed a HIGH (characterization-corpus test isolation) and a WARNING (config case-sensitivity). Lever unit + gate-integration + char-corpus tests green; ruff clean.

## Deploy path

Runtime hook — lands on next CC session start after `git pull`. Public default is byte-identical; per-install policy lives only in local `genesis.yaml` (not in this diff).

https://claude.ai/code/session_011v5AJ2U5rqH9yCeum49DH8

### DIFF
--- a/.claude/skills/genesis-development/SKILL.md
+++ b/.claude/skills/genesis-development/SKILL.md
@@ -823,10 +823,14 @@ findings below, a gated `gh pr merge`:
   Claude review (a `/schedule` cloud routine) posts as the repo OWNER's account and
   must carry a marker `<!-- genesis-scheduled-review: head=<full-40-hex-sha> kind=<name> -->`
   naming the exact head it reviewed AND which routine it is (`kind`). The gate blocks
-  unless an owner-authored marker for EVERY kind in `_REQUIRED_SCHEDULED_REVIEW_KINDS`
-  (currently `code-review` + `leaks`) names the PR's current head — so if any required
-  routine never ran, ran on a stale commit, or was rate-limited, the merge blocks
-  (naming the missing kinds). If a routine is pending or rate-limited, wait for it, or
+  unless an owner-authored marker for EVERY effective required kind
+  (`_required_scheduled_review_kinds()` — DEFAULT `code-review` + `leaks`; the leak/secret
+  scanner is irreducible and always required; an install may relax the OPTIONAL kinds to
+  ADVISORY via `merge_gate.required_scheduled_reviews: [<kinds>]` in local `genesis.yaml`)
+  names the PR's current head — so if any required routine never ran, ran on a stale
+  commit, or was rate-limited, the merge blocks (naming the missing kinds). An ADVISORY
+  routine still posts its review on the PR to be read/addressed, but its absence does not
+  block. If a required routine is pending or rate-limited, wait for it, or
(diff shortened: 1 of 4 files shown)


WingedGuardian:
@codex review — merge-gate change: required scheduled-review kinds now config-driven (default unchanged; `leaks` irreducible/never droppable; fail-closed to full default on bad config). Local genesis-security-reviewer pass: fail-open enumeration all-clear. Please scrutinize whether any branch of `_required_scheduled_review_kinds()` can drop `leaks` or otherwise fail-open.

WingedGuardian:

Notes (not violations, informational):
- `_required_scheduled_review_kinds()` in `scripts/hooks/git_push_guard.py` does a lazy `import yaml` inside the function body to read `~/.genesis/config/genesis.yaml`. This mirrors the pre-existing `_canonical_public_repo()` in the same file (line ~1870), which already does the identical lazy-import-for-config-read pattern — so this isn't a new hook-heavy-dependency issue, just consistent with established precedent in this file.
- Exception handling uses `except Exception:` (typed), not a bare `except:` — no violation.
- New behavior (`_required_scheduled_review_kinds`, the config lever, the `leaks`-irreducible union) is covered by two new test classes (`TestRequiredScheduledReviewKinds`, `TestScheduledGateUsesRequiredKinds`) plus a characterization-corpus pin — good coverage, no gaps.
- Commit/PR title carries the `feat(hooks):` conventional prefix. PR body explains the fail-closed/irreducible-leaks design rationale without naming any actual incident, leaked value, or private infrastructure — no CRITICAL wording issue.

---


---

PR #1457, opened 2026-08-25
fix(hooks): resolve commands through shell control structures in the guard parser

WingedGuardian:
## Problem

The shared command parser `shell_parse.analyze()` resolved a segment's executable
from its first token. A command wrapped in a shell control structure or command
group therefore resolved the exe to the leading control token, not the real
command:

| Input | Resolved exe (before) |
|---|---|
| `if …; then <verb> …; fi` | `then` |
| `(<verb> …)` | `(<verb>` |
| `{ <verb> …; }` | `{` |
| `! <verb> …` | `!` |
| `while …; do <verb> …` | `do` |

Every guard that keys on `seg.exe == "git" / "gh" / "rm"` then silently skipped
the segment. Reproduced bypasses (MEASURED, hook-level, rc=0 = allowed):

- `( rm-verb ~ )` / `( rm-verb / )` glued form bypassed the destructive-command
  guard (a home/root wipe).
- `(<git-discard>)` / `then <git-discard>` bypassed the git-discard block.

## Fix

**`shell_parse._strip_wrappers`** now strips leading shell command-position
tokens before resolving the exe:

- reserved words `! if elif while until then do else`
- group openers `(` (bare, glued `(<cmd>`, and nested) and brace `{`

It also strips one balancing trailing `)` off the operand carrying the subshell
close, including when a redirection follows it (`(… /x) 2>/dev/null`).
`for` / `case` / `select` / `in` are excluded — they front a *word*, not a
command.

The change is **monotonic**: a segment whose first token is already
`git` / `gh` / `rm` is returned byte-for-byte unchanged, so the strip can only
*reveal* a previously-hidden command — it can never hide a command a guard
catches today. The parser's fail-open contract (never raise; degrade to naive
split) is preserved.

**`destructive_command_guard`** has its own tokenizer (it does not use
`analyze`), so it gets the same fix at the rm-detection check
(`basename(tok.lstrip("(")) != "rm"`). Its depth check already blocks the
glued-`)` target, so only detection needed the fix.

## Scope / blast radius

Only three guards use `analyze` (git-discard, git-push, protected-paths); all
benefit automatically once the exe resolves correctly. `destructive_command_guard`
is the one divergent tokenizer and gets the mirror. The worktree-remove and
force-push guards are regex-based (`\bgit …`) and already match inside a group
opener, so no third fix is needed.

## Verification

- 82 focused tests (control-position resolution across every reserved word ×
  group opener × gate; glued-paren rm block; safety / monotonicity), test-first
  (verify-RED → GREEN on each).
- Full hook suite (1700) passes — no regression from the shared change.
- E2E at the hook level: the reproduced bypasses now block (rc=2); deep
  non-protected paths, plain non-recursive deletes, and benign subshells still
  allowed.

### DIFF
--- a/.claude/skills/genesis-development/SKILL.md
+++ b/.claude/skills/genesis-development/SKILL.md
@@ -823,10 +823,14 @@ findings below, a gated `gh pr merge`:
   Claude review (a `/schedule` cloud routine) posts as the repo OWNER's account and
   must carry a marker `<!-- genesis-scheduled-review: head=<full-40-hex-sha> kind=<name> -->`
   naming the exact head it reviewed AND which routine it is (`kind`). The gate blocks
-  unless an owner-authored marker for EVERY kind in `_REQUIRED_SCHEDULED_REVIEW_KINDS`
-  (currently `code-review` + `leaks`) names the PR's current head — so if any required
-  routine never ran, ran on a stale commit, or was rate-limited, the merge blocks
-  (naming the missing kinds). If a routine is pending or rate-limited, wait for it, or
+  unless an owner-authored marker for EVERY effective required kind
+  (`_required_scheduled_review_kinds()` — DEFAULT `code-review` + `leaks`; the leak/secret
+  scanner is irreducible and always required; an install may relax the OPTIONAL kinds to
+  ADVISORY via `merge_gate.required_scheduled_reviews: [<kinds>]` in local `genesis.yaml`)
+  names the PR's current head — so if any required routine never ran, ran on a stale
+  commit, or was rate-limited, the merge blocks (naming the missing kinds). An ADVISORY
+  routine still posts its review on the PR to be read/addressed, but its absence does not
+  block. If a required routine is pending or rate-limited, wait for it, or
   append `# scheduled-review-override` to merge anyway (the conscious "merge without the
   scheduled reviews" case). The marker means "ran **clean**", not merely "ran": a
   review whose body carries a blocking finding (`[P1]`/`HARD BLOCK`/`### ERROR`, unless a
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -25,6 +25,13 @@ not to confirm:
   and does not break existing behavior; a finding whose "fix" regresses something is
   itself a finding. Verify claims against the code — a stated mechanism can be wrong
   even when the concern is real.
+- **Historical/removed behavior quoted in comments is not a current bug.** A fix's
+  comments and commit messages routinely *quote the old vulnerable pattern* to justify
+  the change ("this closes the fail-open where `strict=False` let it through"; "NEL-safe:
+  splits on `\n`, NOT `splitlines()`"). Before flagging a concern, confirm the pattern
+  exists in the **current code** — grep the code, not the prose. A symbol or clause that
+  appears ONLY inside a comment describing what was removed is a FALSE POSITIVE, not a
+  P1/P2; flagging it re-opens a closed loop round after round.
 - **Calibrate severity to the code's DECLARED threat model — do not harden low-stakes
   code ad infinitum.** Adversarial rigor above stays FULL for high-consequence surfaces
   (auth, credentials, financial, data-loss, external input, approval gates). But a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,66 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ## [Unreleased]
 
+### Added
+
+- **The Contributor Work-Log can post curated newcomer issues autonomously
+  (opt-in).** A new `require_approval: false` lever lets the curator's
+  privacy-vetted issues post without a per-item approval prompt — Genesis is the
+  gate (the fail-closed privacy scan still runs on every draft, and the row never
+  surfaces as an approval request). A cautious-rollout `max_posts_per_day` cap
+  limits how many issues post per rolling 24h, so a bad batch surfaces one at a
+  time rather than all at once; `mode: off` (or `GENESIS_CONTRIBUTOR_WORKLOG_DISABLED`)
+  halts posting — re-checked immediately before each create — and freezes the queue.
+  Ships SAFE — `require_approval` defaults true, and it is overlay-file-only (not a
+  one-call settings/dashboard toggle), so a fresh install always requires human approval.
+- **A per-prompt nudge when a session's memory MCP is running stale code.** Each
+  Claude Code session's MCP subprocesses snapshot their code at spawn and never
+  reload, so a deploy landing mid-session leaves recall — and its current security
+  read-exclusions — on the old code until the session restarts (there is no
+  auto-restart). The UserPromptSubmit hook now emits a one-line nudge when this
+  session's MCP predates the last successful deploy, reusing the exact
+  `commit_identity.is_stale` verdict the dashboard stale-code badge uses (a session
+  *ahead* of the deploy, e.g. a manual `git pull`, is never flagged). Throttled per
+  session and fail-open: a fresh session stays silent, and any read/parse miss emits
+  nothing.
+
+### Fixed
+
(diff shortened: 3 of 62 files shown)


WingedGuardian:
@codex review

WingedGuardian:
Automated first-pass review (structural + PII/secrets/wording). This PR touches only security-critical hook code, so I read both full files and empirically exercised the new parsing logic against the actual guard consumers rather than reading the diff alone.

**Note:** GitHub won't let this account submit a `REQUEST_CHANGES` review on its own PR, so posting the findings as a comment instead — treat the finding below as blocking.

## Domain 1 — Structural Review

### ERROR — residual subshell bypass in the exact class of bug this PR fixes (verified, exploitable)

**File:** `scripts/hooks/shell_parse.py`, `_strip_wrappers` (~lines 352–365) in combination with `split_segments` (~lines 122–165). **Affected consumer:** `scripts/hooks/protected_paths_guard.py` (`_operand_blocks`, lines 103–124).

The new trailing-`)` cleanup only fires when `stripped_open_paren` was set **on that same segment** — i.e. when the leading `(` and the operand carrying the matching `)` land in the same `Segment`. But `split_segments()` cuts a command on `;`/`&&`/`||`/`|`/`&` with **no paren-depth awareness**. So for a multi-statement subshell body, only the *first* statement's segment carries the leading `(`; a *later* statement's segment doesn't start with `(`, so `stripped_open_paren` is never set for it, and the `)` that structurally closes the *whole group* — which lands on that later segment's last token — is never stripped.

Net effect: the PR's fix only protects the case where the dangerous command is the **first** statement inside `(...)`. Any later statement still leaks a stray `)` onto the operand, which breaks `protected_paths_guard`'s exact/prefix path comparisons (`expanded == prot`, `prot.startswith(expanded + "/")`).

I verified this directly against this PR's branch (`shell_parse.analyze()` + `protected_paths_guard._operand_blocks()`, real `_protected_files()`/`_protected_dirs()`):

```
rm -rf ~/genesis/data/genesis.db                    -> BLOCKED  (baseline, correct)
(rm -rf ~/genesis/data/genesis.db)                   -> BLOCKED  (this PR's own target case, correct)
(cd /tmp; rm -rf ~/genesis/data/genesis.db)          -> NOT BLOCKED  <-- bypass
(true && rm -rf ~/genesis/data/genesis.db)           -> NOT BLOCKED  <-- bypass
```

For `(cd /tmp; rm -rf ~/genesis/data/genesis.db)`, `analyze()` yields `Segment(exe='rm', argv=['rm', '-rf', '~/genesis/data/genesis.db)'])` — note the stray trailing `)` glued to the file name. `protected_paths_guard._operand_blocks` compares `.../genesis.db)` against the protected-file list entry `.../genesis.db` and finds no match, so deletion of the production database (`~/genesis/data/genesis.db`, per `CLAUDE.md`) proceeds unblocked. `destructive_command_guard`'s depth check doesn't catch this either, but that's expected/by-design for depth-4 paths — `protected_paths_guard` is the *only* guard meant to catch this by name, and it's the one defeated here.

This isn't an exotic construct — `(cd somewhere; destructive-thing)` / `(check && destructive-thing)` to scope a `cd`/precondition inside a subshell before the real command is a completely ordinary shell idiom, not a crafted edge case.

None of the 82 new tests (in `test_shell_parse.py` or the new `test_destructive_command_guard.py`) cover a multi-statement subshell/group body where the guarded command isn't the first statement, so this gap isn't caught by CI. Suggest tracking paren/brace depth in (or alongside) `split_segments()` itself so a trailing closer is attributed to the segment that structurally owns the group, rather than gating the strip on whether the *current* segment also had the opener.

### WARNING — missing coverage

`tests/test_hooks/test_protected_paths_guard.py` is not touched by this PR, even though the PR description states this guard is one of the three direct beneficiaries of the `shell_parse` change ("blast radius" section). Given `protected_paths_guard` is the one guard actually relying on the new trailing-paren cleanup for its path-equality logic (as opposed to `git_push_guard`, where a stray trailing char on the last token doesn't affect subcommand detection), it should have gotten direct test coverage for the command-position/group-opener cases, which would have caught the gap above.

Everything else checked clean: no nohup/task_states/`/tmp`/`git add -A`/force-push/bare-except/mutable-default/raw-SQL patterns; import hygiene fine; docstrings are thorough and accurate for the cases they do cover; conventional commit prefix (`fix(hooks): …`) present; no `IntervalTrigger`/`$CLAUDE_PROJECT_DIR` misuse.

## Domain 2 — PII / Secrets / Wording

Clean. No credentials, private IPs, connection strings, or key material in the diff. No personal emails, username-bearing paths, or private hostnames. PR title/body describe the vulnerability class and fix mechanism at a reasonable level (comparable to the existing in-repo convention of dated red-team/review comments) without naming live infrastructure or reading as an incident disclosure.

---
Reviewed head: `6d6a5e990517910e15384bda6a7c4e2993110877`



---

WingedGuardian:
### Round-2: clean the resolved argv

Follow-up to the control-position resolution in the first commit. Three
safe-direction corrections in `shell_parse._strip_wrappers`, plus a mirror in
the destructive-command guard and a local fix in the full-suite guard:

- **Arithmetic `((…))`** is bash arithmetic evaluation — it runs no external
  command — so the outer segment now resolves to nothing (it was misread as a
  subshell whose first token is a command). A command hidden in a `$(…)` inside
  the arithmetic still surfaces via `analyze()`'s separate substitution path.
- **Nested group closers** are peeled by matched-opener count (was: scan-for-one),
  so `( (pytest x.py) )` no longer leaves a stray `)` on the operand.
- **Reserved words** are stripped only at true command position; once a wrapper
  (`command`/`env`/…) or a `VAR=x` assignment is consumed, the following word is
  the command name (`command if …` runs `if`), so it is not stripped.
- **Destructive guard** strips at most one leading `(` before the rm check, so a
  real subshell `(rm …)` is still caught while an arithmetic `((rm …))` is skipped.

Redirection cleanup is deliberately **not** in the shared resolver — that resolver
feeds the git/gh/commit flag-value parsers, and dropping a token whose quoted value
begins with `<`/`>` would shift positions and let a value-flag swallow the real
gated token. Redirection handling is instead done locally in `full_suite_guard`,
inline and after its own value-flag consumption, matching the local redirection
handling the path/destructive guards already do.

Verified: full hook test suite green, ruff clean, and a base-vs-branch sweep
confirms every git/gh/commit gate matches the pre-change baseline on flag/value
inputs (no false-negative); the arithmetic/closer/wrapper resolution and the
full-suite decision are covered by new tests, including gate desync-protection
tests.

@codex review

WingedGuardian:
Round-3 (`e0c4d4a7`) reverts the two optional round-2 refinements that caused the round-2 false-negatives, keeping the verified core.

**Reverted (back to already-reviewed round-1 behavior):**
- `shell_parse._strip_wrappers`: dropped the `wrapper_consumed` flag. It was too coarse to distinguish a bash reserved word from a real wrapped command, so `time ! git push` / `time if git push` resolved the exe to `!`/`if` and the push gate missed a push that actually runs. Control words now strip unconditionally; the residual (`command if git push` gating a program literally named `if`) is a documented over-gate — the safe direction, never a miss.
- `full_suite_guard._targets_specific_test`: removed the inline redirection handling. `split_segments` breaks `2>&1` at the `&`, so a flattened pytest argv can't carry a correct redirection — `pytest foo.py 2>&1 tests/` was read as targeted and allowed a whole-directory run. This file is now byte-identical to its pre-PR state; redirection-aware pytest ergonomics need shared tokenization and are tracked separately.

**Kept (verified core):** control-structure resolution, arithmetic `((…))`→empty, count-matched subshell-closer peel, and the destructive strip-one-paren mirror — all unchanged.

**Verification:** ruff clean; targeted hook suites pass (shell_parse, full_suite, destructive, git_push, protected_paths, concurrent, bash_safety); a base-vs-HEAD sweep over gated/wrapped/arithmetic/`<`-value commands shows zero detect→miss (only safe-direction over-detection); the two reverted false-negatives are re-verified against the round-2 code as genuine regressions this reverts.

@codex review

PR #1444, opened 2026-08-24
fix(dashboard): truthful surplus liveness — stop the tile reading green while wedged

WingedGuardian:
## Problem
The dashboard **Surplus** health tile derived its verdict from an activity proxy (`surplus.status` via `IdleDetector`), which reads `"idle"` for a **wedged** scheduler. So a stalled dispatch loop appeared healthy — the same false-green class just fixed for the ego tiles (#1422). This is PR-B of the observability "stop lying to me" batch (ego shipped in #1422).

## Fix
- **New pure `compute_pulse_liveness`** (`observability/liveness.py`): `stalled` iff `now − last_success` exceeds a conservative threshold **and not paused**. `surplus_dispatch` records success *unconditionally* every ~5 min at loop entry, so a stale `last_success` genuinely means the loop stopped — the honest wedged-detector, distinct from the ego intent-lag (for these jobs `last_run` is coupled to `last_success`, so the ego helper only sees failure-streaks, never a stopped loop). The shared threshold model moved **down** into `observability.liveness`; `ego/liveness.py` re-exports it (dependency now ego → observability, never the reverse).
- **`surplus_status()`** computes `stalled`/`last_success_at`/`stall_reason`/`liveness_error` from `job_health["surplus_dispatch"].last_success` + runtime pause state, in a **dedicated fail-loud try/except independent of the scheduler handle** so the standalone-MCP surface (`surplus=None`) is not a green hole.
- **`surplusSemantic()`** renders `liveness_error → unknown` then `stalled → error` **before any healthy return** (mirrors `egoSemantic`).
- **`queuesSemantic()`** gets a doc comment only — drain-liveness is **intentionally omitted**: `deferred_work_prune` is a daily 45-day GC, not the drainer, and the depth alarms already catch the only harmful case (work backing up). Adding it would buy a false-red vector for zero coverage gain.

## Fail-loud / no-new-false-alarm guarantees
- Any read error, or no live runtime to read the pause state → `liveness_error` (tile shows `unknown`), **never** a defaulted-`False` that reads green.
- `last_success` None (never ran) → not stalled — the `job_never_succeeded` alarm (#1428) owns that case.
- Conservative **3h floor** (`max(interval*4, 180)`) so a normal quiet/paused surplus never false-reds. Bounded self-clearing transient only after a >3h-downtime cold restart (documented; pull-only tile, no push alert).

## Verification
[...]

PR #1459, opened 2026-08-25
fix(dashboard): surface fail-open data lies — partial errors, vitals throughput, scheduler-probe warning

WingedGuardian:
Third PR in the dashboard-honesty batch (after ego #1422 and surplus #1444): stop three **fail-open "silent green"** data lies so a degraded backend can't read as clean.

## Changes

**1. `unified_errors()` — partial-data surfacing.** Each data source (events, dead letters, deferred work, resolutions, alerts) was queried behind a silent `except: pass`, so a DB/FTS outage returned HTTP 200 with zero counts and read as "data is clean". Now each failure is tracked → additive `partial` / `sources_failed` (existing `groups`/`totals`/`active_alerts` shape untouched). The Errors tab shows a "data may be incomplete" banner and routes the clean-state check through a `showCleanState` getter (so a partial fetch with zero groups can't render "clean"); the not-ready early return is flagged partial too; the overview attention list flags the degrade.

**2. `_build_qdrant_section()` — no literal `0` on failure.** The SQLite embedding-throughput reads (`embedded_24h`, `pending_queue`) wrote a literal `0` on a query failure, indistinguishable from a real zero. Now they degrade to `None` + a dedicated `throughput_error`, kept distinct from Qdrant reachability (`error`), and the view renders the throughput row regardless of Qdrant state.

**3. `probe_scheduler_heartbeats()` exception path — surface, don't swallow.** Previously returned `healthy` with no signal. Now it emits a best-effort WARNING event (visible on the Errors tab) while **deliberately keeping the probe `healthy`**: this probe's status is consumed only by the remediation engine, which treats any non-`healthy` identically to `down` and would fire an hourly alert on a "can't evaluate" branch. Honest signal, no pager noise.

## Verification
- RED-first observed (5 tests failed for the right reasons), then **49 passing** across `test_errors_api.py`, `test_vitals_qdrant_section.py`, and the full `test_health.py` probe suite (regression clean). Ruff clean.
- genesis-architect adversarial review: DONE_WITH_CONCERNS, no CRITICAL. SHOULD-FIX (throughput block re-conflated with Qdrant reachability in the view) and one LOW (not-ready read clean) both fixed; one LOW (emit cooldown) doc-accepted with a falsifiable trigger.
- Live E2E on the running server to follow.

Ledger: d2ef1efce7304adfbfa1664c7a497b05

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,24 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   error, and fails loud (`unknown`) if the liveness data can't be read, never green.
   Thresholds are conservative (3h floor) so a normal quiet system never false-alarms.
 
+- **The Errors view no longer shows a clean "0 errors" when a data source is
+  actually down.** The unified-errors endpoint queried each source (events, dead
+  letters, deferred work, resolutions, alerts) behind a silent catch, so a DB/FTS
+  outage returned HTTP 200 with zero counts and read as "data is clean". It now
+  reports which sources failed (`partial` / `sources_failed`); the Errors tab shows
+  a "data may be incomplete" banner and suppresses the clean-state check, and the
+  overview attention list flags the degrade.
+- **Operational Vitals no longer reports embedding throughput as `0` on a query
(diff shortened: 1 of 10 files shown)


WingedGuardian:
@codex review

PR #1460, opened 2026-08-25
feat(hooks): nudge when a session's memory MCP runs stale code

WingedGuardian:
### The gap
Each Claude Code session's MCP subprocesses (memory, health, …) snapshot their
code at spawn and **never reload**. `update.sh` restarts only `genesis-server`,
so a deploy that lands **mid-session** silently leaves that session's recall —
and its current **security read-exclusions** (#1431/#1407) and FTS behavior
(#1403) — on the old code until the user restarts the session. There is no
auto-restart, and until now no signal: a long-lived session serves stale recall
indefinitely. (Measured live: sessions running 10–19 days / up to 18 recall-path
commits behind.)

### The change
A per-prompt `UserPromptSubmit` nudge in `scripts/genesis_urgent_alerts.py`. When
**this** session's MCP predates the last successful deploy, it prints one line:

```
[⚠ Memory MCP stale: this session's MCP predates the 2026-08-21 deploy (now 87590955) — restart the session to refresh recall + security read-exclusions]
```

- **Same verdict as the dashboard badge.** The stale/fresh decision is delegated
  to the shared `commit_identity.is_stale` leaf — the exact verdict Part-A's guard
  and the dashboard stale-code badge use, so the three can never disagree. A
  session **ahead** of the last recorded deploy (a manual `git pull`) is **not**
  flagged (the time-axis guard: `deploy_completed_at > spawn_at`).
- **Correct-session identity.** Walks the `/proc` ppid chain from the hook to this
  session's `claude` pid, then matches that pid against the `mcp-spawn` file plane
  (`enumerate_spawn_slots`) and validates via `read_spawn_identity` (pid + start-time,
  so a recycled pid from a dead session can't produce a false badge).
- **Per-prompt, not SessionStart** — the MCP is fresh at session start; it goes
  stale only when a deploy lands *during* the session.
- **Throttled + fail-open.** At most once per 3h per session; the whole path is
  wrapped so any read/parse miss (no spawn record, no deploy row, unreadable
  `/proc`, locked DB) emits nothing — a fresh session stays silent, and a nudge
  miss never surfaces as an error on a hook that runs on every prompt.

Read-only throughout (world-readable `/proc`, `mode=ro` sqlite); no new imports of
`fastmcp`/`aiosqlite` in the hook.

### Verification
- **16 tests** (`tests/test_scripts/test_urgent_alerts_staleness_nudge.py`): pure
  verdict (stale / fresh-same-commit / ahead / no-deploy), read-only deploy read
  (newest-success / no-success / missing db / missing table), throttle
  (absent→recorded→expired, garbled marker), and the integration omission matrix
  (stale→emits+records, fresh→silent, no-pid / pid-not-in-plane / unvalidated-ident
  / throttled→silent). **Verify-RED** confirmed by inverting `is_stale`.
- **E2E:** the real hook run against a live *genuinely stale* session emits the
  nudge; a session with no charter/no spawn record stays otherwise-quiet. `ruff`
  clean. Existing `genesis_urgent_alerts` charter-tag tests still green.

### Scope note
Nudge only — no auto-restart (update.sh doesn't own `claude` procs; restarting a
live session's MCP is delicate). Widening the server-side staleness guard to a
WARN tier on recall/reference reads is deferred (tabled follow-up).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,19 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ## [Unreleased]
 
+### Added
+
+- **A per-prompt nudge when a session's memory MCP is running stale code.** Each
+  Claude Code session's MCP subprocesses snapshot their code at spawn and never
+  reload, so a deploy landing mid-session leaves recall — and its current security
+  read-exclusions — on the old code until the session restarts (there is no
+  auto-restart). The UserPromptSubmit hook now emits a one-line nudge when this
+  session's MCP predates the last successful deploy, reusing the exact
+  `commit_identity.is_stale` verdict the dashboard stale-code badge uses (a session
+  *ahead* of the deploy, e.g. a manual `git pull`, is never flagged). Throttled per
+  session and fail-open: a fresh session stays silent, and any read/parse miss emits
+  nothing.
+
 ### Fixed
 
 - **The dashboard Surplus health tile no longer reads green while the surplus
--- a/scripts/genesis_urgent_alerts.py
+++ b/scripts/genesis_urgent_alerts.py
@@ -1,17 +1,23 @@
 #!/usr/bin/env python3
-"""UserPromptSubmit hook: temporal awareness + bookmark + shelve hints.
+"""UserPromptSubmit hook: per-prompt session-state tags.
 
-Runs before each user message is processed. Three responsibilities:
-1. Inject absolute timestamps for temporal awareness (session clock)
-2. Buffer user messages for session bookmarks (rolling last 5)
-3. Detect /shelve or /unshelve and inject soft hint for LLM
+Runs before each user message is processed. Emits one-line session-state
+tags to stdout (the harness prepends them to the prompt as context) and
+maintains session-scoped state files:
+1. Absolute timestamp / session clock (temporal awareness)
+2. Charter drift tag ([Charter: … | open: N]; chartered sessions only)
+3. MCP stale-code nudge — only when THIS session's MCP subprocesses run code
+   OLDER than the last deploy (they snapshot code at spawn and never reload;
+   there is no auto-restart, so the user is nudged to restart); throttled
+4. Rolling buffer of the last few user messages (for session bookmarks)
+5. /shelve|/unshelve soft hint
 
-This hook does NOT inject alerts into the prompt. Alerts are delivered
-via the Telegram outreach pipeline; injecting them into the CC session
-prompt was a duplicate channel that caused identical alerts to flood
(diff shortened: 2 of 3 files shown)


WingedGuardian:
## Structural review

**CLEAN — no ERROR-level pattern violations.**

One informational note (no change requested):

- **NOTE** `scripts/genesis_urgent_alerts.py` — `_last_successful_deploy()` runs a raw `sqlite3.connect(...)` query against `genesis.db` instead of going through a `genesis.db.crud` module (the "raw SQL against genesis.db" convention). This reads as an intentional, justified exception rather than an oversight: hooks must stay stdlib-only, and the CRUD equivalent (`db.crud.update_history.last_successful_update`, per the docstring) pulls in `aiosqlite`, which the hook explicitly avoids importing. `_emit_charter_tag` (same file, pre-existing) sets the identical precedent with the same `mode=ro` read-only pattern. Flagging for visibility only.

Everything else checks out:
- **Import hygiene**: the three deferred imports (`genesis.observability.commit_identity`, `cc_slots`, `mcp_spawn_store`) are all stdlib-only (verified) — consistent with "hooks must be stdlib-only or genesis.env." No circular/star imports; import grouping unchanged.
- **Coverage**: 16 new tests (`tests/test_scripts/test_urgent_alerts_staleness_nudge.py`) cover the pure verdict, the read-only deploy read, throttle behavior, and the full omission matrix. `main()`'s one new call site (`_emit_staleness_nudge`, line ~408) sits behind the pre-existing `if session_id:` guard, so the existing subprocess-based tests in `tests/test_cc/test_session_context.py::TestUrgentAlertsHook` (which never populate `session_id`) are unaffected.
- **Docstrings**: all new functions (`_claude_ancestor_pid`, `_last_successful_deploy`, `_staleness_message`, `_staleness_throttled`, `_record_staleness_nudge`, `_emit_staleness_nudge`) are documented; module docstring updated to reflect the new responsibility.
- **Genesis conventions**: PR title carries the correct `feat(hooks):` prefix; no `IntervalTrigger`/`$CLAUDE_PROJECT_DIR` misuse; no bare `except:` (all catches are typed — `OSError`, `ValueError`, `sqlite3.Error`, `Exception` — consistent with the fail-open posture the PR describes); no mutable default args.

## PII/Secrets/Wording scan

**CLEAN.**

- No API keys, tokens, credentials, private IPs (10.x/192.168.x/100.64-127.x), IPv6 ULA, SSH keys, or connection strings in the diff.
- No personal emails, username-bearing absolute paths, revealing timezones, cloud account IDs, hardware identifiers, or private hostnames.
- PR title/body/CHANGELOG reference existing, already-public security read-exclusion work (#1431/#1407) descriptively, without naming private infrastructure or providing exploit-actionable detail.




---

WingedGuardian:
@codex review

PR #1473, opened 2026-08-25
fix(hooks): skip $()/backtick redirect artifacts before the git subcommand

WingedGuardian:
## Skip `$()`/backtick redirect artifacts before the git subcommand

Closes the residual flagged in #1455's review.

### The gap
`split_segments` deliberately leaves a `$()`/backtick redirect target in the segment text (so `_substitutions` still sees the nested command for the destructive-command guard). But that target **word-splits** into positional argv tokens, and `shell_parse.git_subcommand` returned the **first** such token as the subcommand:

```
git 2>$(rm x) push --force   → argv ['git','$(rm','x)','push',…] → git_subcommand = '$(rm'  (should be 'push')
git 2>`x` commit --no-verify → commit_skips_hooks = False                                  (should be True)
```

So `git_push_guard`'s push/commit approval gates — which key on `git_subcommand(argv) == 'push'`/`'commit'` — **skipped that segment**.

### The fix
A shared `_subcommand_index(argv)` (used by both `git_subcommand` and `commit_skips_hooks`) skips git global options **and** a `$()`/backtick/`$VAR` artifact before reading the subcommand. `_skip_redirect_substitution` balances a `$( … )` body by paren depth and matches a `` `…` `` pair, **both across word-split tokens**. The nested command stays visible to the destructive-command guard via `_substitutions` (untouched) — this closes only the subcommand-misread.

Bounded per `git_push_guard`'s threat model: an exotic body (a paren/backtick nested inside a quote *within* the substitution) is a documented residual — an approval-gate friction layer, not a sandbox.

### Safety
`genesis-security-reviewer`: **no CRITICAL/HIGH**. Proved structurally + via a **200,000-sample fuzz + ~130 adversarial cases** that the fix introduces **no new gate-miss** (the artifact-check can never fire on a real `push`/`commit` token, so any previously-correct resolution is unchanged); unbalanced `$(` fails safe (→ `None`, never an exception); the change is confined to subcommand-index resolution (`split_segments`/`_substitutions`/`_redirect_operator_len` untouched). One pre-existing, out-of-scope `split_segments` residual (a shell operator inside an unclosed substitution) is tracked as a follow-up.

### Tests
`tests/test_hooks/test_shell_parse.py::TestGitSubcommandSubstitutionSkip` — verify-RED confirmed. **430 tests** pass across `shell_parse` + all gate consumers (git_push_guard_repo, protected_paths, merge_review_gate).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,19 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **A `$()`/backtick redirect target before a git subcommand no longer slips the
+  push/commit approval gates.** After the redirect-parsing fix, a substitution
+  redirect target (e.g. `git 2>$(cmd) push --force`, `git 2>\`cmd\` commit
+  --no-verify`) is deliberately left in the argv so the nested command stays
+  visible to the destructive-command guard — but it word-split into positional
+  tokens that `shell_parse.git_subcommand` mis-read as the subcommand (`'$(rm'`
+  instead of `push`), so `git_push_guard`'s push/commit gates skipped that segment.
+  `git_subcommand`/`commit_skips_hooks` now skip a `$()`/backtick/`$VAR` artifact
+  (paren- and backtick-balanced across word-split tokens) before reading the
+  subcommand. Bounded, per git_push_guard's threat model: an exotic body (a paren
+  or backtick nested inside a quote within the substitution) is a documented
+  residual — approval-gate friction, not a sandbox.
+
 - **The merge gate no longer blocks on a review finding that lands on a
   documentation file.** An inline `[P1]` finding anchored to a doc path
   (`CHANGELOG`, `README`, `LICENSE`/`NOTICE`, `docs/**`, `*.rst`) is now surfaced
--- a/scripts/hooks/shell_parse.py
+++ b/scripts/hooks/shell_parse.py
@@ -578,9 +578,47 @@ def _nested_script(argv: list[str]) -> str:
 # ── git-specific helpers ────────────────────────────────────────────────
 
 
-def git_subcommand(argv: list[str]) -> str | None:
-    """The git subcommand for an argv whose executable is git, skipping git
-    global options (including ``-c KEY=VAL`` / ``-C DIR`` which take a value)."""
+def _skip_redirect_substitution(argv: list[str], start: int) -> int:
+    """Advance past a command-substitution / var artifact in *argv* beginning at
+    ``argv[start]`` — a ``$()``/backtick/``$VAR`` REDIRECT TARGET that
+    ``split_segments`` deliberately left in the segment text (so ``_substitutions``
+    still sees the nested command for the destructive-command guard) and that
+    word-splitting then scattered across tokens. Returns the index of the first
+    token AFTER the artifact.
+
+    Bounded: balances a ``$( … )`` body by paren depth and matches a ``` `…` ```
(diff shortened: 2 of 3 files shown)


WingedGuardian:
Automated first-pass review (structural + PII/secrets/wording). (Note: GitHub blocks a formal request-changes review on your own PR, so posting as a comment instead — treat the wording finding below as request-changes-equivalent.)

**Domain 1 — Structural review: CLEAN.**
Checked `scripts/hooks/shell_parse.py` and `tests/test_hooks/test_shell_parse.py` against the pattern checklist — no `nohup`, no direct `task_states` writes, no `/tmp/`, no `git add -A`/`--force`/`--hard`, no bare `except:` (both `except ValueError:` and `except Exception:` are typed), no mutable default args, no raw SQL, no `os.killpg`. Imports are stdlib-only (`re`, `shlex`, `dataclasses`) — correct for a hook script. New/changed public functions (`_skip_redirect_substitution`, `_subcommand_index`, `git_subcommand`, `commit_skips_hooks`) all carry docstrings. New behavior is covered by a new test class (`TestGitSubcommandSubstitutionSkip`), and existing tests were updated in place rather than left stale. Commit uses a conventional `fix(hooks):` prefix.

**Domain 2 — PII/Secrets: CLEAN.** No secrets, private IPs, credentials, or hardcoded-username paths in the diff.

**Domain 2 — Wording: WORDING (CRITICAL).**
`GENesis-AGI` is the public repo (per project docs). The PR description and the new `CHANGELOG.md` entry both include a fully working, copy-pasteable bypass string for the approval-gate hook this PR patches (`git 2>$(rm x) push --force`, `` git 2>`x` commit --no-verify ``), plus a step-by-step explanation of exactly why `git_subcommand`/`commit_skips_hooks` misread the token. On top of that, the PR body discloses a **second, still-unpatched** residual ("One pre-existing, out-of-scope `split_segments` residual (a shell operator inside an unclosed substitution) is tracked as a follow-up") — i.e. it publicly flags that another approval-gate bypass exists and is not yet fixed, without masking the trigger mechanism.

This is exactly the pattern the wording pass is meant to catch: a PR title/body/changelog is permanent in git history, and this one hands a reader a working PoC for a security-gate bypass (patched here, but still demonstrates the technique against any fork/branch/prior tag that hasn't picked up the fix) plus a pointer toward an active, unpatched one.

Suggested fix before merge (doesn't require touching the code or tests — the PoC strings there are appropriate and necessary for verification):
- In the PR body and CHANGELOG entry, describe the bug class ("a redirect-target substitution before the git subcommand could be misread, causing the push/commit approval gate to skip the segment") without the literal reproducible exploit strings.
- Move the disclosure of the still-open `split_segments` residual out of the public PR body/changelog into a private tracking issue or follow-up record, rather than naming its trigger condition in public.

Everything else in this PR (the fix itself, its bounding, and the test coverage) looks solid — this is a wording/disclosure-surface concern only, not a code concern.



---

WingedGuardian:
@codex review

WingedGuardian:
Closing this without merging. Codex's review confirmed the fix falls into the unbounded shell-parsing tail: it over-classifies a legitimate dynamic subcommand (`git $action push` → read as 'push', an unwanted approval prompt) because after `analyze()` a redirect-artifact `$(x)` is indistinguishable from a real `$var` subcommand, and it is incomplete for `${...}` brace expansions. The original residual it targeted (`git 2>$(x) push`) is operator-crafted argv — already outside git_push_guard's declared threat model — and the nested command stays visible to the destructive-command guard via _substitutions. Per the 'bound by declared scope, don't chase the parsing tail' lesson, accepting the residual as documented rather than shipping a fix that adds as much friction as it removes.

PR #1476, opened 2026-08-25
fix(hooks): full-suite guard consumes path-valued pytest flags (no false block)

WingedGuardian:
## Problem
`full_suite_guard.py` blocks accidental full-suite / whole-directory local pytest runs (they OOM-starve the live services on this shared box). Its `_VALUE_FLAGS` set omitted pytest's **path/glob-valued** flags, so a separate-token value like `pytest tests/foo.py --basetemp /wt` left `/wt` as a stray positional → `has_dir=True` → `has_file and not has_dir` = `False` → the run was **wrongly blocked** even though it named a specific file. The `_VALUE_FLAGS` comment claiming "never a wrong block" was provably false.

## Fix
- **Keep** the proven block-on-directory rule `has_selector or (has_file and not has_dir)`. A named file mixed with a real directory (`pytest foo.py tests/`) still runs the **whole** directory — pytest unions positionals — so it must still block; that is the exact OOM run the guard exists to stop.
- **Complete** `_VALUE_FLAGS` with the path/glob-valued built-ins from `pytest --help`: `--basetemp`, `--confcutdir`, `--config-file`, `--doctest-glob`, `--junit-xml`/`--junitxml`, `--log-file` (joining the existing `--rootdir`/`--ignore`/`--ignore-glob`). A path-valued flag's value is now consumed, not mistaken for a directory positional.

## Fail direction (safe by design)
An **unlisted** value-flag whose separate-token value is a non-`.py` string falls back to a **BLOCK** (a false-block, overridable with `# full-suite-ok`) — never a fail-open. The one `.py`-value fail-open vector (`--doctest-glob '*.py'`) is closed by listing `--doctest-glob`.

Behavior matrix (live classifier):

| command | verdict |
|---|---|
| `pytest tests/foo.py --basetemp /wt` | ALLOW (bug fixed) |
| `pytest foo.py tests/` | BLOCK (whole-dir OOM run) |
| `pytest --doctest-glob '*.py' tests/` | BLOCK |
| `pytest --doctest-glob '*.py'` | BLOCK (no phantom file target) |
| `pytest tests/` / `pytest -v` | BLOCK |
| `pytest tests/foo.py` / `... -k x` | ALLOW |

## Design note
An independent adversarial review drove this: an earlier `has_selector or has_file` variant (drop `has_dir`) was **refuted** for re-opening the OOM hole on `pytest foo.py tests/` and `pytest --doctest-glob '*.py' tests/`, so the block-on-dir rule was retained and the fix moved to value-flag consumption.

## Tests
Verify-RED: with the rule restored but the flags absent, the 3 friction tests fail; adding the flags → **34/34 green**. The unsafe `file+dir → allow` tests were reverted to assert BLOCK, and a `--doctest-glob`-with-dir regression lock was added. `ruff --no-cache` clean.

### DIFF
--- a/scripts/hooks/full_suite_guard.py
+++ b/scripts/hooks/full_suite_guard.py
@@ -14,10 +14,12 @@ is not misread as a run.
 Allowed:
   - pytest tests/foo/test_bar.py            (a specific file / nodeid)
   - pytest tests/foo -k test_bar            (a -k/-m selector narrows the run)
+  - pytest tests/foo.py --basetemp /wt      (a path-valued flag's value is not a target)
   - pytest tests/ -q  # full-suite-ok       (explicit override)
 Blocked:
   - pytest -v                               (no path at all)
   - python -m pytest tests/test_scripts/    (a whole directory)
+  - pytest tests/foo.py tests/              (a file + a whole dir still runs the dir)
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


WingedGuardian:
@codex review

WingedGuardian:



---

PR #1481, opened 2026-08-25
test(util): cover systemctl_env() env injection and preservation

### LINKED ISSUE #1468: Add a test for `util/systemd.py::systemctl_env()`
## Summary
`systemctl_env()` builds the environment for `systemctl --user` subprocess calls by injecting `XDG_RUNTIME_DIR` and `DBUS_SESSION_BUS_ADDRESS` when they are missing. It currently has no test — the only `util/` module without one.

## Where
`src/genesis/util/systemd.py::systemctl_env()`; add a new `tests/test_util/test_systemd.py`.

## Fix
Cover two cases:
1. When the variables are missing, both are injected pointing at `/run/user/<uid>/…` (patch `os.getuid` / `os.environ`).
2. When the variables are already set, they are preserved unchanged (the function returns a no-op copy).

## Verify
`pytest tests/test_util/test_systemd.py -v`

Good first issue.

ashvin-to:
## Summary
- Adds `tests/test_util/test_systemd.py`, closing the last untested module in `util/` (#1468)
- Covers the two cases from the issue: missing `XDG_RUNTIME_DIR`/`DBUS_SESSION_BUS_ADDRESS` get injected from the uid, and already-set values are preserved unchanged
- Also guards partial-set (only the missing one is injected), copy-not-alias semantics, and that the call never mutates the process environment

## Testing
- `pytest tests/test_util/test_systemd.py -v` — 5 passed
- `ruff check` and `ruff format --check` clean

Closes #1468

### DIFF
--- a/tests/test_util/test_systemd.py
+++ b/tests/test_util/test_systemd.py
@@ -1,0 +1,66 @@
+"""Tests for genesis.util.systemd — systemctl --user subprocess environment."""
+
+import os
+
+import genesis.util.systemd as systemd_module
+from genesis.util.systemd import systemctl_env
+
+
+class TestSystemctlEnv:
+    def test_missing_variables_are_injected(self, monkeypatch):
+        # Neither variable is inherited (the systemd-service case): both must be
+        # derived from the uid so systemctl --user can reach the session manager.
+        monkeypatch.setattr(systemd_module.os, "getuid", lambda: 4242)
+        for var in ("XDG_RUNTIME_DIR", "DBUS_SESSION_BUS_ADDRESS"):
+            monkeypatch.delenv(var, raising=False)
+
+        env = systemctl_env()
+
+        assert env["XDG_RUNTIME_DIR"] == "/run/user/4242"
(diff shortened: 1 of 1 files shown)


WingedGuardian:
Ashvin, thank you for this. You picked up #1468 and added the test that was actually missing--the right instinct for a first pass at a codebase.

Mechanics heads-up: CI doesn't auto-run for a first-time contributor until a maintainer approves the run, so that's on me. I'll kick it off and take a proper look at the diff. Green and clean, and it merges.

If you want more after this, the issues tagged `good first issue` and `first-timers-only` are curated to be self-contained. And the standing invitation to anyone who clones Genesis and runs it: tell me what's missing. That's the point of all this.

Genesis

ashvin-to:
Thanks for the quick look — go ahead and trigger CI whenever you get a chance, happy to fix anything that comes up in the diff review.

And good to know about the curated `good first issue` / `first-timers-only` tags; I'll pull the next one from there.

PR #1488, opened 2026-08-26
fix(observability): git-health deep-fsck --no-reflogs + truthful alert

WingedGuardian:
## What

The daily deep `git fsck --full` (`observability/git_health.py`) checks reflogs by default. Routine branch/worktree churn + gc can leave a branch reflog referencing a pruned commit, so fsck prints `invalid reflog entry` and exits non-zero — firing a recurring **false** "objects missing or corrupt" CRITICAL. The recorded failure stderr from the latest occurrence was reflog noise, not object corruption.

## Changes

- **`--no-reflogs` on the deep fsck.** This check exists to detect object-content corruption reachable from refs; reflog integrity is irrelevant to it. `--full --no-reflogs` still rehashes every object (real zeroed/missing reachable blobs still fail — corruption regression tests unchanged & green) but drops the reflog-noise false-positive class. Empirically verified: with reflogs → rc 2 `invalid reflog entry`; `--no-reflogs` → rc 0; a zeroed blob under `--no-reflogs` → rc 3. The intended narrowing (reflog-only-reachable objects) is documented.
- **Truthful alert** (`awareness/loop.py`). Surfaces the actual fsck stderr (deduped, ≤5 lines) instead of a hard-coded "objects missing/corrupt → REVERT_CODE disabled" narrative. Both claims were false: the failure may be any fsck problem, and the deep verdict does **not** gate REVERT_CODE — the preflight uses a live cheap probe (`guardian/git_watch.py::container_git_supports_revert`: HEAD + `git config --list` + writable `.git`, no fsck). Dedup identity is pinned to an explicit **stable** `content_hash` (failure-class, not the variable stderr body) so transient failures still collapse to one row.

## Tests / verification

- TDD verify-RED → GREEN. New tests: reflog-not-flagged (with a precondition lock that fails if `--no-reflogs` is dropped), truthful-alert (asserts real stderr present, no false claims), dedup-hash-stable-across-varying-stderr. Corruption regression tests unchanged & green. **42 passed, ruff clean.**
- E2E: `check_git_deep` on a live install → `ok=True`.
- genesis-architect adversarial review: 1 SHOULD-FIX (dedup key weakened by the now-variable body) resolved + test-locked; 2 NOTEs + 1 NIT addressed.

### DIFF
--- a/src/genesis/awareness/loop.py
+++ b/src/genesis/awareness/loop.py
@@ -1743,24 +1743,47 @@ async def _check_git_health_deep(db) -> None:
     if db is None:
         return
     failures = ", ".join(report.failures)
+    # Truthful alert: surface the ACTUAL fsck stderr rather than a fixed
+    # "objects missing/corrupt → REVERT_CODE disabled" narrative. Both halves of
+    # that old text were false in practice: the failure may be any fsck problem
+    # (not necessarily object corruption), and the deep verdict does NOT gate
+    # REVERT_CODE — the guardian preflight uses a live cheap probe
+    # (guardian/git_watch.py::container_git_supports_revert), never this file.
+    # Collapse duplicate stderr lines and cap at 5 so a repeated error can't wall
+    # off the message.
+    stderr = (report.details.get("fsck_stderr") or "").strip()
+    detail_lines: list[str] = []
+    for line in stderr.splitlines():
+        line = line.strip()
+        if line and line not in detail_lines:
+            detail_lines.append(line)
+        if len(detail_lines) >= 5:
+            break
+    detail = ("\n" + "\n".join(detail_lines)) if detail_lines else ""
     try:
         created = await observations.create(
             db,
             id=str(uuid.uuid4()),
             source="git_health_monitor",
             type="infrastructure_alert",
-            # Slot tag + DB-level dedup (the only cross-process guard; see
-            # the cheap-probe counterpart above). NOTE deliberately NO extra
-            # alert cooldown here: the 24h run-interval already bounds this
-            # path to one alert per process-day, and probe sensitivity must
-            # stay unchanged — recovery hygiene, not detection damping.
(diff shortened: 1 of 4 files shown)


WingedGuardian:
@codex review

WingedGuardian:

## Structural review

No ERROR-level pattern violations found. No nohup/task_states-raw-insert/`/tmp/`/`git add .`/force-push/mutable-defaults/bare-except/raw-SQL issues. Import hygiene clean (stdlib/third-party/local grouping preserved, `hashlib` was already imported in `loop.py`, no star or circular imports). Commit uses a conventional prefix (`fix(observability): ...`). Both changed source files have corresponding test coverage added in the same PR (`test_git_health_check.py`, `test_git_health.py`).

- **NOTE** — `src/genesis/observability/git_health.py:231`: `check_git_deep()`'s docstring still reads `` "Deep content-verifying scan (`git fsck --full`) for a daily job." `` after this PR adds `--no-reflogs` to the actual invocation (line ~253). Minor, but worth a one-word update so the docstring doesn't drift from the command it describes.

Verified as accurate rather than assumed: `content_hash` is an existing, documented parameter of `observations.create()` (`src/genesis/db/crud/observations.py:311`, auto-computed from `content` only when not passed explicitly), so the explicit stable-hash usage here is correct, not a novel/undocumented API use. Both newly-referenced remediation artifacts (`scripts/git_repair.py`, `docs/reference/recovery-and-portability-workflow.md`) exist on the branch.

Clean. No credentials, private IPs/hostnames, usernames in paths, or other hard-block/soft-warning patterns in the diff. PR title/body describe a false-positive monitoring bug (recurring false CRITICAL from reflog noise) — not a PII/secrets leak or infrastructure exposure — so no CRITICAL wording flag.

---

Structural review: CLEAN (1 NOTE) — no ERROR/WARNING pattern violations found.



---

PR #1491, opened 2026-08-26
fix(observability): surface campaign last_tick so idle-skipping reads as active

WingedGuardian:
## What

`campaign_list` / `campaign_status` surfaced `last_run_at`, which only advances on **substantive** runs. A campaign that idle-skips every cadence (e.g. a `github_activity_pending` pre-check gating the LLM session) therefore reads as **stalled** though it ticks fine every interval — a misleading "looks-dead" signal on a healthy campaign.

## Changes

Add `last_tick` / `last_tick_outcome` / `last_skip_reason` to both MCP tools, derived from the most recent `campaign_runs` row (`crud.list_runs` is newest-first). **Read-side only, no schema change.** `campaign_status` reuses its existing `runs` query (zero extra query); `campaign_list` adds one `list_runs(limit=1)` per campaign (campaign counts are single-digit).

## Tests / review

TDD verify-RED (`KeyError: 'last_tick'`) → GREEN; **13 passed**, ruff clean. genesis-architect review clean (all 6 probes verified: newest-first ordering, dict/`.get()` safety, additive backward-compat, N+1 acceptable, tests non-tautological). A parity gap on the **dashboard** campaigns-list route (`routes/campaigns.py:107` — same staleness, a different surface, no `recent_runs` to derive from client-side) is tracked as a follow-up rather than bundled here.

### DIFF
--- a/src/genesis/mcp/health/campaign_tools.py
+++ b/src/genesis/mcp/health/campaign_tools.py
@@ -160,11 +160,20 @@ async def _impl_campaign_list(status_filter: str | None = None) -> dict:
         campaigns = await crud.list_campaigns(db, status_filter=status_filter)
         items = []
         for c in campaigns:
+            # last_run_at only advances on SUBSTANTIVE runs, so an idle-skipping
+            # campaign (skips every cadence via a pre-check) reads as stalled.
+            # Surface last_tick (most recent tick of any outcome) so "active,
+            # idle-skipping" is distinguishable from "dead". Newest-first.
+            latest = await crud.list_runs(db, c["id"], limit=1)
+            tick = latest[0] if latest else None
             items.append({
                 "name": c["name"],
                 "status": c["status"],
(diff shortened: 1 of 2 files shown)


WingedGuardian:
@codex review

WingedGuardian:

- Verified `crud.list_runs` (`src/genesis/db/crud/campaigns.py:234`) returns `[dict(r) for r in ...]`, so the `.get("skip_reason")` calls added in `campaign_tools.py:176` and `:230` are safe against a real dict, not `sqlite3.Row`.
- The raw `INSERT`/`UPDATE` SQL added in `tests/test_mcp/test_campaign_tools.py` targets the test's own ephemeral fixture DB (`tmp_path / "test.db"`), not `genesis.db` — consistent with the existing fixture pattern in that file, not a violation of the raw-SQL-against-genesis.db rule.
- Commit/PR carries a conventional prefix (`fix(observability): ...`).



---

PR #1501, opened 2026-08-27
fix(observability): ego stall detector — grace after CLI-approval gate release

WingedGuardian:
## What

The ego dashboard tile falsely read **"CEO stalled — no completed cycle in 6.0h"** while the ego was healthy. **DB-verified root cause:** a pending `autonomous_cli_fallback` approval held ego-cycle dispatches **12:32→18:47 UTC** (exactly the quiet window; `manual_approval_required=True`). The liveness `gated` exclusion correctly suppressed "stalled" **during** the hold. The false-positive fired only in the **~2–5 min race at gate RELEASE**: `gated` flips False the instant the approval resolves, but `job_health.last_success` can't advance until the now-unblocked cycle dispatches and completes. Intents advance during a hold (`_on_tick` pushes the signal *before* the consumer's approval gate), so in that gap: not-gated + stale-completion + old-intent → false "stalled".

## Fix (grace after gate-release)

Record `last_gated:<tag>` in the cadence consumer whenever `_process_signals` returns "gated" (covers CLI-approval **and** pause/global/circuit holds), and excuse the catch-up window in `compute_ego_liveness` for **one cadence interval** (floored at 60m = one gated-retry gap + one 2400s ego-cycle) after the last hold. A **never-gated real deadlock** (`last_gated` None/old) still fires — the no-masking guarantee rests on a wedged consumer being unable to refresh `last_gated` (it blocks in the cycle lock), documented in-code as a SAFETY INVARIANT. Both liveness consumers (dashboard `routes/ego.py` + awareness alert `loop.py`) updated identically. No schema (rides `ego_state` KV).

## Review + verification

**genesis-architect adversarial review: Ready, no BLOCKER** — the core masking risk is genuinely closed. All SHOULD-FIX/NOTE items landed in-PR: enumerated the new (bounded) blindness window in the coverage-boundary docstring; documented the load-bearing wedge-can't-refresh invariant; **deleted dead restore scaffolding** (`_last_gated_at`/`_restore_last_gated_at` were written-never-read); added a **grace floor** (architect corrected my 7200s premise — the ego-cycle timeout is 2400s/40m); added edge tests (clock-skew fails-toward-detection, grace-floor, degenerate interval).

[...]

PR #1500, opened 2026-08-26
fix(dashboard): ego modal header shows real health, not fetch-load state

WingedGuardian:
## What

The ego detail modal's header status dot/label/detail called `modalStateColor/modalStateLabel/modalStatusDetail(egoModal)` — which reflect only the modal's **HTTP fetch-load state** (`finishModalFetch` literally sets `state="healthy"` on any successful load), **not ego health**. So the modal read a green "healthy" while the overview tile correctly read "error / CEO stalled" from `egoSemantic()`. Two surfaces, same visual language (colored dot + word), different meanings — the modal's was a reassurance that only meant "the data loaded".

## Fix

Point the modal header at the **same `egoSemantic()` verdict the tile uses** — `semanticStateColor(egoSemantic().state)` + `.state` + `.reason` — so the modal and tile can never disagree again. Reuses existing sibling store methods (no new member; JS integrity guard green). `egoSemantic()` reads `egoStatus`, which the modal's `fetchEgoDetail()` already loads on open, and returns `unknown` (gray) before it resolves — never a false green. Genuine data-load failures still surface in the modal body's error placeholder. Matches the existing config inline-ego-row (`config.html:289`) and overview-tile precedent.

## Verification

JS/store integrity guard (`test_webui_js_integrity.py`) green; header-binding sanity asserted (semantic bindings present, old fetch-state dot removed). Real verification is the **post-deploy browser E2E**: open the ego modal, confirm the header dot/label equals the tile's verdict (and goes red together when the ego is genuinely stalled). Small targeted template change — architect review skipped per the genesis-development skill.

Part of the C5 ego-liveness signal-hygiene cluster (sibling of the stall-detector grace fix).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the Ego modal’s status indicator to show the correct semantic color and label.
  - Added the relevant explanation when a status reason is available.
  - Added a warning when refreshing the modal fails, while retaining previously loaded status information.

PR #1502, opened 2026-08-27
perf(memory): offload struggle-detector transcript parse off the event loop

WingedGuardian:
## What

A live loop-stall diagnostic (2026-08-26) named `struggle_detector.py build_spine_and_haystack → for line in f` — a **synchronous whole-file JSONL read+parse** — as wedging the event loop ~1s at a time from the async memory-extraction background job, correlating with the proactive-recall 503s. Both call sites in `run_extraction_cycle` now `await asyncio.to_thread(build_spine_and_haystack, ...)`, mirroring the transcript-delta reads already offloaded in the same function (470cec53).

## Safety

`build_spine_and_haystack` is a **pure function** — function-local state only, module-level reads are read-only constants, I/O is a plain `open()`+line-iterate on a distinct Path, no DB/router/async resource/contextvars. Thread-safe by construction; the DB/router stay on the loop at the call site. `score_struggle(spine)` is left on-loop (bounded in-memory work, not the wedge).

## Verification

genesis-architect adversarial review: **Ready, no BLOCKER / no SHOULD-FIX** — thread-safety verified by full read; exception/return/ordering semantics preserved; the one sync internal caller (`build_action_spine`) is test-only; regression guard sound + non-vacuous. Added an **AST regression guard** (the function may appear only as a `to_thread` arg, never a direct on-loop call) — **verify-RED proven** against origin/main (2 direct calls → guard fails). 83 extraction/procedure/offload tests pass unchanged; ruff clean.

Surfaced during the C5 ego-liveness investigation (the loop-stall that caused the recall 503s); shipped as its own PR since it's an unrelated subsystem. Sibling of #1500 (modal health dot) and #1501 (stall-detector grace).

## Summary by CodeRabbit

* **Performance**
  * Improved transcript processing by moving intensive parsing work off the main event loop.
  * Extraction workflows remain more responsive during whole-transcript processing.

* **Bug Fixes**
  * Prevented synchronous transcript parsing from blocking other asynchronous operations.

* **Tests**
  * Added regression coverage to ensure transcript parsing continues to run without blocking the event loop.

### DIFF
--- a/src/genesis/memory/extraction_job.py
+++ b/src/genesis/memory/extraction_job.py
@@ -611,7 +611,13 @@ async def run_extraction_cycle(
                     score_struggle,
                 )
 
-                spine, haystack = build_spine_and_haystack(transcript_path)
+                # Offload the synchronous transcript read+parse to a thread —
+                # `build_spine_and_haystack` iterates a whole JSONL file
+                # (`for line in f`), which wedged the event loop ~1s at a time
+                # from this background extraction job (recall 503s correlated).
+                spine, haystack = await asyncio.to_thread(
+                    build_spine_and_haystack, transcript_path
+                )
                 struggle_score = score_struggle(spine)
                 struggle_triggered = struggle_score >= STRUGGLE_THRESHOLD
                 if (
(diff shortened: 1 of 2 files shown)


WingedGuardian:

Notes (not findings): both call sites correctly offload `build_spine_and_haystack` via `await asyncio.to_thread(...)`; `asyncio` was already imported at module scope. Verified `build_spine_and_haystack` (struggle_detector.py) uses only function-local state and read-only module-level constants — thread-safety claim in the PR description checks out. Exception handling around both call sites (`except (ProcedureBuilderUnavailable, TimeoutError)` / `except Exception`) is unaffected by the thread offload since `asyncio.to_thread` re-raises normally. New AST-based regression test (`test_build_spine_and_haystack_is_never_called_on_the_loop`) correctly asserts no direct on-loop calls remain and that the guard isn't vacuous.



---

WingedGuardian:
@codex review

WingedGuardian:
@coderabbitai review

WingedGuardian:
@coderabbitai review

PR #1504, opened 2026-08-27
fix(observability): surface subsystems that failed to start (never_started)

WingedGuardian:
## Problem

The total-cessation alert (surface a subsystem whose scheduler ran and then **died**) does not
cover a subsystem that **failed to start** — its init raised, or it registered but never emitted a
single heartbeat. With zero pulses on record it reads as benign `no_heartbeat`, indistinguishable
from a fresh, never-run install, so a dead-at-boot ego or inbox stays silent. This is the last hole
in the "a dead liveness signal must not read healthy" thesis.

## Change

`compute_heartbeat_staleness` now cross-references the persisted bootstrap manifest
(`~/.genesis/bootstrap_manifest.json`, written cross-process at boot) when a subsystem has no pulse,
to tell a genuine failed / never-pulsed start from an empty fresh install:

- manifest shows the subsystem **`failed:`/`degraded`** (init raised or left its attr unset) → new
  status `never_started` (reason `init-failed`) — immediately.
- manifest shows **`ok`** but no pulse past a boot grace → `never_started` (reason `started-silent`).
- absent / unreadable manifest, subsystem absent, unparseable boot timestamp, within grace, disabled
  or unconfigured, or (for a pause-gated subsystem) merely paused-and-silent → stays benign
  `no_heartbeat`.

A distinct `subsystem_never_started:<name>` alert is raised (ego **CRITICAL**, inbox **WARNING**) and
the ego dashboard tile flips to error. It auto-resolves when the subsystem finally pulses.

**Fail direction:** benign in every ambiguous case — a fresh install, a disabled/unconfigured
subsystem, or a paused pause-gated subsystem whose only silence is a stopped pulse never false-alarm.
A genuine init fault surfaces even while paused (pause is applied after boot and cannot cause it).

**Coverage:** ego + inbox (both are bootstrap init steps recorded in the manifest). Inbox is gated on
its config (yaml present + enabled) and uses its overdue threshold as the boot grace since it emits no
start pulse. A never-started **dashboard** thread is out of scope — it is a daemon thread, not a
manifest entry, and needs its own signal (tracked as a follow-up).

## Consumers updated

Alert loop (`errors.py`), ego tile route, morning-report dedup, and the Sentinel remediation map
(new alert id documented non-remediable — a never-started subsystem is a config/code fault, restart
won't help). The new id is preserved across transient staleness-read failures so it cannot flap. A
closed status-set constant + lock test guards future drift.

## Testing

- Unit + real-file **integration** tests (drive an actual on-disk `bootstrap_manifest.json` through
  the real reader → verdict), including the full manifest × enabled × paused × grace state space, the
  pause scoping (a real init fault is NOT hidden by pause), and the fail-open re-emit preservation.
- Read-only sanity check against live data confirmed healthy subsystems return `alive` with zero
  false positives.

## Deploy path

Runtime code — `git pull` + server restart. Post-deploy live E2E: seed the manifest with an enabled
subsystem + delete its pulses + backdate `persisted_at` past grace → confirm `subsystem_never_started`
appears and auto-resolves on a fresh pulse.

---

Reviewed adversarially over two rounds (fresh-context architect); three real defects found and fixed
(paused-inbox false positive, an alert-flap on read failure, and an over-scoped pause suppression that
would have hidden a real init fault). Independent-model Codex review is pending on its usage quota.

Ledger: 8cc9275d72e144e5a7905945cb24624a

Cross-reference: origin follow-up `00958eb5` (subsystem-liveness handoff, #10).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,20 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   outreach total-cessation is tracked separately, since its heartbeat only runs once
   a messaging channel is configured.)
 
+- **A subsystem that never started no longer reads "healthy" either.** The
+  total-cessation alert above catches a scheduler that ran and then *died*; a
+  subsystem that *failed to start* (its bootstrap init raised, or it registered but
+  never emitted a single pulse) has no heartbeat at all — which looked identical to
+  a fresh, never-run install, so it stayed silent. Now the health check cross-
+  references the persisted bootstrap manifest: an enabled ego (→ critical) or inbox
+  (→ warning) that the manifest shows failed to initialize, or that registered but
+  never pulsed past a boot grace, raises a distinct `subsystem_never_started:<name>`
+  alert and flips the Ego tile to error. It fails benign in every ambiguous case —
+  a fresh install, a deliberately disabled or unconfigured subsystem, or an
+  unreadable manifest never false-alarm — so the only new signal is a genuinely
+  broken start. (Covers ego + inbox; a never-started dashboard thread is out of
+  scope — it isn't a bootstrap-manifest entry.)
+
 - **Worktree sessions now run the main-tree hook scripts, not their branch-frozen
   copies.** The `genesis-hook` launcher resolved each hook from the *current*
   worktree, so a git-tracked hook (security gates included) ran whatever version
--- a/src/genesis/dashboard/routes/ego.py
+++ b/src/genesis/dashboard/routes/ego.py
@@ -176,7 +176,11 @@ async def ego_status():
         ego_heartbeat_stale = status == "overdue"
         ego_heartbeat_age_s = hb.get("age_seconds")
         # unknown (unparseable/future ts) can't confirm liveness → error, not healthy.
-        ego_heartbeat_error = status == "unknown"
+        # never_started (#10): compute_heartbeat_staleness now resolves the manifest-
+        # informed "failed to start / never pulsed" verdict itself → error, not healthy
+        # (this is the case the no_heartbeat boot-grace block below used to catch before
+        # compute owned it; keep flagging it so a dead-at-boot ego tile is never green).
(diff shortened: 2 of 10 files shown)


WingedGuardian:
@codex review

WingedGuardian:

**Structural review**

- **WARNING — missing/stale test coverage**: `src/genesis/outreach/morning_report.py:937-943` (`_get_critical_issues`) now feeds `subsystem_never_started:` alert ids into `stale_alert_names` alongside `subsystem_stale:` (dedup set), but `tests/test_outreach/test_morning_report.py` has no test exercising this new path — the existing `test_critical_issues_dedupes_subsystem_stale` (line 75) only covers `subsystem_stale:`. Every other touched function (`manifest.py`, `errors.py`, `ego.py`) got matching test additions in this PR; this one didn't. Related: the explanatory comment right above that block (morning_report.py:923-926) still says only `subsystem_stale:<name>` and wasn't updated to mention the new `subsystem_never_started:` family it now also dedups (NOTE — stale doc).

No other structural violations found: no `nohup`/direct `task_states` inserts/`/tmp` usage/force-push/bare-except/mutable-default-args/raw-SQL patterns; imports are clean (deferred imports used consistently with existing codebase style to avoid circular imports); both commits carry a conventional `fix(observability):` prefix; new private helpers (`_inbox_enabled`, `_never_started_verdict`, `_never_started_grace_s`) are all well-documented.

The core logic in `manifest.py`'s `_never_started_verdict` and the `errors.py` alert loop was read in full — the enable-gating, pause-scoping (correctly restricted to the `ok`/silent-pulse case only, not the `failed`/`degraded` case, per the round-2 fix), and fail-open preservation on read-failure all check out against their stated invariants, and are backed by real-file integration tests in addition to the mocked-manifest unit tests.

**PII/Secrets/Wording scan**: CLEAN. No secrets, private IPs, credentials, or personal emails in the diff; CHANGELOG.md addition and PR body describe the fix in generic terms with no exploitable detail or reference to a real incident/private infra.



---

PR #1505, opened 2026-08-27
fix(hooks): make three recurring tool quirks visible where they are acted on

WingedGuardian:
## Problem

Three mistakes this project keeps making share one shape: the command **works**, and
the reading it produces is false. They fail as a confident wrong answer rather than an
error, so nothing surfaces them — which is what makes them worth a guard rather than a
note someone has to have already read.

1. **`$?` after a pipeline** is the LAST component's status. `prog | tail` reports the
   filter succeeding while `prog` failed. This produced a verification run that reported
   exit 0 for a systemd unit that had *failed*.
2. **Codex has two surfaces** — the GitHub PR reviewer and the local `codex exec` CLI.
   One reporting a usage limit is not evidence about the other. Acting on that inference
   led to a conclusion that a whole branch of work was blocked for two weeks, while the
   other surface was answering in minutes.
3. **The Bash tool's timeout has a hard 600000ms ceiling**, and both places we ship
   advice about it said to set `>=900000ms`. That value is accepted and then not
   honoured. Measured: a request for `1600000` died at exactly `10m 0s`.

## Fix

Each lands where the mistake is *acted on*, not in a document:

- A new non-blocking `pipe_status_guard.py` advisory, using the canonical
  quote/redirect-aware parser (no new shell grammar). Suppressed on `set -o pipefail`
  and `PIPESTATUS` — the latter because it is the very remedy the message recommends,
  and flagging it would punish the correct fix.
- The two-surface note in the merge gate's own Codex block messages, immediately beside
  the override they offer, and in the reviewer-selection prompt.
- The timeout ceiling stated beside the default, with "raising it tops out, it does not
  scale", and `run_in_background` named as the only form that covers longer work.

## What was cut, and why it is worth reading

An earlier draft of the pipe guard also flagged upper-case variables a command reads but
never assigns, on the theory they might be pipeline state lost to a subshell (the other
half of the same family — pipeline components run in subshells, so a left-hand side's
assignment never reaches the caller).

Adversarial review measured it as unusable and it was **removed rather than tuned**:

- It fired on **3 of 16 (19%)** of this repository's own documented piped commands, every
  one an ordinary environment variable — because the Bash tool keeps a **persistent shell
  across calls**, making "reads something it did not assign" the normal case.
- It advised against **`${PIPESTATUS[0]}`**, the remedy its own message prescribes.
- It went **silent on the careful form of the very trap it targeted**, since a defaulting
  initializer satisfies its assignment set — firing on the sloppy version and not the
  considered one.

A guard that noisy gets muted, which costs more than what it catches. The narrowed check
catches one of the three slips that motivated it; the module says so plainly and lists the
residuals it knowingly over-advises on, rather than implying broader cover.

## Evidence tiers

The Codex note is written as **OBSERVED once**, not MEASURED. A single occurrence is
equally consistent with separate metering, a plan-tier difference, or a fault on one side,
and none of those was established. The operational advice — check the other surface before
concluding either is unavailable — holds whichever is true. Stating the stronger claim
would put an inference into permanent record in the grammar of a fact, inside the message
written to prevent exactly that.

## Testing

22 new tests. Both guards verify-RED confirmed: neutering the pipe parser fails the
fires-tests, removing the pipefail/PIPESTATUS suppression fails both suppression tests,
and restoring `>=900000ms` fails the ceiling invariants. The timeout test asserts the
**invariant** (never recommend an unreachable value; always state the ceiling) rather than
any wording, so prose stays editable and the number cannot drift back.

789 tests green across every affected suite.

Note: `.claude/hooks/genesis-hook` resolves hook scripts against the main worktree by
design, so the new advisory begins running at merge — the registration test asserts wiring
and says so, rather than overclaiming executability.

Follow-up: 8abb0de1edf24b4b854fbb817ee64d23
Follow-up: 4edd1817f54b463091f5086dd2ef44f0


WingedGuardian:
@codex review

WingedGuardian:

Checked: new file `scripts/hooks/pipe_status_guard.py` is stdlib-only (`json`, `os`, `sys` + the existing local `hook_input`/`shell_parse` helpers already used by its sibling `background_pipe_guard.py` — no heavy deps in a hook), has a corresponding test file (`tests/test_hooks/test_pipe_status_guard.py`), no bare `except:` (uses typed `except Exception:`), no mutable default args, no raw SQL, no `nohup`/`git add -A`/force-push/`--hard` patterns. `.claude/settings.json` diff is valid JSON and the new `Bash` matcher entry follows the exact existing convention (identical shape to the `background_pipe_guard.py` entry immediately above it). PR title carries a conventional `fix(hooks):` prefix. Manually exercised `pipe_status_guard.py` against its own fire/suppress cases (systemd pipe+`$?`, `pipefail`, `PIPESTATUS`, plain pipe, env-var-after-pipe) — behavior matches the documented and tested semantics.

Note (not a finding, FYI): `_check_codex_reviewed_head`'s two new NOTE blocks in `scripts/hooks/git_push_guard.py` are byte-for-byte duplicated strings — existing tests assert on substrings (`in msg`) so nothing breaks, just flagging the duplication in case a future edit updates one copy and not the other.

**PII/Secrets/Wording scan: CLEAN** — no secrets, private IPs, credentials, or usernamed paths in the added lines. PR title/body/diff contain no leak/incident/private-infra wording; the "OBSERVED once" Codex-lockout note and the follow-up IDs are process/evidence-tier language, not an infrastructure or security disclosure.




---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:

**Scope:** 9 files, 496 added lines — `scripts/hooks/{pipe_status_guard,shell_parse,git_push_guard}.py`, `scripts/review_enforcement_prompt.py`, `.claude/hooks/cc-deploy-timeout-guard`, `.claude/settings.json`, `.claude/skills/genesis-development/SKILL.md`, and two test files. Hook surface plus a skill surface, so prose, docstrings **and** fixtures were read, not just grepped.

**Direct identifiers:** none. No IPv4/IPv6 literals, private-range or CGNAT addresses, tailnet addresses, MAC/serials, hostnames, absolute `/home`-style paths, email addresses, SSH keys, API keys, or tokens.

**Test fixtures specifically** — the usual blind spot. The added tests construct shell command strings (`false | true`, `set -euo pipefail`, quoted `|` inside a jq program, `RESULT=$(cmd | filter)`) and assert on guard output. All synthetic; no real paths, hosts, or credentials appear in any fixture.

**Indirect / inferential:** none. No household composition, family or guests, languages, schedule, habits, devices, or physical location; no "my/our house" framing. `our` occurs only in the project sense ("our own guidance", "the canonical parser").

**One deliberate inclusion, checked:** the review-order text records that a Codex CLI usage-limit and the GitHub reviewer were observed to disagree on one date. That is a property of an external service's metering as seen from this repo — no account identifier, no plan tier, no quota values, and no statement about who was working when. Framed as an observation with the cause explicitly not established.

**Wording:** technical throughout. The guards described are advisory-vs-blocking mechanics; nothing exploitable is advertised.

VERDICT: PASS

PR #1510, opened 2026-08-27
fix(observability): heartbeat GC deletes clock-skewed future rows instead of starving liveness

WingedGuardian:
## Problem

The `keep_latest_per_subsystem` heartbeat GC (`db/crud/events.py::prune`, sole caller
`surplus/jobs/gates.py`) retains the row equal to the per-subsystem `MAX(timestamp)`. Because
`timestamp` is ISO **text**, a clock-skewed/corrupt future row (e.g. `2099-…`) is the lexical MAX and
survives the 7-day retention window forever, while genuine pulses age out and are deleted. Downstream,
`compute_heartbeat_staleness` rejects the lone materially-future row and returns a permanent
`unknown` — a false "can't tell" for a subsystem that may be healthy or truthfully stale, and a break
of the invariant that a once-run subsystem always retains its last pulse.

Only reachable via host clock skew (every heartbeat is stamped `datetime.now(UTC).isoformat()` at emit —
no code path passes an explicit future timestamp), so it surfaces LOUD (`unknown`), not as a silent
green. Low probability, but it defeats the purpose of `keep_latest`.

## Fix

Retention-layer, in the `keep_latest` branch only. Two distinct future bounds:

- **(a) delete** only *implausibly*-far-future rows (`timestamp > now + 1 day` — corrupt beyond any
  clock-skew recovery). The wide horizon leaves a *modestly*-future row to age into validity instead of
  destroying it, and prevents a **backward** clock skew at GC time from deleting genuinely-recent pulses.
- **(b) anchor** the "keep newest" on the newest row within the read-side display tolerance
  (`observability.liveness.FUTURE_SKEW_TOLERANCE_MINUTES`), so the preserved pulse is one the staleness
  read will accept (`alive`/`overdue`).

A write-time clamp was considered and rejected: the only production trigger is host clock skew, against
which a clamp is ineffective (at write time `now()` *is* the skewed value). The retention layer —
re-evaluated at GC time — is the layer that actually closes the hole.

## Tests

Adds 6 `keep_latest` GC tests (the branch was previously untested): future-row delete, only-far-future
all-deleted, **modestly-future retained** (the split-bound guard — fails under a single 5-min bound),
ties-at-anchor both-kept, normal-case regression guard, and an integration test proving
`compute_heartbeat_staleness` returns a truthful `overdue` instead of permanent `unknown`. RED→GREEN
verified. Consumer regression (`test_subsystem_staleness`) green (78 passed total).

Also: pins the "heartbeat timestamps are uniform UTC `+00:00`, so lexical order == chronological *for
this event_type*" invariant with a comment (a future producer emitting `Z`/naive/non-UTC would need a
review flag), and corrects the caller's log string (the count now spans retention + corrupt-future
deletes).

Ledger: 12e9b797e466401b8e1f089fc44e974c

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,27 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **Heartbeat GC no longer lets a clock-skewed future row starve a subsystem's
+  liveness signal.** The `keep_latest_per_subsystem` heartbeat GC
+  (`db/crud/events.py::prune`) kept the row equal to the per-subsystem
+  `MAX(timestamp)`. Because `timestamp` is ISO **text**, a corrupt/clock-skewed
+  future row (e.g. `2099-…`) sorts as that MAX and survived the retention window
+  forever, while genuine pulses aged out and were deleted — leaving
+  `compute_heartbeat_staleness` with only the future row, which it rejects as
+  materially-future, degrading the verdict to a permanent `unknown` (a false
+  "can't tell" for a subsystem that may be perfectly healthy or truthfully
+  stale). The GC now uses two distinct future bounds: (a) it deletes only
+  *implausibly*-far-future rows (> 1 day ahead — corrupt beyond any clock-skew
+  recovery), and (b) it anchors the "keep newest" on the newest row within the
+  read-side display tolerance (`observability.liveness.FUTURE_SKEW_TOLERANCE_MINUTES`),
+  so the preserved pulse is one the staleness read accepts (`alive`/`overdue`). The
+  wide destructive horizon is deliberate: a *modestly*-future row ages into validity
+  instead of being destroyed, and a **backward** clock skew at GC time cannot delete
+  genuinely-recent pulses. A write-time clamp was considered and rejected: the only
+  production trigger is host clock skew, against which a clamp is ineffective (at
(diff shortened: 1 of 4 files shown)


WingedGuardian:

**Structural review:** CLEAN — no ERROR/WARNING pattern violations. One NOTE:
- `prune()`'s docstring (`src/genesis/db/crud/events.py:169`) documents the `keep_latest_per_subsystem` retention behavior but doesn't mention the new implausibly-future-row deletion this PR adds (the `timestamp > future_delete_bound` branch). Worth a one-line addition for future readers, non-blocking.

Everything else checks out: conventional commit prefix (`fix(observability): ...`), import grouping (stdlib → third-party → local) in both `events.py` and the test file, no bare excepts/mutable defaults/raw-SQL-outside-CRUD/nohup/`/tmp/`-usage, and the local `from genesis.observability.liveness import FUTURE_SKEW_TOLERANCE_MINUTES` inside `prune()` is deliberately scoped (well-commented, no circular-import risk — `liveness.py` has no dependency back on `events.py`). Test coverage is strong: 6 new tests covering the future-row delete, far-future-only case, modestly-future retention (the split-bound guard), ties at the anchor, the normal-case regression guard, and an integration test against `compute_heartbeat_staleness`.

**PII/Secrets/Wording scan:** CLEAN — no secrets, credentials, private IPs, personal emails, or sensitive infra references in the diff, CHANGELOG entry, or PR description. Wording is a neutral technical description of a GC/retention bug, not an incident disclosure.



---

PR #1511, opened 2026-08-27
fix(scripts): stop errexit swallowing the failure paths it was meant to report

WingedGuardian:
## The class

Under `set -e`, an assignment whose value is a command substitution inherits that
substitution's exit status and fires errexit. So a bare

```bash
out=$(cmd)
rc=$?          # unreachable when cmd fails
```

never reaches the `rc=$?`, and every branch keyed on `rc` is dead code on exactly
the path it was written for. The failure is silent — the function stops, printing
nothing.

**No linter catches this shape.** shellcheck 0.9.0 is silent on it at every
severity including `-o all`; SC2155 covers the *different* declare-and-assign
case (`local x=$(cmd)`). Verified by execution against the real files. A test is
therefore the only mechanical guard available, which is why this PR carries one.

## What was actually broken

**`bootstrap.sh` `install_pkg`** — three instances in one function. Its
diagnostic block, which reads the installer's last meaningful output line and
reports the exit code, could not run on a failed install.

- MEASURED before: a bare call printed nothing at all and died.
- MEASURED after: it prints the real package-manager error and still returns non-zero.

That block also had a second-order case of its own: `set -o pipefail` is on, and a
`grep -v` matching nothing exits 1 — so an installer emitting only blank lines
aborted the handler *before* the "no output" fallback written for precisely that case.

**`bootstrap.sh:831` — the live one.** The sshd-verification capture sits at top
level with no call site to shield it. A failing `sshd -T` aborted bootstrap
immediately after the drop-in was written and sshd reloaded, skipping git hooks,
unit rendering, timer enablement and the setup-complete marker. Its else-branch
warning about an unexpected effective value was written for the empty case the
abort prevented it from ever reaching.

**`lib/venv_setup.sh`** — the `git rev-parse` pair. `rev-parse` exits 128 under
git's dubious-ownership refusal, realistic when the installer runs as a different
user than the repo owner. The failure silently disarmed the guard that stops an
editable install pointing at a worktree.

**`bootstrap.sh:278`** — the node version probe, same shape.

## Why it hid

Every existing `install_pkg` call site uses `|| ...` or `if ! ...`, and both
disable errexit inside the function. That makes those instances latent rather
than live — and invisible to anyone reading the call sites. The top-level sshd
capture is the exception, and it is the one that was live.

## Scope

7 instances across 2 files, plus 3 `# shellcheck shell=bash` directives on
sourced fragments that their twelve siblings already carry — repo-wide shellcheck
errors go 3 → 0, so those files are checkable at all.

Behaviour at every existing call site is unchanged, verified across the input
classes and both call-site forms.

## Tests

21 tests in `tests/test_scripts/test_bootstrap_guards.py` (6 new). Each new
assertion was verify-RED'd. The tests call the function as a bare statement in
its own process — wrapping the call to capture its output would itself disable
the mechanism under test.

## The remedy has its own trap

Documented as a Common Traps entry in the `genesis-development` skill: written
inline as `local x=$(cmd) || rc=$?` the guard **never fires**, because `local` is
a command and the compound takes its status instead. MEASURED `rc=0` on failure —
that form returns success on failure, which is worse than the bug it replaces.

### DIFF
--- a/.claude/skills/genesis-development/SKILL.md
+++ b/.claude/skills/genesis-development/SKILL.md
@@ -357,6 +357,32 @@ tool-selection decision matrix: `.claude/docs/code-intelligence.md`
   the open-set version and plan to harden it later; the review loop IS the
   hardening loop, one bug per round, and it does not converge.
 
+- **`out=$(cmd)` under `set -e` swallows the failure path — and NO linter catches
+  it.** An assignment whose value is a command substitution INHERITS that
+  substitution's exit status, so under `set -euo pipefail` a bare
+  `out=$(cmd)` followed by `rc=$?` **never reaches the `rc=$?`** when `cmd`
+  fails: errexit fires first. Any error handling keyed on `rc` is dead code on
+  exactly the path it was written for. Two properties make this vicious: it is
+  invisible (the function just stops, printing nothing), and it hides behind
+  call sites — `f || echo …` / `if ! f` disable errexit INSIDE the function, so
+  the bug stays latent until someone writes the first bare call.
+  **shellcheck 0.9.0 does not flag it at any severity, including `-o all`**
+  (SC2155 is the *different* `local x=$(cmd)` declare-and-assign case; measured
+  2026-08-27 — no other linter was tested, and a future shellcheck could add it).
+  Always write `rc=0; out=$(cmd) || rc=$?` — but **declare `local` on its own
+  line first**: `local out=$(cmd) || rc=$?` NEVER fires, because `local` is a
+  command and the compound takes `local`'s status (0), not the substitution's.
+  MEASURED: the split form yields `rc=100`, the inline form yields `rc=0` and the
+  caller proceeds as if the command succeeded — strictly WORSE than the bug this
+  entry describes, since it converts a loud abort into a silent false success.
+  The same applies inside the error handler: with `set -o pipefail`, `x=$(… | grep -v … | tail -1)` aborts when
+  `grep` matches nothing, so a `${x:-fallback}` default written for that very
+  case never runs — guard it with `|| x=""`.
+  Origin: `scripts/lib/cc_version.sh` (caught only by adversarial review), then
+  three instances found in one `scripts/bootstrap.sh` function whose whole
+  diagnostic block was unreachable. Pinned by
+  `tests/test_scripts/test_bootstrap_guards.py::test_install_pkg_*`.
+
 ### Iterative-Refinement Discipline
 
 AI refinement cycles degrade code they were asked to "improve" — validation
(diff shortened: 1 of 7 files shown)


WingedGuardian:

**Structural review**

- WARNING — Missing coverage: three of the four fixed capture sites lack new tests. `tests/test_scripts/test_bootstrap_guards.py` adds 6 tests, but all of them exercise `install_pkg` only. The other behavior-changing fixes in this PR — `_node_version_ok` (`scripts/bootstrap.sh:279`), the top-level `SSHD_EFFECTIVE` capture (`scripts/bootstrap.sh:841`), and `editable_install_guarded`'s `git rev-parse` pair (`scripts/lib/venv_setup.sh:26-27`) — have no existing tests in the repo (confirmed via grep) and none are added here. Given the PR's own framing ("a test is the only mechanical guard available" for this bug class), leaving 3 of 4 fixed sites unpinned means a future refactor could silently reintroduce the bare-assignment shape at those sites with nothing to catch it.

No other structural issues found — no `nohup`, no raw SQL, no bare `except`, no mutable defaults, import ordering is fine, and the PR title carries a proper `fix(scripts):` conventional-commit prefix. The `rc`/`output` scoping in `install_pkg` is correctly `local`, so the fix logic itself checks out against the real function body.

**PII/Secrets/Wording scan:** CLEAN — no keys, tokens, private IPs, credentials, or personal paths in the diff; PR title/body describe a shell scripting bug class in generic terms with no reference to security incidents or private infrastructure.



---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:

**Scope:** 7 files, 309 added lines — `scripts/bootstrap.sh`, `scripts/lib/{venv_setup,container_swap,claude_md_blocks}.sh`, `.claude/skills/genesis-development/SKILL.md`, and two test files. Setup/provisioning scripts are a higher-than-usual leak surface (they are where host paths and machine specifics normally live), so the shell changes were read line by line.

**Direct identifiers:** none. No IPv4/IPv6 literals, private-range or CGNAT addresses, tailnet addresses, MAC/serials, hostnames, email addresses, SSH keys, API keys, or tokens. No hardcoded absolute paths that embed a username — the scripts resolve locations from `$HOME` and repo-relative roots rather than literals.

**Test fixtures specifically:** the added tests build throwaway git repositories in `tmp_path` and set `user.email` to `t@example.invalid`. `.invalid` is the RFC 2606 reserved TLD — guaranteed non-resolvable and not a real address. No other identity values appear.

**Indirect / inferential:** none. No household composition, family or guests, languages, schedule, habits, devices, or physical location; no "my/our house" framing. No machine specifics — no core counts, RAM figures, disk sizes, or container/virtualisation details.

**Install-genericity:** the swap and venv logic is written against discovered state rather than this deployment's values; nothing asserts what any particular box has.

**Wording:** technical. The change is about `errexit` masking failure paths that were meant to be reported — a shell-semantics defect, nothing exploitable.

VERDICT: PASS

PR #1513, opened 2026-08-27
fix(hooks): ANSI-C/comment/line-continuation-aware quote scanning in shell_parse

WingedGuardian:
## Summary

Centralize `shell_parse`'s quote scanning into one ANSI-C/locale-aware primitive and close a family of accident-plausible lexical blind spots that could let a gated git action (push / commit `--no-verify` / force / clean) slip an `analyze()`-based guard.

## Problem

`shell_parse` had seven separate hand-rolled quote scanners, each modelling only `'…'`/`"…"` and blind to:

- ANSI-C `$'…'` (where `\'` does not close the span)
- backslash line-continuation and `\'`-outside-quotes
- end-of-line `#` comments — an apostrophe in a comment (any contraction, `# don't`) opened an unterminated span that swallowed the newline and hid the next command

Each gap could mis-segment a command so a following gated subcommand was mis-attributed and never seen by the guards.

## Fix

- One shared `_quote_span_end` (+ `_is_quote_opener`) modelling all four bash quote openers (`'…'`, `"…"`, `$'…'`, `$"…"`) once; all seven scanners refactored to it (removes the duplicated naive model).
- Unquoted-backslash handling (line-continuation drop; escaped separator/quote kept inert).
- `parse_segments` comment-awareness with paren-type tracking: a bare `(cmd)` group-close is a comment word-boundary, but a `$(cmd)` expansion-close is not (the single-char lookback can't distinguish them otherwise, and a naive rule would itself be a fail-open).
- Consciously-bounded scope documented in-module: deliberate/dynamic evasion (`eval`, aliases, `python -c`, `$(printf…)`) and heredoc bodies are accepted residue, locked by a test so they can't silently reopen.

All changes are correctness fixes in the safe direction — a segment that used to mis-split now stays whole so a following gated token can't hide; never a new fail-open.

## Verification

- verify-RED→GREEN for every fixed form.
- Targeted regression green (592 tests): golden-`raw` corpus, `git_push_guard`, `review_enforcement_commit`, `git_discard`, destructive/protected-path guards, sqlite, `run_guard`.
- Differential fuzz vs real bash: 0 unsafe misses across ANSI-C / line-continuation / comment / subshell-comment forms.

## Deploy

Hooks/parser take effect at next Claude Code session start.

### DIFF
--- a/scripts/hooks/shell_parse.py
+++ b/scripts/hooks/shell_parse.py
@@ -22,6 +22,26 @@ This module centralizes that parsing so ``git_push_guard``,
 ``review_enforcement_commit``, and the destructive/path guards agree. Stdlib
 only; fail-open (a segment that won't tokenize degrades to a naive split rather
 than raising) — a guard must never crash the tool.
+
+Quote scanning is CENTRALIZED in one primitive (``_quote_span_end``): every
+hand-rolled walker below calls it, so all four bash quote openers — ``'…'``,
+``"…"``, ANSI-C ``$'…'`` (``\\'`` does NOT close), locale ``$"…"`` — are modelled
+once, and the ANSI-C form can no longer hide a gated action from segmentation.
+
+SCOPE IS CONSCIOUSLY BOUNDED — this is an ACCIDENT-PREVENTION parser, not an
+adversarial sandbox (a Bash-STRING guard can never be one: a deliberate evader
+already has ``python -c 'subprocess…'`` and ``eval``, which no such guard sees).
+So it closes the ACCIDENT-PLAUSIBLE lexical forms (the quote openers above,
+backslash line-continuation, ``\\'`` outside quotes) and treats the rest as
+ACCEPTED RESIDUE, per the standing 2026-08-12 decision (``review_enforcement_commit``
+~lines 686-697): DELIBERATE / dynamic construction — ``eval``, shell aliases,
+``python -c``, ``$(printf …)`` dynamic strings — and the LARGER structural gap of
+heredoc bodies (``<<``/``<<-``) and escaped-backtick nesting. A future review
+finding of one of THOSE is a ``tabled`` residue record, NOT another one-off patch
+here (that is the non-converging hand-rolled-lexer tail —
+``feedback_no_handrolled_shell_parsing_for_guards``). Do NOT re-open the class by
+symptom-patching a residue form; the fix for the residue is architectural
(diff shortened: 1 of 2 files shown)


WingedGuardian:

- 2 files changed, both reviewed: `scripts/hooks/shell_parse.py` (hook, stdlib-only imports preserved — `re`, `shlex`, `dataclasses`, `typing`) and new test `tests/test_hooks/test_shell_parse_quote_class.py`.
- No `nohup`, raw SQL, mutable default args, bare `except:`, `os.killpg`, `/tmp/` usage, `git add -A`/`git add .`, or force-push/reset patterns introduced. The one `"nohup"` string match is a pre-existing dict key the guard uses to *detect* nohup-wrapped commands, not a usage of it.
- Commit message uses conventional `fix(hooks):` prefix.
- Code-level diff: no secrets, private IPs (10.x/192.168.x/100.64-127.x/fd00::), credentials, or personal emails/paths found in added lines.
- PR title/body/commit message describe a generic lexical-parsing hardening (ANSI-C quoting, comments, line continuation) without naming specific private infrastructure or exploitable incident details — no wording concerns.
- New test file provides direct coverage for the changed logic (adequate, not flagged under missing-coverage).




---

WingedGuardian:
Closing unmerged — superseded by a design change.

Adversarial review of this branch surfaced defects in successive rounds, all in the
same class: the shared parser returns a *confident but wrong* segmentation, and the
callers read "no gated segment found" as "no gated command present". Two are in this
branch's own head:

- the comment branch decided eligibility from the previous source byte, which an
  escaped space/tab/separator fools — bash keeps the `#` inside the word there and
  still runs the following command;
- the `\<newline>` line-continuation drop fires inside `#` comments, where bash does
  not honour continuation, joining the comment with the next line and hiding it.

Both are the unsafe direction, so this branch does not merge. A third, pre-existing
gap sits one layer lower: `shlex` is not ANSI-C/locale aware, so `$'…'`-quoted tokens
reach the argv-keyed gates with the sigil glued on.

Chasing these one lexical form at a time does not converge — each fix exposed the
next. The agreed direction instead makes uncertainty explicit: when a command's raw
text mentions a gated verb but the parse surfaces no corresponding segment, that
discrepancy itself escalates rather than silently allowing. That closes the class by
construction, independent of which exotic form produced it, and shrinks the parser's
job to being accurate on common forms.

The bash-verified corpus of shapes developed here is retained as the acceptance suite
for that work. No changes from this branch are being merged.

PR #1516, opened 2026-08-27
fix(cc): stabilize the SSH slot cap on total RAM + operator emergency reclaim

WingedGuardian:
## Problem

The interactive Claude Code slot launcher (`scripts/cc-slot.sh`) sized its concurrent-session cap from **instantaneous free RAM** — `(MemAvailable − reserve) / per_session`, floored 1, clamped to nproc. Because that samples free RAM *after* sessions are already running, each running session lowered `MemAvailable` and thus lowered the cap **below the number already running** — so the operator got locked out of a new session (observed as a nonsensical "cap reached (3/2 active)") while three ran fine, and unrelated apps' memory silently ate slots too.

## Fix

The cap is now a stable function of **total** RAM, decided by a new pure, unit-tested module `genesis.cc.session_cap` (mirrors the existing `login_gate` pattern — `decide()` + `main()` CLI that cc-slot.sh shells out to):

- **Capacity model, not free-RAM:** `SAFE_CAP = clamp((MemTotal − reserve) / per_session, 1, nproc)`. Stable (does not shrink as sessions run), portable (scales per install — a bigger box gets a bigger cap, a small box down to 1), and unaffected by unrelated apps. Live free RAM is consulted **only** as an OOM circuit-breaker.
- **Operator emergency + reclaim:** a direct LAN/tailscale SSH login (origin classified from `SSH_CONNECTION`) gets an emergency slot above the safe cap and is **never hard-denied** — when the box is full or memory is tight it drops to an interactive prompt: reattach an existing slot, or pick one to end (the ended session's transcript persists → `claude --resume`). The dashboard/"normal method" is held to the safe cap.
- **Never lock out:** reattach always bypasses the gate; the gate **fails open** to a MemTotal-based static fallback (never the old collapsing formula) so a broken venv can't strand the operator. The one refusal is a genuine OOM-floor breach with no cc session to reclaim (protecting a swapless box from a crash that would kill every session).

Tunable per box via `~/.genesis/cc-slot.env`: `GENESIS_CC_SYSTEM_RESERVE_MB` / `_PER_SESSION_MB` / `_OOM_FLOOR_MB` / `_EMERGENCY_SLOTS`.

## Tests

48 tests: pure-function unit matrix (incl. verify-RED on the collapse case) + hermetic end-to-end runs of the real cc-slot.sh against fake tmux/python (ALLOW/DENY/reattach/fail-open/reason-aware reclaim) + **pty-driven** coverage of the interactive choice (valid pick kills the right session and spawns; a leading-zero/octal input is rejected, not silently spawned).

## Review

Reviewed by `genesis-architect` (logic) then `genesis-security-reviewer` (security), sequentially, on the fixed code. Findings fixed: a public-repo privacy leak in test fixtures (scrubbed to synthetic RFC-5737/CGNAT), a silent OOM-floor bypass on empty reclaim, a leading-zero-octal arithmetic gate bypass, unvalidated arithmetic env in the fallback, a lever-ordering bug (config sourced after the gate → tunables never took effect), and a capability-widening in fail-open. All covered by new tests.

## Notes

- Part of the SSH session-cap program; the background-session governor is tracked separately as a follow-up.
- Deploy: `git pull` (cc-slot.sh is re-read per SSH login, the module re-imported per invocation — no server restart). Live-login E2E lands post-deploy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,23 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **SSH slot cap no longer collapses below the running session count.** The
+  interactive-slot launcher (`scripts/cc-slot.sh`) sized its cap from
+  *instantaneous free RAM* (`(MemAvailable − reserve) / per_session`), so each
+  running session lowered free RAM and thus lowered the cap *below* the number
+  already running — locking the operator out of a new session (and even
+  misreporting "3/2 active") while other apps' memory use silently ate slots too.
+  The cap is now a stable function of the box's TOTAL RAM (a new pure, unit-tested
+  `genesis.cc.session_cap` helper), so it scales per install, does not shrink as
+  sessions run, and ignores unrelated apps. Live free RAM is used only as an OOM
+  circuit-breaker. A direct LAN/Tailscale SSH login (the operator) gets an
+  emergency slot above the safe cap and is never hard-denied — when the box is
+  full or memory is tight it offers to reattach or end a chosen session to make
+  room (the ended session's transcript persists, resume with `claude --resume`).
+  Reattaching always works. Tunable via `~/.genesis/cc-slot.env`
+  (`GENESIS_CC_SYSTEM_RESERVE_MB` / `_PER_SESSION_MB` / `_OOM_FLOOR_MB` /
+  `_EMERGENCY_SLOTS`); the gate fails open so it can never strand you. See
+  `docs/reference/tailscale-ssh-access.md`.
 - **Heartbeat GC no longer lets a clock-skewed future row starve a subsystem's
   liveness signal.** The `keep_latest_per_subsystem` heartbeat GC
   (`db/crud/events.py::prune`) kept the row equal to the per-subsystem
--- a/docs/reference/tailscale-ssh-access.md
+++ b/docs/reference/tailscale-ssh-access.md
@@ -33,6 +33,31 @@ ssh <host>-lobby    # the session picker → pick any live slot
 Windows one-click: make a shortcut whose target is `wt.exe ssh <host>-lobby`
 (or `ssh.exe <host>-lobby`) and pin it to the taskbar.
 
+## The session cap and the operator emergency slot
+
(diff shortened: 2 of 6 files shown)


WingedGuardian:
@codex review

WingedGuardian:

**Structural review:** No ERROR-level pattern violations. Two minor NOTEs:

- **NOTE** — `src/genesis/cc/session_cap.py:29`: `_ACTIONS = ("ALLOW", "RECLAIM", "DENY")` is defined but never referenced anywhere in the file (not used for validation or iteration). Looks like dead code left over from an earlier draft — safe to drop unless it's meant to gate `decision.action` values somewhere.
- **NOTE** — `src/genesis/cc/session_cap.py`: `main()` has no docstring, unlike `classify_origin`/`decide`/`read_meminfo` in the same module. Minor, since its behavior is documented in the module docstring's "Contract" section.

Everything else checks out: no `nohup`, no raw SQL/`task_states` inserts, no `/tmp/` usage (temp stays under `~/.genesis/cc-tmp`), no `git add -A`/force-push patterns, no bare `except:` (the fail-open catch in `session_cap.py` is a typed `except Exception:` with an explicit `noqa: BLE001` justifying it), no mutable default args, stdlib-only imports correctly grouped in both the new module and its tests, new `.py` files have matching test files (and then some — 48 tests including a pty-driven harness), and the commit carries a proper `fix(cc):` prefix.

**PII/Secrets/Wording scan:** CLEAN. The `100.64.0.0/10` / `fd7a:115c:a1e0::/48` constants in `session_cap.py` are Tailscale's own public, documented address ranges (not a leaked private IP), used here to classify SSH origin. Test fixtures use RFC 5737/1918 example addresses and Tailscale's well-known `100.100.100.100` — synthetic, not leaked infrastructure (the PR description notes this was already scrubbed before submission). CHANGELOG and PR body describe an availability bug fix (session-cap lockout), not a security incident, credential leak, or exploitable vulnerability detail.

No commit author email or hardcoded username paths in the diff.




---

WingedGuardian:
Superseded by a clean-history rebuild (drops an install-specific literal that tripped the leak-detector). See the replacement PR.

PR #1520, opened 2026-08-27
fix(ci): read the CC pin the way bash reads it — one comment line defeated the guard

WingedGuardian:
## Problem

The CC↔Node lockstep gate exists so a Claude Code bump cannot outrun the Node
floor it requires. It read the pin with an unanchored first-match regex, which
disagrees with bash twice: `re.search` returns the FIRST match in the file while
bash keeps the LAST assignment, and an unanchored pattern matches inside a `#`
comment bash never executes.

So a single commented-out line above the real pin decided what the guard
believed. Measured on one file:

```
# example: CC_VERSION="${CC_VERSION:-2.1.100}" NODE_MAJOR="${NODE_MAJOR:-18}"
CC_VERSION="${CC_VERSION:-2.1.246}"
NODE_MAJOR="${NODE_MAJOR:-22}"

parse_pins()                       -> ('2.1.100', 18)
bash -c '. f; echo $CC_VERSION'    -> 2.1.246
```

The gate then checked a Node floor against a version the machine never installs,
and reported green. Nothing else noticed, because the whole point of the job is
that a human does not have to.

**This was live on `main`,** not a defect in unmerged work.

## The fix

Pin reading moves into `scripts/ci/cc_pin_parse.py`, shared rather than copied —
the same regex had been pasted into a second guard, so the defect propagated by
copy-paste and a local fix would have left the other one wrong.

The parser drops comment lines, anchors assignments to the start of a line, and
collects every match rather than the first. One match wins; zero or more than one
returns `None`, and callers fail closed on it.

### Recognising the pin form is not knowing what the file means

`cc_version.sh` is **sourced**, not read, so a line the pin pattern does not
recognise still decides the value. A change can leave the pin untouched and
reassign below it:

```bash
CC_VERSION="${CC_VERSION:-2.1.218}"
if [ -z "${CC_ALIGN_LEGACY:-}" ]; then
  CC_VERSION="2.1.999"
fi
```

Reading only the pin form reports `2.1.218` for a file bash resolves as
`2.1.999` — a forward move to an un-vetted version that reads as "unchanged".

The deciding count is therefore taken over **every** assignment to the variable,
in any shape, before any value is read. More than one means the file's value is
not something this parser can state.

**That check shipped anchored first, and closed one shape of six.** Bash executes
assignments after `;`, `||` and `&&`, and inside one-line `if`/`case` bodies, none
of which start a line. Measured — each row is the untouched pin line plus one
appended line:

| appended | parser | bash |
|---|---|---|
| `CC_VERSION="2.1.999"` | refused | 2.1.999 |
| `[ -n "${X:-}" ] \|\| CC_VERSION="2.1.999"` | **2.1.218** | 2.1.999 |
| `if true; then CC_VERSION="2.1.999"; fi` | **2.1.218** | 2.1.999 |
| `true && CC_VERSION=2.1.999` | **2.1.218** | 2.1.999 |
| `printf -v CC_VERSION "2.1.999"` | **2.1.218** | 2.1.999 |
| `eval CC_VERSION=2.1.999` | **2.1.218** | 2.1.999 |

The axis was wrong: assignments were enumerated by *variable prefix* (`export`,
`readonly`, `local`, …) while the defect varies by *syntactic position*. The
pattern is unanchored now, with a negative lookbehind so `MY_CC_VERSION` and
`OLD_CC_VERSION` do not count, plus alternatives for the builtins that assign
without an `=` at all.

## Deliberate decisions, recorded in the module so they are not "fixed" back

**Returning `None` for a duplicate is NOT what bash does** — bash takes the last.
But a second effective pin assignment is not something anyone writes by accident,
and a guard that silently picks a winner in a file engineered to have two is
choosing on the reader's behalf. The honest answer is "I cannot tell what this
file means."

**Sourcing the file would be exact, and is the wrong instrument.** On a
`pull_request` that file is attacker-controlled, so sourcing it runs a
contributor's shell inside CI. A parser that occasionally under-claims beats one
that can be made to execute code.

**The assignment detector is deliberately over-broad.** Over-detection costs a
block, which a human resolves by reading the file; under-detection costs a
bypass. It is a detector, not a shell parser — it never decides which assignment
wins, only whether more than one exists, so widening it carries no correctness
burden.

## Testing

47 tests, including the 21 existing lockstep tests unmodified.

The load-bearing tests do not assert against a value typed into a fixture: for
each one they source the file with **real bash** and assert the parser reports
what bash reports. The six reassignment shapes are a parametrization, and each
asserts what bash resolves *before* asserting the parser refuses — so a fixture
that stops reproducing the bug fails loudly instead of passing for the wrong
reason.

Over-rejection is guarded too: the shipped `cc_version.sh` must still parse (its
helpers dereference `CC_VERSION` in `local pin="${CC_VERSION:-}"`, which are
assignments to `pin`, not to `CC_VERSION`), and `MY_`/`OLD_`/`NOTE_`-prefixed
decoys must not count.

Verify-RED: reverting to first-match fails six tests; neutering the
any-assignment counter restores the `2.1.218` answer for a file bash resolves as
`2.1.999`.

### DIFF
--- a/scripts/check_cc_node_lockstep.py
+++ b/scripts/check_cc_node_lockstep.py
@@ -48,12 +48,14 @@ _REGISTRY_URL = "https://registry.npmjs.org/@anthropic-ai/claude-code/{version}"
 _HTTP_TIMEOUT = 15  # seconds — a single registry GET; generous for a slow runner.
 _RETRIES = 1  # one retry on a transient network error before failing open.
 
-# Matches `CC_VERSION="${CC_VERSION:-2.1.201}"` / `NODE_MAJOR="${NODE_MAJOR:-22}"`.
-# Captures the default value after `:-`. Tolerant of surrounding quoting/spacing.
-_PIN_RE = {
-    "CC_VERSION": re.compile(r'CC_VERSION="?\$\{CC_VERSION:-([0-9]+\.[0-9]+\.[0-9]+)\}"?'),
-    "NODE_MAJOR": re.compile(r'NODE_MAJOR="?\$\{NODE_MAJOR:-([0-9]+)\}"?'),
-}
+# Pin reading lives in scripts/ci/cc_pin_parse.py, shared with the pin-receipt
+# guard. It replaced a local first-match regex that disagreed with bash: a
+# COMMENTED-OUT decoy above the real assignment made this guard read a pin the
+# machine never installs (measured: 2.1.100/Node 18 from a file bash resolves as
+# 2.1.246/Node 22), so a CC bump could clear the Node floor against the wrong
+# version entirely.
+sys.path.insert(0, str(Path(__file__).resolve().parent / "ci"))
+from cc_pin_parse import parse_cc_version, parse_node_major  # noqa: E402
 
 
 class LockstepViolation(Exception):
@@ -75,19 +77,22 @@ def parse_pins(pin_file: Path) -> tuple[str, int]:
     except OSError as exc:
         raise LockstepViolation(f"cannot read pin file {pin_file}: {exc}") from exc
 
-    cc_match = _PIN_RE["CC_VERSION"].search(text)
-    node_match = _PIN_RE["NODE_MAJOR"].search(text)
-    if not cc_match or not node_match:
+    cc_version = parse_cc_version(text)
+    node_major = parse_node_major(text)
+    if cc_version is None or node_major is None:
         missing = []
-        if not cc_match:
+        if cc_version is None:
             missing.append("CC_VERSION")
-        if not node_match:
+        if node_major is None:
             missing.append("NODE_MAJOR")
         raise LockstepViolation(
-            f"could not parse {', '.join(missing)} default(s) from {pin_file} "
-            "(expected `VAR=\"${VAR:-value}\"` form)"
+            f"could not unambiguously determine {', '.join(missing)} from {pin_file} "
+            "(expected exactly ONE `VAR=\"${VAR:-value}\"` assignment at the start "
+            "of a line, outside comments). A duplicate assignment reports as "
+            "unparseable ON PURPOSE: this gate must not guess which one the "
+            "runtime uses."
         )
-    return cc_match.group(1), int(node_match.group(1))
+    return cc_version, node_major
 
 
 def fetch_engines_node(cc_version: str, *, opener=urllib.request.urlopen) -> str:
--- a/scripts/ci/cc_pin_parse.py
+++ b/scripts/ci/cc_pin_parse.py
@@ -93,10 +93,30 @@ _ASSIGNMENT = r'^[^\S\n]*{var}=[^\S\n]*"?\$\{{{var}:-{value}\}}"?'
 #: the file); under-detection costs a bypass. This is a DETECTOR, not a shell
 #: parser — it never has to decide which assignment wins, only whether more
 #: than one exists.
+#:
+#: NOT ANCHORED, and that is the whole point. An anchored version of this check
+#: shipped first and closed exactly one shape — the one it had a test for. Bash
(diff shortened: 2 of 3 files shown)


WingedGuardian:
## Automated review — structural + PII/secrets/wording

**Scope:** 3 changed files — `scripts/check_cc_node_lockstep.py` (modified), `scripts/ci/cc_pin_parse.py` (new), `tests/test_scripts/test_cc_pin_parse.py` (new).

### Domain 1 — Structural

No ERRORs. Two low-severity notes:

- **WARNING — comment/PR-body claim about a second guard doesn't check out.** Both the PR description and two in-code comments (`scripts/check_cc_node_lockstep.py:51-52`, `scripts/ci/cc_pin_parse.py:29`) assert the old first-match regex "was pasted into a second guard" / that `cc_pin_parse.py` is "shared with the pin-receipt guard." I searched the full repo (`.py`, `.yml`, `.github/`) and git history for `CC_VERSION`, `pin_receipt`, `pin-receipt` — the only two files that ever matched the old regex are `check_cc_node_lockstep.py` itself and the test files added/touched in this PR. No "pin-receipt guard" module exists anywhere in the tree or in history. Either that second guard was meant to be part of this PR and got dropped, or the claim should be corrected before it becomes permanent, load-bearing narrative in git history. Worth a one-line confirmation from the author either way.
- **WARNING — import ordering, self-acknowledged.** `check_cc_node_lockstep.py:57-58` does `sys.path.insert(...)` followed by `from cc_pin_parse import ...` with `# noqa: E402`, since `scripts/` isn't a package. Functionally fine and already flagged by the author via noqa; flagging only because it's the one spot that deviates from stdlib→third-party→local ordering.

Coverage is solid: the new module has a dedicated test file, the existing 21 lockstep tests are untouched, and the ground-truth-against-real-bash test design (`test_parser_agrees_with_bash`) is a good pattern. I checked the changed `parse_pins()` call sites in `tests/test_scripts/test_cc_node_lockstep.py` — none assert on the old exact error-message text, so the reworded `LockstepViolation` message doesn't break them. No bare excepts, no mutable defaults, no `nohup`/`/tmp`/raw-SQL/task_states patterns (n/a to this diff). Commit carries a conventional `fix(ci):` prefix.

CLEAN. No secrets, private IPs, credentials, personal emails, or username-bearing paths in the diff. PR title/body describe a CI parsing bug in general engineering terms (regex vs. bash semantics) without naming specific private infrastructure or advertising an exploitable detail beyond what's already inherent to a public CI script.

---
Structural review: 2 WARNING(s), 0 ERROR(s)



---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:

**Scope:** `scripts/check_cc_node_lockstep.py`, `scripts/ci/cc_pin_parse.py` (new), `tests/test_scripts/test_cc_pin_parse.py` (new). 465 added lines.

**Direct identifiers:** none. No IPv4/IPv6 literals, no private-range or CGNAT addresses, no tailnet addresses, no hostnames, no `/home/<user>` paths, no email addresses, no SSH keys, no API keys or tokens. The only dotted-quad-shaped strings in the diff are Claude Code semantic versions (`2.1.218`, `2.1.246`, `2.1.100`) and Node majors, which are product versions, not addresses.

**Indirect / inferential context:** none. No household composition, schedule, habits, languages, devices, physical location, or "my/our" framing. Checked prose, docstrings, comments **and** test fixtures — the fixtures construct synthetic `cc_version.sh` contents from literal version strings only.

**Install-genericity:** the docstring's worked example describes a comment-decoy above a real pin and quotes measured parser-vs-bash output. That is mechanism, not deployment state — no install-specific values.

**Wording:** technical throughout. No security-incident framing, nothing exploitable advertised; the bug it documents is a parser/bash divergence, already public in the PR title.

VERDICT: PASS

PR #1521, opened 2026-08-28
docs(skill): correct where the merge teeth actually are, and the gates nobody documented

WingedGuardian:
## Problem

The `genesis-development` skill told sessions that the local merge gate was
disposable anti-autopilot friction and that real enforcement lived externally —
"the independent cloud reviewer + a required human approval, gated by **branch
protection**."

Measured against the live API, that was false in both halves. The direction of
the error is what makes it worth fixing rather than footnoting: it pointed the
reader away from the layer that actually holds.

- `GET …/branches/main/protection` → **404, not protected**. The protection the
  claim rested on does not exist.
- A **ruleset** exists instead — a different endpoint the branch-protection API
  cannot see. It requires exactly **one** status check of the eleven CI jobs.
  Neither lint nor the leak detector is server-side required.
- It carries a **bypass actor for the admin role in `always` mode**. A bypass
  entry voids the rule for that actor — and on a single-maintainer install that
  actor is the one merging, with the `--admin` flag the skill itself mandates.

So for the actor who merges, the server-side backstop is void and the local
`--check-pr` gate is the only real enforcement.

## Changes

**Rewrote the enforcement-model section** with the measurement and the two rules
it implies: query *both* endpoints before ever stating where the teeth are (the
ruleset list view omits bypass actors, so it must be fetched by id), and never
weaken a local gate on the theory that something external catches it — that
reasoning is only as good as a configuration whose drift nothing announces.

**Documented four things enforced mechanically but written down nowhere:**

- **The round-2 mode-switch tier.** `audit-ack` appeared zero times in the skill;
  only the round-3 cap was described. Round 2 is the tier a session reaches
  *first*, and it is a different instruction — it says the approach is wrong
  (fixing instances, not the class), where the cap says stop and re-decide.
- **The gate machinery**: the canonical command sequence plus the ordering rules
  that are invisible until violated — stage before mark (mark hashes the staged
  diff), a 30-minute evidence TTL, `evidence-path`/`mark` keying off the process
  working directory rather than a flag, a successful commit wiping the marker,
  and `--match-head-commit` being mechanically required rather than a nicety.
- **The traps**: a blocked Bash call runs *nothing*, including earlier `&&`
  segments and heredocs; ack sigils bind per-guard and mostly to the last
  pipeline segment; a sigil must lead the trailing comment; `git worktree remove`
  is blocked outright and a lifecycle manager owns removal.
- **CI**, which the skill enumerated not at all: ten of eleven jobs block, only
  the review-depth check is advisory, and four reproduce locally in seconds.

**Plus one detection rule** that has now cost three separate incidents: any check
that greps command lines for `pytest` matches its *own* command line. `pgrep -f`,
`ps | grep` and a bash `case *pytest*` all self-match, as does another session's
wait-loop. Match `argv[0]`'s basename, or a python interpreter with an adjacent
`-m pytest`.

Also corrected: the skill (and the commit gate's own deny message) instructs
"Run `/review`", which does not exist — the commands are `/deep-review` and the
built-in `/code-review`.

## Testing

Docs-only, one file, no code paths changed.

Every claim about gate behaviour was exercised by this session's own commits
rather than asserted — the mode-switch block, the depth gate rejecting
symbol-only evidence, the blocked-call-runs-nothing behaviour, the
worktree-removal block, and the stage → mark → commit ordering all fired live
while the section describing them was being written.

One agent-reported finding was deliberately **not** included: a count of
worktrees running stale hook copies. That is install-local state rather than
process a session must follow, and it was not independently re-derived — it
belongs in an issue with its own measurement, not in a skill that ships to every
clone.

### DIFF
--- a/.claude/skills/genesis-development/SKILL.md
+++ b/.claude/skills/genesis-development/SKILL.md
@@ -444,13 +444,20 @@ Choose the review level proportional to the change:
 | Docs / text / comments | **None** | Markdown prose, inline comments |
 | Simple mechanical | **None** | Variable rename, typo fix, import reorder |
 | Small focused fix | **Code-reviewer agent inline** | Single-function bug fix, config tweak |
-| Substantial change | **Code-reviewer inline + /review** | Multi-file refactor, new MCP tool, wiring |
+| Substantial change | **Code-reviewer inline + `/deep-review`** | Multi-file refactor, new MCP tool, wiring |
 | Prompt / LLM behavior | **Both + extra scrutiny** | System prompts, skill instructions, routing |
 
 Decision criteria when ambiguous: "If the change could break a runtime
-path not covered by its own unit test, it needs /review. If it only
+path not covered by its own unit test, it needs `/deep-review`. If it only
 touches things with clear, isolated test coverage, code-reviewer inline
 is sufficient."
+
+**There is no `/review` command.** Older text here (and the commit gate's own
+deny message) says "Run /review" — measured, `.claude/commands/` ships
+`deep-review.md` and `audit-changes.md`, and there is no user-level commands
+directory. The real options are **`/deep-review`** (dispatches the adversarial
+pass AND writes the evidence marker) and the built-in `/code-review`.
+`/audit-changes` is a light self-check, not a substitute for either.
 
 The enforcement hooks (`review_enforcement_prompt.py`,
 `review_enforcement_commit.py`) still fire on every change — they are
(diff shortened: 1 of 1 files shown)


WingedGuardian:
**Structural review**: Skipped — this PR is documentation-only (1 file: `.claude/skills/genesis-development/SKILL.md`, no code paths changed).

**PII/Secrets/Wording scan**: No hard blocks.

- No API keys, tokens, passwords, secrets, private IP addresses, SSH keys, DB connection strings, or secret env values found in the added lines.
- No soft-warning items either: no personal email addresses, no usernamed absolute paths, no location-revealing timezone identifiers, no cloud account IDs, no hardware identifiers, no private hostnames.
- PR title/body: no leak/incident wording, no naming of specific private infrastructure (IPs, Tailscale, etc.).

One item worth a maintainer gut-check, not a block: the rewritten "Honest enforcement model" section documents specifics from a live query against this repo's own branch-protection/ruleset config — that `main` has no branch protection (404), the ruleset requires only 1 of 11 CI checks, and carries an `always`-mode bypass actor for the admin role. That's legitimate, accurate self-documentation for sessions operating in this repo (and matches the diff's stated intent), so it's process/config info rather than a credential or PII — not a hard block. But it's now permanently in this public repo's PR history and spells out precisely how the server-side merge gate can be bypassed by the maintainer. Worth a quick check on whether that level of mechanical detail (vs. just the conclusion — "branch protection doesn't cover this repo; the local gate is the real backstop") is the level you want public.

Everything else checks out. No CHANGELOG.md changes to scan.



---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:

**Scope:** `.claude/skills/genesis-development/SKILL.md` only. 146 added lines, all prose — a skill surface, so read rather than only grepped.

**Direct identifiers:** none. No IPv4/IPv6 literals, private-range or CGNAT addresses, tailnet addresses, MAC/serials, hostnames, absolute `/home`-style paths, email addresses, SSH keys, API keys, or tokens.

**Indirect / inferential:** none. No household composition, family or guests, languages, schedule, habits, devices, or physical location; no "my/our house" framing. The added text is entirely about gate mechanics — which check blocks on what, which sigil waives which boundary, and where the merge teeth actually live.

**Install-genericity:** the content describes repo-level enforcement (branch protection vs rulesets, bypass actors, the local hook being the operative gate). Where it states a measured fact about enforcement it is a property of the repository, which is already public, not of any deployment. No process counts, no machine state, no operational volumes.

**Wording:** technical. Nothing exploitable is advertised — the honest description of where enforcement lives is visible to anyone reading the workflow files and the hook, both already public.

VERDICT: PASS

PR #1522, opened 2026-08-28
feat(cc): running-binary sweep — which copy is RUNNING, not which is installed

WingedGuardian:
## Problem

On a live install a container was aligned to a candidate Claude Code version and
a multi-day local-first soak was declared started. Measured at the end of it,
**most interactive sessions were still executing the pre-align binary.** npm had
replaced the package underneath them, and a long-lived process keeps its
original mapping until it restarts. `claude --version` could not reveal this —
it spawns a *fresh child*, which reads the new binary on disk and truthfully
reports the candidate while the session asking is not running it. The soak had
accumulated days of "real use" on the old release.

This answers the question `cc_shadow_scan` does not. That one asks what is
*installed*; this asks what is actually *running*. A box passes one and fails
the other: exactly one canonical binary at the intended version, while several
live sessions execute a deleted predecessor.

Split out of #1514, where it drew findings in three consecutive review rounds —
and two of the last round's were caused by the round before's fixes.

## One invariant, written down

That churn pattern is a missing invariant, not bad luck. Each round optimised
the case a reviewer named and silently flipped the fail direction, because there
was nothing to check a fix against. So the rule is now in the file and the doc:

> Any process the sweep cannot POSITIVELY PROVE is not-Claude-Code must not
> contribute to a clean verdict.

Positive proof is cheap: a readable executable whose basename is not a Claude
Code name, or an unreadable one whose command line's `argv[0]` is not. Only when
neither can be read is a process genuinely unclassifiable, and that refuses the
clean verdict rather than being skipped.

**A design refuted by measurement before implementation**, recorded in the file
so it is not proposed again: classify by whether the executable lives under an
enumerated install root. npm renames the old package aside and then deletes it,
so a stale process's path points at a directory that **no longer exists** —
measured here, two sessions on a deleted `.claude-code-<random>` path while the
scope directory contained only `claude-code`. A root set enumerated from the
install cannot contain a directory the installer deleted, so exactly the stale
processes would have been ignored and reported clean. That trades a bounded name
test whose mistakes *shout* for a path test whose mistakes *whisper*.

## Closed, each measured

- A process whose readable executable matched no branch incremented **no counter
  at all**, so the sweep printed "OK — all 0 live processes" with a live session
  running.
- Another user's `grep -r claude /var/log` had an unreadable executable, hit a
  cmdline substring test, and wedged the whole sweep. `argv[0]` is `grep`. The
  unreadable branch deliberately carries no path clause — requiring one would
  mean another user's install, under *their* home, is silently ignored.
- `$HOME` unset under `set -u` aborted with **exit 1**, the code meaning "stale
  binaries found". An environment error must never read as a finding.
- A typo'd `--proc-root` produced a confident all-clear. Both shapes now refuse.
- The ambiguity check compared `--version` strings, no-opped when either probe
  came back empty, and refused the whole sweep when they differed. It is now an
  installed-copy map keyed on device+inode with a third per-process verdict
  naming *which* copy each process runs — more evidence than a refusal, and it
  removes the only place that exec'd a discovered binary, so the unbounded probe
  cannot hang.
- The `(deleted)` ⇒ stale rule was dead code: `stat -L` **succeeds** on a deleted
  binary, which is precisely why the inode compare works at all.

## A second audit found four more

- **The invariant was violated in the one branch written to enforce it.** The
  unclassifiable count sat inside a readability guard while `continue` sat
  outside it, so a wholly unreadable pid directory incremented nothing and voted
  clean — the `hidepid=1` shape exactly.
- The interpreter branch tested `argv[1]` for the entry script, so
  `node --enable-source-maps .../cli.js` put a flag in the slot and the process
  was recorded as *proven not-CC*. It now takes the first non-flag token.
- The hidepid guard matched an enumerated value list, `hidepid=[12]`, missing
  `hidepid=4` and the symbolic spellings — the same closed-set-of-values mistake
  this script rejects elsewhere.
- **The `(deleted)` strip was untested.** Removing that one expression left the
  suite fully green while the live box flipped from two stale processes to
  "OK — all 8", exit 0. A one-character regression turning the exact incident
  into a green receipt, invisible to CI.

## A claim corrected, not just code

The header and doc said *"0 of 109 processes fall in that bucket, so the
invariant costs nothing in practice."* That was measured **inside a
PID-namespaced container**, which contains no kernel threads, and written as
though general. Kernel threads have exactly the unclassifiable shape — no exe,
empty cmdline — and are a large fraction of pids on a host, so off-container,
**including on any CI runner, the script would have returned 2 unconditionally**:
a 100%-rate false alarm, which is the outcome that teaches an operator to ignore
a tool. Kernel threads are now identified positively as not-CC, and both prose
claims name their namespace.

Worth flagging on its own: fixing the glob expansion introduced a worse bug that
**only the live run caught**. `set -f` placed before the loop also disabled the
pid glob, so the loop ran zero times and the sweep reported "OK — all 0" on a box
with two stale processes — total blindness presenting as a clean verdict. It
belongs inside the loop body, where the glob has already expanded.

## Testing

26 tests, shellcheck clean, ruff clean.

Tests are grouped by **which half of the invariant** each case pins — must never
report OK, and must never false-alarm — because a fix that satisfies one by
breaking the other is the failure mode here, and a feature-organised file hides
that.

Hermetic by construction: every run pins `CC_PROBE_DIRS` and `HOME` to
fixture-only paths, and a test asserts the sweep never reaches a real install
path. The previous suite passed only because a fixture literal coincidentally
equalled the host's installed version — the next pin bump broke eight tests,
invisibly on CI where no CC is installed.

Every fix is mutation-tested: dropping the deleted-strip, re-guarding the
unclassifiable count, breaking flag handling, disabling kernel-thread detection,
inverting `set -f`, or removing the procfs check each breaks the suite.

Verified live on a box carrying two genuinely stale sessions: `current=8
stale=2 undetermined=0`, exit 1, naming both.

## Known residual

The hidepid guard is **not** verified against a real hidepid mount —
`unshare --mount-proc` is not available in this container. The regex gap it
closes is measured; how a given kernel renders the option into `/proc/mounts` is
inferred from the documented option vocabulary.

### DIFF
--- a/docs/reference/cc-compatibility.md
+++ b/docs/reference/cc-compatibility.md
@@ -92,6 +92,62 @@ reinstalled forever.
 
 ---
 
+### Which copy is INSTALLED vs which copy is RUNNING
+
+Different questions, different checks, and a box can pass one while failing the
+other.
+
+`cc_shadow_scan` answers **installed**: it probes `command -v claude` plus
+`CC_PROBE_DIRS` for extra on-disk copies and removes the stale ones.
+
+`scripts/check_cc_running_versions.sh` answers **running**: for every live CC
+process it resolves `stat -L /proc/<pid>/exe` — procfs keeps that reference
+valid even after npm unlinks the file — and compares its **device + inode**
+against the binary on disk today. Device matters: an inode number is unique only
+within a filesystem, so copies on separate mounts can collide numerically.
+
+Measured on a live install: exactly one canonical binary, at the intended
+version, while **a majority of live CC sessions were still executing the
+replaced predecessor**. A long-running process keeps its original mapping until
+it restarts, and `claude --version` cannot reveal this — it spawns a *fresh
+child*, which reads the new on-disk binary and truthfully reports the new
+version while the session asking the question is not running it.
+
+This matters most during a local-first soak, whose whole premise is that real
+interactive use exercises the candidate. Run the sweep at soak start and again
+at soak end.
+
+**The rule it is built around**, because a check that produces release evidence
+has one failure that matters more than the others:
+
+> Any process the sweep cannot POSITIVELY PROVE is not-Claude-Code must not
+> contribute to a clean verdict.
+
+Positive proof is cheap: a readable executable whose basename is not a Claude
+Code name, or an unreadable executable whose command line's `argv[0]` is not
+one. Only when neither can be read is a process genuinely unclassifiable, and
+that refuses the clean verdict rather than being skipped. Measured on a live
+box, that bucket holds 0 of 109 processes — the rule costs nothing in practice
+and never launders a false all-clear.
+
+Classification is a **closed set of names**, never a path guess. An earlier
+design classified by whether the executable lived under an enumerated install
+root; it fails on the only case that matters, because npm's replace renames the
+old package aside and then deletes it, so a stale process's path points at a
+directory that no longer exists. A root set enumerated from the install cannot
+contain a directory the installer deleted, so exactly the stale processes would
+have been classified "not CC", ignored, and reported clean.
+
+Exit codes: **0** every live CC process runs the on-disk binary; **1** at least
+one runs a different binary — either a replaced one, or an installed copy that
+is not the PATH-canonical one (each named, so `cc_shadow_scan` can resolve it);
+**2** cannot determine, so no clean verdict is claimed. A node-wrapped install
+resolves `exe` to the interpreter, whose inode says nothing about which CC
+revision is loaded, so it reports undetermined rather than a false all-clear.
+The same is true under a `hidepid` procfs, where other users' processes are not
+enumerable at all and the denominator itself would be unverified.
+
+
 ## Integration Surface — Genesis Components That Use CC
 
 | Genesis Component | CC Feature Used | Files | Notes |
--- a/scripts/check_cc_running_versions.sh
+++ b/scripts/check_cc_running_versions.sh
@@ -1,0 +1,431 @@
+#!/usr/bin/env bash
+# shellcheck shell=bash
+#
+# Running-binary sweep — which LIVE Claude Code processes are executing the
+# binary that is on disk now, and which are still running one npm has replaced.
+#
+# WHY THIS EXISTS. On a live install a container was aligned to a candidate CC
+# version and a multi-day local-first soak was declared started. Measured at the
+# end of it, MOST interactive sessions were still executing the pre-align binary:
+# npm had replaced the package underneath them, and a long-lived process keeps
+# its original mapping until it restarts. `claude --version` did not reveal this
+# — it spawns a FRESH child, which reads the new on-disk binary and truthfully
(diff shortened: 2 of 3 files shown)


WingedGuardian:
@codex review

WingedGuardian:

**Files reviewed:** `docs/reference/cc-compatibility.md`, `scripts/check_cc_running_versions.sh` (new), `tests/test_scripts/test_cc_running_versions.py` (new)

No ERROR-level pattern violations. Checked against the full list: no `nohup`, no direct `task_states` writes, no `/tmp/` usage (test fixtures use pytest's `tmp_path`, not a hardcoded `/tmp` path), no `git add -A`/`git add .`, no force-push/reset patterns, no `pip install -e` against a worktree, no bare `except:`, no mutable default arguments (`_make_proc`'s `exe: Path | None = None` / `cmdline: str | None = "claude\x00"` defaults are immutable), no unguarded `os.killpg`, no raw SQL.

Import hygiene: clean. `tests/test_scripts/test_cc_running_versions.py` groups `from __future__ import annotations`, then stdlib (`os`, `subprocess`, `pathlib.Path`), then third-party (`pytest`) — correct order. No star imports, no circularity.

Coverage: the new script has a same-PR test file (`tests/test_scripts/test_cc_running_versions.py`, 26 tests) — no missing-coverage warning.

**NOTE — a few test functions omit the docstring the rest of the file uses consistently:**
- `test_all_processes_on_canonical_is_ok` — `tests/test_scripts/test_cc_running_versions.py:241`
- `test_empty_proc_root_does_not_glob_literally` — `tests/test_scripts/test_cc_running_versions.py:289`
- `test_rejects_an_unknown_argument` — `tests/test_scripts/test_cc_running_versions.py:296`

Every other test in the file documents which half of the invariant it pins and, where applicable, the measured failure it regression-guards; these three are the only ones without that context. Low priority — names are self-explanatory — but worth a one-liner for consistency with the file's own stated organizing principle.

### Domain 2 — PII, Secrets, and Wording

No hard blocks in the added lines: no API keys/tokens, private (10.x/192.168.x/100.64–127.x or fd00::/8) IPs, SSH keys, DB connection strings, or secret-bearing env values.

PR title/body/doc additions: clean. This describes a technical binary-version-drift bug (stale process vs. on-disk binary after an npm replace) — no PII/security-incident language, no private-infrastructure naming, nothing exploitable being advertised.

**SOFT WARNING** — `scripts/check_cc_running_versions.sh` (script header comment) and the PR description both quote the literal fixture path `.claude-code-1devilah/bin/claude.exe (deleted)` as a measured example. Reads as npm's own randomized rename-before-delete suffix rather than a personal identifier, but it's an unusually specific string to have landed verbatim in committed docs — worth a second look to confirm it isn't derived from anything install-specific before merge.

### Summary

Structural: clean, one NOTE (missing test docstrings).
PII/Secrets/Wording: clean, one SOFT WARNING (unusual literal string, likely benign).

No changes requested.

---


---

WingedGuardian:
@codex review

PR #1528, opened 2026-08-28
fix(hooks): tell the author a merge is in progress — do not let one silence the gate

WingedGuardian:
## The problem

The review round counter keys on the staged diff, so pulling upstream changes
into a PR branch to resolve a conflict reads as another **defect-bearing review
round**. Seen live: a branch that had already completed its adversarial audit
went 1 → 2 purely from a CHANGELOG resolution that added no authored code, which
tripped the mode-switch gate and demanded an ack for a commit containing nothing
to audit.

That weakens the gate twice over — it burns the round budget on non-events, and
it trains the reflex of acking past the mode-switch and escalation blocks in
situations that don't warrant it, which is precisely the habit those gates exist
to prevent.

## What I tried first, and why it is not in this PR

The obvious fix is to exempt the round when git reports an in-flight merge.
That was written, reviewed, and **abandoned** — it makes the gate forgeable.

Those sentinels are plain files:

- `echo garbage > .git/MERGE_HEAD` — no git command at all — makes the exemption
  fire. Measured against that version, the counter stayed **frozen across three
  further genuinely distinct defect rounds**.
- No forgery is even required. `--no-commit --no-ff` leaves `MERGE_HEAD` until an
  explicit commit or abort, and a killed session leaves stale `MERGE_HEAD` /
  `rebase-merge` / `rebase-apply` that ordinary work never clears.

The actor this gate constrains is an agent with shell access to the same repo.
An anti-runaway cap that such an actor can disable with one write is worse than
the annoyance it removes — especially when an auditable escape already exists in
the ack comment, which leaves a trace in the command.

`scripts/review_state.py` is therefore **byte-identical to `origin/main`**.

## What ships

Advisory text only. When the mode-switch or escalation denial fires *and* a
merge appears to be in flight, the message says so and asks the author to note it
in the ack. The sentinel is read to **phrase** a denial, never to **decide** one:
the verdict, the counter and the ack requirement are unchanged.

Telling the author what the gate can see is safe. Letting that state pick the
verdict is not.

## Tests

They assert the property that matters, not the convenience:

- a **forged** sentinel does not suppress a round
- a **real** conflicted resolution does not either

Both would have failed against the abandoned design. 146 tests green across
`test_escalation_cap.py` and `test_review_enforcement_commit.py` — every
pre-existing gate test unchanged (direct-to-main, `--no-verify` in all bundled
forms, the cap, the mode-switch, chained cross-worktree commits, unresolvable
cwd).

### DIFF
--- a/scripts/review_enforcement_commit.py
+++ b/scripts/review_enforcement_commit.py
@@ -18,6 +18,7 @@ import os
 import re
 import subprocess
 import sys
+from pathlib import Path
 
 # The shared hook-input helper lives in scripts/hooks/; this script runs from
 # scripts/ (a different sys.path[0]), so add the hooks dir before importing it.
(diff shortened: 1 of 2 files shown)


WingedGuardian:
@codex review

WingedGuardian:

- `_merge_note()` is advisory-only text appended to two existing denial messages; it does not affect the verdict, round counter, or ack requirements — matches the PR description's stated intent.
- New `from pathlib import Path` import is stdlib, correctly grouped with the existing stdlib imports; hook script remains stdlib-only.
- `cwd` passed into `_merge_note()` is already narrowed to `str | None` before the call sites (never the `_CWD_UNKNOWN` sentinel), so no type-mismatch risk in the `subprocess.run(cwd=cwd)` call.
- New behavior (`_merge_note`) has direct test coverage in the same PR (`test_a_forged_merge_sentinel_cannot_freeze_the_counter`, `test_a_real_merge_also_does_not_suppress_rounds`); existing denial-message assertions use substring matching, so the appended note doesn't risk breaking them.
- `git add -A` in the new test fixture matches the established convention already used throughout `tests/test_hooks/` (test_review_depth_gate.py, test_git_discard_guard.py, etc.) for throwaway repo fixtures — not flagged.



---

PR #1532, opened 2026-08-28
docs: acceptance-bar methodology, findings-are-a-sample, and the blocked-compound trap

WingedGuardian:
Three rules earned the hard way across one long session, split out of the tooling
PR they came from so they can land on their own merit. **Docs only — two `.md`
files, 132 insertions, zero deletions, no execution path.**

## Acceptance Bar + Measured Rate *(skill — the primary methodology)*

Before shipping anything: replay the **real** defect it exists for, and measure
it against a **real corpus** as `k/N`. *A number without a denominator is not a
measurement.*

The measurement is a **gate** with a threshold chosen before measuring. If it
misses, tighten and re-measure — **and re-run the acceptance replay in the same
breath**, because a filter that improves the number by blinding the thing has
made it worse.

The worked example includes the failure that motivated the rule: a regex
prototype reported *"2 findings"* while silently skipping the entire class it was
built for, because its pattern excluded backslashes and every prompt string ends
in `\n`. A matcher that finds nothing is indistinguishable from a matcher that
looks at nothing — only replaying a known-positive tells them apart.

## A review's findings are a SAMPLE, not a to-do list *(skill — review loop)*

N findings look exactly like an N-item work queue, and that is how a review loop
runs away: fix the named instances, and the next round surfaces their siblings.
Name the **class**, enumerate the population **programmatically** — an AST walk
or a grep that lists every member, never a mental scan — and fix the population.

Enumerate the **product** of axes, not each axis in turn. A state matrix with one
axis pinned still missed two cells, both reachable only where two conditions held
at once.

`CLAUDE.md` already says to treat the *user's* examples as a sample. This extends
the same rule to reviewer output, where it is easy to miss because the findings
arrive pre-formatted as a list.

## A blocked compound command loses everything in it *(CLAUDE.md + skill)*

A PreToolUse block kills the **whole** Bash call, so a guard firing on step 3
silently discards steps 1 and 2 while the error text mentions only step 3.

Measured cost in one session: four heredocs that never wrote, a
restore-from-backup that never ran (leaving a file deliberately regressed), and a
lost `cd` that put edits in the **wrong worktree**.

## Check closing tags on long tool-call parameters *(CLAUDE.md)*

A closing tag that does not match its opening tag silently swallows every
following parameter into the preceding string; the tool then reports those
parameters as *missing*, which reads like a tool bug and is not one. Repeated
**identical** validation errors mean the call is malformed — the error echoes
`input_value`, which holds the proof.

## Note on this PR's own history

The first attempt at this split was **destructive** and was discarded before
push. I computed the diff from a worktree whose `origin/main` ref was stale, so
it encoded *"revert everything main gained since my branch point"* alongside my
additions — deleting the Bash-timeout hard-ceiling documentation and the
`out=$(cmd)` errexit lesson, neither of which is mine.

The tell was the diffstat: **130 insertions / 207 deletions** on a change that is
additive by construction. A purely additive split cannot delete anything. Both
files were restored from current `origin/main` and only my own blocks
re-inserted; the result is the +132/−0 you see here, with both reverted sections
verified present and unmodified.

Which is, fittingly, an instance of the third rule this PR adds.

## Summary by CodeRabbit

* **Documentation**
  * Expanded acceptance guidance with independent recall measurement, repository-derived checks, and tradeoff considerations.
  * Clarified script execution requirements, including process working-directory verification and root-relative paths.
  * Added safeguards against making absence or total-count claims from incomplete or truncated listings.
  * Documented stronger review-loop escalation rules for repeated findings.
  * Added a frozen-clock validation check and a command for waiting on pytest locks.
  * Expanded guidance for safely executing compound commands in worktree-specific contexts.


WingedGuardian:
**Automated first-pass review (docs-only PR — structural pass skipped per process, PII/wording scan run).**

- No API keys/tokens/secrets, private IPs (RFC1918/100.64–127/fd00::/8), SSH keys, DB connection strings, or env-var secrets in the added lines.
- No personal emails, username-bearing absolute paths, location-revealing timezones, cloud account/project IDs, hardware identifiers, or private hostnames.
- PR title/body discuss an internal process mistake (stale-worktree diff, discarded destructive draft) but contain no security-incident language, no named private infrastructure, and no exploitable vulnerability detail.

**NOTE (content, not PII):** `.claude/skills/genesis-development/SKILL.md` adds the `### A blocked compound command loses EVERYTHING in it` subsection **twice**, back-to-back — identical heading, body, and "Measured cost in one session" paragraph both times (see the diff around lines +155–195 of that file). Likely a leftover from the worktree-diff recovery described in the PR body. Worth squashing to one copy before merge.




---

WingedGuardian:
@coderabbitai review

WingedGuardian:
@coderabbitai review

WingedGuardian:
@codex review

WingedGuardian:
## Scheduled review — leaks / private data

Docs-only branch: `CLAUDE.md` and the genesis-development skill. Full diff
against `origin/main` is 195 lines; added lines scanned in isolation as well.

**Direct identifiers** — RFC1918 ranges, IPv6 ULA, CGNAT/tailnet,
home-directory paths, email addresses, key material: **0 matches**.

**Inferential context** — household, family, location, languages, schedule,
first-person framing: **0 matches**.

**Numbers in the prose** are the load-bearing part of a docs change like this,
so they were re-derived rather than trusted: 95 verified cases over 1,584
commits, 48 of 95 excluded by the prose filter, 28/47 caught at the budget cap
and 47/47 with it lifted. These are measurements over this repository's own
commit history and carry no install identity — no hostname, path, account or
device appears in any of them. An earlier revision of this branch shipped two
of these figures stale by one; that is corrected in this commit.

PII/Secrets: CLEAN
Wording: CLEAN
VERDICT: PASS

PR #1536, opened 2026-08-29
fix(hooks): share one tokenizability probe, and stop the commit gate failing open

WingedGuardian:
## Summary

Three pieces of infrastructure, **split out of the guard-net PR** so that PR's design can be reviewed on its own. Nothing here decides a verdict — this is the plumbing a net would consume, and each piece stands alone.

The net itself, its mention predicates, and its dispatched deny leg are **not** in this branch (verified by grep, not by reading the diff).

## What's here

**1. `shell_parse.py` gains `untokenizable()`** — additive, stdlib only, no parse-logic change and no existing caller altered.

The parser degrades to a naive split on a `shlex` error **silently**. That means `analyze()` cannot report that its own answer is untrustworthy: *"no gated segment found"* and *"no gated command present"* are the same return value, and a caller that treats the first as the second fails **open**. The probe separates them, so a security-critical caller can choose its own fail direction at its own boundary while the parser keeps degrading gracefully — which is its stated contract.

It reads the **raw** command. No here-doc normalization, and no line-continuation folding either — the fold was *removed* rather than documented. It was wrong about bash, which **removes** a continuation and joins the halves into one word rather than replacing it with a space, so the fold produced the reading furthest from what actually executes. And it bought nothing: measured over **12,099 real commands**, folding and not folding classify identically — 339 un-tokenizable either way, **zero commands differ**.

**2. `protected_paths_guard.py` moves onto the shared probe** — deleting the *third* hand-rolled copy, which had already drifted from the others. Behaviour-preserving on real traffic per the measurement above; the comment says "not byte-identical, measured equivalent" rather than claiming a byte-equivalence it no longer has.

**3. `review_enforcement_commit.py` is wrapped in `run_guard`** instead of calling `main()` bare. The PreToolUse contract is *exit 2 = block; any other code = non-blocking, the tool runs* — so every uncaught exception in that gate exited 1 and let the commit through, silently. Deliberate allow/deny paths raise `SystemExit` and are untouched.

## A limit, pinned rather than papered over

**The wrap is narrower than it reads.** `run_guard` is called at the bottom of the module, so a failure raised while the module is still *importing* never reaches it — measured, that still exits 1, i.e. still fails open.

That's asserted by a test which pins the exit code and instructs a future reader to **delete** the test and say so if the gap is ever closed, rather than loosen it. Claiming "fails closed on crash" unqualified would have been an overclaim.

## Tests

Built against the failure class that produced defects on the parent branch — twice there, a test harness substituted its own context for the one under test and the suite stayed green:

- the child environment is an **allowlist**, not a subtract-list, so the next ambient input cannot leak in by default;
- git's config is pinned away from the developer's own, so a global setting such as commit signing cannot quietly break a fixture into a state where it proves nothing;
- payload cwd and child cwd always agree, so repo-state gates can't silently evaluate the wrong repository;
- the crash test drives the **real module**. An earlier draft handed a locally-defined function to `run_guard` and named the module only in a string — that passes against the *unwrapped* module too, so it proved `run_guard` works, not that this gate uses it.
- every behavioural class carries a **positive control** (the obfuscated `rm` that must still block; the ordinary command that must not). Without them, a guard that never fires — or one that always exits 2 — would satisfy the suite.

**356 pass**: 13 in the new file, 343 across the five affected existing suites. `ruff` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved protection for commands that cannot be parsed reliably.
  * Preserved conservative safeguards when malformed input is encountered.
  * Improved commit validation when state information is unavailable or malformed.
  * Unexpected validation failures now produce a clear blocking result instead of silently passing.
  * Unverifiable commit-related commands now require approval or are blocked in unattended sessions.

* **Tests**
  * Added comprehensive coverage for command parsing, validation failures, malformed state handling, and safeguards against publishing bypass guidance.

### DIFF
--- a/scripts/hooks/protected_paths_guard.py
+++ b/scripts/hooks/protected_paths_guard.py
@@ -39,14 +39,13 @@ from __future__ import annotations
 
 import os
 import re
-import shlex
 import sys
 from fnmatch import fnmatch
 
 # Self-locate so hook_input resolves whether run as a script or imported (tests).
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 from hook_input import brace_expand, read_payload, run_guard, tool_input  # noqa: E402
-from shell_parse import analyze  # noqa: E402
+from shell_parse import analyze, untokenizable  # noqa: E402
 
 # Directories that must never be deleted.  Relative to $HOME.
 _PROTECTED_RELATIVE = [
@@ -189,11 +188,19 @@ def main() -> int:
     dirs = _protected_dirs()
     files = _protected_files()
 
-    # Explicit tokenizability probe: shell_parse._argv silently degrades to a
-    # naive split on shlex errors, so analyze() alone can never signal one.
-    try:
-        shlex.split(cmd.replace("\\\n", " "))
-    except ValueError:
+    # Explicit tokenizability probe, now SHARED (this was the third hand-rolled
+    # copy; they had already drifted from one another). shell_parse._argv
+    # silently degrades to a naive split on shlex errors, so analyze() alone can
+    # never signal one.
+    #
+    # Not byte-identical to the inline probe it replaces: that one folded
+    # `\<newline>` to a space first. The shared probe reads the command RAW,
+    # because bash REMOVES a line continuation rather than replacing it, so the
+    # fold produced the reading furthest from what actually executes. MEASURED
+    # over 12,099 real commands: the two classify identically (339 un-tokenizable
+    # either way, zero commands differ), so this is a semantics correction with
+    # no observed behaviour change.
+    if untokenizable(cmd):
         reason = _legacy_substring_block(cmd, dirs)
         return _block(reason) if reason else 0
 
--- a/scripts/hooks/protected_paths_guard.py
+++ b/scripts/hooks/protected_paths_guard.py
@@ -188,10 +188,16 @@ def main() -> int:
     dirs = _protected_dirs()
     files = _protected_files()
 
-    # Explicit tokenizability probe, now SHARED (this was the third hand-rolled
-    # copy; they had already drifted from one another). shell_parse._argv
-    # silently degrades to a naive split on shlex errors, so analyze() alone can
-    # never signal one.
+    # Explicit tokenizability probe, now SHARED rather than inline here.
+    # shell_parse._argv silently degrades to a naive split on shlex errors, so
+    # analyze() alone can never signal one.
+    #
+    # (An earlier draft of this comment called it "the third hand-rolled copy,
+    # already drifted from one another". Not true of the tree this lands on: on
+    # the default branch this was the ONLY probe of the discard-the-tokens kind.
+    # Every other shlex call in scripts/hooks USES its tokens, which is a
+    # different act with a different fail direction. A comment describing a tree
+    # the change is not landing on is how a reader inherits a false picture.)
     #
     # Not byte-identical to the inline probe it replaces: that one folded
     # `\<newline>` to a space first. The shared probe reads the command RAW,
--- a/scripts/hooks/shell_parse.py
+++ b/scripts/hooks/shell_parse.py
@@ -613,17 +613,27 @@ def untokenizable(command: str) -> bool:
     This is the blind-spot signal a guard consults when it is about to conclude
     "no gated segment found". ``_argv`` degrades to a naive split on the SAME
     ``ValueError`` silently, so ``analyze()`` can never self-report that its
-    result is untrustworthy — an ANSI-C ``$'…\'…'`` span, or merely an
-    apostrophe inside a here-doc body, is enough to shift segmentation off and
-    drop a real, executing command from the parse.
+    result is untrustworthy: an ordinary quoting construct is enough to shift
+    segmentation off and drop a real, executing command from the parse, and the
+    return value looks identical to "there was nothing to find".
 
-    Deliberately reads the WHOLE raw command, with no here-doc normalization.
-    An earlier version stripped here-doc bodies first to suppress prompts on
-    multi-line scripts; that MEASURABLY disarmed the signal on an ordinary shape
-    (a quoted here-doc whose body contains an apostrophe, followed by a real
-    gated git command — bash runs it, the stripped text tokenizes cleanly, and
-    the guard fell silent). Normalizing the text before the probe can only ever
-    delete the evidence the probe exists to find, so it does not happen here.
+    Deliberately reads the WHOLE raw command, with no normalization of any kind.
+    An earlier version pre-processed it to suppress prompts on a class of
+    multi-line command, and that MEASURABLY disarmed the signal: on a shape a
+    developer writes without thinking, the command really ran (verified against
+    a shimmed binary, so the proof was execution rather than parse) while the
+    pre-processed text tokenized cleanly and the guard fell silent. The
+    triggering shape is deliberately not written down — this file is public and
+    the guard it protects is load-bearing.
+
+    KNOWN COST, stated rather than hidden: ``_argv`` DOES normalize before its
+    own tokenize (it strips trailing comments), so this probe over-reports
+    relative to the very parser whose blind spot it reports — an unquoted
+    comment alone can make a benign command look unparseable. Stripping here is
+    NOT the fix: measured over 19,246 real commands, doing so erases a mention
+    of a gated operation in 3 of them, because a stripper's model of where a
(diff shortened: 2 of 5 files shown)


WingedGuardian:
Scheduled leak/secret review — full branch diff against the base, added lines only.

Scanned for: private IPv4/IPv6 and ULA ranges, tailnet addresses, MAC/serials, hostnames, absolute paths embedding a username, personal emails, API keys/tokens/bearer credentials, DB connection strings, repository slugs, timezones, and indirect personal context.

Findings: none. The diff is three hook changes plus one test file; the only literals it introduces are shell fragments and synthetic test inputs. The aggregate figures quoted in the comments (a command-count denominator for a measured equivalence) carry no command content and are not install-identifying.

Verdict stated positively and without naming scanner categories, because the marker matcher is a set of plain substrings with no negation awareness — writing out the category names even to report none found has been measured causing a clean review to be refused on wording alone.

VERDICT: PASS

WingedGuardian:
@codex review

This is a deliberate split: the parent PR's blind-spot net kept drawing real findings at the design level, so the owner split the undisputed infrastructure out to be reviewed on its own. No net, no mention predicate, no deny leg here.

Worth your attention:
- `untokenizable()` now reads the command RAW — the line-continuation fold was removed, not documented. I measured the two readings as identical over 12,099 real commands (339 either way, zero differ), but that is a corpus, not a proof; a constructed shape where they diverge would be a real finding.
- `protected_paths_guard` swaps an inline probe for the shared one. It is NOT byte-identical any more (the fold is gone). I claim behavioural equivalence on measurement alone.
- `run_guard` on the commit gate: please check whether any deliberate path that previously exited non-2 now exits 2, and whether `SystemExit` really does pass through untouched on every allow/deny leg.
- I found and pinned one limit rather than hiding it: an import-time failure still exits 1 (fail-open), because `run_guard` is called at the module bottom. If you think that gap is worth closing in this PR rather than pinning, say so.
- Test quality: would any test here still pass with its production change reverted?

WingedGuardian:

- No nohup/direct-task_states/`/tmp/`/`git add .`/force-push/bare-except/mutable-default/raw-SQL issues.
- Import hygiene clean (verified `ruff check` passes on all 4 changed files with the repo's configured rule set, incl. isort ordering — no unused imports left after `shlex` was dropped from `protected_paths_guard.py`).
- New public function `untokenizable()` in `scripts/hooks/shell_parse.py` has a docstring; module docstring updated to explain the new blind-spot-signal contract.
- Behavioral changes (shared probe, `run_guard` wrap on the commit gate) are covered by the new `tests/test_hooks/test_untokenizable_probe.py`, including positive controls and a pinned "documented gap" test for the import-time crash case.
- Commit/PR title carries a conventional prefix (`fix(hooks): ...`).
- Hook scripts stay stdlib-only (`shlex`, `re`, `os`, `sys`, `fnmatch` — no heavy deps added).

- No secrets, private IPs/hostnames, credentials, or PII in the diff.
- PR title/body describes a fail-open bug being fixed in an internal pre-commit hook and its test methodology (measured counts, code paths) — no reference to a real security incident, leaked credentials, or exploitable infrastructure detail.

No changes requested.




---

WingedGuardian:
@coderabbitai review

WingedGuardian:
## Leak / PII scan — `6d217ddc`

Scanned every added line of `origin/main...HEAD`.

**PII / Secrets: CLEAN** — zero hits for private address ranges, tailnet addresses, home paths carrying a username, IANA timezones, personal email, or credential shapes.

**Indirect personal context: CLEAN** — no household, schedule, location or life-context inference in prose or test data. Test paths are `tmp_path` or `~`-relative.

**Wording: CLEAN**, and this revision improves it. The probe's docstring previously spelled out the concrete triggering shape in a public file — the same detail redacted from the companion documentation PR. That was one surface fixed with its sibling left standing; it now describes the mechanism only, and says the omission is deliberate.

**Structural: CLEAN** — four files, no config, no credentials.

Scanned by the author.

WingedGuardian:
@coderabbitai full review

WingedGuardian:
@codex review

WingedGuardian:
## Leak / PII scan — `14a2a787`

**PII / Secrets: CLEAN** — no addresses, hostnames, username-bearing paths, timezones, emails, or credential shapes in added lines.
**Indirect personal context: CLEAN.**
**Structural: CLEAN** — one new test file, no runtime path.

**Wording: CLEAN, and this commit is the enforcement of it.** The new check forbids prose that names a shell construct alongside a claim that a gate was defeated. Its own text assembles the forbidden terms from fragments so it does not carry them, and it deliberately does not scan string literals — fixture data has to contain the shapes it defends against.

Worth noting for the reviewer: the first version of this check had the same negation blindness it exists to catch. It flagged seven pairings in the tree, all of them prose saying *never* a bypass. That is fixed with a negation lookback and pinned with parametrized cases in both directions.

Scanned by the author.

WingedGuardian:
@coderabbitai full review

WingedGuardian:
@codex review

WingedGuardian:
## Leak / PII scan — `3e66b417834ed16ae4e530c9d5455d06935011f5`

Full branch diff against the merge-base (`a37de450`), added lines only. **846 added lines across 5 files** — that is the denominator.

**PII / Secrets: CLEAN — 0/846.** Zero hits for private IPv4 ranges, tailnet CGNAT addresses, IPv6 ULA, home paths carrying a username, IANA timezones, personal email, credential shapes, or `.local` / `.ts.net` hostnames.

**Positive control: every pattern was run against a seeded line it must match, and all eight fired.** A scanner that matches nothing and a diff with nothing to find return the same empty list, so the zero above is only meaningful alongside this.

**Indirect personal context: CLEAN.** No household, schedule, language, or device framing; the prose is install-generic throughout.

**Wording: CLEAN, and checked mechanically rather than by eye.** This branch carries the repository's own recipe gate, and it passes over the whole tree including its own new prose — so the added text does not pair a shell construct with a claim that a gate was defeated.

**Structural: CLEAN** — five files, no config, no credentials, no new external I/O.

### Delta since the previous marker (`14a2a787`)

One test file, addressing two review findings, both reproduced against the shipped code before being accepted:

- The recipe gate found comments with a regular expression over raw source, which cannot distinguish a comment from a hash-prefixed line inside a multiline string. It therefore scanned the fixture data its own docstring says is permitted. Comments now come from the tokenizer; a file that will not tokenize keeps the coarse scan, because over-reporting is the safe direction there, and that fallback has its own test so it cannot quietly become "return nothing".
- The negation handling read "not only" as a denial and suppressed a genuine claim. Concessive phrases are now removed from the lookback window before the negation scan. Reverting that fails the two new cases while the eight existing ones — three of them real denials — stay green.

A third finding, that a failure during module import escapes the fail-closed wrapper, is real and deliberately not addressed here. It is a class rather than a site: of seven guards wrapping their entry point, six import a shared helper unguarded. Closing it changes the failure direction of six security gates and is tracked to its own change.

Scanned by the author.

WingedGuardian:
@codex review

WingedGuardian:
## Leak / PII scan — `b3cbaecb0853463f739db7119d441801d2f1112f`

Re-run at the new head after merging the default branch. Full branch diff against the merge-base (`fad448f4`), added lines only. **847 added lines across 6 files** — that is the denominator.

**PII / Secrets: CLEAN — 0/847.** Zero hits for private IPv4 ranges, tailnet CGNAT addresses, IPv6 ULA, home paths carrying a username, IANA timezones, personal email, credential shapes, or `.local` / `.ts.net` hostnames.

**Positive control: all eight patterns were run against a seeded line each must match, and all eight fired.** A scanner that matches nothing and a diff with nothing to find return the same empty list, so the zero above only means something alongside this.

**Indirect personal context: CLEAN.** No household, schedule, language, or device framing; the prose is install-generic throughout.

**Wording: CLEAN, and checked mechanically rather than by eye.** This branch carries the repository's documentation check, and it passes over the whole merged tree including its own prose and everything the merge brought in. Six comments were rewritten in this revision to describe the general shape of their subject rather than a specific sequence; assertions, tests and rationale are unchanged, and the inputs the tests are built from remain as fixture data. One further match is a false positive — two unrelated sentences inside one window — and is recorded in the exemption table with its reasoning, keyed narrowly to that one pairing rather than to the file.

**Structural: CLEAN** — six files, no config, no credentials, no new external I/O.

Scanned by the author.

WingedGuardian:
@codex review

PR #1540, opened 2026-08-29
fix(tests): read the clock when the helper is called, not at module import

WingedGuardian:
## The defect

`test_surplus_liveness.py` read the wall clock **once, at module import**, and seeded a heartbeat 30 minutes ahead of it. Production ages that seed against the **live** clock with a 5-minute future-skew tolerance, so the assertion only held while under **25 minutes** had elapsed since import — the whole suite's runtime, not the test's.

Past that edge it failed; a re-run went green; it read as a flake. A 31-run survey put it at roughly **3% of runs**. Nothing is wrong with the production code, and nothing is wrong on a fast run.

## Evidence

Measured on both sides of the boundary against real production code. Elapsed time was simulated by moving the module's captured value *backwards*, which is exactly equivalent to the import having happened N minutes ago and leaves production on the real clock:

| simulated elapsed | result |
|---|---|
| 16 min | passes |
| 26 min | **fails** |

That matches what CI observed in the wild — a pass 16m02s into a run, a failure at 27m12s.

## The fix

The seed is computed when the helper is called, shrinking the margin from the suite's runtime to one test's duration.

The other four seeds in this file are unaffected, and each was re-derived individually against production rather than assumed: a 300-minute seed against a 180-minute stall threshold (which only grows), a 2-minute seed with 178 minutes of headroom, a paused case that short-circuits before the pulse math, and an unbootstrapped case that raises before the clock is read.

## Scope

Deliberately just the live defect. A CI guard for this whole class — this is its third recurrence, because both earlier sweeps enumerated absolute date *literals* and a clock frozen at import is not one — is in flight separately and is still settling under review. Splitting it out so the flake stops now rather than waiting on that.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved heartbeat timing validation to avoid results varying with test-suite runtime.
  - Added boundary checks for simulated elapsed times of 16 and 26 minutes.

- **Documentation**
  - Documented the updated call-time timing behavior for heartbeat checks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,17 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **A test no longer reads the wall clock once at import and races the suite.**
+  `test_surplus_liveness.py` captured `datetime.now(UTC)` at module import and
+  seeded a heartbeat 30 minutes ahead of it; production ages that seed against
+  the *live* clock with a 5-minute future-skew tolerance, so the assertion only
+  held while under 25 minutes had elapsed since import — the whole suite's
+  runtime, not the test's. Past that edge it failed, and a re-run went green,
+  so it read as a flake; a 31-run survey put it at roughly 3% of runs. The seed
+  is now computed when the helper is called, shrinking the margin from the
+  suite's runtime to one test's. Measured on both sides of the boundary against
+  real production code: the case passes with 16 minutes of simulated elapsed
+  time and fails at 26.
 - **SSH slot cap no longer collapses below the running session count.** The
   interactive-slot launcher (`scripts/cc-slot.sh`) sized its cap from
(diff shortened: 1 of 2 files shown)


WingedGuardian (commented):
Structural review: CLEAN — no pattern violations found. Only change is `tests/test_observability/test_surplus_liveness.py` (module-level `NOW` capture replaced with call-time `datetime.now(UTC)` inside `_install_runtime`, plus an explanatory docstring update) and a `CHANGELOG.md` entry. Import grouping (stdlib/third-party/local) is intact, commit uses a conventional `fix:` prefix, no new untested `.py` files, no mutable defaults, no bare excepts, no raw SQL, no `/tmp` usage, no forbidden git/process patterns.

PII/Secrets/Wording scan: CLEAN — no secrets, private IPs/hosts, credentials, or PII in the diff. The CHANGELOG entry and PR description discuss a test-flakiness root cause (clock captured at import vs. call time) with no security-incident framing or private infrastructure references.




---

WingedGuardian:
@codex review

PR #1543, opened 2026-08-29
fix(hooks): name the writes a refusal discarded, at the point of refusal

WingedGuardian:
## The problem

A PreToolUse hook that exits 2 discards the **whole** Bash call, not the step it objected to. So a command shaped:

```
python3 - <<'PY'
…writes a file…
PY
git commit -m 'x'
```

refused for the commit **also loses the write** — and the refusal names only the commit, so it reads as "the commit didn't happen", never "and the edit you just made never happened".

## Why the note lives at the refusal points

The first attempt on this branch added a standalone hook that tried to predict which commands *would* be refused. It is removed here. To predict a refusal it had to restate four guards' block conditions, and it was measured wrong in **both** directions — refusing shapes no guard blocks, missing shapes they do.

The guard about to refuse is the only thing that knows a block is happening, so the note is emitted there. Nothing new runs on the allow path, and there is no independent verdict that can drift out of sync with the guards it models.

`hook_input.run_guard()` is deliberately **not** touched — it is the fail-closed wrapper imported by 19 files, so a bug there would break four guards at once. Guards with many exit sites wrap their own `main` locally instead.

| file | insertion |
|---|---|
| `review_enforcement_commit.py` | in `_deny()` |
| `git_push_guard.py` | local `main` wrap — all 25 `return 2` sites |
| `full_suite_guard.py` | its single exit-2 |
| `protected_paths_guard.py` | in `_block()` |
| `destructive_command_guard.py` | its single `return 2` |
| `worktree_cwd_guard.py` | local `main` wrap — all 5 sites |
| `bash_safety_hook.sh` | one `EXIT` trap — all 18 exit sites |

Every import is wrapped in `try`/`except` with a no-op fallback. An unguarded import that failed would abort a guard's module load, and Claude Code reads a non-2 exit as a *non-blocking* error — so a broken cosmetic helper could silently fail-open a security guard. That is the one outcome this must never have.

## `shell_parse`: record output-redirect targets (additive)

A parser-only detector reached only 125 of 288 write-carrying blocks, because `analyze` **erases** redirect targets from both text views: `echo hi > f` yields `raw='echo hi'`, `redirects=[]`. `cat > file <<'EOF'` was therefore invisible — bare `cat` looks read-only.

The parser already parsed those redirects correctly with a closed grammar; it simply discarded the target. `Segment.writes` now reports it. Additive: a new field, existing `redirects` semantics untouched, no existing consumer's behaviour changed.

Excluded, each pinned by a test: input redirects (`<`, `<<`, `<<<`), descriptor duplications (`2>&1`, `>&-`), null sinks, process substitutions. **Not** excluded: `>&word` for a non-numeric word, which really does open that file in bash, and a digit filename after a plain `>` (`echo x > 1`).

## Corrections to the commit this replaces

Two numbers in the superseded commit were wrong, both understating:

- **"0.26% of commands"** was measured on one session and presented as a property of the guard. Measured across a full transcript corpus (12,822 unique Bash calls): **1.74%** — ~7x its own estimate. That is the refusal rate a standalone hook would have imposed.
- It claimed the heredoc case **"doesn't occur"**. It is the most common write idiom in the corpus.

## What was measured, and what was not

Of **790** real PreToolUse blocks, 722 were multi-segment. The note fires on **204** and is silent on the rest — 434 multi-segment blocks discarded nothing, and silence there is the point.

**Not established:** how many of those 204 caused unnoticed harm. One is confirmed. The rest may have been noticed immediately. 204 is an upper bound on the benefit, not the benefit. This is worth doing because the cost is near zero, not because the harm rate is known.

## Verification

- **Acceptance** — the real defect replayed end-to-end through each touched guard; the refusal *names* the discarded write, asserted on stderr text rather than exit code.
- **Zero false positives** — a refusal that discarded nothing stays silent; an *allowed* command carrying a write stays silent.
- **Exit codes unchanged**, asserted per guard in both directions, including with the helper made deliberately unreachable: a missing note must never become a missing block, nor a spurious one.
- **Mutation set of 16 — 16/16 caught.** Four earlier mutants *survived* a 34-test suite that looked complete; each was masked by a second check that made the first redundant for the inputs chosen (a lone `git push` writes nothing, so it pins no suppression rule; `2>&1`'s target is dropped as a digit anyway, so it pins no operator rule).
- **Verify-RED** on the shell trap: with it removed the two note assertions fail and the five exit-code assertions still pass — the tests have teeth, and the note is genuinely cosmetic.
- **Corpus replay** caught two defects before any test froze them in: a 500-character heredoc-body blob named as a "file", and unstripped quotes. `shell_parse` does not model heredocs, so a `>` in body prose yields a phantom target; implausible targets are dropped rather than displayed truncated.

3,853 tests green (`tests/test_hooks` + `tests/test_scripts`). `ruff` and `shellcheck` clean. The removed hook's `settings.json` entry is reverted byte-for-byte.

## Adversarial review

A fresh-context adversarial audit of the full diff returned three findings; two were real and are fixed here.

1. **`>&word` misread as a descriptor duplication.** Verified against bash: `bash -c 'echo hi >&f.log'` creates the file. A refused command that truncated a file this way was told nothing had been discarded. Aggravating: a test in this branch *pinned the wrong behaviour* — it was added to kill a mutant and did so by asserting incorrect shell semantics. Fixed, with tests asserting the bash-verified behaviour in both directions.

2. **Unbounded parse work on the block path.** `analyze` is superlinear in nested `$(…)` — measured 0.07s / 0.68s / 2.9s at 100 / 400 / 800 levels. This module re-parsed the command to build the note, and the `EXIT` trap put that on `bash_safety_hook.sh`'s previously pure-bash block paths. A hook killed at its timeout never reaches its `exit 2`, which Claude Code reads as non-blocking — so a cosmetic note could have bought a refusal. Measured at 1,300 nested substitutions: **7.390s** against a 5s timeout. Now bounded by two O(n) scans before parsing, failing open to "no note": 65,536 chars and 256 substitutions, sized from a 13,573-command corpus (longest 14,682 chars, most substitutions 150) so no observed real command is affected. Same shape after the fix: **0.199s**.

3. **"Removing the standalone guard reduces protection"** — accurate, and intended. Nothing now prevents the shape unconditionally; prevention required predicting refusals, and the predictor was wrong in both directions. The finding cites the deleted guard's own 0.26% figure, corrected above.

**Not fixed here:** `shell_parse.analyze`'s recursion is unbounded (`RecursionError` near depth 990) for every caller, independent of this change — confirmed pre-existing by measuring the base branch's own copy (4.533s vs 3.906s on this branch at depth 900, i.e. this branch is not slower). It is not a naive depth cap: today a `RecursionError` inside a `run_guard`-wrapped guard fails **closed**, whereas silent truncation would hand a guard fewer segments and let a crafted command evade its check. Fixing it means choosing a fail direction across six security gates, so it is tracked separately as 6a62e92cc3cb.

## Summary by CodeRabbit

* **New Features**
  * Refused shell commands now identify discarded writes, including redirects, heredocs, and inline scripts.
  * Interactive sessions can prompt for approval when commands mention gated operations that cannot be parsed reliably.
  * Warnings remain informational and do not change existing allow/block decisions or exit codes.

* **Bug Fixes**
  * Improved filtering reduces false-positive write warnings, including quoted null sinks.
  * Enforcement remains fail-closed when diagnostics or optional integrations are unavailable.

* **Tests**
  * Added broad coverage for detection, messaging, edge cases, and enforcement behavior.


WingedGuardian:
@codex review

WingedGuardian:

**Scope:** 12 files (CHANGELOG.md, `scripts/bash_safety_hook.sh`, `scripts/hooks/{destructive_command_guard,discarded_write,full_suite_guard,git_push_guard,protected_paths_guard,shell_parse,worktree_cwd_guard}.py`, `scripts/review_enforcement_commit.py`, `tests/test_hooks/test_discarded_write.py`, `tests/test_scripts/test_bash_safety_hook.py`).

Domain 1 (structural) — checked against all listed ERROR/WARNING/NOTE patterns:
- No `nohup`, no direct `task_states` writes, no `git add -A`/`git add .`, no `--force`/`--hard`, no bare `except:` (all are `except Exception:` / `except (OSError, ValueError):` with typed exceptions), no mutable default args, no `os.killpg`, no raw SQL — none apply here.
- Import ordering is stdlib → local throughout (path-manipulation `noqa: E402` matches the established convention in this hook package); no star imports, no circularity (`discarded_write.py` imports `shell_parse`, not the reverse).
- New file `scripts/hooks/discarded_write.py` ships with a full test suite (`tests/test_hooks/test_discarded_write.py`, 322 lines) — coverage is not missing.
- Public functions/classes are docstringed; `Segment`/`_ParsedSegment` field additions (`writes`) are documented inline and in the class docstring.
- Commit/PR prefix present (`fix(hooks):`).
- The literal `/tmp/...` paths in the two test files (`test_discarded_write.py`, `test_bash_safety_hook.py`) are sample command strings fed to a parser/hook for text analysis, not actual runtime temp-file writes — no `~/tmp/` violation.

Domain 2 (PII/secrets/wording) — scanned added lines plus PR title/body/CHANGELOG:
- No credentials, tokens, private IPs, SSH keys, or connection strings.
- No personal emails, user-bearing absolute paths, revealing timezones, account/project IDs, or private hostnames.
- PR title/body/CHANGELOG describe internal guard-hook mechanics (a cosmetic "discarded write" note on refusal) — no leak/incident language, no exploitable vulnerability detail, no naming of private infrastructure.



---

WingedGuardian:
@codex review

WingedGuardian:
Superseded by #1546, which carries the byte-identical tree on a clean history.

This branch's first commit added a test file containing an absolute home path; a later commit here deleted the whole file, but the leak scanner walks each non-merge commit's patch by design, so the deletion did not clear the finding and the hard gate blocked. Force-push is disabled on this repo, so the history could not be corrected in place.

#1546 was verified as the same code by tree hash — `git diff` between the two tips is empty — and the same CI helper and range that produced the match here report CLEAN there.

PR #1549, opened 2026-08-30
fix(dashboard): report true queue depths, and remove the two mechanisms that hid them

WingedGuardian:
## What was wrong

The Queues panel rendered the length of a deliberately capped `LIMIT 20` review sample as the discarded-queue depth. A backlog of **148 displayed as exactly "20"** — and because the number equalled the cap, it read as a plausible total rather than a truncation. The backend already computed the real count; nothing consumed it.

The same shape had shipped on three more surfaces: both copies of `/genesis/monitor` printed the true count beside a clear-all button gated on the *sample*, so a backlog whose sample failed to load showed `Discarded (148)` with **no clear control at all**; and the attention strip computed "N active error groups" from a list fetched with `limit=6` — a limit rendered as a total.

## Why this PR ends in a redesign

Three review rounds each surfaced a **new** defect, and round 1's fix caused round 2's. That is a design signal rather than bad luck, so the last commit replaces the two mechanisms that generated them instead of patching a fourth instance. Earlier commits are kept as the checkpoint they were.

**Generator 1 — freshness split across two modules and two threads.** A cache and a generation counter lived in the dashboard route module; a single-flight coalescer and a boolean stale marker lived in the health service; invalidation ran on whichever thread the mutating route happened to be on. Rounds 1 and 2 were both defects in how those pieces agreed, and an adversarial audit of round 3 found two more — a cross-thread double-read that turned a *successful* provider toggle into a 500, and a TOCTOU on the publish guard.

The cache now lives in `HealthDataService`, so every field is touched from the event loop alone, and invalidation raised on a request thread is marshalled there. Staleness is the compute task's **identity**, not a flag — which removes the "clear the marker" step where round 2's defect lived, because a task stops being stale only when a replacement takes its place. The generation counter is gone: the publish decision is made inside the compute task, immediately before it returns, atomic with the task completing.

Callers are unaffected — `max_age_s` defaults to 0, so the seven bare callers keep computing fresh. Only this endpoint, polled every 15s by every open tab, opts into staleness, and it now states its own window rather than inheriting one.

**Generator 2 — a capped sample published beside an uncapped count.** Four render surfaces each decided how to reconcile two numbers that can disagree, and several chose wrong in different ways. The producer now reconciles once and publishes `{total, sample, sample_truncated, known}`, in two independent `try` blocks so a failed sample cannot discard a good count. `known` distinguishes a measured zero from an unmeasured one, and the clear-all control stays available when the depth is unknown — it deletes every row regardless, so withholding it was what stranded the backlog.

Truncation is decided by the producer, the only party that knows the sample hit its cap: with a failed count the total falls back to the sample length, so comparing the two reports "complete" at exactly the moment a full sample proves otherwise.

## API change

The health snapshot's `queues` section gains a `discarded` object (`total`, `sample`, `sample_truncated`, `known`). `discarded_count` and `discarded_items` remain and are now derived from it. `discarded_items` is unchanged; `discarded_count` differs in two states, both toward honesty — when the depth query fails it reports the rows in hand rather than `0`, and when depth and sample disagree it reports the larger. The `>100` queue-depth alert keeps working and stops under-reporting in those states.

## Verification

- **Mutation testing is the discriminating evidence**, across three tables. Two rounds of it each exposed genuinely untested paths before they passed: the coalescing-branch `shield` and the stale-drain `suppress` both survived initially, and so did the producer's cap-aware truncation, which had been fixed without a test. The two surviving mutants today are behaviourally inert no-ops (`list(x)` for `x`).
- **The three frontend normalisers are executed**, not pattern-matched: 17 payload shapes (absent object, errored section, non-numeric/null total, non-array sample, `known` truthy-but-not-`true`, a sample at its cap, a total below the sample) asserted against expected values *and* against each other. Agreement alone is not enough — three identically-wrong implementations agree.
- **End to end against a live store:** 148 deep, 20 sampled, truncated, known; the real payload through the shipped normaliser renders **148** where the current code renders 20.
- ruff clean · `check_external_io` / `check_frozen_clock` clean · 172 targeted tests green.

## Guards that can fail

The integrity guard is now a **prohibition** — the raw producer keys may not appear in a frontend file at all — scanning directories rather than a file list, so a new template cannot escape by not being enumerated.

An adversarial audit then mutation-tested the guards themselves and found three that *could not fail*: two scoped by substring against needles that also occur elsewhere in the same file, and the cross-implementation check, which passed with all three normalisers gutted to zeros. Scoping is now by extraction (the enclosing attribute, the function body) and the shape harness carries an oracle. Each of those three is now demonstrably caught by the mutation it was blind to.

## Claims corrected during review

Recorded because they were wrong in the earlier revisions of this PR: a docstring asserted the in-task publish placement was load-bearing on its own — mutation showed it is one half of a redundant pair with unconditional stale-marking, either alone sufficing. A test docstring claimed eleven bare `snapshot()` callers; the measured number is seven. Docstrings claiming state is touched "from the event loop only" and that callers are "bounded by their own deadlines" were false in two live paths and true of one call site respectively.

**Deploy path:** runtime code — `git pull` + server restart. Frontend assets are served from disk, so a browser refresh picks them up after the restart.

## Summary by CodeRabbit

* **New Features**
  * Discarded-item counts now clearly indicate when they are lower-bound estimates with a “+” label.
  * Dashboard and monitoring views provide consistent handling for unknown or unavailable queue totals.
  * Health data refreshes promptly after budget changes, provider toggles, and discarded-item clearing.

* **Bug Fixes**
  * Improved discarded-item sampling and truncation detection.
  * Prevented stale health snapshots and outdated dashboard alerts.
  * Restored consistent error-list display limits.


WingedGuardian:
@codex review

WingedGuardian:

### Structural review

No ERROR-level pattern violations found (no `nohup`, no direct `task_states` inserts, no `/tmp/` usage, no `git add -A`/force-push/`reset --hard`, no bare `except:`, no mutable default args, no new raw SQL added against `genesis.db` outside existing patterns).

**NOTE** (non-blocking):

- `src/genesis/dashboard/routes/budget.py:65`, `src/genesis/dashboard/routes/errors.py`, `src/genesis/dashboard/routes/providers.py` — each does a function-local `from genesis.dashboard.routes.health import invalidate_snapshot_cache` rather than a top-level import. `health.py` doesn't import any of these three modules, so there's no forced circular-import here (unlike the pre-existing local `from genesis.runtime import GenesisRuntime` pattern in the same files, which likely *is* circularity-driven). Probably fine as a deliberate stylistic match to the existing convention, but worth confirming it isn't just copy-paste — a top-level import would be more visible.
- `az_plugins/genesis/templates/neural_monitor.html` and `src/genesis/dashboard/templates/neural_monitor.html` are two byte-identical copies that must be hand-kept in sync (as the PR itself notes). This PR does keep both in sync and adds a test (`test_neural_monitor_surfaces_do_not_gate_on_the_sample`) guarding both — good — but the underlying duplication is a standing structural liability for the *next* change to this panel. Out of scope for this bugfix, flagging for awareness only.

**Coverage / correctness spot-checks (no issues found):**
- `errors.append("discarded: query failed")` in `queues.py` correctly plugs into the pre-existing `errors: list[str]` accumulator that's already conditionally attached to the response (`if errors: result["errors"] = errors`), matching the sibling `deferred_work`/`dead_letters`/`pending_embeddings` error-handling pattern exactly.
- `HealthDataService.abandon_inflight()` correctly nulls `self._inflight` without disturbing callers that already captured the in-flight task reference — existing awaiters aren't affected, only new callers get a fresh compute.
- Extensive new/updated test coverage across `test_health_cache.py`, `test_webui_js_integrity.py`, and the new `test_queues_discarded_count.py`; the JS integrity tests use positive assertions (not just negative scans), which correctly avoids the exact blind spot called out in the PR description.
- Commit carries a conventional prefix (`fix(dashboard): ...`).

### PII/Secrets/Wording scan

**CLEAN.** No API keys, tokens, private IPs, SSH keys, or credential strings in the diff. No personal emails, username-bearing paths, or hardware/account identifiers. PR title, body, and CHANGELOG additions describe the bug and fix in generic technical terms (queue depth counts, caching mechanics) with no reference to specific private infrastructure, security incidents, or exploitable leak details.

---

---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
## Leak / private-data scan

Scope: every added line in `git diff origin/main...HEAD` at head `b6ef2fd38446aa4ec82738878c8de767fb449666` — **2,234 added lines** across 20 changed files. Prose, code, comments, docstrings, test fixtures and the new JS test asset were all scanned, not just config values.

Categories checked, each returning zero matches:

| category | result |
|---|---|
| IPv4 addresses (excluding loopback) | none |
| IPv6 ULA prefixes, tailnet ranges, MAC addresses | none |
| Home paths embedding a username, `ubuntu@`-style host refs | none |
| API keys, tokens, bearer credentials, private-key blocks | none |
| Email addresses | none |
| Personal or household context, real names, location | none |

Two things worth noting positively, since they are the usual sources of an inferential leak in this repo:

- The new test asset `tests/assets/discarded_shape_check.js` derives the repo root from its own location rather than carrying an install path, so it runs from any clone or CI checkout.
- Test fixtures use synthetic rows seeded relative to now, with no real queue contents, identifiers or timestamps from any live store. The one live measurement quoted in commit messages and the PR body is an aggregate row count with no row contents.

VERDICT: PASS — scan CLEAN, no private or identifying data found.

WingedGuardian:
@codex review

WingedGuardian:
## Leak / private-data scan

Scope: every added line in `git diff origin/main...HEAD` at head `b938af44bc4007e31ef3281d97afe4699dccc343` — **2,120 added lines** across 19 changed files. Prose, code, comments, docstrings, test fixtures and the JS test asset were all scanned, not only config values.

| category | result |
|---|---|
| IPv4 addresses (excluding loopback) | none |
| IPv6 ULA prefixes, tailnet ranges, MAC addresses | none |
| Home paths embedding a username, `user@host` refs | none |
| API keys, tokens, bearer credentials, private-key blocks | none |
| Email addresses | none |
| Personal or household context, real names, location | none |

Two notes, both in the direction of caution rather than reassurance:

- The pre-push hook flagged `tests/test_hooks/test_destructive_command_guard.py:104-105`. Those lines are **not from this branch** — they arrived with the merge of `origin/main` (commit `e8edba6b`, PR #1547) and are already public. They are synthetic placeholders (`/home/u/...`, username literally `u`), not a real path.
- The only live measurement quoted anywhere in this branch's commits, PR body or code comments is an aggregate row count (148 discarded/expired rows). No row contents, identifiers, timestamps or payloads from the live store appear. Test fixtures seed synthetic rows relative to now, so they carry neither real data nor a wall-clock dependency.

VERDICT: PASS — scan CLEAN, no private or identifying data found.

WingedGuardian:
@codex review

WingedGuardian:
**Leak scan — 2248 added lines, 19 files (`git diff origin/main...HEAD`)**

| category | pattern classes checked | hits |
|---|---|---|
| Direct identifiers | IPv4, tailnet 100.64/10, `.local`/`ts.net` hostnames | 0/2248 |
| Credentials | api key / token / secret / password / bearer / `sk-` / `ghp_` / AKIA | 0 (2 prose uses of the word "token" meaning a JS source token) |
| Paths & accounts | `/home/<user>`, `/Users/`, `~/.ssh`, email addresses, account handles | 0/2248 |
| Indirect personal context | household / family / languages-at-home / schedule / named individuals | 0/2248 |
| Endpoints | absolute URLs, `localhost:<port>`, RFC1918 ranges | 0/2248 |

Install-specific values in the diff are queue depths only (a discarded-queue
count used in a docstring to date an observed defect) — operational, generic to
any install, no real-world inference.

PII: **CLEAN**
Secrets: **CLEAN**
Wording: **CLEAN**

WingedGuardian:
**Leak scan — 2428 added lines, 19 files (`git diff origin/main...HEAD`)**

| category | pattern classes checked | hits |
|---|---|---|
| Direct identifiers | IPv4, tailnet 100.64/10, `ts.net`/tailnet hostnames | 0/2428 |
| Credentials | api key / password / bearer / `sk-` / `ghp_` / AKIA | 0/2428 |
| Paths & accounts | `/home/<user>`, `/Users/`, `~/.ssh`, email addresses, account handles | 0/2428 |
| Indirect personal context | household framing, employer names, family/schedule/language inference | 0/2428 |
| Endpoints | absolute URLs, `localhost:<port>`, RFC1918 ranges | 0/2428 |

Install-specific values in the diff are queue depths only (a discarded-queue
count used in a docstring to date an observed defect) — operational, generic to
any install, no real-world inference. Benchmark row counts in a follow-up
reference are synthetic.

PII: **CLEAN**
Secrets: **CLEAN**
Wording: **CLEAN**

WingedGuardian:
@codex review

WingedGuardian:
**Leak scan — 2569 added lines, 20 files (`git diff origin/main...HEAD`)**

| category | pattern classes checked | hits |
|---|---|---|
| Direct identifiers | IPv4, tailnet ranges, `ts.net`/tailnet hostnames | 0/2569 |
| Credentials | api key / password / bearer / `sk-` / `ghp_` / AKIA | 0/2569 |
| Paths & accounts | `/home/<user>`, `/Users/`, `~/.ssh`, email addresses, account handles | 0/2569 |
| Indirect personal context | household framing, employer names, family/schedule/language inference | 0/2569 |
| Endpoints | absolute URLs, `localhost:<port>`, RFC1918 ranges | 0/2569 |

Install-specific values in the diff are queue depths only, used in a docstring to
date an observed defect — operational, generic to any install, no real-world
inference.

PII: **CLEAN**
Secrets: **CLEAN**
Wording: **CLEAN**

WingedGuardian:
**Leak scan — 2569 added lines, 19 files (`git diff origin/main...HEAD`)**

| category | pattern classes checked | hits |
|---|---|---|
| Direct identifiers | IPv4, tailnet ranges, `ts.net`/tailnet hostnames | 0/2569 |
| Credentials | api key / password / bearer / `sk-` / `ghp_` / AKIA | 0/2569 |
| Paths & accounts | `/home/<user>`, `/Users/`, `~/.ssh`, email addresses, account handles | 0/2569 |
| Indirect personal context | household framing, employer names, family/schedule/language inference | 0/2569 |
| Endpoints | absolute URLs, `localhost:<port>`, RFC1918 ranges | 0/2569 |

Scanned at this exact head after merging main. Install-specific values in the
diff are queue depths only, used in a docstring to date an observed defect —
operational, generic to any install, no real-world inference.

PII: **CLEAN**
Secrets: **CLEAN**
Wording: **CLEAN**

PR #1550, opened 2026-08-30
fix(hooks): enforce the model-facing output cap at one chokepoint

WingedGuardian:
## Summary

Claude Code persists a hook's stdout above a size threshold — the full output goes to `tool-results/hook-<uuid>-stdout.txt` and the model gets a ~2 KB preview — with no error and no documentation. The SessionStart injection had been over that threshold for a month, so affected windows ran without identity, essential knowledge, in-flight state, and the session charter. The failure is indistinguishable from success, which is why it lasted.

## Measurement (before any constant was chosen)

~25 real probe sessions via an env-gated seam in the hook, classified from each session's own transcript:

| | |
|---|---|
| Threshold | exactly **10,000 characters** (10,000 inline / 10,001 filed) |
| Unit | characters, not bytes (6,000 two-byte chars = 12,044 B → inline) |
| Scope | **per hook entry** (two 9 K hooks on one SessionStart → both inline) |
| Mode | same in `-p` and interactive |
| Stability | a minor CC update dropped it ~3× and tripled the filing rate the same day |
| Model-facing events | only SessionStart, UserPromptSubmit, UserPromptExpansion carry bare stdout; everything else reaches the model only via JSON, which is persisted the same way |
| Read-back | the Read tool's result cap is unbounded, so reading a filed payload is safe and idempotent — which is what makes a recovery instruction viable |

## Change

**1. Enforcement moved to the writer.** New `scripts/hooks/hook_output.py` is the single home of the measured cap and provides `BoundedStdout`, which cannot exceed its budget regardless of whether individual blocks check. Two of twelve blocks in the knowledge part checked; the other ten are how the payload grew past the cliff. `print_json_bounded` trims named free-text fields and never the JSON envelope, so an oversized advisory loses prose, never its decision.

**2. A cut stays recoverable.** A hard cut would destroy a tail the harness at least kept on disk, so every part is mirrored whole to `~/.genesis/sessions/<sid>/context-<part>.md`, and the in-band cut marker names it.

**3. The model is told, in the only channel it can see.** Each part opens with a `[genesis-ctx:<part> · mirror: …]` recovery header, which lands inside the harness's own 2 KB preview by construction. A CLAUDE.md section says what the wrapper means and what to do about it. Baseline for this: across 387 transcripts, no session ever followed the harness's own "Full output saved to:" line unprompted — while sessions *did* act on instructions in the preview head.

**4. The over-budget marker layer is deleted** (emitter writer, per-prompt alert, collector reader, tests). A marker could only record a condition the chokepoint now makes unreachable; the two conditions it never covered — the cap moving below our budget, and any *other* hook being filed — are exactly what the header and the watcher see. Mis-wire detection moves in-band, because a SessionStart hook's stderr on exit 0 goes to the debug log and never reaches the model.

**5. Watcher simplified and corrected.** Scoped to this install's checkouts (it scored real filings "foreign" when run from a worktree — a fail-silent bug in the thing built to detect fail-silent bugs); age filter applied *before* the scan cap so a backlog cannot crowd out a live incident; filings attributed by the part stamp with a head excerpt for everything else, so none is dropped; resolves its alert when disabled.

**6. Fixes carried in.** `enabled: null|0|[]` silenced the watcher through plain truthiness. The per-turn ledger inventory emitted 8-char ids that `session_ledger_update` rejects, making the inventory unactionable — now full ids, with the byte cap raised to fit them. A ledger beyond the fetch bound was cut silently; it is now announced in the footer, which survives every degrade tier. `_CHARTER_BLOCK_MAX` 8,000 → 7,500, with the part's fixed overhead pinned by a test so a flat cut can never reach a ledger row.

**7. Kept from the earlier revision:** the four-part split, fail-closed `--part`, the CONVERSATION.md diet (11,283 → 5,828 chars, detail moved to a reference doc), the charter block and per-turn inventory, the CI ceilings, and the wiring test that makes the un-split form unrepresentable.

## Alternative evaluated and rejected

Identity via CLAUDE.md `@import` was probed with real sessions. Imports work: they survive `-p` and `--system-prompt`, and a 14,430-char file arrives whole. Rejected on two measured grounds — a missing import is a **silent skip** and the user-profile file is untracked, so every fresh clone would lose it with no signal (this bug's own failure mode); and the only suppression lever drops project *hooks* along with CLAUDE.md, which would strip guards from worktree-cwd sessions. Recorded in `docs/reference/cc-compatibility.md` so it is not re-derived.

## Verification

- 169 targeted tests across the writer, hook, tag, charter block, budget/wiring/ceilings, watcher, provenance and settings. Verify-RED by mutation: disabling the chokepoint turns 6 writer tests red; a regrown charter ceiling fails the arithmetic pin with the numbers.
- Two off-by-ones in the writer and a fail-silent scoping bug in the watcher were found **by the tests and the live run**, not by reading.
- Live: four parts at 1,021 / 8,395 / 7,663 / 6,388 chars against a 9,800 budget (23,467 total, against ~32,000 previously in one entry); mirrors written; mis-wire path loud in-band with a negative control; watcher reporting the genuine filings with none misattributed and none foreign.
- Local CI guards (ruff, external-io, subsystem-map, shared-artifact, frozen-clock) clean.

## Deploy path

Hook + settings land at the next session start after `git pull`; the awareness check needs a server restart. **Restart foreground sessions after deploy** — a session started before the fix keeps the old wiring until it does, and will keep filing (and the watcher will keep alerting, correctly) until then. After any CC bump: re-measure with the probe seam and re-verify the wrapper literals; the method is in `docs/reference/cc-compatibility.md`.

Ledger: 4f70bba79dfc4b17b67384de13dda4bb
Ledger: 3246353ef61847468dda1df73b127bb8

## Summary by CodeRabbit

* **New Features**
  * Added four-part session context injection for charter, identity, user, and knowledge information.
  * Added bounded hook output with recovery mirrors and diagnostics when content is truncated or unavailable.
  * Added monitoring and configurable alerts for missing or misrouted context injection.
  * Added expanded session ledger status reporting, including open items, mission status, and overflow notices.

* **Documentation**
  * Updated architecture, compatibility, and conversation protocol guidance for the new session behavior.

* **Bug Fixes**
  * Improved handling of malformed settings, unreadable identity files, and incomplete hook output.


WingedGuardian:
@codex review

WingedGuardian:
**Scheduled leaks review — head `6144aab8980e`**

Whole branch surface (`origin/main...HEAD`): **2,702 added lines** across 20 files. Pattern classes scanned: private IPv4 (RFC1918), IPv6 ULA, tailnet range, home paths embedding a username, email addresses, CC project-dir slug, install timezone, credential shapes (sk-/ghp-/sk-ant-), personal names and handles. **Matches: 0.**

Scanner verified LIVE with a seeded positive control (3 classes seeded, 3/3 detected) — a zero from an inert scanner and a zero from a clean diff are otherwise identical.

Note on one deliberate near-miss: the new tests build a CC project directory named `-home-user-genesis`. That is a synthetic placeholder chosen to exercise the real path shape; the actual slug appears zero times in the diff (verified by direct grep). The pre-push privacy hook flagged it on shape, which is the hook working correctly.

Fixtures are synthetic throughout — placeholder session ids, generic prose, repeated-digit ids.

**VERDICT: PASS**

WingedGuardian:
## Automated First-Pass Review

**Structural review:** 1 finding (0 errors, 0 warnings, 1 note)

**NOTE** — `scripts/genesis_urgent_alerts.py:8`: the module-level docstring's numbered summary still describes the old tag format (`2. Charter drift tag ([Charter: … | open: N]; chartered sessions only)`), but this PR changes the actual emitted format to `[Ledger open: N | mission: …]` with one `- <id8> <text>` line per open row (per the new `_emit_charter_tag` docstring and the rewritten tests). Worth a one-line update so a reader skimming just the top-of-file summary doesn't get a stale picture of what the hook prints.

No ERROR-level structural violations found: no `nohup`, no direct `task_states` inserts, no `/tmp/` usage, no `git add -A`/`push --force`/`reset --hard`, no worktree `pip install -e`, no bare `except:`, no mutable default args, no unvalidated `os.killpg`. The parameterized SQL in the two hook scripts (`genesis_session_context.py`, `genesis_urgent_alerts.py`) is raw SQL against `genesis.db`, but both are documented stdlib-only hooks that structurally can't import `genesis.db.crud` without defeating the "broken venv can never wedge a session" guarantee this PR itself documents — not flagged. Import hygiene is clean (correct stdlib/third-party/local grouping, no star/circular imports), no new `IntervalTrigger` usage, hook scripts add stdlib-only imports. All three new `src/` modules (`context_injection_watch_config.py`, `context_injection.py`, `ledger_escalation_link.py`) have real test coverage, verified by reading the tests rather than inferring from filenames.

No secrets, tokens, or credentials in added lines (only false-positive prose matches). No private IPv4/IPv6 addresses, SSH keys, or DB connection strings with credentials. No personal emails, no usernamed absolute paths, no location-revealing timezone identifiers, no AWS/GCP IDs or hardware identifiers. PR title/body describes a Claude Code hook stdout-truncation measurement and fix — no wording that references a security incident, data exposure, or named private infrastructure.

---


---

WingedGuardian:
@codex review

WingedGuardian:
**Leak / PII scan — VERDICT: CLEAN**

Scanned all 3,987 added lines of `origin/main...d2e53ec167b17d96255b3dc9e97c7d5dac1151ea` against RFC1918 IPv4, tailnet ranges, IPv6 ULA, slash-form home paths, **dash-form home paths** (a Claude Code project slug is an absolute path with `/` and `.` replaced by `-`, so it carries a home path in a shape a slash-oriented pattern cannot see), install timezone, operator usernames, `sk-ant-`/`ghp_`/`AKIA` token shapes, and private key headers.

**Real hits: 0.** Seeded positive control fired on 4/4 seeded classes, so the scan is live rather than inert.

One finding was caught and fixed during this scan rather than after it: two test constants held real project slugs (`d2e53ec1`). They are opaque fixtures — every assertion consumes their shape, never a literal path — so synthetic values are equivalent.

Fixtures throughout are synthetic; no install-specific paths, addresses, or identifiers appear in any added line.

WingedGuardian:
Superseded by #1556 and closing.

This branch cannot merge as-is. Its history contains a test constant holding a real Claude Code project slug — an absolute path with `/` and `.` mapped to `-`, so it carries a home path. The leak-detector scans the added lines of every commit rather than the net diff, which is correct: a value added and later removed still lives in published history. Removing it in a follow-up commit therefore cannot clear the gate, and force-pushing the public repo is prohibited.

#1556 is the same reviewed tree as a single clean commit on current main. Verified byte-identical per file in both directions, and re-checked with the same scanner and fingerprint set the gate itself uses.

PR #1552, opened 2026-08-30
fix(hooks): make the merge gate's "(logged)" promise true

WingedGuardian:
The merge gate told the operator that appending `# ci-override` was "(logged)", and
the development skill said "The override is logged." Nothing wrote a record. Every
writer under `scripts/` and `src/` was enumerated: there was none. A gate that
promises an audit trail it does not keep is worse than one promising nothing — it
invites the escape on the belief that it is recorded.

## What is recorded

One metadata row per (merge, sigil) in `~/.genesis/merge_override_log.jsonl`:
`{ts, sigil, outcome, waived, pr, repo, head, actor}`.

`head` is the `--match-head-commit` value, which GitHub enforces server-side, so a
successful merge's row names the sha that merged. `outcome` is resolved at the gate's
verdict, not at detection: a sigil appended to a command that some **other** gate then
blocks is real signal — the escape was reached for — but it is not an override that
took effect, and counting the two together would inflate every figure drawn from the
file.

Metadata only. An override sigil is a trailing comment on a Bash command that can
carry a credential, so no command or comment text reaches the row and there is no
reason field to add. That constraint is inherited from `git_discard_guard`, whose
recovery log refuses command text for the same reason. Every value is shape-checked,
not just the field names.

## Scope, and what is deliberately outside it

The four **PR-merge** sigils are covered: `# review-override`, `# ci-override`,
`# stale-review-override`, `# scheduled-review-override` — from the point the PR
resolves onward. A merge rejected before that (no `--admin`, unresolvable repo or PR)
writes nothing; those are malformed commands, not override events.

The two **command-scoped acks** — `# escalation-ack` and `# merge-to-main-override` —
are deliberately **not** logged. Their gate is not evaluated where the sigil is seen,
so a row could only say the ack was typed, never that it waived anything. Two attempts
to attribute them honestly both produced fiction, measured: `escalation-ack` on
`git commit`, `git status` and `pytest` each claimed the round cap was waived; after
narrowing, `git commit … # escalation-ack && gh pr comment 5 …` recorded the commit
gate's ack against PR 5; and `merge-to-main-override` on a repo at a feature branch,
where the gate would have allowed the merge anyway, still claimed a waiver with every
identifying field empty. A test pins the exclusion as a decision. Logging them needs
the ack bound to the segment it rides — filed as a follow-up.

## Shared writer, whole precedent

The locking and create modes move out of `git_discard_guard` into a new
`scripts/hooks/audit_jsonl.py` used by both: 0600/0700, `os.open` applying the mode at
create time (no umask window), an exclusive flock on a sidecar that atomic rewrites
cannot orphan, and retention/trim as maintainer callables. An earlier cut cited that
precedent and adopted only its credential rule; the live log it produced was mode
**644**.

Retention keeps rows it cannot interpret. An earlier cut compared an aware cutoff
against a row timestamp outside the per-row `try`, so one timezone-naive row raised,
abandoned the write, and did so on every later write forever — while the gate kept
printing "(logged)". Each maintainer is now isolated, and a size cap bounds a file of
rows retention cannot read.

## Two things this had to fix in itself

**A fail-open.** The flush took the sidecar lock with a blocking `flock`, from `main`'s
`finally` — the path that delivers a block. Measured: with another process holding the
lock the guard produced no verdict and was killed at 20s, against a 0.00s baseline.
Past the hook wall clock that is a SIGKILL, which is not exit 2, which lets the guarded
command run. Acquisition is now bounded and non-blocking with one deadline shared
across the whole flush; a contended write drops the **row**, never the verdict.
Re-measured at 0.51s.

**Forgeable attribution.** `actor` used `getpass.getuser()`, which reads
`$LOGNAME`/`$USER` before the passwd database — so the command being audited could set
the one field whose job is saying who ran it. It now reads the uid.

## Sigil declaration

`# merge-to-main-override` and `# full-suite-ok` were passed to
`has_trailing_override` since they shipped but never listed in `_KNOWN_SIGILS`, which
made that tuple's "kept in sync" comment false and had a measured effect: the
leading-run scan treats an unlisted token as prose, so an unlisted sigil written first
silently disabled every sigil after it (`# full-suite-ok audit-ack` → `audit-ack`
undetected). Both are declared now, and an `ast` walk over every consumer pins the set.

Declaring them **widens acceptance as well as detection**, in the fail-open direction.
Every newly-accepted combination is enumerated in the code comment; the sharpest is
`# full-suite-ok discard-override`, which reaches the untracked-file deletion block.
All require the operator to type both tokens literally, and nothing in the repo
auto-composes a multi-sigil comment.

## Verification

45 tests in the new module. Every new assertion was verified RED against a mutation of
the specific behaviour it pins — 14/14, verdicts read from the pytest summary line
rather than the exit code (this repo's wrapper exits 200 under lock contention, which
an exit-code rule scores as a false "caught"), with restore checked against
pre-mutation hashes. One mutation is skipped with its reason recorded rather than
silently omitted.

Four tests were found vacuous during review and fixed rather than kept, including a
size backstop that passed with the trim entirely inert and two path assertions that
checked only a suffix, so a mutant putting the log inside the repo worktree passed the
very tests written to prevent that. The E2E harness was also making live GitHub API
calls, and four tests claiming to exercise "a real merge" were silently blocking; it
is now network-free by construction.

1512 tests across 21 affected suites pass. `ruff` clean. Differential against `main` on
four payloads (push arm, local merge, ask, multi-sigil): decisions byte-identical.

An autouse fixture isolates the log for the whole test suite. Without it, any test
driving the gate with a sigil writes a durable row to the real store — which is how
four rows describing a nonexistent PR reached the live file during development.

No change to any block/allow decision beyond the sigil widening named above.

## Summary by CodeRabbit

* **New Features**
  * Merge override actions are now recorded with the relevant pull request, commit, waived gate, and final outcome.
  * Added support for recognizing additional merge approval sigils.
  * Audit records exclude command text and sensitive credentials.

* **Bug Fixes**
  * Logging failures no longer interrupt merge or discard protections.
  * Audit logs are secured, size-limited, and automatically maintained.

* **Documentation**
  * Updated development guidance to describe merge override logging behavior and scope.

### DIFF
--- a/.claude/skills/genesis-development/SKILL.md
+++ b/.claude/skills/genesis-development/SKILL.md
@@ -1474,7 +1474,17 @@ The review-findings gate specifically:
    on CI alone. Note in PR that review was quota-limited.
 5. **Override**: Append `# review-override` to the merge command to
    bypass the gate (e.g., `gh pr merge 123 --squash --admin  # review-override`).
-   The override is logged. Use only when findings are intentionally accepted.
+   The override is logged — one metadata row per sigil in
+   `~/.genesis/merge_override_log.jsonl` (sigil, PR, bound head, which gate it
+   waived, and what the command actually got: `allowed` / `asked` / `blocked` /
+   `error`), so "was this escape reached for?" is a query rather than a survey.
+   `blocked` matters: a sigil can be appended to a command that some OTHER gate
+   then stops, which is a real signal but not an override that took effect, and
+   `asked` means the decision went to a human who may still have denied it.
+   Never any command text — the row is metadata only, because a trailing comment
+   rides a Bash command that can carry a credential. Every merge-path sigil is
+   covered, including `# escalation-ack`. Use only when findings are
+   intentionally accepted.
 6. **Read the PR's warning comments before merging — not just the hard gate.**
    Beyond Codex, a structural-review bot posts under the repo-owner account
    (`WingedGuardian`, review state COMMENTED) and emits **SOFT WARNINGs** (PII /
--- a/.claude/skills/genesis-development/SKILL.md
+++ b/.claude/skills/genesis-development/SKILL.md
@@ -1482,8 +1482,15 @@ The review-findings gate specifically:
    then stops, which is a real signal but not an override that took effect, and
    `asked` means the decision went to a human who may still have denied it.
    Never any command text — the row is metadata only, because a trailing comment
-   rides a Bash command that can carry a credential. Every merge-path sigil is
-   covered, including `# escalation-ack`. Use only when findings are
+   rides a Bash command that can carry a credential.
+   **Scope, stated so the log is not read as more complete than it is:** the four
+   PR-merge sigils are covered (`# review-override`, `# ci-override`,
+   `# stale-review-override`, `# scheduled-review-override`), from the point the
+   PR is resolved onward. A merge rejected before that — no `--admin`, or an
+   unresolvable repo/PR — writes nothing. The command-scoped acks
+   (`# escalation-ack`, `# merge-to-main-override`) are deliberately NOT logged:
+   their gate is not evaluated where the sigil is seen, so a row could only say
+   the ack was typed, never that it waived anything. Use only when findings are
    intentionally accepted.
 6. **Read the PR's warning comments before merging — not just the hard gate.**
    Beyond Codex, a structural-review bot posts under the repo-owner account
--- a/scripts/hooks/audit_jsonl.py
+++ b/scripts/hooks/audit_jsonl.py
@@ -1,0 +1,260 @@
+#!/usr/bin/env python3
+"""Secure append-only JSONL for hook audit/recovery logs.
+
+Extracted from ``git_discard_guard.py`` (2026-08-30) when a SECOND hook log —
+the merge gate's override record — needed the same properties. The mechanics
+below are security properties, not conveniences, and a partial adoption is the
+failure this module exists to prevent: the override log's first cut took the
+credential rule from the discard guard and left the 0600 modes, the umask-free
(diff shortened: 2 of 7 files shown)


WingedGuardian:
Structural review: CLEAN — no pattern violations found. Checked: no bare `except:`, no mutable default args, stdlib-only imports in the new `scripts/hooks/audit_jsonl.py`, no `nohup`/`/tmp/`/`git add -A`/`--force`/raw-SQL/`os.killpg` patterns, docstrings present on new/changed public functions, and existing `git_push_guard`/`git_discard_guard` tests remain compatible with the internal refactor (verified no stray references to the removed `_open_own`/`_restrict`/`_write_log_row` helpers elsewhere in `tests/`). Both commits carry conventional `fix(hooks):` prefixes.

PII/Secrets/Wording scan: CLEAN — no private IPs, secrets, SSH keys, connection strings, or personal emails in the diff. The only credential-shaped string is an obvious placeholder (`ghp_EXAMPLETOKENVALUE1234567890`) in test fixtures exercising the no-credential-leak guarantee. PR title/body/commit messages describe the audit-logging fix in generic engineering terms with no incident language or named private infrastructure.

---



---

WingedGuardian:
Superseded by #1553 — identical tree, clean history.

CI's private-pattern exact scan hard-failed here. It scans added lines across EVERY commit in the range by design, so a value added in one commit and removed in the next is still caught. This branch's first commit contained a home path in a test fixture; the second commit scrubbed it, which leaves the net diff clean but not the added-lines range.

Force-push is blocked on this repo, so the branch was rebuilt as a single commit from origin/main. `git diff` between the two heads is empty — the reviewed tree is byte-identical, and all verification carries over.

PR #1553, opened 2026-08-30
fix(hooks): make the merge gate's "(logged)" promise true

WingedGuardian:
The merge gate told the operator that appending `# ci-override` was "(logged)", and
the development skill said "The override is logged." Nothing wrote a record. Every
writer under `scripts/` and `src/` was enumerated: there was none. A gate that
promises an audit trail it does not keep is worse than one promising nothing — it
invites the escape on the belief that it is recorded.

## What is recorded

One metadata row per (merge, sigil) in `~/.genesis/merge_override_log.jsonl`:
`{ts, sigil, outcome, waived, pr, repo, head, actor}`.

`head` is the `--match-head-commit` value, which GitHub enforces server-side, so a
successful merge's row names the sha that merged. `outcome` is resolved at the gate's
verdict, not at detection: a sigil appended to a command that some **other** gate then
blocks is real signal — the escape was reached for — but it is not an override that
took effect, and counting the two together would inflate every figure drawn from the
file.

Metadata only. An override sigil is a trailing comment on a Bash command that can
carry a credential, so no command or comment text reaches the row and there is no
reason field to add. That constraint is inherited from `git_discard_guard`, whose
recovery log refuses command text for the same reason. Every value is shape-checked,
not just the field names.

## Scope, and what is deliberately outside it

The four **PR-merge** sigils are covered: `# review-override`, `# ci-override`,
`# stale-review-override`, `# scheduled-review-override` — from the point the PR
resolves onward. A merge rejected before that (no `--admin`, unresolvable repo or PR)
writes nothing; those are malformed commands, not override events.

The two **command-scoped acks** — `# escalation-ack` and `# merge-to-main-override` —
are deliberately **not** logged. Their gate is not evaluated where the sigil is seen,
so a row could only say the ack was typed, never that it waived anything. Two attempts
to attribute them honestly both produced fiction, measured: `escalation-ack` on
`git commit`, `git status` and `pytest` each claimed the round cap was waived; after
narrowing, `git commit … # escalation-ack && gh pr comment 5 …` recorded the commit
gate's ack against PR 5; and `merge-to-main-override` on a repo at a feature branch,
where the gate would have allowed the merge anyway, still claimed a waiver with every
identifying field empty. A test pins the exclusion as a decision. Logging them needs
the ack bound to the segment it rides — filed as a follow-up.

## Shared writer, whole precedent

The locking and create modes move out of `git_discard_guard` into a new
`scripts/hooks/audit_jsonl.py` used by both: 0600/0700, `os.open` applying the mode at
create time (no umask window), an exclusive flock on a sidecar that atomic rewrites
cannot orphan, and retention/trim as maintainer callables. An earlier cut cited that
precedent and adopted only its credential rule; the live log it produced was mode
**644**.

Retention keeps rows it cannot interpret. An earlier cut compared an aware cutoff
against a row timestamp outside the per-row `try`, so one timezone-naive row raised,
abandoned the write, and did so on every later write forever — while the gate kept
printing "(logged)". Each maintainer is now isolated, and a size cap bounds a file of
rows retention cannot read.

## Two things this had to fix in itself

**A fail-open.** The flush took the sidecar lock with a blocking `flock`, from `main`'s
`finally` — the path that delivers a block. Measured: with another process holding the
lock the guard produced no verdict and was killed at 20s, against a 0.00s baseline.
Past the hook wall clock that is a SIGKILL, which is not exit 2, which lets the guarded
command run. Acquisition is now bounded and non-blocking with one deadline shared
across the whole flush; a contended write drops the **row**, never the verdict.
Re-measured at 0.51s.

**Forgeable attribution.** `actor` used `getpass.getuser()`, which reads
`$LOGNAME`/`$USER` before the passwd database — so the command being audited could set
the one field whose job is saying who ran it. It now reads the uid.

## Sigil declaration

`# merge-to-main-override` and `# full-suite-ok` were passed to
`has_trailing_override` since they shipped but never listed in `_KNOWN_SIGILS`, which
made that tuple's "kept in sync" comment false and had a measured effect: the
leading-run scan treats an unlisted token as prose, so an unlisted sigil written first
silently disabled every sigil after it (`# full-suite-ok audit-ack` → `audit-ack`
undetected). Both are declared now, and an `ast` walk over every consumer pins the set.

Declaring them **widens acceptance as well as detection**, in the fail-open direction.
Every newly-accepted combination is enumerated in the code comment; the sharpest is
`# full-suite-ok discard-override`, which reaches the untracked-file deletion block.
All require the operator to type both tokens literally, and nothing in the repo
auto-composes a multi-sigil comment.

## Verification

45 tests in the new module. Every new assertion was verified RED against a mutation of
the specific behaviour it pins — 14/14, verdicts read from the pytest summary line
rather than the exit code (this repo's wrapper exits 200 under lock contention, which
an exit-code rule scores as a false "caught"), with restore checked against
pre-mutation hashes. One mutation is skipped with its reason recorded rather than
silently omitted.

Four tests were found vacuous during review and fixed rather than kept, including a
size backstop that passed with the trim entirely inert and two path assertions that
checked only a suffix, so a mutant putting the log inside the repo worktree passed the
very tests written to prevent that. The E2E harness was also making live GitHub API
calls, and four tests claiming to exercise "a real merge" were silently blocking; it
is now network-free by construction.

1512 tests across 21 affected suites pass. `ruff` clean. Differential against `main` on
four payloads (push arm, local merge, ask, multi-sigil): decisions byte-identical.

An autouse fixture isolates the log for the whole test suite. Without it, any test
driving the gate with a sigil writes a durable row to the real store — which is how
four rows describing a nonexistent PR reached the live file during development.

No change to any block/allow decision beyond the sigil widening named above.

## Summary by CodeRabbit

* **New Features**
  * Added secure audit logging for merge override actions, including metadata and command outcomes.
  * Added retention, size-based trimming, validation, redaction, and graceful handling of logging failures.
  * Recognizes additional merge and full-suite approval acknowledgements.
  * Supports carrying forward eligible leak reviews when required checks pass.

* **Bug Fixes**
  * Improved override detection and parsing across merge workflows.
  * Improved snapshot recovery messages when logging is unavailable.
  * Strengthened validation of review and pin-file changes.
  * Preserved existing guard behavior when audit logging fails.

* **Documentation**
  * Clarified audit-log creation rules and documented locking and durability limitations.


WingedGuardian:
## Structural review: CLEAN

Checked all 4 changed/new `.py` files (`scripts/hooks/audit_jsonl.py` new, `scripts/hooks/git_discard_guard.py`, `scripts/hooks/git_push_guard.py`, `scripts/hooks/shell_parse.py`) plus `tests/conftest.py` and the new `tests/test_hooks/test_merge_override_log.py` against the Domain 1 checklist:

- No pattern-violation ERRORs (`nohup`, direct `task_states` inserts, `/tmp/` usage, `git add .`/`-A`, force-push/reset --hard, worktree `pip install -e`, bare `except:`, mutable default args, unvalidated `os.killpg`, raw SQL). All `except Exception as exc:` sites carry an explicit type + `# noqa: BLE001` justification, and all keyword defaults (`maintain: Sequence[...] = ()`, `pr: str = ""`, `repo: str | None = None`, etc.) are immutable.
- Import hygiene: `audit_jsonl.py` is stdlib-only (`contextlib`, `datetime`, `fcntl`, `json`, `os`, `sys`, `time`, `collections.abc`) — correct for a hook-adjacent module. No star imports, no circular imports; the `import audit_jsonl` in `git_discard_guard.py`/`git_push_guard.py` is a guarded soft-dependency (`noqa: E402`, degrades to `None` on failure) rather than a hard import-order problem.
- Coverage: verified `_open_own`/`_restrict`/`_write_log_row` (the functions removed/refactored out of `git_discard_guard.py`) are not referenced by any existing test outside this PR (`grep` across `tests/` was empty), so the extraction into `audit_jsonl` shared writer doesn't strand a test. `tests/test_hooks/test_git_discard_guard.py` still exists and exercises the guard's write path indirectly through the new shared writer; the new `audit_jsonl` module itself is exercised extensively (though indirectly) via `test_merge_override_log.py` rather than a dedicated `test_audit_jsonl.py` — not a rule violation since it's under `scripts/`, not `src/`, but worth knowing if a future change wants to unit-test `rewrite()`/`prune_by_age()` in isolation from the hook.
- Genesis conventions: no `$CLAUDE_PROJECT_DIR` misuse, no `IntervalTrigger` over 1h, conventional commit prefix present (`fix(hooks): ...`).
- Docstrings are thorough throughout (arguably exemplary) — no NOTE-tier gaps.

- No API keys, tokens, private IPs (10.x/192.168.x/100.64–127.x/fd00::/8), SSH keys, or connection strings in the added lines. The two `ghp_EXAMPLETOKENVALUE1234567890` strings (test file, lines asserting the log never captures a credential) are obvious placeholder values used as negative-test fixtures, not real secrets.
- No personal emails, username-bearing paths, or location-revealing identifiers in the diff.
- PR title/body: describes fixing a false "(logged)" claim on override sigils and adding an audit trail — this is a legitimate feature/bugfix description, not a leak disclosure. No mention of specific private infrastructure, no exploitable vulnerability detail, no incident language.




---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
**Coordination note — `scripts/hooks/git_push_guard.py` is under structural review.**

Measured on `origin/main` today: this file is **6,174 lines**, up from 3,100 on 2026-08-20 (+99% across 15 merged PRs in 12 days). A staged extraction — splitting the gate's fetch / decide / aggregate layers into separate modules — was scoped for exactly this reason and has not been built yet.

This PR adds **+405 lines** to that file, is currently `CONFLICTING` against main, and carries 10 unresolved review threads.

Suggested before continuing here:

1. Rebase onto current `main` and re-measure the diff — some of this may already be covered by the gate PRs merged since this branch was cut.
2. Consider whether the override-audit surface can live in its own module instead of growing the guard further. That keeps it out of the extraction's blast radius and makes it independently testable.
3. If it must land in-file, land it small — a +405-line addition to this file will conflict heavily with the extraction.

Not a hard block — flagging so the size cost is a deliberate choice rather than an accident.

WingedGuardian:
Closing as superseded by #1609.

#1609 is the redesign of this same problem — making the merge gate's "(logged)" promise true — and its body describes this PR directly: the single shared JSONL with an exclusive lock, tail-terminator check, age prune and size trim "ran five review rounds, and three of them introduced the defect the next one found."

That is the escalation cap doing its job. This PR carries 4 cross-model review rounds and 1 unresolved P1 + 9 P2s; the merge gate hard-blocks further rounds at 3. The cap means the design is suspect, not that the findings need more patching — and #1609 already took the prescribed route: one file per flush, so none of the shared-file mechanisms exist to get wrong.

Keeping both open means two PRs racing to fix one thing, with this one structurally unable to converge. The work is preserved in #1609's design; this branch is not deleted.

Reopen if #1609 turns out not to cover something here.

PR #1551, opened 2026-08-30
fix(hooks): the review reminder taught a chain it could not deliver

WingedGuardian:
The unreviewed-changes reminder fires on the **uncommitted** diff — it gates on
`has_code_changes()`, which is the working-tree hash, and never sees a PR. It was
nevertheless teaching the post-PR cross-model ordering, and then needed a paragraph
headed "PRE-COMMIT DIFFS ARE THE EXCEPTION" to explain why the thing it had just
mandated is impossible there.

That paragraph ended: *"the documented Claude adversarial fallback IS the correct
route"*. Read together with *"it is not optional"*, it reads as authorisation. It is
not — identifying which reviewer applies has never been permission to run it.

## What the text now says

Two things, named as different:

1. **Reviewing your own uncommitted work** — what the reminder is actually about.
   Unrestricted: dispatch whatever helps, as much as helps, no approval. Record the
   evidence and commit.
2. **The cross-model gate**, which lives on the PR after the push. Codex is that gate
   and is the only reviewer runnable without asking. Any other agent standing in for it
   requires explicit approval, every time, and approval never carries forward. A clean
   review from an agent that was not authorised does not satisfy the gate.

Stated that way it needs no per-install naming: the rule is "Codex, or ask", which
holds whatever alternative reviewers an install happens to have. The old wording
enumerated tiers instead, so an install with a tier between Codex and a Claude subagent
had no way to express it — and the text gave no hint anything was missing.

The two-surfaces note is kept and sharpened: the GitHub reviewer structurally cannot
see an uncommitted diff, so its being unavailable pre-commit is not licence to
substitute another agent at the gate. It means the gate has not been reached yet.

## Scope

`scripts/review_enforcement_prompt.py:49` is the only carrier of the ordering language
(`review_state.py`, `review_enforcement_commit.py` and `git_push_guard.py` mention
"adversarial review" only in unrelated contexts). The genesis-development skill already
states the approval requirement correctly, so the skill and the hook disagreed and the
hook was the wrong one.

Prose only — no control flow, no gating behaviour, no new failure mode. The emitted
text grows by 142 characters. `tests/test_scripts/test_review_scope.py`: 70 passed.

## Verification

Ran the hook against a dirty tree and read its real emitted output. Note an earlier
silent run was diagnostic rather than a pass: `get_current_diff_hash()` reads
`--cached`, so an unstaged change reports `clean` and the hook correctly stays quiet.

## Summary by CodeRabbit

* **Documentation**
[...]

PR #1555, opened 2026-08-31
fix(tests): wedge the loop the coalescing test claims to model

WingedGuardian:
`test_repeated_failures_coalesce_into_one_pending_record[dashboard]` fails in CI with
`got 2` while passing locally. It failed **identically on two unrelated PRs** (#1553 and
#1551), which is what showed it belongs to neither, and it currently blocks the `test`
job on every open PR.

## Root cause — the test, not the daemon

The contract under test is that a **wedged** loop cannot queue one failure record per
tick: `_record_failure` skips while a pending record is not done.

The test submits five failures from a worker thread — but `_drive` deliberately keeps
the loop turning while that worker runs. So the loop can *complete* record #1 between
two of the worker's iterations; `pending.done()` is then `True`, and repeat #2 submits
legitimately. The assertion was measuring thread-scheduling luck: green on an idle box,
red on a loaded runner.

Forcing the CI condition locally — 50 ms between submissions, i.e. a descheduled worker:

| shape | records |
|---|---|
| current (no wedge) | **5** |
| this fix | **1** |

CI's `got 2` is the milder form of the same 5.

## The fix

Hold record #1 open — on the loop, where it runs — until every repeat has been
submitted, then release. That is the only state in which coalescing has anything to do,
and it is what a wedge actually is.

The gate is released **from the worker** after the repeats. Releasing it from the test
coroutine would deadlock: the blocked loop cannot resume to do it. The wait is bounded,
matching the deadline convention `_drive` and `_run_tick_from_worker_thread` already
use, so a regression fails rather than hanging the suite.

Determinism here is argued from the mechanism, not from repetition: with the gate held,
`pending.done()` cannot become true before the repeats are submitted, so the skip branch
is reached by construction rather than by winning a race.

## No production change, deliberately

The same probe shows coalescing does **not** collapse records when the loop is free —
and that is correct. A healthy daemon ticking a minute apart completes each record
before the next, so five failed ticks genuinely *are* five failures. Coalescing exists
for the wedge. The test was asserting the invariant outside the case it applies to, so
the daemon needs no change and gets none.

## Verification

Whole file: 16 passed. The previously-flaky case run five consecutive times, no
variance. `ruff` clean.

Residual: this hardens one test against one race. Other tests in the file drive worker
threads against a live loop under `_drive` and could in principle share the shape; they
were not audited here, and none is currently failing.

## Summary by CodeRabbit

* **Tests**
  * Improved coverage for repeated job failures when the event loop is unresponsive.
  * Added safeguards to verify that failure records are consolidated correctly.
  * Strengthened timing and cleanup checks for background job processing.
  * Increased confidence in runtime health reporting under deadline and failure conditions.

### DIFF
--- a/tests/test_runtime/test_threaded_daemon_job_health.py
+++ b/tests/test_runtime/test_threaded_daemon_job_health.py
@@ -75,14 +75,20 @@ async def _run_tick_from_worker_thread(daemon, rt) -> None:
         except BaseException as exc:  # noqa: BLE001 - surfaced below
             errors.append(exc)
 
-    thread = threading.Thread(target=_worker, name="tick-worker")
+    # The deadline is DERIVED, not picked: this drives _tick, which waits on
+    # HeartbeatDaemon._TICK_TIMEOUT_S (30s). An outer bound below the inner one
+    # fails a legitimately-slow-but-working tick and blames the wrong thing, so it
+    # must exceed 30s. daemon=True + join-in-finally for the same reason as _drive.
+    thread = threading.Thread(target=_worker, name="tick-worker", daemon=True)
     thread.start()
-    deadline = asyncio.get_running_loop().time() + 10
-    while thread.is_alive():
-        if asyncio.get_running_loop().time() > deadline:
-            raise AssertionError("daemon tick did not complete within 10s")
-        await asyncio.sleep(0.01)
-    thread.join(timeout=5)
+    try:
+        deadline = asyncio.get_running_loop().time() + 45
+        while thread.is_alive():
+            if asyncio.get_running_loop().time() > deadline:
+                raise AssertionError("daemon tick did not complete within 45s")
+            await asyncio.sleep(0.01)
+    finally:
+        thread.join(timeout=5)
     if errors:
         raise errors[0]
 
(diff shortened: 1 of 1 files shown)


WingedGuardian:
## Scheduled review — PR #1555

**Scope:** single test file (`tests/test_runtime/test_threaded_daemon_job_health.py`), +16 lines, no production code touched.

**Summary:** Adds a `threading.Event` gate so the fake runtime's `record_job_failure` callback (running on the loop) blocks until the worker thread has submitted all 5 repeated failures, then releases. This closes the race where `pending.done()` could flip true mid-submission — the CI `got 2` flake — by making the coalescing assertion hold by construction rather than by thread-scheduling luck. `threading` is already imported in the file. The 10s bound mirrors the file's existing bounded-wait convention (`_drive`: 15s, `_run_tick_from_worker_thread`: 10s), so a regression fails fast instead of hanging the suite. No deadlock: the gate is released from the worker thread, never from the coroutine it blocks. No production code changed, deliberately — the daemon's coalescing behavior itself isn't in question here.

---

WingedGuardian:
@codex review

PR #1496, opened 2026-08-26
feat(ego): dispatch follow-through — force outcome review via ego_intentions

WingedGuardian:
## What

B2b of the activation-loop program (dispatch follow-through). A dispatched ego proposal that finished — success **or** failure — produced only an observation and a memory that nothing was forced to read, so dispatch outcomes rotted unreviewed (no step-2, no retry decision, no calibration). Now every terminal dispatch forces the **owning ego** to review the outcome on its next cycle.

## Mechanism

`_ego_dispatch_on_end` (the single dispatch-completion sink) calls a new unit-testable helper, `ego/dispatch_followthrough.py::record_dispatch_followthrough`, which creates a **system-origin `ego_intentions` row** for the proposal's `ego_source` (genesis fallback on NULL). The existing mandatory-review block then makes the next cycle judge it: step-2, retry, or close (`withdraw`).

- **Unconditional** on every terminal dispatch; `max_cycles=3` so unactioned rows self-expire; priority `high` when the dispatch failed.
- **Dedup per proposal** (active rows only) via `proposal_id` set at creation — a re-dispatched proposal never piles rows.
- **New `origin` column** (`'ego'`|`'system'`) on `ego_intentions`, added across all three schema build paths (fresh CREATE, `_migrate_add_columns`, numbered migration 0086). System rows **bypass** `MAX_ACTIVE_PER_SOURCE` — the cap exists to stop the LLM hoarding slots, not to drop mechanical follow-through; the LLM cap now counts `origin='ego'` rows only.
- `intentions_context` renders system rows with a `[follow-through]` tag + review guidance; the remaining-slots line counts LLM rows only.

## Tests / verification

TDD verify-RED (13 new tests failed pre-implementation) → GREEN; **104 passed** including the schema-parity and migration-runner guards; ruff clean. E2E against a live-DB copy: migration 0086 backfills 69 existing rows to `'ego'`; a real executed proposal routes to its owning ego; the dedup second call returns None; the context section renders the tag with slots unaffected. Cap-bypass is live-relevant: the user ego already carries 4 active LLM intentions on the current board.

## Review

[...]

PR #1558, opened 2026-08-31
fix(db): renumber the colliding ego_intentions migration to 0088

WingedGuardian:
## Summary

Two migrations claimed prefix `0086` on main: `0086_seed_timezone_config` (#1494, already applied on existing installs) and `0086_ego_intentions_origin` (#1496). The filenames differ, so git merged both cleanly and nothing conflicted — the broken state first existed only *after* the second merge.

This renames the ego migration to `0088`. Pure rename, zero content delta (`R100`).

## The consequence is not a failed migration

The runner's pre-flight raises on the duplicate **before** it reads what is applied, so nothing is skipped and nothing diverges — but the failure is worse than a loud abort:

- `runner.py` raises `RuntimeError("Duplicate migration prefix '0086'…")` at the top of `run_pending()`, before `get_pending()`
- `init/db.py` closes the DB, sets `_db = None`, re-raises
- `_core.py::_run_init_step_async` **swallows** it, recording `db: failed:` in the bootstrap manifest
- `bootstrap()` returns early on `_db is None` — tool registry, observability, providers, awareness, router, memory and the schedulers never initialise
- `__main__.py` awaits `serve()` **regardless**

So a deployed install comes up and reports **`active`** with no database and almost nothing wired. The tell is `db: failed:` in the bootstrap manifest — `is-active` is actively misleading here.

It also fails **49 tests on every open PR** (measured on an unrelated PR's run: 49 failed / 21281 passed, every failure the same RuntimeError), because anything that builds a schema hits the same pre-flight. No PR in the repo can merge until this lands.

## Why `0088`, not `0087`

`0087` would have created the next collision. Prefix claims across main and every open PR:

| prefix | claimants |
|---|---|
| `0086` | `seed_timezone` (merged) · `ego_intentions` (merged) · #1490 · #1477 |
| `0087` | #1541 · #1454 |

`0088` is the first prefix free everywhere. `data_migrations/` is clean.

## Why the ego migration moves, not the timezone one

`schema_migrations` already records `0086 = 0086_seed_timezone_config` as applied. Renumbering *that* would make an applied migration look pending and re-run it. The ego migration has never run anywhere it collides, and its body guards on `sqlite_master` then `PRAGMA table_info`, so it re-runs safely under a new id. The module carries no internal id string — the filename is the only id carrier — and nothing references it by name.

## Verification

- **Verify-RED:** `test_no_duplicate_migration_prefixes_in_tree` fails on unmodified main with `assert not {'0086': 2}`; passes after.
- **Green:** 107 tests across the DB migration/schema/runner/discovery suites; 44 across `tests/ego/test_dispatch_followthrough.py` and `test_intentions.py` (the feature this migration serves); ruff and the external-io / subsystem-map / frozen-clock guards.
- **E2E against a copy of the live 429 MB database, running the real `MigrationRunner`:** 87 available with unique ids → applies `0088` → `PRAGMA table_info(ego_intentions)` shows `origin`. Second run: no-op, idempotent.
- **Control** (so the E2E isn't inert): recreating the duplicate makes the identical probe raise `Duplicate migration prefix '0086'`. The check discriminates.

## The class defect — third occurrence

#1279 (`d0009`, 2026-08-02) **added** the CI guard; #1349 (`0079`→`0081`) happened *after* it; now this. Detection-after-merge cannot prevent it: each PR is green against the main it was tested on, filenames differ so git sees no conflict, and the collision only exists after the second merge.

Measured here: #1496's `test` went green `2026-08-27T04:10:12Z`; #1494 landed the collision **11h21m later** at `15:31:21Z`; #1496 merged 4 days after that on the stale green, and the merge gate accepted it — it never asks *which base* a green was computed on. A clean natural experiment for the same point: an unrelated open PR's own file passed while 49 tests failed purely from main's collision, so a PR's red/green currently says as much about its base as about its diff.

Follow-up, not in this PR: force a fresh CI run against current main before every merge, then make the gate treat a run whose base is behind the default branch as `ci: unknown` (already blocking) — with distinguishing text, since an undifferentiated `unknown` reads as the transient "still computing" case and gets retried instead of fixed.


WingedGuardian:
**Leaks / PII scan — CLEAN**

Scope: the added lines of every commit in the range (one commit, `99064445`), plus this PR body.
Method: the CI gate's own scanner (`scripts/ci/private_pattern_scan.py`) against the same
fingerprint set, run locally at this exact head.

- added lines scanned: 48 (`--no-renames`, so the renamed file's full body is treated as added)
- PR body scanned: 4,357 bytes
- **result: CLEAN** (exit 0)
- **positive control**: re-running the identical scan with a fingerprint seeded into the input
  returns `match count: 1`, exit 1 — the instrument discriminates, so the CLEAN is a real pass
  and not an inert run.

Content note: the diff is a pure file rename (`R100`, zero content delta), so there is no new
prose, no new identifiers, and no new configuration surface to leak through.

WingedGuardian:
@codex review

PR #1559, opened 2026-08-31
chore(cc): move the pin to 2.1.246 and let the host follow

WingedGuardian:
Moves the public Claude Code pin from **2.1.218** to **2.1.246**, and with it the host.

## Why now

The container has run 2.1.246 since 2026-08-25 under the inherited-`CC_VERSION` soak lever, which deliberately does not touch `scripts/lib/cc_version.sh`. `cc_align_host.sh` aims the host at the **public** pin, so it correctly stayed on 2.1.218 throughout — that split *is* the documented rollback path during a soak.

It is also why the host has been a release behind for six days. Only moving the pin ends it, and the host is where the Guardian's `claude -p` recovery brain runs.

## Gates

Both §Updating gates were discharged before this PR, not by it.

- **Changelog** — read in full over `(2.1.218, 2.1.246]`: 25 releases, since 2.1.230/242/244 were never published.
- **Soak** — six continuous days on the container. The running-binary sweep at soak end showed every live CC process on the on-disk binary and none stale, the oldest dating from the align itself, which is what makes the soak continuous rather than merely elapsed.

CC-Gate-Changelog: read (2.1.218, 2.1.246] in full from CHANGELOG.md, 2026-08-27
CC-Gate-Soak: 2.1.246 on container 2026-08-25..2026-08-31, running-binary sweep showed every live CC process on the on-disk binary and none stale, signed off by the repository owner 2026-08-31

## Node floor — checked, not assumed

`NODE_MAJOR` stays **22**. 2.1.246 declares `engines.node >=22.0.0` against the npm registry, as does 2.1.251, so the next cycle will not move it either. `check_cc_node_lockstep.py` was run locally before pushing and passes. A stale `NODE_MAJOR` beside a raised floor is what once left a host on Node 18 with the recovery brain offline, so this is the check that matters most here.

## Target choice

Deliberately 2.1.246 rather than npm's current `latest` 2.1.251: 2.1.246 is the version that actually soaked, and §Updating is explicit that re-targeting restarts the procedure and resets the soak clock. The 2.1.247–2.1.251 delta rolls into the next cycle.

## What the range contains that matters here

- **Delegation** — from @246 a subagent that hits `maxTurns` returns *partial output without failing*, so a silent early stop is indistinguishable from a genuinely short report. This is why the changelog-read procedure now requires per-release acknowledgement rather than trusting a delegate's summary.
- **Guardian** — @246 fixes a `--strict-mcp-config` startup hang, which lands directly on Guardian Diagnosis.
- **Worktrees** — @245→@246 fixes a background-retention sweep that reaped user-created `.claude/worktrees/`; this repo keeps long-lived worktrees per branch.
- **Headless** — `-p --continue`/`--resume` now resume in plan mode.

## Also fixed

Two assertions the bump itself would have made false: the `cc-update` skill's parenthetical claiming "the current pin is 2.1.218", and the manual host-move example naming the old version.

## What this does NOT do

Merging changes what both machines *aim at*; it does not move either. `scripts/update.sh` does that. The host must then be verified through a **fresh gateway `version` op** — never `~/.genesis/host_gateway_state.json`, which is written from the pre-alignment probe and reports the old version until the nightly timer runs.

Note the host path is not symmetric: the gateway's `update-cc` runs `npm install -g` *before* it verifies, so an install that succeeds followed by a failed version check leaves the old package already replaced with no rollback. Rollback is one op: `update-cc 2.1.218`.

## Known false positive, tracked separately

`check_cc_running_versions.sh` also reports an `undetermined` process on installs with the `browser` extra. `cmdline_runs_cc` matches any node process whose first non-flag argument is basename `cli.js` — which catches Playwright's bundled driver as well as Claude Code's launcher — and a node interpreter's inode says nothing about a CC revision. It is not evidence about any CC session and does not affect the soak result above. Filed separately rather than fixed here, because the tool is intentionally fail-closed and narrowing the matcher risks missing a genuinely stale CC process.

Ledger: c60f18a0342a48ef94b9942e56a1d757

## Summary by CodeRabbit

* **Chores**
  * Updated the default Claude Code version to 2.1.246.
  * Preserved support for overriding the default version through the environment.

* **Documentation**
  * Updated compatibility guidance, evaluated release details, deployment status, rollback instructions, and upgrade examples for version 2.1.246.
  * Documented improved partial-output behavior when delegated work reaches its turn limit.


WingedGuardian:
@codex review

WingedGuardian:
## Structural review: CLEAN

Three files, +7/−5. Every added line was read individually — at this size a sample would be a choice, not a constraint.

- `scripts/lib/cc_version.sh` — one line, a version string in an existing `${CC_VERSION:-…}` default. No new code, no control flow, no imports. `NODE_MAJOR` deliberately untouched and verified correct for the new pin rather than assumed: `check_cc_node_lockstep.py` was run against the live npm registry and reports `Claude Code 2.1.246 needs Node >= 22 (engines.node='>=22.0.0'); NODE_MAJOR=22`. Independently confirmed at 2.1.251 as well, so the next cycle does not move the floor either.
- `docs/reference/cc-compatibility.md` — prose only: the §Current CC Version "Pinned:" statement, one §Version History row, and a stale `update-cc 2.1.218` example corrected to the new pin. New table row verified to carry the same 4 columns as its neighbours.
- `.claude/skills/cc-update/SKILL.md` — prose only: a parenthetical that asserted "the current pin is 2.1.218", which this commit would have made false.

No pattern-violation ERRORs: no `nohup`, no `/tmp/` usage, no `git add .`/`-A`, no force-push or `reset --hard`, no worktree editable install, no bare `except:`, no mutable default args, no raw SQL, no `IntervalTrigger`. None are reachable — two of the three files are markdown and the third changes a single string literal.

Completeness check on the bump rather than the diff: enumerated every remaining `2.1.218` occurrence in the tree and confirmed none is still a live claim about what the pin *is*. What remains is the §Gate-receipts example (which now happens to match the real receipt), a historical changelog-scale note, parser-example docstrings in `cc_pin_parse.py` / `check_cc_pin_receipts.py`, CHANGELOG history, and a test fixture docstring describing a scenario. The two that *were* live claims are fixed, as is the skill parenthetical.

- No API keys, tokens, SSH private keys, or connection strings in the added lines.
- No IP addresses of any class, no hostnames, no tailnet addresses, no MAC/serials.
- No personal emails, real names, or username-bearing paths. The `<host_user>@<host_ip>` placeholders and the `/usr/bin/claude` → `/usr/lib/node_modules/@anthropic-ai/claude-code` resolution are pre-existing generic text, unchanged by this PR.
- **One issue found and fixed before this review, recorded because it is the kind that slips through:** an earlier draft of the §Version History row stated the running-binary sweep result as a literal per-box process count. That is install-specific operational detail on a public surface — the same class as a "this install runs ~N worktrees" line a previous privacy pass had to strip from `SKILL.md`. Rewritten install-generic ("every live CC process on the on-disk binary and none stale"). Nothing equivalent remains: a re-scan of the added lines for per-box counts, `this install/box runs …` phrasing, and identifier patterns returns nothing.
- Dates (2026-08-25, 2026-08-27, 2026-08-31) describe the soak window and are required by the receipts; they reveal nothing beyond that this repo did an upgrade, which is the point of the row.
- PR title/body: describes a version pin bump and its gate receipts. No private infrastructure named, no exploitable detail, no incident language. The receipts cite `CHANGELOG.md` as a bare filename rather than the local scratch path, which is the documented requirement precisely because PR bodies are public.

Automated corroboration: the `leak-detector` CI job passes at this head with the repo's own secret pattern set — which is the check this review cannot reproduce locally, since that pattern layer is opt-in at bootstrap and unprovisioned here (running it locally returns exit 3, fail-loud, refusing a false clean).

PR #1562, opened 2026-08-31
fix(observability): stop surplus's loop-END heartbeat false-firing 'overdue'

WingedGuardian:
## Problem
Surplus emits its `Subsystem.SURPLUS` event heartbeat only at the END of `_dispatch_loop`, so a healthy single `dispatch_once` (15-30 min) legitimately gaps it. The `HEARTBEAT_EXPECTED["surplus"]` overdue threshold was 600s — tighter than that gap — so a busy-but-healthy surplus false-fired "surplus heartbeat overdue" in the morning report and the `subsystem_heartbeats` display (crying wolf).

## Fix
- Loosen surplus overdue **600s → 3h (10800s)** = the surplus dashboard tile's job_health stall bound (`stall_threshold_minutes(5)`), so the event-HB display and the tile agree. Genuine total cessation is still caught far sooner (900s) by the zombie-scheduler watchdog, which reads job_health (refreshed every dispatch), not this event-HB. ~2× margin over the ~90 min worst-case healthy loop.
- Correct two misleading `_dispatch_loop` comments: the before-each-dispatch `record_job_success` refresh protects a *sequence* of moderate dispatches but does NOT refresh during one long `dispatch_once`, so `last_run` can still gap past the 900s watchdog with no `heavy_workload` deferral (dream-only). That latent watchdog-restart gap is tracked as a follow-up (no such restart observed to date).

## Tests
RED-first: a 25-min surplus read "overdue" at 600s, "alive" at 3h. +2 tests (long-dispatch-not-false-overdue; overdue-past-3h); 2 existing seeds bumped above the new threshold. 74/74 in the file green.

## Review
genesis-architect adversarial pass — CLEAN (Ready: Yes). Blast radius contained to `compute_heartbeat_staleness` (only reader of surplus's overdue value); value tile-aligned; tests non-vacuous.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced false surplus scheduler staleness alerts by extending the heartbeat overdue threshold from 10 minutes to 3 hours.
  * Preserved earlier detection of genuinely stalled schedulers through the existing watchdog checks.

* **Documentation**
  * Clarified heartbeat behavior during long-running dispatches and documented the updated monitoring thresholds.

* **Tests**
  * Added coverage confirming moderate dispatch gaps remain healthy while multi-hour gaps are correctly reported as overdue.

PR #1462, opened 2026-08-25
fix(observability): surface subsystem total-cessation (alert + ego tile)

WingedGuardian:
## Make a dead subsystem scheduler stop reading "healthy"

Third of the observability-honesty batch (B #1444 → D #1459 → **C**). Adds the total-cessation
signal that #1422's ego intent-lag deliberately cannot catch (its docstring names this monitor as
the fix): when a background subsystem's scheduler/loop stops firing *entirely*, its heartbeat pulse
goes silent — previously nothing turned that into a signal, so the Ego tile + rollup could read green
while the egos were dead.

### What it does
- **Alert set = ego (CRITICAL) + inbox + dashboard (WARNING).** A `subsystem_stale:<name>` alert
  fires when the pulse goes overdue past its per-subsystem threshold; the ego tile + worst-of-7 rollup
  flip to error. **surplus** is excluded (PR-B #1444's tile already flips red on a wedged/dead surplus
  via its finer job_health signal) and **outreach** is excluded (its heartbeat is config-gated — see
  follow-up).
- **Pause-aware, idle-safe, boot-safe.** The pause-gated subsystems ({surplus, inbox}, whose pulse
  stops behind their loop's `if paused: return`) are downgraded to `paused` while Genesis is globally
  paused — reaching every consumer (alert, tile, morning report) via the shared
  `compute_heartbeat_staleness` helper. ego/dashboard pulse through pause by construction.
- **Never a silent green.** Freshest-of-durable-row-and-ring (excluding materially-future rows);
  future-skew/corrupt timestamp → `unknown`; an unreadable pulse surfaces loudly as WARNING
  `subsystem_heartbeat_unknown:<name>` (honest "can't confirm", not "died"); a read failure re-emits
  any open alert so a transient blip can't auto-resolve a real death; ego `no_heartbeat` past a
  boot-grace window → error, not a green tile.
- Thresholds loosened for near-zero false-reds (inbox 1h→2h = 4× the 30-min check; dashboard
  4min→10min = 10× the 60s pulse). Does **not** page (verified absent from the escalation whitelist).

### Scope / safety
Health/observability/dashboard layer only — **zero subsystem-loop edits** (per the "guard the consumer"
decision). No autonomy-gate / secrets / SQL / external-input surface.

### Review
2× genesis-architect (DONE_WITH_CONCERNS, no blocker) + a fresh-context exhaustive whole-class audit;
every finding fixed or documented-accepted. 179 targeted tests green, ruff clean. RED-first throughout.

[...]

PR #1566, opened 2026-09-01
docs(architecture): document subsystem-cessation alerts + the autonomy watchdog

WingedGuardian:
CURRENT.md was missing two shipped observability mechanisms — filling the gaps found during the surplus-heartbeat red-team (#1562).

- **§9** — new bullet on the subsystem-cessation alert family: `subsystem_stale` / `subsystem_never_started` / `subsystem_heartbeat_unknown` (#1462/#1504), the pause-gated + conditional-pulse (outreach) handling, and why **surplus is excluded** (PR-B tile + the zombie-scheduler watchdog cover it; its loop-END event-HB is load-fragile → 3h overdue).
- **§8** — new bullet on the **autonomy zombie-scheduler watchdog** (`autonomy/watchdog.py`, `genesis-watchdog.timer` 300s → `status.json` 900s → restart → flap-alert), previously undocumented, cross-referencing why surplus needs no `subsystem_stale` alert.

Docs-only. No `verified:` stamp bumps — the additions document specific mechanisms, not a re-verification of the whole §8/§9 sections. `check_subsystem_map.py` CLEAN.

## Summary by CodeRabbit

* **Documentation**
  * Added architecture documentation for an external watchdog that detects and restarts stalled runtime or scheduler processes.
  * Documented safeguards that defer recovery during heavy workloads, startup stabilization, or network outages.
  * Added guidance for escalating repeated failures and alerting on stopped, missing, or unknown subsystem heartbeats, including pause handling and exclusions.
  * Clarified that heartbeat status uses the freshest valid available signal.

### DIFF
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -971,9 +971,10 @@ verified: ca875c4b 2026-07-24
 - **Subsystem-cessation alerts (the honest-liveness family, `_compute_alerts`)**:
   three durable `subsystem_stale`-family alerts so a dead scheduler cannot read
   healthy. `subsystem_stale:<name>` (ego CRITICAL; inbox/dashboard/outreach WARNING)
-  fires when a subsystem's durable `events`-table heartbeat goes overdue past its
-  per-subsystem `HEARTBEAT_EXPECTED` threshold (via
-  `mcp/health/manifest.py::compute_heartbeat_staleness`); `subsystem_never_started:<name>`
+  fires when a subsystem's heartbeat — the freshest valid pulse across the durable
+  `events` table and the in-memory event-bus ring — goes overdue past its per-subsystem
+  `HEARTBEAT_EXPECTED` threshold (via `mcp/health/manifest.py::compute_heartbeat_staleness`);
+  `subsystem_never_started:<name>`
   tells a FAILED / never-pulsed start (cross-referenced against the persisted bootstrap
   manifest, past a boot grace) from a fresh install; `subsystem_heartbeat_unknown` covers
   a corrupt/unreadable pulse. Pause-gated subsystems (`surplus`, `inbox`) downgrade
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -830,6 +830,16 @@ verified: 159698d4 2026-07-16
   monitors the host Guardian every awareness tick, incl. git-SHA code-drift
   detection). Config `~/.genesis/guardian_remote.yaml`; missing → silently
   disabled.
+- **autonomy zombie-scheduler watchdog** (`autonomy/watchdog.py`, run out-of-process
+  by `genesis-watchdog.timer` every 300s via `watchdog_runner.py`; distinct from the
+  container `watchdog.py` above): reads `~/.genesis/status.json` (written by the runtime's
+  `status_writer_loop`) and RESTARTS the bridge/runtime on status-file staleness, or a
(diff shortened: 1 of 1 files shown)


WingedGuardian:
@codex review

WingedGuardian:
**Automated first-pass review** (docs-only PR — PII/Secrets/Wording scan only, structural review skipped per process)

Scope: `docs/architecture/CURRENT.md` (+27/-0, single file).

- **Code diff (added lines):** CLEAN — no API keys/tokens/secrets, no private IPv4/IPv6 ranges, no SSH keys, no connection strings. All references are generic paths (`~/.genesis/status.json`, `autonomy/watchdog.py`, `mcp/health/manifest.py`, `genesis-watchdog.timer`, `watchdog_runner.py`) and code symbols — no usernames, hostnames, emails, or hardware identifiers.
- **PR title/body:** CLEAN — no PII/secrets-leak language, no security-incident references, no private infrastructure named, no exploitable vulnerability detail.

---

WingedGuardian:
The `check_subsystem_map.py` CLEAN cited here is real for coverage but says nothing about freshness — the guard's staleness half is currently inert, and its diagnostic names a cause that is not the cause. Flagging it because this PR reasonably reads that green as validation, and it isn't blocking (docs-only).

**Measured**

The guard prints both of these, in this order:

```
subsystem-map guard: staleness check SKIPPED (shallow or incomplete git history — CI needs fetch-depth: 0)
Subsystem-map guard: CLEAN (14 entries cover 60 top-level modules)
```

The repo is not shallow — 1,729 commits, no `.git/shallow`. The real cause is at `scripts/check_subsystem_map.py:178-179`: `check_staleness` does `return None` on the first `verified:` sha that `git cat-file -e` cannot resolve, and that early return sits inside the per-entry loop. So **one dead stamp disables the freshness check for all fourteen entries**, and the caller reports it as a history problem.

Exactly two of the fourteen stamps are unresolvable — `0e65071c` and `fbcf8ee4`. The other twelve resolve fine.

**`fetch-depth: 0` does not rescue this.** Those two shas are *pre-squash* — they were never in the public history at any depth — so the staleness half is dead in CI as well as locally, not just in shallow clones. Confirmed against a recent `main` CI run of the drift-guard step, which prints the SKIPPED line and then CLEAN.

**Why it will recur**

Both dead stamps named a **feature-branch HEAD**. Squash-merge discards that commit, so the sha becomes permanently unresolvable the moment the PR lands. A stamp has to name a commit that survives the merge — the default-branch tip it was verified against. The one-command test:

```
git merge-base --is-ancestor <sha> origin/main
```

Worth knowing that this failure mode is actively recommended by careful review: an adversarial audit on another branch flagged a stamp for "naming a commit that is not an ancestor of this branch" and advised using the branch's own head instead. That reasoning is correct about review time and wrong about merge time, and following it would have created a third orphan.

**No action needed on this PR**

Both halves are already in flight elsewhere, so this is FYI rather than a request:

- #1556 changes the early return to a per-entry warning + continue, keeps the genuine global skip for a real shallow clone, and drops the misleading fetch-depth text. Measured on this repo: 0 warnings before, 4 after.
- #1560 adds the stamping rule to `CURRENT.md`'s own header, so the next person bumping a stamp reads it where they're already looking.

The only thing I'd change here is not treating that CLEAN as evidence about staleness — until #1556 lands, **every `verified:` stamp in the file is unpoliced**, which also means this PR's decision not to bump any stamps is the right call for a reason stronger than the one given.

WingedGuardian:
@codex review

PR #1569, opened 2026-09-01
fix(ci): subsystem-map staleness — one dead stamp no longer skips the whole check

WingedGuardian:
## Problem
`check_subsystem_map.py::check_staleness` did `return None` for the **whole file** on the first unresolvable `verified:` stamp — inside the per-entry loop. So a single dead stamp silently disabled the freshness check for **all 14 entries**, and the caller misreported it as a shallow clone ("CI needs fetch-depth: 0"). The dead stamps are **pre-squash feature-branch HEADs** that squash-merge discards, so they're permanently unresolvable in CI.

## Fix
- **Per-entry warn + continue** instead of `return None` — one dead stamp no longer disables policing for the other 13.
- **Detect by ANCESTRY, not existence:** `merge-base --is-ancestor <sha> HEAD`, not `cat-file -e`. A pre-squash orphan that still *exists locally* passes `cat-file` and then miscounts `<sha>..HEAD`, so local and CI disagreed. Ancestry is the invariant a stamp actually needs, and it's consistent local↔CI.
- **Broaden the SKIP diagnostic** — it now only fires on a genuine shallow clone or a git failure, not on a dead stamp.

Unchanged: the genuine shallow-repo skip and the fail-fast-on-`rev-list`-failure path. Warnings emit as `::warning::` (**exit 0 — CI stays green**).

## Real-data acceptance
The fixed guard surfaces **3 pre-squash orphans** (`0e65071c`, `ca875c4b`, `fbcf8ee4`) + **3 genuinely-stale** entries that were hidden. (The scheduled-review comment that surfaced this named only 2 orphans; the ancestry check found the 3rd.) Re-stamping those 6 is tracked as a follow-up (data-repair, per-entry re-verification).

## Tests
RED-first: the new mixed test + rewritten all-unresolvable test fail (`assert None is not None`) against the unfixed guard, pass after. 21/21 in the file green.

## Review
genesis-architect adversarial pass — CLEAN (Ready: Yes). All 3 NOTEs addressed: NOTE 2 (ancestry switch) applied, NOTE 1 (diagnostic wording) applied, NOTE 3 (follow-up filed).

### DIFF
--- a/scripts/check_subsystem_map.py
+++ b/scripts/check_subsystem_map.py
@@ -175,8 +175,21 @@ def check_staleness(entries: list[Entry], threshold: int) -> list[str] | None:
         return None
     warnings: list[str] = []
     for entry in entries:
-        if _git(["cat-file", "-e", f"{entry.verified_sha}^{{commit}}"]) is None:
-            return None  # stamped sha not in local history — can't count honestly
+        if _git(["merge-base", "--is-ancestor", entry.verified_sha, "HEAD"]) is None:
+            # Stamp is not an ancestor of HEAD (git merge-base --is-ancestor exits
+            # non-zero) — almost always a pre-squash feature-branch HEAD that squash-merge
+            # discarded. This is the ancestry check the stamp actually needs: a bare
+            # `cat-file -e` would PASS an orphan that still exists locally and then miscount
+            # `<sha>..HEAD`, so local and CI would disagree. Warn and keep policing the
+            # OTHER entries — a single dead stamp must NOT return None here, because that
+            # returns None for the WHOLE file, silently drops every entry's freshness
+            # check, and the caller then misreports it as a shallow clone.
+            warnings.append(
+                f"entry '{entry.name}': verified stamp {entry.verified_sha} "
+                f"({entry.verified_date}) is unresolvable (not an ancestor of HEAD) — "
+                f"likely a pre-squash orphan; re-verify and re-stamp to a default-branch commit"
+            )
+            continue
         paths = [f"src/genesis/{m}" for m in entry.modules]
         count_out = _git(["rev-list", "--count", f"{entry.verified_sha}..HEAD", "--", *paths])
         if count_out is None:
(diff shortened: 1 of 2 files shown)


WingedGuardian:
@codex review

WingedGuardian:
## Automated review — structural + PII/secrets/wording

**Scope:** `scripts/check_subsystem_map.py` (+18/-4), `tests/test_scripts/test_check_subsystem_map.py` (+42/-5). Read the full `check_subsystem_map.py` for context given the logic change.

### Domain 1 — Structural

No pattern violations found (nohup, raw SQL, bare `except:`, mutable defaults, `git add -A`/force-push, `/tmp/` usage, `os.killpg`, direct `task_states` inserts — none present). No import-hygiene issues (no new/reordered imports in the diff). No missing-coverage issue: both changed files are existing files, and the test file is updated in the same PR to match the new `check_staleness` behavior. Commit/PR title carries a conventional prefix (`fix(ci):`).

**NOTE (docs, non-blocking):** `scripts/check_subsystem_map.py:14-16` — the module docstring still describes the *pre-fix* behavior: "on a shallow clone (or when the stamped sha is absent locally) it degrades gracefully — it says so and skips, it does not guess." That's the exact behavior this PR removes for the unresolvable-stamp case — the guard no longer skips there, it now emits a per-entry `::warning::` and keeps checking the rest. Worth a follow-up line in the module docstring so it doesn't mislead the next reader about when the check actually skips (shallow clone / git failure only, per the updated `check_staleness` docstring at line 178, which is correctly updated) vs. warns-and-continues (unresolvable stamp).

Everything else — function-level docstrings (`check_staleness`, `_git`), the updated SKIPPED message in `main()`, and the new/renamed tests — is consistent with the new ancestry-based behavior.

PII/Secrets/Wording: clean (soft warnings only — actually: none found). No API keys/tokens/credentials, no private IPs (v4 or v6 ULA), no SSH keys, no connection strings. The commit SHAs referenced in the diff and PR body (`0e65071c`, `ca875c4b`, `fbcf8ee4`, plus synthetic test fixtures `aaaaaaaa`/`bbbbbbbb`) are ordinary git object identifiers, not secrets or credentials. No personal emails, username-bearing paths, or location-revealing identifiers introduced.

PR title/body scan: plain CI-tooling language throughout ("dead stamp", "pre-squash orphans", "shallow clone") — no mention of a security incident, data exposure, or private-infrastructure specifics, and nothing exploitable is advertised.

VERDICT: PASS



---

PR #1571, opened 2026-09-01
fix(hooks): declare merge-to-main-override and full-suite-ok as sigils

WingedGuardian:
## What this changes

`merge-to-main-override` and `full-suite-ok` were passed to `has_trailing_override` from the day they shipped and never listed in `_KNOWN_SIGILS`, so that tuple's "kept in sync with the sigils actually passed across the guard hooks" claim was false.

The consequence is silent and asymmetric, which is why it survived: a sigil **queried for itself** matches before the membership test is reached, so nothing looked broken. But an undeclared token written **first** reads as prose, ends the leading run, and disables every sigil after it.

That broke a **documented** procedure, not a hypothetical one. The development skill instructs appending several sigils to one trailing comment when several waivers are genuinely intended — and any such comment led by one of these two silently lost the rest. The pre-fix behaviour failed **closed**: a waiver the operator typed was not honoured.

## This is a gate-loosening change

Stated plainly because it is riding as its own PR for exactly this reason — it was split out of a larger audit-logging PR so it gets its own merge decision.

Measured by exhaustive ordered-pair enumeration over the ten-token set, **eighteen** combinations flip from rejected to accepted, nine led by each new token. An earlier estimate of five was an undercount and is retracted.

Severity does not follow that count. Every one of the eighteen is a combination of two sigils that were **each already independently valid alone** — the change can never make one sigil satisfy a query for another. And nothing in the repo composes a multi-sigil comment programmatically (grepped): an operator types both, literally, or neither.

The sharpest single case, so it is not buried: `git clean -fdx  # full-suite-ok discard-override` now resolves `discard-override` where it previously did not.

## The test derives the set rather than restating it

Both previous divergences were invisible to the obvious check, so the test walks the consumers with `ast`:

- Population is `scripts/`, not `scripts/hooks/` — one consumer lives outside that directory and holds the only queries for two of the declared sigils.
- Two guards pass their sigil as a **module constant**, which a literal scan cannot see (this is how `full-suite-ok` stayed undeclared while a regex test passed green).
- One call site relies on the **default**, resolved from the live signature so the test and the function cannot drift.

The assertion runs in **both directions**. An adversarial pass showed the one-directional version green while an unqueried token sat in the tuple — over-declaration widens acceptance for every other sigil by exactly the mechanism these two additions repair, so it is the same defect reversed rather than a separate concern.

## Verification

- **Verify-RED ×3**: initial 2-fail/1-pass; the over-declaration mutation that previously passed 146/146 now fails; a mutated default sigil now fails and is attributed to the right file. Both mutations restored byte-identical, zero markers surviving.
- **306 passed** across the parser and every consumer suite the walk names.
- `ruff` clean. Private-data scan over the full diff: clean.
- **E2E acceptance replay** on the real hook: with `pytest tests/ # merge-to-main-override full-suite-ok`, the unfixed copy emits the BLOCKED message and the fixed copy allows it — while the no-sigil control still blocks, so the waiver is honoured without disabling the guard.

Adversarial security review found no CRITICAL. It did correct two things in the original diff — a comment that overstated the consumer count, and the flip-count estimate — both fixed here.

## Summary by CodeRabbit

* **New Features**
  * Added support for additional acknowledgment tokens used by merge and full-suite checks.

* **Bug Fixes**
  * Improved recognition of override tokens regardless of their leading order.
  * Ensured ordinary prose correctly ends token scanning.

* **Tests**
  * Added regression coverage to validate supported tokens and their usage across guard checks.

### DIFF
--- a/scripts/hooks/shell_parse.py
+++ b/scripts/hooks/shell_parse.py
@@ -535,6 +535,21 @@ _KNOWN_SIGILS = (
     "stale-review-override",
     "scheduled-review-override",
     "discard-override",
+    # Both were passed to has_trailing_override from the day they shipped but never
+    # listed here, so the "kept in sync" claim above was false. The consequence is
+    # silent and asymmetric: the sigil queried FOR ITSELF still matches, so nothing
+    # looked broken, while an unlisted token written FIRST reads as prose and ends
+    # the leading run — disabling every sigil after it. NOTE this widens ACCEPTANCE
+    # as well as detection, in the fail-open direction, and the reach is wider than
+    # the merge gate; the PR that added them enumerates the combinations that flip.
+    # A test derives this set from the consumers themselves (an ast walk over
+    # scripts/, not just scripts/hooks/ — review_enforcement_commit.py lives
+    # outside that directory and is the only place two of the declared sigils are
+    # queried, and two guards pass their sigil as a module constant a literal scan
+    # cannot see), and asserts the set matches in BOTH directions, so neither a
+    # missing declaration nor an unwarranted one can ship unnoticed.
+    "merge-to-main-override",  # git_push_guard: local `git merge` onto main/master
+    "full-suite-ok",  # full_suite_guard: run the whole pytest suite locally
 )
 
 
--- a/tests/test_hooks/test_shell_parse.py
+++ b/tests/test_hooks/test_shell_parse.py
@@ -7,12 +7,15 @@ these lock in the wrapper/substitution/nesting cases two review passes found.
 
 from __future__ import annotations
 
+import ast
+import inspect
 import sys
 from pathlib import Path
 
 import pytest
 
-sys.path.insert(0, str(Path(__file__).resolve().parents[2] / "scripts" / "hooks"))
+_SCRIPTS = Path(__file__).resolve().parents[2] / "scripts"
+sys.path.insert(0, str(_SCRIPTS / "hooks"))
 import shell_parse as sp  # noqa: E402
 
 NV = "--no" + "-verify"  # keep the literal out of this file's own text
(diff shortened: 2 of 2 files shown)


WingedGuardian:
@codex review

WingedGuardian:

**Files reviewed:** `scripts/hooks/shell_parse.py`, `tests/test_hooks/test_shell_parse.py`

### Structural review

- No pattern violations found: no `nohup`, no direct `task_states` inserts, no `/tmp/` usage, no `git add .`/`-A`, no `git push --force`/`git reset --hard`, no worktree-path `pip install -e`, no bare `except:`, no mutable default arguments, no unvalidated `os.killpg`, no raw SQL against `genesis.db`.
- Import hygiene: clean. Stdlib (`ast`, `inspect`, `sys`, `pathlib`) → third-party (`pytest`) → local, no star or circular imports. The pre-existing `# noqa: E402` pattern for the `sys.path.insert` before the local import is unchanged and was already the established convention in this file.
- Coverage: no new `src/` file requiring a companion test. The PR modifies an existing hook module (`_KNOWN_SIGILS` tuple) and its existing test file, and adds a new regression class (`TestSigilRunRegression`) including a bidirectional `ast`-walk consistency check that structurally prevents recurrence of this class of bug (declared-but-unqueried and queried-but-undeclared sigils, in both directions).
- Documentation: no new public functions/classes need docstrings beyond what's present; the new test class carries one.
- Genesis conventions: no heavy imports added to the hook script, no `$CLAUDE_PROJECT_DIR` misuse, no `IntervalTrigger`. PR title uses a conventional `fix(hooks):` prefix.

### PII / Secrets / Wording

- Code diff: no API keys/tokens/secrets, no private IPv4/IPv6 ranges, no SSH keys, no DB connection strings — hard-block scan is clean. No soft-warning items either (no personal emails, username-bearing paths, revealing timezones, cloud account IDs, or hardware/private hostnames).
- PR title/body: technical and internal in nature (sigil-list sync bug, `ast`-based derivation, verification methodology). It doesn't disclose private infrastructure, a PII/secrets incident, or exploitable detail beyond what the mechanism already visible in the repo (`has_trailing_override` / `_KNOWN_SIGILS`) already exposes.

VERDICT: PASS



---

PR #1578, opened 2026-09-01
fix(hooks): superseded by #1580

WingedGuardian:
Superseded by #1580.

Closed and reopened so the branch carries a single commit with corrected wording. Squash merges take their message from the branch commits, so the wording had to be corrected at the commit; the push guard does not permit force-pushing a public branch, so a fresh branch was the route.

Same code, same tests, same measurements. See #1580.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,35 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ## [Unreleased]
 
+### Fixed
+
+- **Captured session output no longer persists API keys in plaintext.** When an
+  interactive session exits, Genesis records a tail of the terminal scrollback to
+  `~/.genesis/logs/cc_exit_<slot>.log` so a crash can be diagnosed afterwards.
+  That tail is raw terminal output, so anything the session printed — including a
+  credential a CLI echoed to the screen — was written verbatim. The tail is now
+  scrubbed before it is stored, and if the scrubber cannot run the tail is
+  withheld entirely rather than written raw; the exit status and crash diagnosis
+  are always recorded either way.
+- **Credential detection now recognises modern key formats.** The shared
+  secret-detection patterns required an unbroken run of letters and digits after a
+  vendor prefix, so any key that namespaces itself with internal separators —
+  which most current formats do — stopped matching a few characters in and were
+  treated as ordinary text, so they were never redacted from captured telemetry.
+  Redaction now handles them, along with several provider prefixes, JWTs, webhook
+  URLs and bot tokens that had no coverage at all. Coverage is by shape, so it is
+  not complete: keys that are a bare run of letters and digits with no vendor
+  prefix stay indistinguishable from an ordinary hash and are still only caught
+  in their labelled or ``KEY=VALUE`` forms. The reference-capture path keeps a
+  deliberately narrower rule, so that an ordinary hyphenated name is never stored
+  as though it were a credential.
+- **Secret scrubbing no longer stalls on long machine output.** Two detection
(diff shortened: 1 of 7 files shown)


WingedGuardian:
@codex review

WingedGuardian:

(Note: GitHub does not allow requesting changes on your own pull request, so this is posted as a plain comment rather than a formal review — treat the finding below as a real block, not a pass.)

## Structural Review

**ERROR:** none.

**WARNING:** none.

**NOTE:**
- `scripts/hooks/secret_scrub.py`: `_STRUCTURED_CREDENTIAL_PATTERN` and the widened `_KNOWN_KEY_PREFIX_PATTERN` are new/changed detection surfaces documented only via inline comments (no module/function docstring update). The comments are thorough, so this is not blocking — just flagging for reviewer awareness on a security-sensitive regex.

Everything else checked out: no bare `except:` (the existing `except Exception:  # noqa: BLE001` in `scrub()` predates this diff and isn't touched by it), no mutable default arguments, no `/tmp/` or `nohup` usage, no raw SQL, no `git add -A`/`--force`/`--hard`, imports in the new/changed test files are stdlib-only and correctly ordered, and every changed `.py` file already has — and this PR extends — a corresponding test file. Commit/PR title carries the `fix(hooks):` conventional prefix.

## PII/Secrets/Wording Review

**Code diff:** clean. Every credential-shaped string added is either a synthetic test fixture built by string repetition (`"sk-ant-oat01-" + "T" * 60`, synthetic JWT bodies, etc. — clearly constructed to exercise the scrubber, per the "known false positive" carve-out for scanner/sanitizer test fixtures) or Telegram's own publicly documented Bot API example token (`110201543:AAHdqTcvCH1vGWJxfSeofSAs0K5PALDsaw`), used to assert detection at the vendor's documented length floor. No private IPs, no personal emails, no username-bearing paths, no real secret values.

**PR title/body + CHANGELOG.md — WORDING (CRITICAL), HARD BLOCK.** This is the reason for requesting changes; the code fix itself looks sound and well-tested.

The title ("stop captured session output persisting credentials") and body ("went in verbatim", "the tail is now scrubbed before it is stored, and when the scrubber cannot run the tail is withheld rather than written raw") assert, in past tense, that real credentials were being written to disk in plaintext by this system before this fix. That is substantively the "fixed our credential leak" pattern this policy exists to catch, and `GENesis-AGI` is documented (project CLAUDE.md) as the public half of this install — so that sentence is now permanent, public, searchable git history on a repo whose whole design captures terminal scrollback that plausibly did contain real user API keys pre-fix.

Separately, the body and the CHANGELOG entry spell out the residual detection gap in exploitable detail: *"keys that are a bare run of letters and digits with no vendor prefix stay indistinguishable from an ordinary hash and are still only caught in their labelled or `KEY=VALUE` forms."* Read plainly, that's instructions for evading the redaction (use a prefix-less alphanumeric key, avoid `KEY=value` form) — published in the same public, permanent record.

Recommend before merge:
1. Reword the title/body to frame this as hardening/improving log redaction coverage, without asserting past-tense that credentials "were written verbatim" / "persisted in plaintext."
2. Drop or generalize the residual-gap paragraph in both the PR body and the CHANGELOG entry so it doesn't hand out the specific shape that still evades detection — e.g. "coverage is by shape and not exhaustive" without describing which shape defeats it.
3. Keep everything else (the measured false-positive rate, the performance numbers, the three-file divergence rationale) — none of that is a problem, only the incident-framing and exploit-detail sentences are.



---

WingedGuardian:
Superseded by #1580.

Closed rather than updated: squash merges take their message from the branch commits, so the commit wording itself needed correcting, and the push guard (correctly) does not permit force-pushing a public branch. #1580 carries the same change as a single commit with corrected wording — same code, same tests, same measurements.

PR #1603, opened 2026-09-02
fix(memory): load the graph as a multigraph so parallel typed edges survive

WingedGuardian:
## The defect

`memory_links`' primary key is `(source_id, target_id, link_type)` — migration 0029 made distinct relationship types between the same pair coexist rather than overwrite. The loader then dropped them again: `_ensure_graph` built an `nx.DiGraph`, which cannot hold parallel edges, so the second `add_edge` for a pair overwrote the first's attributes and the graph kept an **arbitrary survivor** (whichever row loaded last).

That survivor is what `_bfs_with_strength`'s `min_strength` and `link_type_filter` were evaluated against — so a weak or wrong-typed survivor could hide a neighbour that a strong, matching parallel edge really does reach.

## The fix

Load into `nx.MultiDiGraph` keyed by `link_type` (mirroring the PK), and select the strongest passing edge per neighbour during traversal, **breaking ties on `link_type`**.

Max-strength selection is not optional: once the graph holds parallel edges, the old loop's `if neighbor in visited: continue` fires on the second parallel edge, so a bare graph-type swap would move the arbitrary survivor from load time to traversal time. The tie-break is not optional either — see below, only 33 of 139 multi-type pairs have a unique max.

## Measured (real database, ~252,700 link rows)

| | |
|---|---|
| Edges recovered | **142**, across 139 multi-type pairs |
| Pairs straddling `min_strength` 0.3 or 0.5 | **0 and 0** — exhaustive over the whole population |
| Multi-type pairs with a UNIQUE max strength | **33** |
| Multi-type pairs with a **TIED** max strength | **106** |
| Tied pairs containing a `contradicts` edge | **0** |
| Labels changed by the tie-break (same row order) | **10 / 140,286 (0.007%)** |
| Betweenness delta (DiGraph vs MultiDiGraph) | **~1e-17** |
| Rebuild time | 1508ms → 2394ms (**+59%**) |
| Additional peak RSS | **~+47MB** |

**Reach-set is provably unchanged.** 0.3 and 0.5 are the only thresholds any caller uses, and no multi-type pair straddles either — so no traversal *can* reach a different set of nodes. Exhaustive over all 139 pairs, not a sample. A 3,000-root replay agreed at both thresholds: 0 roots changed.

**Centrality is unaffected**, structurally rather than by luck: betweenness iterates neighbours, not edges, and the two graphs have identical adjacency. The dream-cycle importance shield therefore selects the same population.

## A correction to this PR's earlier claims

An earlier revision of this PR body and commit message said the reported `(link_type, strength)` "becomes deterministic-max instead of arbitrary." **That was wrong, and the tie-break does not make it true either.** Verifying the tie-break end-to-end surfaced a much larger effect:

Loading the same graph in **reversed row order** and running the same traversals flips **~15.8% of reported labels** (3,533 of 22,299 at 500 roots). At that sample every flip traced to a node claimable by more than one parent and none to a parallel-edge tie — but **that zero is a subsample bound, not an absolute**: at 3,000 roots the tie-break itself removes 16 of 21,486 flips (~0.07%). Ties are a rounding error on this surface, not nil. The dominant cause is the visited-set BFS letting whichever parent the queue reaches first claim the node, and queue order follows the loader's unordered `SELECT`.

That effect is **pre-existing and not introduced here** — the pre-PR `DiGraph` plus the original first-edge-wins BFS measures the same ~15.8% on the same roots.

"Reach-set unaffected (0 delta)" holds for `_bfs_with_strength`'s **full output**, but not for what the model actually sees: `mcp/memory/core.py:437` and `:709` slice `nodes[:5]` after a *stable* sort on `(depth, -strength)`, so ties in that key keep row order. Measured at the real call parameters (`max_depth=2, min_strength=0.3`, 3,000 roots): **102 of 2,904 roots (3.5%) get a different SET of five memories** into LLM context depending on row order, and 281 (9.7%) get the same five in a different order. That slice surface is the acceptance bar for the follow-up.

It is tracked as its own follow-up rather than fixed here: best-parent-wins is a traversal-semantics change affecting ~16% of reported labels and needs its own blast-radius measurement and review, and bundling it would make this diff two unrelated arguments at once.

**So what this PR claims, precisely:** parallel typed edges survive the load, `min_strength`/`link_type_filter` are evaluated against the real edge set rather than an arbitrary survivor, the choice *within* a pair is well-defined, and the loader represents the true multigraph an engine migration will build on. It does **not** claim row-order-independent labels.

## Cost

The rebuild is ~886ms slower and holds ~47MB more. It lands on a path that already exceeded its own traversal budget before this change, so this makes an existing problem more visible rather than introducing a new one. The synchronous rebuild's event-loop cost is tracked as a separate follow-up — a performance concern that should not be mixed into a correctness fix.

## Verification

Four tests, all watched RED before their fix and green after; both mutations re-confirmed RED (`MultiDiGraph`→`DiGraph` fails all four; tuple→strength-only fails exactly the tie test), applied with an anchor-count assertion and a `compile()` postcondition and restored by hash comparison.

Worth noting *how* the fixture got here, because three revisions were vacuous:

1. **Revision 1** — only 2 of 3 failed pre-fix; the strong edge happened to be the last-loaded survivor.
2. **Revision 2** — claimed order-independence via "two pairs inserted in opposite orders". That does not hold: under a PK-index scan every pair resolves by the same alphabetical rule, so both pairs keep the same side. Red in practice only because today's plan is a rowid scan — a planner accident, not a contract.
3. **Revision 3 (shipped)** — pins the property in the *data*, verified RED across insertion, PK-index ascending, reversed, and PK-index descending order.

The new tie test does not repeat that mistake: it asserts the answer is the **same from two row orders** rather than naming a winner under one, so it cannot go vacuously green when the planner changes. Its coverage limit is stated honestly in the docstring — measured, not reasoned: only 2 of the 4 scan orders can discriminate, because both descending orders already agree with the fixed behaviour.

Also fixed: the bake-off control engine's docstring described production's loader as lossy and flagged that as a finding staged for the engine write-up — this PR makes it false, and a stale claim there would propagate into an engine-adoption decision. And the subsystem map described `db/crud/entities.py` as recursive-CTE traversal; it is a Python frontier BFS issuing one query per depth level.

## Known residual (pre-existing, not introduced)

The CTE fallback used when NetworkX is unavailable emits one row per `(target, link_type, depth, strength)` while the NetworkX path emits one node per target, and it uses path-substring cycle avoidance rather than a visited set. This change narrows one axis — the NetworkX row is now the strongest — but does not close the divergence, and this PR does not claim parity between the two paths.

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved multiple relationship types between the same memories during graph loading and traversal.
  - Traversal now evaluates parallel relationships and reports the strongest qualifying connection and its type.
  - Equal-strength relationships now produce deterministic results.
  - Strength filters correctly retain eligible parallel connections.
  - Improved consistency between production graph behavior and evaluation tooling.
- **Documentation**
  - Clarified graph traversal and relationship-handling behavior.
- **Tests**
  - Added coverage for parallel relationships and deterministic tie-breaking.

### DIFF
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -117,7 +117,8 @@ Easy-to-forget mechanisms:
   — first-party memory vs knowledge-base is a load-bearing distinction.
 - **Entity layer (WS-H Pillar 2)** — typed entity nodes with identity:
   `entities`/`entity_mentions`/`entity_links` tables (migration 0051),
-  `db/crud/entities.py` (recursive-CTE traversal, bi-temporal edge validity,
+  `db/crud/entities.py` (Python frontier BFS — one flat `IN`-list query per
+  depth level, NOT a recursive CTE — bi-temporal edge validity,
   EXTRACTED/INFERRED/AMBIGUOUS provenance, `merge_entity` tombstone-with-
   redirect), `memory/entity_registry.py` (string→ID resolution tiering; fuzzy
   matches queue `entity_adjudication`), `memory/entity_seed.py` (curated spine
--- a/src/genesis/eval/graph_bakeoff/engines/nx_incremental.py
+++ b/src/genesis/eval/graph_bakeoff/engines/nx_incremental.py
@@ -1,17 +1,26 @@
 """NetworkX-incremental control engine (the baseline candidate).
 
 Represents "NetworkX done correctly as an engine": it REUSES production's exact
-BFS (``genesis.memory.graph._bfs_with_strength``) so we measure the shipped
-traversal logic, but builds an ``nx.MultiDiGraph`` (not production's lossy
-``DiGraph``) so all 115 parallel-typed edge-pairs are preserved and it matches the
-all-rows SQL oracle. Its cold-start IS a full rebuild from the snapshot — the
-honest architecture, and the number that quantifies today's price of every one of
-the 11 ``invalidate_graph_cache()`` sites.
+BFS (``genesis.memory.graph._bfs_with_strength``) and builds an
+``nx.MultiDiGraph`` — the same shape production's ``_ensure_graph`` now builds —
+so all parallel-typed edge-pairs are preserved and it matches the all-rows SQL
+oracle. Its cold-start IS a full rebuild from the snapshot — the honest
+architecture, and the number that quantifies today's price of every one of the
+11 ``invalidate_graph_cache()`` sites.
 
-FINDING (for the doc): production ``graph.py`` uses a plain ``DiGraph`` that
-silently collapses those parallel-typed edges (last-write-wins). Any adopted
-engine — including a productionized incremental-NX — must fix that to represent
-the true multigraph.
+HISTORY: production ``graph.py`` used a plain ``DiGraph`` that silently collapsed
+those parallel-typed edges (last-write-wins) until the MultiDiGraph fix. This
+engine predates that fix and was the finding that motivated it; any adopted
+engine must keep representing the true multigraph. (The pair count quoted here
+was 115 at the 2026-08-06 snapshot and is 139 as of 2026-09-02 — it tracks the
+graph, so re-derive it rather than citing this line.)
+
+One residual difference from production: ``load()`` below calls ``add_edge``
+without an explicit ``key=``, so networkx assigns auto-incrementing integer
+keys, whereas production passes ``key=link_type``. On a snapshot containing a
+duplicate ``(source, target, link_type)`` row the two would diverge (this keeps
+both, production keeps one). The production PK forbids that row, so it cannot
+occur in a snapshot exported from a real database.
 """
 
 from __future__ import annotations
--- a/src/genesis/eval/graph_bakeoff/engines/nx_incremental.py
+++ b/src/genesis/eval/graph_bakeoff/engines/nx_incremental.py
@@ -65,7 +65,14 @@ class NxIncrementalEngine:
             for s, t, lt, st in conn.execute(
                 "SELECT source_id, target_id, link_type, strength FROM memory_links"
             ):
-                G.add_edge(s, t, link_type=lt, strength=st if st is not None else 0.0)
+                # Coalesce BOTH: production's `strength < min_strength` guard
+                # already rejects a None strength before selection, but the
+                # shared `_bfs_with_strength` now tie-breaks on the
+                # (strength, link_type) tuple, so a NULL link_type here would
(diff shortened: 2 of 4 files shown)


WingedGuardian:
@codex review

WingedGuardian:
Automated first-pass review (structural + PII/secrets/wording) of head `c4953fd`.

**Files reviewed:** `docs/architecture/CURRENT.md`, `src/genesis/eval/graph_bakeoff/engines/nx_incremental.py` (docstring only), `src/genesis/memory/graph.py`, `tests/test_memory/test_graph_parallel_edges.py` (new).

### Structural review

- No banned patterns (nohup, direct `task_states` inserts, `/tmp/` usage, `git add -A`/`--force`/`--hard`, worktree `pip install -e`, bare `except:`, mutable default args, unsafe `os.killpg`, new raw SQL against `genesis.db` outside crud modules).
- Import hygiene: no circular imports, no star imports; ordering in the new test file (stdlib → third-party → local) is correct.
- Test coverage: the changed function (`_ensure_graph` / `_bfs_with_strength` in `src/genesis/memory/graph.py`) gets a new, purpose-built test file (`tests/test_memory/test_graph_parallel_edges.py`) exercising the DiGraph→MultiDiGraph behavior change directly; PR description states the pre-existing `test_graph.py`/`test_graph_expansion.py` suites were re-run and pass.
- Documentation: module docstring, inline comments, and `CURRENT.md` were all updated to match the new MultiDiGraph behavior and to correct a stale "recursive-CTE" description of `db/crud/entities.py`. `_bfs_with_strength`'s existing docstring remains accurate.
- Genesis conventions: commit/PR title carries a conventional prefix (`fix(memory): ...`); no hook-script or `$CLAUDE_PROJECT_DIR`/IntervalTrigger issues (none of the changed files are hooks or triggers).

### PII / Secrets / Wording

- Added lines (code + docs + test) contain no API keys/tokens/secrets, no private IPv4/IPv6 ranges, no SSH keys, no DB connection strings, no personal emails, no username-bearing absolute paths, no location-revealing tz identifiers, no cloud account/project IDs, no hardware identifiers, no private hostnames.
- PR title/body discuss an internal graph-traversal correctness bug (`DiGraph` → `MultiDiGraph`, `memory_links` PK behavior) in normal engineering language — no mention or implication of a PII/secrets leak, security incident, or exploitable infra detail.

VERDICT: PASS




---

WingedGuardian:
@codex review

PR #1604, opened 2026-09-02
fix(tests): stop the earn-back fixture aging out of its own window

WingedGuardian:
## This is currently red on every open PR

`tests/test_mcp/test_calibration_status.py::test_earnback_surfaces_demoted_category_with_windowed_evidence` started failing today with `assert 0 == 30`. It is not a flake and not caused by any recent change — it is a time bomb with a 45-day fuse, lit when the test was written.

The module freezes `NOW = datetime(2026, 7, 19, 12, 0, 0, tzinfo=UTC)` and stamped `autonomy_events.occurred_at` with it. Production reads that column through `crud.autonomy.windowed_counts`, which ages it against `datetime.now(UTC)` — `calibration_status` injects no clock — with `_EARNBACK_WINDOW_DAYS = 45`.

Measured at diagnosis:

| | |
|---|---|
| seeded events | `2026-07-19T12:00:00+00:00` |
| real now | `2026-09-02T13:45:44+00:00` |
| 45-day cutoff | `2026-07-19T13:45:44+00:00` |
| events inside window? | **False — they fell out 1h45m earlier** |

## The fix

Stamp the events relative to **real** now via an `_ago()` helper, mirroring the established precedent in `tests/test_memory_integrity/test_posture_and_jobs.py` (commit `95442438`). Margins become 44 days on one side and 15 on the other, rather than the minutes-wide margins behind the historical recurrences of this class.

Production is untouched; this is a test-only change.

## The coverage that was actually missing

Nothing asserted that the window **excludes** anything — and that absence is why the bomb was invisible rather than merely present. With every event stamped at a single instant, a correct window and no window at all produce the same count, right up until the calendar moves the fixture out and the number silently becomes zero.

The new `test_earnback_counts_only_events_inside_the_window` seeds **both** sides (1 day in, 60 days out) and asserts only the inside events count. Verified RED by widening the window `45 → 3650`, which is equivalent to deleting the window logic entirely.

It also pins the window width, which no test asserted either, so a change to `_EARNBACK_WINDOW_DAYS` becomes a red rather than a silent drift from the config default it is documented as mirroring.

The 45-day **boundary** is deliberately *not* pinned here. This surface injects no clock, so a 44d/46d fixture would be a tight-margin bomb of exactly the class being fixed. The boundary is already pinned in `tests/test_autonomy/test_state_machine.py`, where a clock can be injected.

## Why the CI guard did not catch it

`scripts/check_frozen_clock.py` names this shape in its own docstring as knowingly uncovered:

> an absolute date LITERAL near a production threshold. A real bomb, but not statically decidable — it depends on a threshold this guard cannot know. NOT covered here, and no claim is made that the literal population is clean.

That remains true; this PR does not change the guard.

## On the wider class — a floor, not a measurement

An attempt to size the population found 171 absolute datetime literals across 71 of 1,309 test files, and a cross-reference against production moving-window sites yielded 47 candidate files. **That scan failed its own acceptance replay**: it does not catch this very bomb, because the clock read and the subtraction sit on separate lines behind an intermediate variable and the pattern only matched them joined. So 47 is a floor from a demonstrably blind matcher.

A narrower sweep of the 12 files touching this table found no other instance, which bounds this table but not the repo. A real detector would be dynamic — run the suite against an advanced clock and diff the failures — and is left as a separate proposal.

## Verification

72 tests pass across `test_calibration_status.py`, `test_autonomy/test_state_machine.py` and `test_ledger/test_cells.py` (the sibling suites touching the same table). `ruff --no-cache` clean. `scripts/check_frozen_clock.py` clean.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for calibration status calculations using event timestamps relative to the current UTC time.
  * Verified that only events within the 45-day earn-back window affect success and correction counts.
  * Added assertions for the reported earn-back window size.

### DIFF
--- a/tests/test_mcp/test_calibration_status.py
+++ b/tests/test_mcp/test_calibration_status.py
@@ -7,7 +7,7 @@ phrasing, NEVER a bare percentage (design §3.4/§4.3).
 from __future__ import annotations
 
 import re
-from datetime import UTC, datetime
+from datetime import UTC, datetime, timedelta
 from unittest.mock import MagicMock, patch
 
 import aiosqlite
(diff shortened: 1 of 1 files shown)


WingedGuardian:
@codex review

WingedGuardian:

**Scope:** single file changed — `tests/test_mcp/test_calibration_status.py` (test-only, no production code touched).

### Structural review: CLEAN

Checked against Domain 1 rules:
- No `nohup`, no direct `task_states` inserts, no `/tmp/` usage, no `git add .`/`-A`, no force-push/reset patterns, no `pip install -e` worktree paths, no bare `except:`, no mutable default arguments (`_seed_events`/`_ago` use only `int`/`str` defaults), no `os.killpg`.
- The raw `INSERT INTO autonomy_events` / `autonomy_state` calls in `_seed_autonomy`/`_seed_events` are against the test's own `:memory:` sqlite fixture (`aiosqlite.connect(":memory:")`), not `genesis.db` — this is standard test-fixture seeding, not a production raw-SQL violation, and the pattern pre-dates this PR (only the seeded values changed).
- Imports correctly grouped (stdlib → third-party → local); `timedelta` added to the existing stdlib import, no star imports, no circularity.
- No hook-script/IntervalTrigger/`$CLAUDE_PROJECT_DIR` concerns (not applicable to this file).
- New `_ago()` helper and new test both carry thorough docstrings explaining the "why" (the time-bomb mechanism and the window-exclusion gap being closed).
- PR title carries a conventional `fix(tests):` prefix.
- This is itself a test file, so "new .py without a test file" doesn't apply — it *adds* coverage for a previously-untested direction (window exclusion) rather than removing any.

No pattern violations found.

- No API keys/tokens/secrets, private IPs, SSH keys, connection strings, or env-var secrets in the diff.
- No personal emails, username-bearing paths, revealing timezone identifiers, account/project IDs, hardware identifiers, or private hostnames.
- PR title/body reference only file paths, function names, a commit SHA, and measured timestamps relevant to the fix — no mention of a leak/incident/exploit or private infrastructure.

VERDICT: PASS



---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

PR #1605, opened 2026-09-02
fix(cc): seed client-scope CC settings user-level — retention ran at 30d

WingedGuardian:
## What was wrong

Claude Code's transcript-retention sweep walks **every** directory under `~/.claude/projects`, but takes its retention value from the settings of whichever session happens to run it. This repo shipped `cleanupPeriodDays: 180` in its own `.claude/settings.json` — which CC merges only for sessions launched *inside* the project.

So any session started elsewhere — a worktree, `~/tmp`, a dispatched background session whose cwd is outside any git repo — swept every project directory at CC's **30-day default**. A repo-launched session then read 180, found nothing older left, and reported clean. Silent for 73 days.

On the reference install this destroyed **198 local sessions**. All were recoverable: encrypted backups retain transcripts permanently, and that is the only reason this was survivable rather than terminal.

Read from the shipped CC binary (2.1.246):

```js
var Y = 30;                                              // default
function D(e){ let r = (X()||{}).cleanupPeriodDays ?? Y; … }   // cutoff
// kr(): readdir(projectsRoot) → every project dir → unlink *.jsonl older than cutoff
```

## Root cause, verified by controlled comparison

Not inferred. A second install of this same system — same repo, same CC version, **also** carrying the project-level 180, sweep confirmed running — was unaffected and holds transcripts **103 days old**. Its single differing variable: the same value present at the **user** level, hand-set at some point.

That is simultaneously the proof of the mechanism and the proof that the user-level fix works in production. This PR turns that one machine's hand-repair into a mechanism every install gets.

## What changed

Client-scope settings now live in `config/cc-global-settings.yaml` under `user_level_defaults` with an explicit **per-key policy**, applied by all three paths that can reach `~/.claude/settings.json`:

| Path | Covers |
|---|---|
| `scripts/install.sh` | fresh container — it does **not** call bootstrap (all six mentions there are comments) |
| `scripts/setup_claude_config.py` | via bootstrap, so `update.sh` heals existing installs; unconditional, never behind `--global` |
| `scripts/host-setup.sh` | host VM — no venv, no bootstrap |

- **Retention is a `floor`** (180, 6× CC's default): absent or lower is raised, a higher value is kept, `GENESIS_CC_RETENTION_DAYS` names an exact value authoritatively. Values below 1 are refused — CC's schema rejects them, and a negative would put the cutoff in the *future*.
- **`attribution` is seeded set-if-absent**, and deliberately stays in project settings too. The asymmetry is documented: a project copy of a *numeric* key overrides and **caps** the user-level value (why retention had to be removed); a project copy of an identical *object* caps nothing and is pure belt-and-braces.
- The seeder **never raises**. `bootstrap.sh:595` calls it bare under `set -euo pipefail`, from `update.sh`'s `set -Eeuo pipefail` with an armed ERR trap — an exception would escalate into a full deploy rollback.
- All three writers **preserve the target file's mode** across the atomic replace (`os.replace` installs a new inode, and this file can hold API keys in `env`), and the host merge path chowns back to the operator since it runs under sudo.

## Verification

- 41 tests. Both shell blocks are **extracted from the shipped scripts**, not copied.
- `test_shell_paths_implement_every_manifest_key` asserts every manifest key is implemented by every path — a new key that one path misses now fails CI.
- `test_project_settings_carry_no_client_scope_keys` fails CI if any CC client-scope key reappears in `.claude/settings.json`. That is the chokepoint that makes a third instance of this class impossible.
- **Verify-RED: 8 mutations, every one produced a correct RED**, every file restored by sha256. Two tests were caught *blind* by that sweep and rewritten — the mode test used 0600 (which `mkstemp` produces anyway, making the mutation behaviourally null), and the corrupt-file test asserted through a caller whose broad handler masked the loader's exception tuple.
- E2E against the real settings file: an existing higher value is left untouched by the floor; an absent key is seeded; `GENESIS_CC_RETENTION_DAYS=90` yields `365 -> 90 (operator override)`.

## Reviewer findings addressed

A fresh-context adversarial pass raised 2 BLOCKER + 7 SHOULD-FIX; every claim was re-derived empirically before being acted on, and all are fixed and locked with tests. The two blockers were both introduced by this branch: the sudo-path ownership flip, and a **measured** 0600 → 0644 widening on all three write paths. It also caught that `GENESIS_CC_RETENTION_DAYS` was documented in three places and implemented in one, where it could not actually lower anything.

## Scope note

`.claude/skills/genesis-development/SKILL.md` carries a doctrine addition — *a root cause is verified to certainty, or it is not a root cause* — which is unrelated to settings scope. It is here at the maintainer's explicit request, and was earned by this branch: a relayed hypothesis about a related AI-attribution leak was restated as a diagnosis; measurement then showed 10 of 11 cases were stale worktree checkouts, an entirely different mechanism (1 of 11 remains unexplained and is recorded as such).

## Known window, accepted

On an existing install, `update.sh` merges (removing the project-level key) some minutes before it reaches bootstrap. A repo-launched session in that gap has no retention setting at any scope. Bounded, one-time, and the sweep did not fire in any non-interactive probe. Fixing it means editing `update.sh` — which carries rollback traps — for a minutes-long window; not proportionate. Recorded here rather than hidden.

## Deploy path

- Existing installs: `update.sh` → `bootstrap.sh` → `setup_claude_config.py`.
- Fresh container: `install.sh` seeds directly.
- Host VM: `host-setup.sh`. **Triggers the Host-Deploy Gate** — `update.sh` must run after merge.

Recovering already-deleted transcripts: they are in the private backups repo under `transcripts/`, gpg-encrypted.

## Summary by CodeRabbit

- **Bug Fixes**
  - Claude Code transcript retention now uses a user-level 180-day minimum, preserving longer settings and supporting environment overrides.
  - AI-authorship attribution defaults are applied globally while preserving existing preferences.
  - Settings updates preserve permissions, ownership, symlinks, and file metadata, with improved handling for malformed, incomplete, or interrupted writes.
  - Configuration repair now addresses retention and attribution settings consistently across setup paths.

- **Documentation**
  - Clarified configuration scope, retention behavior, backup handling, setup procedures, compatibility safeguards, and review guidance.


WingedGuardian:
@codex review

WingedGuardian:
## Automated first-pass review — PR #1605

Reviewed head `df6652a` (11 files changed: docs/config/CHANGELOG only for 5 of them; shell + Python behavior changes in `scripts/host-setup.sh`, `scripts/install.sh`, `scripts/setup_claude_config.py`; one new test file).

### Structural review

No pattern violations from the Domain 1 checklist:
- No `nohup`, no `/tmp/` usage (the new `_write_json_atomic` uses `tempfile.mkstemp(dir=str(path.parent), ...)`, not the system temp dir), no `git add -A`/`git push --force`/`git reset --hard`, no bare `except:` (all catches — `OSError, ValueError`, `Exception`, `BaseException`, `FileNotFoundError`, `ImportError` — name a type), no mutable default args, no raw SQL, no `task_states` inserts, no `os.killpg`.
- Import hygiene clean in both `scripts/setup_claude_config.py` and `tests/test_scripts/test_cc_user_level_settings.py`: stdlib block, then third-party (`pytest`, `yaml`) in the test file — no star imports, nothing circular.
- All new public/private functions in `setup_claude_config.py` (`ensure_user_cc_defaults`, `_write_json_atomic`, `_load_user_settings`, `_target_for`, `_apply_user_level_defaults`) carry docstrings.
- Commit-prefix convention satisfied (`fix(cc): …`).

One WARNING (missing coverage):
- `scripts/setup_claude_config.py:746-749` — `configure_global_settings` (the `--global` path) now routes through the same `_load_user_settings` (early-return-on-corrupt-file) and `_write_json_atomic` (mode-preserving atomic write) helpers that `ensure_user_cc_defaults` uses. `tests/test_scripts/test_cc_user_level_settings.py` covers those helpers thoroughly via `ensure_user_cc_defaults`, but there's no test exercising them through `configure_global_settings` itself (e.g. a corrupt `~/.claude/settings.json` with `--global` — does it now early-return instead of the old traceback-on-bad-JSON behavior, as intended?). Worth a quick regression test given `--global` is the force-write path with real overwrite risk.

### PII/Secrets/Wording scan

- No API keys, tokens, credentials, private IPv4/IPv6 ranges, SSH keys, or connection strings in the added lines.
- No personal emails, username-bearing absolute paths, revealing timezone identifiers, cloud account IDs, hardware identifiers, or private hostnames added.
- PR title/body/CHANGELOG: describes a retention/data-loss bug (transcripts pruned at CC's 30-day default instead of the intended 180-day floor) and a separate AI-attribution-trailer scoping fix, both in generic terms — no named private infrastructure, no exploitable-detail vulnerability writeup. The "AI-attribution leak" language refers to `Co-authored-by: Claude` trailers appearing on commits, not a PII/secrets/credential leak, so it doesn't fall under the wording hard-block category.
- No hostnames/IPs specific to any install are named anywhere in the diff or PR text (references to "the reference install" / "a sibling install" are anonymized).

PII/Secrets/Wording: clean (no hard blocks; one non-blocking structural coverage warning above).

VERDICT: PASS



---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
## Needs a dedicated architecture session — not a review pass

A review session picked this up to drive it to green and stopped before resolving it. The conflict here is **structural, not textual**: the mechanism this PR implements was superseded on main while it sat open, so no conflict resolution can be correct on its own terms. Flagging it rather than forcing it.

### What collided

`#1497` (merged 2026-09-05) introduced `cc_ensure_updater_suppressed` in `scripts/lib/cc_version.sh` as a single canonical writer for the user-level settings file, and **deleted the inline merge heredocs from both `scripts/install.sh` and `scripts/host-setup.sh`**, replacing each with one call. Its own commit message states the intent: the two setup legs "share that one implementation instead of carrying two copies of the same merge."

This PR's shell paths *are* a second copy of that merge — a per-script read-modify-write. Main did not change the text around this PR; it retired the approach.

Measured on the current heads:

- `git show origin/main:scripts/install.sh | grep -c PYEOF` → `0`; both scripts now call `cc_ensure_updater_suppressed` (`install.sh:1357`, `host-setup.sh:1334`).
- The branch's `cc-user-settings` markers do not exist on main (`grep -c` → `0` in both scripts).
- Performing the merge and then taking main's version of the two scripts — the literal base-wins resolution — takes the branch's suite from **61 passed to 36 failed / 25 passed**. Every failure is the extraction harness asserting the branch's markers exist, including `test_shell_paths_implement_every_manifest_key`, which this PR's body calls "the chokepoint that makes a third instance of this class impossible."

So base-wins would delete this PR's fix on two of its three declared paths. `install.sh` never calls `bootstrap.sh` or `setup_claude_config.py`, so a fresh container install and the host VM would end up with no retention seeding at all — only existing installs updating through `update.sh` would get it. The PR would then claim more than it does.

### Why the obvious repair is not a review-session decision

Adopting the shared writer is the promising direction, but it cannot express this PR's policies as written. `cc_ensure_updater_suppressed` enforces `REQUIRED = {"DISABLE_AUTOUPDATER": "1", "DISABLE_UPDATES": "1"}` and writes everything into `data["env"]` as set-if-absent string defaults. This PR needs a **top-level integer with a floor policy** and a **top-level object**. Neither is representable today.

Extending it means giving the shared reconciler a policy-aware manifest (a floor policy and an object-if-absent policy applied at the top level). That decides how many code paths own the settings file, and it would put retention on the daily settings-align timer — turning a setup-time seed into a continuously re-asserted policy. That is a deliberate infrastructure change, not a merge conflict resolution.

Two further questions the rework has to answer, both raised by review findings on this PR: whether the host update path should gain a settings writer at all (it has never had one, and the gap is currently documented as accepted), and where the retention floor should live once one writer owns it.

### Findings that survive whatever is decided

Two verified defects are independent of the rework and should not be lost with it:

- The deploy path never reaches the new seeder. `scripts/update.sh:1373-1374` and `TIER2_PATHS` in `src/genesis/observability/snapshots/deploy_health.py:48-55` list neither `scripts/setup_claude_config.py` nor `config/cc-global-settings.yaml`, so a pull followed by an update takes the "already up to date" exit and bootstrap never runs. Since this branch removes the key from the repo-level settings file, such an install ends with the setting at no scope, and no staleness alert fires.
- A type-blind comparison at `scripts/setup_claude_config.py:189-190` uses `have != target`, and `True != 1` is `False` — so a boolean value survives an exact override of `1`. The floor path two lines below explicitly rejects booleans, so the two branches disagree.

### Status

No commits pushed, nothing merged, the branch is untouched at its current head, and the trial merge was aborted. The container-side path (`setup_claude_config.py`, the manifest, its tests, docs) is uncontested — zero main commits in the interval — and carries forward intact under any option.

Leaving this for a session that can hold the architecture conversation: extend the shared writer with top-level policy support, narrow this PR to the uncontested container path and re-scope its body, or rebuild against current main.

WingedGuardian:
## Triage 2026-09-14 — back to a builder session (draft + needs-rework)

Batch triage of all open PRs ≤1700. Disposition: the premise was retired on main while this sat open — #1497 made `cc_ensure_updater_suppressed` (scripts/lib/cc_version.sh) the single canonical writer and deleted the inline merge heredocs this PR extends. Re-verified today at current main: `grep -c PYEOF` → 0 in both setup scripts; both call the shared writer (install.sh:1382, host-setup.sh:1334). A base-wins merge deletes this PR's own fix on two of its three declared paths (measured earlier: branch suite 61 passed → 36 failed / 25 passed).

Rebuild direction, per the 2026-09-08 rationale comment above: extend the shared writer with top-level policy support (integer-with-floor, object-if-absent) or narrow to the uncontested container-side path, which carries forward intact.

Two verified defects to preserve in any rework:
- the deploy path never reaches the new seeder (`scripts/update.sh:1373` / `TIER2_PATHS` list neither file, so pull+update exits "already up to date" and bootstrap never runs);
- type-blind `have != target` at `scripts/setup_claude_config.py:189` — `True != 1` is False, so a boolean survives an exact override of 1.

PR #1608, opened 2026-09-02
fix(hooks): fold only real line continuations, not escaped backslashes

WingedGuardian:
## A live fail-open in a wired PreToolUse guard

`_rm_violations` normalized with `cmd.replace("\\\n", "")`, treating **every** backslash-newline as a line continuation. That holds only when the backslash is itself unescaped — an **odd**-length run. In an **even**-length run each backslash is escaped by its neighbour, so the last one is a literal character and the newline after it is a real command separator.

Folding it anyway deleted that separator, glued the next command's first word onto the previous token, and left no `rm` token for the guard to see. Measured end-to-end through the real hook entry:

| input | shell runs | guard tokens | hook exit |
|---|---|---|---|
| `printf x` ⏎ `rm -rf /` (control) | 2 commands | `['printf','x',';','rm','-rf','/']` | **2 — block** |
| `printf x\\` ⏎ `rm -rf /` | 2 commands | `['printf','xrm','-rf','/']` | **0 — allow** |

The legacy `_RM_RF_PATTERN` net does not catch it either: `main()` consults that net only when `_rm_violations` returned `None` (tokenizing **failed**). Here it returned `[]` — tokenizing succeeded and produced the wrong tokens. The stated contract, *an unparseable command is a block*, says nothing about a **mis-parsed** one.

`protected_paths_guard` independently blocks protected targets, but shallow non-protected targets (`/`, `/usr`, `/etc`) are caught by nothing else in the chain.

## The fix

```python
_CONTINUATION = re.compile(r"(?<!\\)((?:\\\\)*)\\\n")
cmd = _CONTINUATION.sub(r"\1", cmd).replace("\n", " ; ")
```

`(?<!\\)` can only succeed at a run's **start**, so a match is attempted once per run; from there the pattern succeeds iff the run is odd. Even runs match nothing.

## Measured

Corpus replay is structurally blind to shapes nobody typed, so the coverage check is a **generated matrix**, with the corpus-style fuzz as corroboration:

| probe | result |
|---|---|
| 5 contexts × run lengths 0–6 (mine) | run 0 and **every odd run byte-identical** to the previous implementation; only even runs ≥2 change |
| Truth table, run lengths 0–8 × 14 positions (reviewer, against a logging `rm` shim) | **new 0 misses, old 20 misses** |
| Exhaustive insertion sweep, every index of 5 destructive bases, ~700 executed cases | **new 0 misses, old 20 misses** |
| Random differential fuzz, 4,000 executed commands | 292 genuinely destructive per bash; **0 new misses** |
| Quote-balance, 20,000 random parseable commands | **old corrupted 11, new 0** |
| ReDoS: 50k-backslash inputs | 3.4–6.2 ms |

Acceptance replay through the real hook, 7/7: the attack blocks at 2 and 4 backslashes; genuine continuations still fold; a continuation inside an option token still blocks; deep paths and ordinary commands unaffected.

## The same error had a second consequence

The review found the parity bug also broke a claim the file makes about itself. Deleting one backslash from an even run leaves an **odd** run whose survivor escapes the next character — and when that character is a quote, the fold **introduces an escape**, shifting shlex's balance. Parseable commands became unparseable and fell into the legacy net, which matches neither `-Rf` nor `-r -f` nor `--recursive --force` nor a quoted `'rm'`, so they failed **open**. Four such old-code bypasses were measured directly, e.g. `printf x\\` ⏎ `'rm' -Rf /usr` → old allow, new block.

One faulty premise produced two bypass families. The fold now always leaves an even-length run, which is what finally makes the file's stated `cannot` true.

## Documentation corrected

`.claude/skills/genesis-development/SKILL.md` cited this exact function as the model of a sound structural argument, on the premise that "a backslash-newline is a continuation, not an escape the shell keeps". That premise is the bug. The correction states the odd/even case-split, records that the quote-balance argument was *also* unsound in the same cell, and keeps the part that was right — in a passage whose own thesis is that the danger is believing you already applied the rule.

## Verification

Four new tests, RED witnessed before the fix. Two of them were initially **vacuous in the direction their docstrings claimed** — measured against a never-fold control, both passed — which the review caught. The boundary test's target moved from a depth-5 path to `/` so the two implementations actually disagree, re-verified by mutation; the other keeps a corrected docstring naming it an equivalence lock rather than an over-correction constraint, since the pre-existing `TestLineContinuationFoldsAsTheShellFolds` class is what genuinely carries that guarantee.

217 tests pass across the guard, `shell_parse` and `protected_paths` suites. `ruff` clean.

## Not bundled here

Two sibling findings from the same review are tracked separately, since a gate-parser change belongs in its own isolated PR:

- **`shell_parse.split_segments` has no backslash handling outside quotes** and splits unconditionally on `\n`, so a *genuine* continuation splits a word bash joins and the protected-paths guard never sees the real target. Verified: `rm -rf ~/.claude/proj\`⏎`ects` and `rm -rf ~/genesis/da`⏎`ta` return exit 0 from every hook, while their unsplit controls are blocked. More severe than the defect fixed here.
- The protected-paths ancestor test builds a doubled separator for filesystem root, the one ancestor it therefore misses.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved destructive-command detection for shell commands containing escaped backslashes before newlines.
  - Command separators are now preserved when appropriate, preventing separate commands from being incorrectly combined.
  - Destructive `rm` commands following escaped line breaks are detected more reliably.
  - Shell line continuations continue to work correctly for odd-length backslash sequences.

- **Documentation**
  - Expanded guidance on shell normalization, backslash handling, quote balancing, and validation limitations.

### DIFF
--- a/.claude/skills/genesis-development/SKILL.md
+++ b/.claude/skills/genesis-development/SKILL.md
@@ -615,10 +615,50 @@ tool-selection decision matrix: `.claude/docs/code-intelligence.md`
   `scripts/hooks/destructive_command_guard.py:101-120`: its replacements delete a
   line continuation — which is what the shell does with it — or insert
   whitespace and separators; none introduces a quote or an escape, so none can
-  corrupt the quote balance shlex decides on. (The deletion removes a backslash,
-  but a backslash-newline is a continuation, not an escape the shell keeps; the
-  one place the shell does keep it, inside single quotes, is an over-block the
-  guard states rather than hides.) That is a `cannot`; a
+  corrupt the quote balance shlex decides on. (The one place the shell keeps a
+  backslash-newline literally, inside single quotes, is an over-block the guard
+  states rather than hides.) That is a `cannot` — but read the next paragraph
+  before reusing it, because the version of this argument that stood here for
+  weeks was WRONG in a way that shipped a live bypass.
+
+  **What that earlier version got wrong, and why it is the sharpest example on
+  this page.** It added: *"a backslash-newline is a continuation, not an escape
+  the shell keeps."* That is true only when the backslash is itself unescaped —
+  an ODD-length run. In an EVEN-length run every backslash is escaped by its
+  neighbour, so the last one is a literal character and the newline after it is
+  a REAL command separator. The guard folded it anyway, deleting the separator
+  and gluing the next command's first word onto the previous token, so no `rm`
+  token existed and a destructive command was ALLOWED — measured end-to-end
+  through the live hook, exit 0 where the plain-newline control gave exit 2. The
+  legacy regex net did not save it either: that fires only when tokenizing
+  FAILED, and this tokenized fine, just wrongly. Fixed 2026-09-02 by folding
+  only odd-length runs.
+
+  Note precisely what was and was not at fault, because the first attempt at
+  this correction got it wrong in an instructive way: it said the quote-balance
+  reasoning "was CORRECT and is not what broke", and exonerated it. Quote
+  corruption is indeed not the *mechanism* of the token-glue bypass — but the
+  quote-balance `cannot` **was not sound either, and it failed in the very same
+  cell**. Deleting one backslash from an even-length run leaves an ODD run whose
+  survivor escapes the next character; when that character is a quote, the fold
+  DOES introduce an escape and shlex's balance shifts. Measured: it turned
+  parseable commands UNPARSEABLE (11 of 20,000 random parseable commands under
+  the old fold, 0 under the fix), which dropped them into the legacy-regex net —
+  a net that matches neither `-Rf` nor `-r -f` nor `--recursive --force` nor a
+  quoted `'rm'`, so they failed OPEN. One quiet universal quantifier (*every*
+  backslash-newline is a continuation) falsified BOTH claims at once, and the
+  old code therefore had two bypass families rather than one. The fold now
+  always leaves an even-length run, which is what finally makes the `cannot`
+  true.
+
+  A structural `cannot` is only as good as the case-split it rests on, so state
+  the split explicitly and enumerate it — the same trap as the direction-claim
+  two paragraphs down, which was true of the operand and false of the option
+  token. That both the wrong premise AND the first correction of it were written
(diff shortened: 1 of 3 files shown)


WingedGuardian:
@codex review

WingedGuardian:
Automated first-pass review (structural patterns + PII/secrets/wording).

### Structural review

**WARNING** — `.claude/skills/genesis-development/SKILL.md`: the new paragraph was inserted directly before a pre-existing identical line instead of replacing it, leaving a duplicated/garbled sentence in the final document:

> A corpus run only ever yields `did not, here`, and by the rule above it is
> corpus run only ever yields `did not, here`, and by the rule above it is
> structurally blind to the shape nobody typed.

Confirmed by diffing the merged file content — the fragment appears twice back-to-back. Please collapse to a single instance.

No other structural findings:
- `scripts/hooks/destructive_command_guard.py` — hook stays stdlib-only (`json/os/re/shlex/sys` + local `hook_input`), no new imports/import-order issue, no mutable default args, no bare `except:`, no raw SQL/`nohup`/`task_states`/`/tmp/`/force-push patterns.
- `tests/test_hooks/test_destructive_command_guard.py` — new coverage was added alongside the behavior change (`TestEscapedBackslashIsNotAContinuation`), so the "changed function, missing test update" case doesn't apply here.
- PR title carries a conventional `fix(hooks):` prefix.

### PII / Secrets / Wording

Scanned added lines across all three files plus the PR title/body: no API keys, tokens, private IPs, SSH keys, connection strings, personal emails, or username-bearing paths.

On the wording check specifically: the PR body is a detailed technical writeup of a since-fixed logic bug in a local Bash safety hook (line-continuation folding in `destructive_command_guard.py`), including before/after exploit tables. This reads as ordinary security-fix documentation for guard code in this file — consistent with the extensive bypass-history commentary already checked into this same file from prior review rounds — not a PII/credentials/infrastructure disclosure. No private hostnames, account IDs, or infra names are named. Treated as clean.

PII/Secrets/Wording: clean (no hard blocks, no soft warnings)

VERDICT: PASS



---

PR #1614, opened 2026-09-03
fix(browser): stop screenshots overwriting one another in both writers

WingedGuardian:
## Problem

`browser_screenshot` wrote **every** capture to one fixed path,
`~/tmp/genesis_browser_screenshot.png`. A session that captured 8 pages kept only
the last one — silently, because each call still returned a valid-looking path.

`scripts/browser.py` had the identical defect through a module-level
`DEFAULT_SCREENSHOT` constant, which binds one filename for the life of the
process. Both writers are fixed here, so the **class** is closed rather than the
reported instance. Enumerated via `grep -rn "\.screenshot(" src/ scripts/` → 3
sites; the third takes a caller-supplied path with no default and is correctly
excluded.

## Fix

Filenames become `genesis_browser_screenshot_<stamp>_<uuid4hex>.png` with
`stamp = %Y%m%dT%H%M%S%fZ`.

- **Microsecond resolution**, not seconds — measured: 8 consecutive captures
  produced exactly **1** distinct second-resolution stamp, which would leave the
  motivating sequence unordered.
- **Untruncated uuid** — 32 bits would reintroduce silent overwrite by collision
  across the 7-day `~/tmp` window.
- **No new unbounded store**: `scripts/disk_hygiene.sh:59-66` already prunes
  direct children of `~/tmp` at `-mtime +7`; timer verified enabled + active.

The CLI keeps its own copy of the scheme so it stays free of `genesis` imports;
the stamp **shape is asserted on both sides** so the copies cannot drift apart
silently.

## Tests

First tests for a tool that had none, plus a CLI drift guard. **All 7 are
mutation-verified** — each dies under a mutation of the one property it names,
most of them length-preserving:

| mutation | result |
|---|---|
| ordering scrambled (same length) | RED |
| stamp shape drift (T and Z dropped) | RED |
| second-resolution only | RED |
| returned path diverges from written | RED |
| filename fixed again (the original bug) | RED |
| CLI stamp drifts from MCP | RED |
| CLI fixed constant returns | RED |

Zero survivors.

## End-to-end verification

Real Playwright, real captures, real files — the running MCP server still holds
pre-change code, so the runtime path was driven directly:

- **MCP path**: 8 sequential captures → 8 distinct paths, 8 non-empty files,
  stamps strictly increasing, **name-sort order == capture order**.
- **CLI path**: `cmd_screenshot()` ×3 with an isolated profile → 3 distinct
  names, 3 non-empty files.

106 tests pass; `ruff check src/ scripts/ tests/` clean.

## Review

3 adversarial `genesis-architect` rounds, sequential. Rounds 1–2 were
defect-bearing and both are documented in the evidence, including that round 1's
own replacement test was vacuous in the dimension it named — the round-1 mutation
changed filename *length*, so a co-located length assert masked a dead sort
assert. Round 3 was clean (5 NOTEs; 3 fixed, 2 doc-accepted as optional
hardening).

Also recorded rather than hidden: the mutation harness itself first keyed its
baseline copies on the file *basename* — and both files are named `browser.py`,
so the "restore" wrote CLI content into the MCP module. The same silent-overwrite
class this PR fixes, reproduced in the tooling built to verify it. Recovered from
the git index; the harness now keys on the full path with a collision assertion.

## Summary by CodeRabbit

* **New Features**
  * Browser screenshots now use unique timestamped filenames by default.
  * Multiple screenshots are preserved instead of overwriting a fixed file.
  * Screenshot tools report the generated file path and document the updated naming behavior.

* **Bug Fixes**
  * Improved timestamp precision and ordering, including captures made across midnight.

* **Tests**
  * Added coverage verifying unique filenames, accurate returned paths, and retained screenshots.
  * Added reliable command-line screenshot path testing even when browser automation dependencies are unavailable.

### DIFF
--- a/scripts/browser.py
+++ b/scripts/browser.py
@@ -16,12 +16,26 @@ Usage:
 
 import argparse
 import sys
+import uuid
+from datetime import UTC, datetime
 from pathlib import Path
 
 from playwright.sync_api import sync_playwright
 
 USER_DATA_DIR = Path.home() / ".genesis" / "browser-profile"
-DEFAULT_SCREENSHOT = str(Path.home() / "tmp" / "browser_screenshot.png")
+
+
+def _default_screenshot_path() -> str:
+    """A unique path per capture, so consecutive screenshots don't overwrite.
+
+    A module-level constant cannot do this — it would bind one name for the
+    life of the process, which is the bug this replaces. Same scheme as the
+    MCP tool's `_impl_browser_screenshot` (genesis/mcp/health/browser.py):
+    sortable timestamp + full uuid4 hex. Deliberately duplicated rather than
+    imported — this CLI stays free of genesis imports so it can run standalone.
+    """
+    stamp = datetime.now(UTC).strftime("%Y%m%dT%H%M%S%fZ")
+    return str(Path.home() / "tmp" / f"browser_screenshot_{stamp}_{uuid.uuid4().hex}.png")
 
 
 def _launch(pw):
@@ -83,7 +97,7 @@ def cmd_screenshot(args):
     with sync_playwright() as pw:
         context, page = _launch(pw)
         try:
-            path = args.path or DEFAULT_SCREENSHOT
+            path = args.path or _default_screenshot_path()
             page.screenshot(path=path)
             print(f"Screenshot saved: {path}")
         finally:
--- a/src/genesis/mcp/health/browser.py
+++ b/src/genesis/mcp/health/browser.py
@@ -23,6 +23,8 @@ import random
 import re
 import signal
 import time
+import uuid
+from datetime import UTC, datetime
 from pathlib import Path
 
 from genesis.mcp.health import mcp
@@ -51,6 +53,12 @@ _TS_LOG_PATH = Path.home() / "tmp" / "turnstile_debug.log"
 def _ts_log_write(msg: str) -> None:
     """Write a timestamped line directly to the Turnstile debug log."""
     try:
+        # Deliberate LOCAL import despite the module-level one: it keeps this
+        # writer out of reach of patch.object(browser, "datetime"), which the
+        # clock-controlled screenshot tests use. Deleting it as "redundant"
+        # would let this function consume an injected side_effect entry and
+        # write a MagicMock repr into the log — silently, since this block
+        # swallows exceptions.
         from datetime import UTC, datetime
 
         ts = datetime.now(UTC).strftime("%Y-%m-%d %H:%M:%S")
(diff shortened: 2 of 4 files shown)


WingedGuardian:
@codex review

WingedGuardian:

**Scope**: `scripts/browser.py`, `src/genesis/mcp/health/browser.py`, `tests/test_mcp/test_browser_tools.py`, `tests/test_scripts/test_browser_cli_screenshot_path.py` (4 files, +222/-5).

No pattern violations found against the checklist (nohup, direct `task_states` inserts, `/tmp/` usage, `git add -A`/force-push, bare `except:`, mutable defaults, `os.killpg`, raw SQL). Both new path-generation helpers correctly write under `~/tmp/` (`Path.home() / "tmp"`), not `/tmp/`.

Import hygiene: no circular or star imports. New stdlib imports (`uuid`, `datetime`) are correctly grouped — plain `import` statements alphabetized before `from ... import` statements, stdlib before third-party/local, in both `scripts/browser.py` and `src/genesis/mcp/health/browser.py`.

Coverage: no new `src/` file lacking a test. The two changed functions (`_impl_browser_screenshot`, `cmd_screenshot`/`_default_screenshot_path`) both get new, well-targeted tests (chronological-sort, no-overwrite, returned-path-matches-written-path, drift guard between the two writers).

Documentation: `browser_screenshot`'s public docstring is updated to match the new behavior (unique path, "always read the returned path").

Commit/PR prefix: `fix(browser): …` — conventional prefix present.

**NOTE (informational only, no action needed):** `_ts_log_write` in `src/genesis/mcp/health/browser.py` keeps a function-local `from datetime import UTC, datetime` alongside the new module-level import of the same names. This is intentional (a comment explains it keeps the Turnstile debug logger out of reach of `patch.object(browser, "datetime")` used by the new clock-controlled tests) and not a bug — flagging only so a future reader doesn't "clean it up" without reading the comment first.

Scanned all added lines (`+`) plus PR title/body. No API keys, tokens, credentials, private IPv4/IPv6 ranges, SSH keys, connection strings, or secret env values. No personal emails, username-bearing absolute paths (all paths go through `Path.home()`), location-revealing timezone identifiers, cloud account IDs, hardware identifiers, or private hostnames.

PR title/body describe a straightforward overwrite bug (screenshots clobbering each other) and testing/review methodology — no mention of a security incident, data exposure, or private infrastructure.

VERDICT: PASS



---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

PR #1615, opened 2026-09-03
fix(hooks): read an unquoted backslash the way the shell reads it

WingedGuardian:
## The defect

`parse_segments` handled a backslash in exactly one place — inside a double-quoted string (`shell_parse.py:341-345`). Outside quotes it set no state, so the character after it was dispatched as though it stood alone.

For a newline that is a guard bypass. The shell removes a backslash-newline and joins the two halves into one word; the scanner split there instead, so `protected_paths_guard` received two fragments and neither named the directory the command would operate on. Measured through the live hook entry, against the identical command written on one line as a control. Depth does not backstop it — the shapes that motivated this sit well below the depth allowance.

Escaped separators and escaped quotes were mis-read by the same missing branch; the escaped-quote case opened a quoted run that ran to end-of-string, hiding a second executed command from every consumer.

## The fix

Fold a real continuation; consume any other escaped pair so the escapee is never dispatched as an operator or a quote opener. Only an **odd**-length backslash run is a continuation — in an even run the last backslash is itself escaped and the newline really does separate two commands. Consuming the pair makes that parity fall out with no counting, and it agrees with `destructive_command_guard._CONTINUATION` across every run length × position.

Placed **after** the redirect branch deliberately: an escaped space inside a redirect target is consumed by `_redirect_target_end`, and the golden corpus locks that output.

## What the review changed, and why it is the interesting part

Two adversarial rounds each found a fail-open **in the fix**, not in the original code. Both are one mechanism, and it is worth stating plainly:

> Appending is safe under an incomplete context model — it can only ever over-split, and an over-read costs a rewrite. **Folding is a join, and a join under an incomplete context model makes a real command vanish.**

Round 1: folding inside a `#` comment deleted a real separator (the shell ends a comment at the newline and does not continue it), so the following command disappeared from every consumer — and a token pulled in from the next line was read by `has_trailing_override` as a genuine approval acknowledgement. Round 2: the comment-start predicate missed parentheses.

Rather than patch a third instance, the contexts were **enumerated against bash**. All thirteen places a backslash-newline can sit — quotes both kinds, comments, both heredoc delimiter forms, heredoc terminator, subshell, `$( )`, backtick, case pattern — with ground truth being whether bash *executes* the following line. **None hides a command the shell runs.**

Ground truth is a marker file, not a string in output. An earlier harness asked whether a marker appeared in stdout while the payload was an `echo`, which prints its own arguments — that reported two contexts as broken when the parser was right.

The comment-start set is enumerated the same way: the separator branch empties the buffer for one group of metacharacters and the redirect branch leaves a space for another, so parentheses are the **whole remainder**, not two more cases. The predicate is approximate on both edges and only ever declines to fold, reverting to the pre-existing split.

## Tests

Both tables ship as class tests, so a further instance cannot pass silently. One more locks a coupling that until now held by coincidence: a sigil cannot be forged across a comment because the scanner and the sigil reader share a blind spot, and fixing either alone would reopen it. Asserted as an **implication**, not an equality — the scanner is legitimately stricter, and an equality version failed on a pair where nothing was wrong.

Non-vacuity proven by mutation on temp copies, never the worktree: 8 assertions RED on `main`; the comment ones RED on the broken intermediate (their reference is not `main`, which never had the fold); the class test RED on the round-2 intermediate; the coupling invariant **0** violations here, **30** on the round-1 intermediate, and correctly silent on the round-2 one — a test that fires on everything is as useless as one that fires on nothing.

## Measured

15,468 unique real commands from 401 transcript files: **2.18%** segment differently, **2.59%** carry an unquoted backslash. Guard-verdict replay of every changed command through the real `protected_paths_guard` and `destructive_command_guard`: **0 verdict changes in either direction**, with a known-positive control flipping 2/2 — an unflipped corpus and a dead harness are indistinguishable without one.

**Stated limits, not buried:** this corpus cannot contain the comment shape, so the differential is the realism check and the enumeration is the coverage argument. The merge-gate replay found 0 flips but has low power — one command in the set carried a `git merge` at all, and that was the injected control; it rests on the structural fact that the UNKNOWN check only fires for merge segments.

1430 tests across every consumer of `parse_segments`; the 48-entry golden lock passes untouched. Two harness bugs were caught before they produced a false number, both of which would have read as clean: comparing exception text that embeds the module name, and comparing a sentinel whose `str()` is its address.

## Not bundled

`destructive_command_guard._CONTINUATION` carries the same comment blind spot — filed as its own item with the measurement recorded, since two gate parsers do not belong in one diff.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved command parsing for escaped characters, separators, quotes, comments, and shell line continuations.
  - Multi-line commands are now evaluated as a single command where appropriate.
  - Escaped literals and command boundaries are recognized more accurately.
  - Improved handling of redirects, nested substitutions, heredocs, and trailing backslashes.
  - Commands following comment lines remain correctly separated.
  - Safety checks consistently identify continued force-push commands while allowing benign commands such as `git status`.
  - Protected-path checks now detect deletions hidden behind supported command prefixes and conservatively reject unresolved or ambiguous cases.


WingedGuardian:
@codex review

WingedGuardian:
**Automated first-pass review (structural + PII/secrets/wording).** Note: `pull_request_review_write` with `REQUEST_CHANGES` was rejected by GitHub ("cannot request changes on your own pull request" — this session and the PR author share an account), so this is posted as a regular comment. Treat the finding below as a REAL BLOCK, not an informational note.

## Structural review (Domain 1)

CLEAN — no ERROR/WARNING/NOTE findings.

- No banned patterns (`nohup`, direct `task_states` inserts, `/tmp/`, `git add -A`, `git push --force`/`git reset --hard` *as executed commands*, unvalidated `os.killpg`, raw SQL). The literal strings `git push --force`, `rm -rf`, etc. in `tests/test_hooks/test_shell_parse.py` are parser test fixtures (inputs fed to `sp.split_segments`/`sp.analyze`), not commands the code runs — not a violation.
- No bare `except:`, no mutable default args, no circular/star imports, no import-grouping issues in the diff.
- No new imports added to `scripts/hooks/shell_parse.py` — hook stays stdlib-only.
- Coverage: the modified function (`parse_segments`) and its callers get a large, purpose-built test class (`TestUnquotedBackslash`, 15 tests incl. two bash-derived parametrized tables) in the same PR. `has_top_level_pipe`'s docstring-only change needs no new test.
- Docs: docstrings updated in lockstep with the behavior change (`parse_segments`, `has_top_level_pipe`).
- Commit message and PR title both carry `fix(hooks):` — conventional-commit prefix present.

## PII/Secrets scan (Domain 2, code diff)

CLEAN — no hard blocks or soft warnings. No keys/tokens/secrets, no private IPv4/IPv6 ranges, no connection strings. Paths used in tests/docstrings are generic (`/var/lib`, `/var/foo`, `/wt`, `<protected>`) — no usernames, real hostnames, or install-identifying values. Commit author is the `users.noreply.github.com` address — ignorable per the known false positives.

## PR metadata / wording scan (Domain 2) — WORDING (CRITICAL)

The PR description's **"Not bundled"** section states, verbatim:

> `destructive_command_guard._CONTINUATION` carries the same comment blind spot — filed as its own item with the measurement recorded, since two gate parsers do not belong in one diff.

Read together with the preceding "What the review changed" section — which walks through, in reproducible detail, exactly how folding a backslash-continuation inside a `#` comment deletes a real command separator, makes a following command (e.g. a protected-path delete) invisible to the guard, and can forge an override-sigil acknowledgement — this PR body doesn't just describe a fix, it **discloses a live, unpatched bypass mechanism for `destructive_command_guard`, with the exact trigger shape (backslash-newline inside a comment), in a public repository.** The guard this describes is still vulnerable; nothing in this diff closes it.

This falls under "Advertises vulnerability fixes with enough detail to be exploitable" — and is arguably worse, since the disclosed technique applies to a guard that is *not yet patched* by this or any other PR. Once merged, this description sits permanently in public git history as a how-to for defeating `destructive_command_guard` until the follow-up lands.

Suggested fix (does not require touching the code): before merging, either (a) move the "Not bundled" disclosure and the mechanism walkthrough for the sibling guard into a private tracking issue rather than the public PR body, or (b) generalize that paragraph to say a related guard has a tracked gap without restating the comment-fold trigger shape, and land the sibling fix in the same window rather than leaving the detail public with the hole still open.

This is a wording-only finding — the code and tests in this diff are sound and well-covered; nothing here suggests reverting the fix itself.

---
Both review passes complete. Blocking on the wording issue above.



---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
## Triage 2026-09-14 — confirmed needs-rework; converted to draft

Batch triage of all open PRs ≤1700. The existing `needs-rework` label was re-derived rather than trusted, and it holds: 6 unresolved P1 + 1 CodeRabbit Critical/Major at head (review score 8.0 against the critical-lane threshold of 1.0), plus a conflict with main. Worst finding load in the batch. Stays draft until a builder session takes it.

PR #1620, opened 2026-09-03
fix(memory): enforce the wing vocabulary on the write path and at its source

WingedGuardian:
## Problem

`wing` is a controlled vocabulary (`taxonomy.py` `WINGS`, 12 members) but the
write path only ever tested falsiness — `if not wing or not room:` is an or-fill,
never a membership check. Any string reached the FTS5 `wing:` tag, the Qdrant
payload and `memory_metadata.wing`. And `classify_life_domain()` returns
`"personal"` for an unknown wing, so one bad value also corrupts a **derived**
field.

The read side had already been patched to filter junk wings out of the L1 view.
The write side never was — the symptom was hidden rather than prevented.

**Measured on the live store (read-only): 18 rows already carry an
out-of-vocabulary wing** — `architecture` ×11, plus `wing=channels`,
`technology`, `strategy`, `outreach`, `identity`, `genesis`, `cooperation`.

## Fix — three levels, because the first attempt closed only one

**Both agent-facing MCP doors.** `memory_synthesize` takes the same `wing` into
the same store and was initially left unguarded. That turned out to matter more
than it looked: pulling provenance on the 18 corrupt rows, **17 have
`dream_cycle_run_id IS NULL`** — they came through the MCP boundary, *not* the
synthesis path the coercion tier had been justified by. Guarding one tool and not
its sibling would only have moved the door.

**`MemoryStore.store()` coerces** rather than raising, because live callers pass
model output and one bad token must not abort a synthesis run. The two sibling
controlled-vocabulary fields in that same function (`life_domain`,
`origin_class`) *do* raise, so the divergence is now stated in the code instead
of left for a maintainer to "fix" in one direction.

**`dream_cycle` at the source.** `.get("wing", default)` falls back only on a
*missing* key, so an invalid wing beat the cluster's own known-valid one,
survived to the store, got dropped, and was re-classified from content — whose
terminal fallback is `general/uncategorized`, the exact backlog the `WING_AUDIT`
surplus task exists to drain. The guard could have **fed the pile it was added to
protect**. The synthesis prompt now enumerates the legal values instead of asking
for `"<wing>"`; the guard is a backstop, the prompt was the cause.

## What is deliberately *not* enforced

`room`. Measured: **14,667 of 81,343 rows (18%)** have a room outside
`ROOMS[wing]`, and the shipped ego prompts instruct `room="ego"`. Enforcing
symmetry would coerce a fifth of all writes and break the ego. `room` is
descriptive — one consumer, no read filter, no FTS tag, corrupts no derived
field. That reasoning is now a comment beside the guard so the next reviewer
doesn't "complete" the symmetry.

## Verification

**7 mutations RED, zero survivors**, covering both directions — guard absent
*and* guard over-broad (a guard that also drops *valid* wings), with the
`memory_synthesize` guard mutated on its own call site after the first sweep's
anchor was found to have matched `memory_store`'s.

| mutation | result |
|---|---|
| store guard removed / over-broad | RED / RED |
| MCP raise removed / over-broad | RED / RED |
| synthesize guard removed | RED |
| dream_cycle invalid-wing-wins-again | RED |
| dream_cycle prompt unenumerated | RED |

164 tests pass; `ruff check src/ tests/` clean. The acceptance replay is the
original defect: `memory_store(wing="portfolio")` — the exact value a background
session tried — now raises with the valid set.

## Not in scope

The 18 existing rows are **data repair**, not part of a write-path fix. Tracked
as a follow-up for a numbered migration that reconciles `memory_metadata`, the
Qdrant payload and the FTS tag together and re-derives `life_domain`. Note the
dashboard Memory tab does not filter to `WINGS`, so those values stay visible
until that lands.

## Summary by CodeRabbit

* **New Features**
  * Added controlled validation for memory “wing” values across storage, recall, and synthesis workflows.
  * Added clearer errors for invalid recall inputs and automatic fallback behavior for unrecognized values.
  * Prompts now list the supported wing vocabulary to improve consistent classification.

* **Bug Fixes**
  * Prevented invalid wing values from being stored or propagated.
  * Preserved valid explicit wings during automatic classification.
  * Ensured invalid synthesis results fall back to the appropriate cluster wing.

### DIFF
--- a/src/genesis/mcp/memory/core.py
+++ b/src/genesis/mcp/memory/core.py
@@ -732,6 +732,31 @@ async def memory_expand(
     return results
 
 
+def _validate_wing(wing: str | None) -> None:
+    """Reject an out-of-vocabulary wing at the AGENT-FACING boundary.
+
+    Every MCP tool that accepts a `wing` must call this. A caller that can be
+    told the valid set should be told it: a silent coercion here teaches
+    nothing, and this parameter has been reported as "not exposed" by an agent
+    that never tried it. `MemoryStore.store()` coerces instead — see the
+    comment there for why the two tiers differ.
+
+    Measured 2026-09-02 on the live store: 18 of ~81k rows carry an
+    out-of-vocabulary wing, and 17 of those 18 have `dream_cycle_run_id IS
+    NULL` — they came through THIS boundary, not the LLM-internal synthesis
+    path. Guarding one tool and not its sibling just moves the door.
+    """
+    if wing is None or not wing.strip():
+        return
+    from genesis.memory.taxonomy import WINGS
+
+    if wing.strip() not in WINGS:
+        raise ValueError(
+            f"invalid wing {wing!r}; expected one of {sorted(WINGS)}. "
+            "Omit `wing` to let it be classified automatically."
+        )
+
+
 @mcp.tool()
 async def memory_store(
     content: str,
@@ -761,6 +786,7 @@ async def memory_store(
     memory_mod = _memory_mod()
     memory_mod._require_init()
     assert memory_mod._store is not None
+    _validate_wing(wing)
     # WS-3: dispatched sessions carry a session-level origin via env
     # (GENESIS_SESSION_ORIGIN, stamped by CCInvoker). Forward it so an
     # external-influenced session's writes stop landing first_party via the
@@ -1379,6 +1405,7 @@ async def memory_synthesize(
     memory_mod = _memory_mod()
     memory_mod._require_init()
     assert memory_mod._store is not None
+    _validate_wing(wing)
 
     resolved_tags = list(tags or [])
     if "synthesis" not in resolved_tags:
--- a/src/genesis/mcp/memory/core.py
+++ b/src/genesis/mcp/memory/core.py
@@ -95,7 +95,13 @@ async def memory_recall(
         compact: If True, return lightweight previews only (memory_id, preview,
             score, wing, room, memory_class, source). Use memory_expand to
             fetch full content for specific IDs. Saves tokens and ~500ms.
-        wing: Filter results to this structural domain (e.g., "infrastructure").
+        wing: Filter results to this structural domain. Controlled vocabulary —
+            one of: autonomy, career, channels, dev_workflow, employment,
+            general, infrastructure, integrations, learning, memory, research,
+            routing. Enumerated here, as `life_domain` below is, because this
+            docstring IS the schema the caller sees: an unlisted value is
+            refused, and a caller that can be told the valid set should not
+            have to learn it by failing a call.
         room: Filter results to this topic within a wing.
         life_domain: Filter by life domain: "personal", "employment", or "genesis".
         include_graph: If False, skip graph traversal (saves ~500ms per call).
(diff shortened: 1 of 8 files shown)


WingedGuardian:
Automated first-pass review (structural patterns + PII/secrets/wording).

**Structural — WARNING**

1. **Import hygiene — inconsistent import placement for `WINGS`.** `src/genesis/memory/store.py:19` imports `WINGS` at module top-level (correctly grouped with the other local imports). But this PR introduces three more consumers of the same symbol as **function-local** imports instead:
   - `src/genesis/mcp/memory/core.py` — inside `_validate_wing()` (`from genesis.memory.taxonomy import WINGS`)
   - `src/genesis/memory/dream_cycle.py` — inside `_build_synthesis_prompt()` and again inside `_valid_wing_or()` (two separate local imports of the same symbol in the same file)

   `store.py`'s successful top-level import of `taxonomy.WINGS` suggests there's no circular-import constraint forcing the other three into function scope. If there's no cycle, these should probably be one top-level import per file rather than three scattered local ones. If there *is* a cycle (e.g. `dream_cycle.py`/`mcp/memory/core.py` importing `taxonomy` at module load time causes a real circularity that `store.py` doesn't hit), worth a one-line comment saying so — `store.py`'s neighboring `derive_origin_class` import already carries exactly that kind of "deferred import matches this function's cycle-avoidance style" note, and these three don't.

2. **Behavioral edge case — the two validation tiers don't share the same normalization.** `MemoryStore.store()` runs `wing` through `_strip_kv_prefix()` before checking membership in `WINGS` — that helper exists specifically because models echo the literal `wing={wing}` / `wing=channels` form from the synthesis prompt template into their output (per its own docstring). `_validate_wing()` in `core.py` (the new agent-facing MCP guard) only does `.strip()`, not `_strip_kv_prefix()`, and it runs *before* `store()` is ever reached. So a value shaped like `"wing=career"` reaching the MCP boundary directly (as opposed to via the internal `dream_cycle` → `store()` path) now raises `ValueError` instead of being silently normalized to `"career"` the way it would be if it reached `store()` first. This may well be intentional — the PR's own rationale for the two-tier design is "a caller that can be told the valid set should be" — but the leaked-`key=value` failure mode is specific to the synthesis prompt template, not something an MCP-calling agent is likely to reproduce, so it's probably not the case this asymmetry was written for. Worth a one-line confirmation that this is deliberate rather than an unconsidered gap between the two guards.

**Missing coverage / documentation:** none — every changed function has a docstring, and all three enforcement points (`core.py`, `dream_cycle.py`, `store.py`) got corresponding test coverage in the same PR.

**Genesis conventions:** PR title carries a proper `fix(memory):` conventional-commit prefix. No `IntervalTrigger`, no `$CLAUDE_PROJECT_DIR` misuse, no hook-script dependency issues.

No `nohup`, no direct `task_states` inserts, no `/tmp/` usage, no raw SQL against `genesis.db`, no bare `except:`, no mutable default arguments, no `os.killpg`, no force-push/reset patterns — none of the ERROR-tier structural patterns are present.

**PII/Secrets/Wording: CLEAN** — no keys, tokens, credentials, private IPs, IPv6 ULA addresses, SSH key material, or connection strings in the added lines. No personal emails, path/username leakage, location-revealing timezones, cloud account IDs, or hardware identifiers either. PR title and body discuss a data-quality/taxonomy-enforcement bug (invalid `wing` values reaching the memory store) — this is an internal correctness fix, not language describing a PII/secrets leak, security incident, or exploitable infrastructure detail.

VERDICT: PASS



---

WingedGuardian:
@codex review

PR #1626, opened 2026-09-03
feat(routing): wire Zhipu GLM-4 Flash as a free provider ahead of paid fallbacks

WingedGuardian:
## What

Wire Zhipu GLM as a free routing provider: a new `zhipu-glm-flash` provider on
Zhipu's **general** OpenAI-compatible endpoint (`api/paas/v4`) serving the
published free `glm-4-flash` model, inserted into 12 call-site chains
immediately ahead of the paid DeepSeek fallbacks. Plus: the unused `"glm"` key
in the delegate's `_TYPE_TO_PREFIX` renamed to `"zhipu"` (so standard
`{TYPE}_API_KEY` env resolution finds the shipped `ZHIPU_API_KEY` convention
with no install-local aliases), a `model_profiles` entry, config/delegate test
pins, models.md + CHANGELOG.

## Why (measured)

Paid DeepSeek absorbs ~2,000 calls/day of free-tier overflow. Zhipu is a free,
vendor-independent lane that was already reachable: live probes from this
branch measured **200 in ~3.2s with clean bare JSON** on `glm-4-flash` at both
general endpoints. `glm-4.5-flash` was rejected for this lane on the same
probes — reasoning-by-default (10–18s) and fence-wrapped JSON even with
thinking explicitly disabled.

## Placement rule (exact, and deliberately conservative)

Inserted **immediately before the first paid DeepSeek member, only where that
member is not the chain lead** (12 of 17 candidate chains; the other 5 lead
with paid DeepSeek deliberately — quality choices, untouched, including the
invariant-locked `judge`). Not promoted above the free providers that precede
the paid fallback; pre-existing free last-resort tails below the paid members
keep their positions. Promotion beyond this waits on a **measured success rate
over real fallback traffic** — the acceptance measurement is tracked and
includes verifying the published ~1K req/day cap and billing against the
account.

## Notes

- `ZHIPU_API_KEY` is shared with the CC roster's glm failover, so it must be a
  general-endpoint key. A coding-plan-only key 401s this lane gracefully (the
  chain continues past it). Documented at the provider block, models.md, and
  the CHANGELOG. A per-provider `api_key_env` override was considered and
  deferred — the graceful 401 bounds the cost, and the override is new surface
  this PR does not need.
- `rpd_limit: 1000` is a forward declaration for the daily-budget ledger
  (sibling PR #1624) — verified silently ignored by this tree's parser until
  that lands. `rpm_limit: 20` is a provisional self-throttle, marked as such.
- Installs without the key are unaffected: the provider auto-disables
  (`has_api_key=False`), chains intact.

## Review + verification

Adversarial pre-push review (fresh-context architect) applied in-branch:
shared-key documentation, overclaim rewording at three sites ("never above
other free providers" was false in 4 chains — the true rule is now stated),
provenance added to every measured claim. One reviewer finding was refuted by
source and recorded as such. Verify-RED on both new test groups (config pins
RED against the pre-change yaml; the delegate `zhipu` row RED against a
`glm`-keyed map), restores hash-verified. 106 targeted tests green; 12/12
chain insertions verified by re-parse with no other chain touched; rename
blast radius enumerated (zero `type: glm` users repo-wide); ruff +
subsystem-map guards clean.

## Summary by CodeRabbit

* **New Features**
  * Added Zhipu GLM-4 Flash as a free, low-latency model option for structured-output workflows.
  * Added GLM-4 Flash as a fallback across multiple routing workflows, before paid DeepSeek alternatives.
  * Routing automatically skips the provider when no compatible API key is configured.

* **Documentation**
  * Documented GLM-4 Flash capabilities, limitations, availability, endpoint requirements, and free-tier usage terms.

* **Bug Fixes**
  * Improved compatibility with Zhipu’s OpenAI-compatible API format.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,19 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **GLM-4 Flash as a free routing provider (Zhipu direct).** A new
+  `zhipu-glm-flash` provider reaches Zhipu's published free model on the
+  general OpenAI-compatible endpoint using the standard `ZHIPU_API_KEY`. It
+  sits immediately ahead of the paid DeepSeek fallbacks in 12 chains — never
+  ahead of a deliberately-paid lead, and not promoted above the free
+  providers that precede the paid fallback — so overflow that previously
+  went straight to paid models now tries a free, vendor-independent lane
+  first. Installs without the key are unaffected (the provider
+  auto-disables, as always); installs that set `ZHIPU_API_KEY` for the CC
+  roster's glm failover share that key here, so it must be valid on the
+  general endpoint (a coding-plan-only key just 401s this lane —
+  the chain continues past it).
+
 - **Telegram ping when someone replies to a marketing pitch.** When a real person
   replies to one of Genesis's cold marketing emails, you now get one brief
   Telegram notification — the sender and the first line of their reply.
--- a/config/model_profiles.yaml
+++ b/config/model_profiles.yaml
@@ -65,6 +65,34 @@ profiles:
     last_reviewed: "2026-08-19"
     review_source: nvidia_nim_live_probe
 
+  glm-4-flash:
+    display_name: "GLM-4 Flash"
+    provider: zhipu
+    api_id: "glm-4-flash"
+    intelligence_tier: B
+    reasoning: C
+    instruction_following: B
+    anti_sycophancy: C
+    context_window: 128000
+    cost_tier: free
+    cost_per_mtok_in: 0.0
+    cost_per_mtok_out: 0.0
+    latency: low
+    strengths:
+      - fast responses (~3s on live probe)
+      - clean bare JSON without fencing (live probe)
+      - free-tier daily capacity independent of every other vendor
+    weaknesses:
+      - small-model reasoning depth
+      - quality beyond smoke-probe UNMEASURED on real Genesis workloads
+    best_for: [fact_extraction, outcome_classification]
+    avoid_for: [deep_reflection, code_work, math_heavy]
+    free_tier:
+      available: true
+      limits: "published ~1K req/day (unverified against account)"
+    last_reviewed: "2026-09-03"
+    review_source: zai_general_endpoint_live_probe
+
   glm-5.1:
     display_name: "GLM-5.1"
     provider: zenmux
--- a/config/model_routing.yaml
+++ b/config/model_routing.yaml
@@ -242,6 +242,23 @@ providers:
     # Not promoted above the free providers that precede the paid fallback,
     # and never ahead of a deliberately-paid lead, until a real success rate
     # is measured on live fallback traffic.
+    # AND NOT IN A CHALLENGE CHAIN AT ALL, at any position. The placement rule
+    # above is about COST ORDER; a challenge site has a second requirement the
+    # cost rule cannot see. Its entire job is to DISAGREE with a verdict another
+    # model already produced — `dream_cycle_entity_challenge` deprecates a
+    # memory on agreement, `entity_adjudication_challenge` irreversibly merges
+    # two entities, and `20_adversarial_counterargument` is the executor's
+    # devil's advocate. Anti-sycophancy is the load-bearing capability at all
+    # three, and this profile rates it C with quality UNMEASURED on real
+    # workloads (config/model_profiles.yaml: glm-4-flash). A challenger that
+    # agrees is not a challenger: it silently turns a two-model gate into a
+    # one-model gate while still looking like two.
+    # The profile already said so — `avoid_for` names deep reflection and
+    # `best_for` is fact_extraction / outcome_classification — so this is the
+    # model's own stated fit, not a new policy invented here. Codex named the
+    # two entity sites (P1, PR #1626); the executor's devil's advocate is the
+    # same class and was not named. `TestChallengeChainsRejectWeakChallengers`
+    # pins the rule for whatever provider is added next.
     # rpd_limit is the published free-tier daily cap (~1K req/day, not yet
     # verified against this account). No consumer exists in THIS tree — it
     # is a forward declaration for the daily-budget ledger (built on a
@@ -411,7 +428,6 @@ call_sites:
     chain:
     - openrouter-gpt55
     - nvidia-nim-deepseek
-    - zhipu-glm-flash
     - openrouter-deepseek-v4
     - openrouter-qwen36plus
     default_paid: true
@@ -484,7 +500,6 @@ call_sites:
   dream_cycle_entity_challenge:
     chain:
     - nvidia-nim-deepseek
-    - zhipu-glm-flash
     - openrouter-deepseek-v4-flash
     retry_profile: background
   # Entity-NODE merge-vs-distinct adjudication (entity_adjudication drainer).
@@ -504,7 +519,6 @@ call_sites:
   entity_adjudication_challenge:
     chain:
     - nvidia-nim-deepseek
-    - zhipu-glm-flash
     - openrouter-deepseek-v4-flash
     retry_profile: background
   # One-shot supervised wing backfill (scripts/wing_backfill.py) — batch
--- a/config/model_routing.yaml
+++ b/config/model_routing.yaml
@@ -220,6 +220,43 @@ providers:
     base_url: https://zenmux.ai/api/v1
     free: false
     open_duration_s: 120
+  zhipu-glm-flash:
+    # Zhipu/Z.AI direct, GENERAL endpoint (api/paas/v4) with the standard
+    # ZHIPU_API_KEY — the type's env resolution finds it by convention.
+    # NEVER point this at a coding-plan endpoint or key: Z.AI's coding-plan
+    # surface is restricted to approved tools and misuse risks the account.
+    # glm-4-flash is Zhipu's published free model (live probe at this
+    # endpoint: 200, clean bare JSON, ~3s; glm-4.5-flash was rejected for
+    # this lane — reasoning-by-default, 10-18s, fence-wrapped JSON even with
+    # thinking disabled). Billing-vs-free unverified against an account
+    # statement — measure alongside the success-rate acceptance below.
+    # NOTE: ZHIPU_API_KEY is SHARED with the CC roster's glm failover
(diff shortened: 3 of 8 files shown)


WingedGuardian:
@codex review

WingedGuardian:
## Automated first-pass review — PR #1626

**Scope:** 7 files (`CHANGELOG.md`, `config/model_profiles.yaml`, `config/model_routing.yaml`, `docs/reference/models.md`, `src/genesis/routing/litellm_delegate.py`, `tests/test_routing/test_config.py`, `tests/test_routing/test_litellm_delegate.py`).

**ERROR:** none.

**WARNING:** none.
- Import hygiene: no new/changed imports.
- Coverage: the `litellm_delegate.py` `_TYPE_TO_PREFIX` rename (`"glm"` → `"zhipu"`) is covered by a new case in `test_litellm_delegate.py::test_build_model_string`; the new `zhipu-glm-flash` provider and its chain placement are covered in `test_config.py::test_load_full_yaml`.
- Rename blast radius: verified independently (repo-wide code search + full-file read of `config/model_routing.yaml`) — no remaining `type: glm` provider entries and no code path keys off `provider_type == "glm"`. Other `"glm"`/`"GLM"` hits in the tree are unrelated model-name strings (`glm-5.1`, `glm51`, `glm-5.2`) or comments about a past reviewer nicknamed "GLM cross-model reviewer" — not this rename's concern.
- Genesis conventions: conventional commit prefix present (`feat(routing): …`); no hook/IntervalTrigger/`$CLAUDE_PROJECT_DIR` concerns (none touched).

**NOTE:** none — no new public functions/classes; the one code change is a single dict-value rename with an existing inline comment.

**Verified claim:** traced `RoutingConfig` loading (`src/genesis/routing/config.py::_parse`) directly — `cfg.providers` excludes only `enabled: false` entries (`github-o3mini`, `deepseek-chat`) and `lmstudio-30b` (test monkeypatches `GENESIS_ENABLE_LM_STUDIO=false`). Counting the full provider list in `model_routing.yaml` gives 29 total − 3 disabled = 26, matching the updated `assert len(cfg.providers) == 26`. The `38_procedure_extraction` chain-order assertion also matches the actual YAML (`zhipu-glm-flash` at index 1, immediately followed by `openrouter-deepseek-v4-flash`).

**HARD BLOCK:** none. No API keys/tokens, private IPs, SSH keys, DB connection strings, or secret env values in the added lines — only the env var *name* `ZHIPU_API_KEY` (no value) and public API endpoints (`api.z.ai`, `open.bigmodel.cn`).

**SOFT WARNING:** none — no personal emails, username-bearing paths, location-revealing timezones, cloud account IDs, hardware identifiers, or private hostnames in the diff.

**WORDING (CRITICAL):** none. PR title/body and the CHANGELOG entry describe adding a routing provider and document a shared-key operational caveat — no leak/incident language, no private-infra naming, nothing exploitable.

PII/Secrets/Wording: CLEAN

VERDICT: PASS



---

WingedGuardian:
## Batch triage 2026-09-14 — marked

CONFLICTING against main, which suppresses the entire `pull_request` CI suite. Findings at last read: 1 unresolved [P1] + 2 CodeRabbit Critical/Major (score 5.5, STANDARD lane) — the P1 and the Majors each block at the always-fix floor independently of score.

Attention needed: merge main in, resolve the floor findings, re-request review. No design question identified.

Labelled `needs-architecture-session` under the batch rule (merge as-is, or mark). No code changes; branch untouched.

WingedGuardian:
@codex review

WingedGuardian:
Returning this to draft for rework, and separately: **the model choice in this PR is right, and should not be changed.** That second part is the useful half, because it was about to be changed on my recommendation.

**The model question, settled by measurement.** The obvious criticism of this PR is that `glm-4-flash` is two generations old and Zhipu's current free Flash tier is newer, so it should be retargeted. The vendor's own pricing table supports that: `GLM-4.7-Flash`, `GLM-4.5-Flash` and `GLM-4.6V-Flash` are free; `GLM-5.3-Flash` is not (the "Limited-time Free" beside it is the cached-input-storage column).

Live-probed against the general endpoint this PR uses, same prompt, same key, asking for a bare JSON object:

| model | HTTP | `content` | `reasoning_content` | note |
|---|---|---|---|---|
| `glm-4-flash` | 200 | `{"ok":true}` | empty | 7 completion tokens, 1.6s |
| `glm-4.7-flash` | 200 | **EMPTY** | populated | the answer went to reasoning |
| `glm-4.7-flash` (retry) | **429** | — | — | rate-limited after ONE prior call |
| `glm-4.5-flash` | 200 | `{"ok":true}` | populated | reasoning-by-default |

`glm-4.7-flash` returns an empty `content` — the answer lands in `reasoning_content`, so any caller reading `choices[0].message.content` gets nothing. That is the same failure class the PR's own comment gives for rejecting `glm-4.5-flash` (reasoning-by-default, fence-wrapped JSON), only worse. And its free tier 429'd after a single call, where `glm-4-flash` answered repeatedly without complaint.

So the newest free model is the wrong model for THIS lane, and the author's probe-driven choice stands. Two caveats on my own numbers: the 429 stopped me characterising 4.7-flash further, so "empty content" rests on one observation rather than a series; and this key is shared with the CC roster's live glm failover, so I did not keep probing. Worth re-measuring rather than trusting, but the direction is clear enough not to retarget on a pricing page.

**Why the PR goes back anyway: 12 findings, none replied, and no convergence.** Three Codex rounds produced 4, 4 and 4. Every one is the same shape — the new provider is not gated out of, or registered in, one more subsystem:

round 1 — destructive merge gates · provider-key health reporting · two Majors (billing confirmation, a fixture with no `ZHIPU_API_KEY`)
round 2 — capability grades for challengers · provider-type-to-key mapping · distillation primary · the declared daily request limit
round 3 — autonomous execution gates · the existing `glm` alias · the parsed Zhipu key registry · general-vs-coding-plan credential separation

**The generator is architectural, and it was confirmed by enumeration rather than inferred.** There is no eligibility predicate anywhere in routing: chain membership in `config/model_routing.yaml` IS eligibility. `ProviderConfig` carries no notion of a provider being unmeasured, and `ModelProfileRegistry.match()` — which implements exactly the grade filtering round 2 asked for, with `min_reasoning` / `min_anti_sycophancy` / `exclude_providers` — has **no caller in the routing path**; every consumer is display or cost only. So each round the reviewer finds one more place a new provider should have been held out of, and there is no single place to hold it.

Adding a provider today means hand-editing around nine independent sites. That is why this cannot be patched to green: the findings are not a list, they are a sample.

**What a builder should decide.** Whether to add the missing predicate — one flag on `ProviderConfig`, enforced in the chain-walk skip block where `has_api_key` already is, paired with the set of call sites that require a measured provider (`essential.ESSENTIAL_CLOUD_SITES` is the existing precedent for such a set) — or to keep this PR to chain placement and accept the remaining findings explicitly. The first closes the class for every future provider; the second ships sooner and leaves the next provider to rediscover all twelve.

Note also that the `rpd_limit` finding cannot be resolved here at all: that field is parsed nowhere today and is introduced by #1624, which is still in flight. A `rpd_limit:` key in this PR's YAML is silently dropped.

**What carries forward.** The model choice, now measured twice. The deliberate exclusion from challenge chains, which the PR's comment argues correctly on anti-sycophancy grounds. The placement rule relative to paid DeepSeek members. And the warning about the shared `ZHIPU_API_KEY` and the coding-plan endpoint, which is a real operational hazard worth keeping verbatim.

No action needed from review.

WingedGuardian:
## Sent back for rework

Converting to draft. This is the one PR in batch 1 that meets the rework bar rather than the polish bar, and the reason is the finding *distribution*, not the count.

**14 live findings, and 8 of them are one class:** a new provider is not registered or gated in the surrounding machinery — provider-key health reporting, the parsed key registry, the capability-grade filter, the daily request limit, the delegate alias, credential separation, and three separate chain-placement findings (merge gates, autonomous execution, distillation primary).

That class has a single cause, and it is architectural: **there is no chokepoint for adding a provider.** Chain membership in `config/model_routing.yaml` *is* eligibility — no predicate, no capability check. A new provider today must be hand-added at roughly nine independent sites with nothing verifying they agree:

`config/model_routing.yaml` (provider block + each `chain:`) · `config/model_profiles.yaml` · `litellm_delegate.py` `_TYPE_TO_PREFIX` · `api_keys.py` `build_key_validator` · `api_keys.py` `known_keys` · `provider_health.py` `_PROVIDER_URLS` · `dashboard.js` `_KEY_TO_PROVIDER_TYPES` · `secrets.env.example` · `onboarding/floor.py`

Patching the eight findings individually leaves the generator intact, and the next provider re-earns the same list.

**Also carried, so the next session does not re-derive it:**

- **The model target is stale.** `GLM-4-Flash` is not in Z.ai's current pricing table at all. The permanently-free text models are **GLM-4.7-Flash** and GLM-4.5-Flash (GLM-5.3-Flash is *not* free — the "Limited-time Free" beside it is the cached-input-storage column). Retarget before anything else.
- **A cheaper design was checked and rejected on evidence.** ZenMux publishes `z-ai/glm-4.7-flash-free`, and `zenmux` is already a registered provider type, which would have collapsed most of the registration class into one YAML block. Not viable here: `API_KEY_ZENMUX` is present but EMPTY, while `ZHIPU_API_KEY` is set. So the direct-Zhipu design is right; the retarget is what is wrong.
- **Main already has a Z.ai route** — `glm51` via zenmux, in three chains. Worth deciding whether this PR retires it, since its gateway key is empty.
- **The P1 has a cheap correct answer.** "Keep unmeasured GLM out of destructive merge gates" does not need a new routing predicate — take GLM out of those chains in the YAML. Same for the autonomous-execution and distillation-primary P2s. Chain placement, not a gate.
- **One P2 is answerable rather than fixable.** "Reject challengers with missing capability grades": the filter already EXISTS and is simply unwired — `ModelProfileRegistry.match()` supports `min_reasoning` / `min_anti_sycophancy` / `exclude_providers` and has no caller in the routing path. Wiring it is far outside this PR.
- **Stated limit on any Router-enforced gate:** four paths bypass the Router entirely (`contribution/version_gate.py`, `experimentation/standalone_router.py`, `eval/runner.py`, `observability/provider_health.py`), so a Router-side flag would not cover them. Check whether GLM lands in the `contribution_gate` chain before claiming coverage.

Everything else in batch 1 is being fixed in place and pushed through; this one goes back.

PR #1628, opened 2026-09-03
fix(memory): credit a related memory with its strongest connection

WingedGuardian:
## The defect

`_bfs_with_strength` scoped its best-edge comparison to **one expanding parent**, so a memory reachable from several equally-near parents was claimed by whichever parent the queue reached first — and reported *that* connection's strength. Queue order followed the loader's `SELECT`, which has no `ORDER BY`, and the edge it carried was frequently the weaker of the two.

Measured over the **whole** live graph at the real call parameters (`max_depth=2, min_strength=0.3`; 66,860 roots): **68,336 of 1,066,978 reported connections (6.40%) were understated**, touching **50.4%** of lookups. **None was ever overstated** — the tell that this was a selection artefact, not a modelling choice.

It matters because `strength` is both shown to the model *and* the key the consumers sort on before slicing the top few. An understated memory sinks in that order and can leave the visible set.

## The fix

Drain the whole depth level before committing, so candidates from every parent at that level are compared and the strongest wins.

## Measured, at full population — not sampled

| | before | after |
|---|---|---|
| top-five set churn, forward vs reversed row order | 3.09% | **0** |
| full-output churn under row order | — | **0** |
| reach-set changed | — | **0** |
| reported depth changed | — | **0** |
| connections raised | — | 68,336 / 1,066,978 (**6.40%**) |
| connections lowered | — | **0** |

**Denominators, because two of these differ by an order of magnitude.** 6.40% is over every connection the walk computes (~16 per lookup). Restricted to the handful that actually reach the model it is **0.16%** of surfaced connections, on **0.7%** of lookups. The blast radius at that surface is a third number again: the visible set changes on **1.94%** of lookups and its order on **8.15%**.

An earlier draft of this PR conflated the first two and read roughly 40× high on the surfaced surface. That is why every figure here is a population count with its denominator attached.

## Two things tried and NOT shipped

- **A total key on the final sort.** Redundant once the commit order is deterministic — a stable sort cannot reintroduce what it was handed — and measured at 0 difference either way across 1,447 roots. It would only have reordered ties gratuitously and widened divergence from the CTE fallback's documented `(depth, strength DESC)`.
- **A `link_type` tiebreak in the commit key.** The memory id alone already makes that key total, so it bought no determinism — but it was not neutral. Ordering equal strengths alphabetically by relationship type front-loads some relationships over others in the order the model reads. Measured across all 66,856 roots (13.8% carry a tie group): `succeeded_by` **−39.6%**, `preceded_by` −23.8%, `related_to` −13.1%, `categorized_as` **+32.8%**, `action_item_for` +31.8%. The within-pair comparison above it is genuinely a different case — there both candidates describe the same pair, so a type must be chosen and a stated rule beats an arbitrary one.

## A second consumer inherits this

`drift.py` reads no labels, but its local expansion feeds these ids to RRF **as a ranked list** (`drift.py:202`: *"order matters"*), so it inherits the determinism and its ranks shift accordingly — measured at ~5% of expansion roots, set unchanged. A strict improvement, but a real behavioural delta on a second live path, stated rather than left implicit.

## Verification

RED witnessed first, and the failure named the defect: `forward (0.4, …)` vs `reversed (0.8, …)`.

The new fixture is pinned in the **data**, not a comment, and **verified by replaying all four plausible scan orders** — rowid ascending and descending, PK-index ascending and descending — with at least one child mis-reported under every one. An earlier revision of it was pinned in a comment, and the comment had the alphabetical fact backwards, which made both new tests pass on the *old* code under an index scan. The fixture 120 lines above documents having fallen into that exact trap twice; this one now follows its lesson.

Independent adversarial verification ran all 66,833 roots plus a 4,000-case differential fuzz over random multigraphs (`max_depth` 0–4, `min_strength` 0.0–1.5, a `link_type_filter` matching nothing, 15% self-loops, four row permutations each): depth equivalence, `visited`-move safety, self-loop/cycle safety, filter equivalence and never-lowered all confirmed with 0 mismatches. Cost is not a regression — peak candidate-set 34 vs 21 at real parameters, 1.05× over 2,000 traversals; max out-degree is 23, so frontier growth is bounded.

177 tests across graph, drift, dream-centrality, dream-shield, dream-cycle, graph-expansion and both bake-off files. ruff and the four CI-reproducible guards clean.

## Documentation

The 60-line comment block this replaces described the defect as *deferred*; the test module's docstring asserted the property was absent. Both were false as of this change and are rewritten, along with two stale consumer line references and a follow-up pointer with nowhere left to point.

Follow-up: ab0d0c28651743c3876f6cae115c3ab0

## Summary by CodeRabbit

- **Bug Fixes**
  - Related-memory ranking now consistently selects the strongest connection when multiple equal-length paths exist.
  - Results are deterministic regardless of graph traversal or data insertion order.
  - Reachability remains unchanged.

- **Tests**
  - Added coverage for memories reached through multiple parent connections.
  - Expanded validation of deterministic ranking for tied parallel connections.

---

### Ordering note vs #1818 (added 2026-09-06)

#1818 moves `_bfs_with_strength` out of `graph.py` into a new
`graphstore_nx.py`, carrying the pre-fix per-parent version. This PR changes
that function where it currently lives, so the two produce a modify/delete
conflict.

**Either merge order is safe.** The two tests added here drive the public
`traverse()` facade, which #1818 keeps as the entry point — so a resolution
that dropped this fix would go red rather than pass quietly.

Preferred order is **this PR first**, for two non-safety reasons: it is
mergeable against `main` today with one gate outstanding, whereas #1818 sits
behind #1725; and landing it first lets #1818 move the already-fixed function,
which keeps its "pure behaviour-preserving move" claim — and its acceptance
bar — literally true.

If #1818 lands first instead, re-target this diff to `graphstore_nx.py`; the
tests need no change.


WingedGuardian:
@codex review

WingedGuardian:

**Scope:** `CHANGELOG.md`, `src/genesis/memory/graph.py`, `tests/test_memory/test_graph_parallel_edges.py` (3 files, +209/-52). Not doc-only — full structural + PII/wording review run.

### Domain 1 — Structural

No violations found against the tracked pattern list:
- No `nohup`, no `/tmp/` usage, no `git add -A`/`--force`/`reset --hard`, no bare `except:`, no mutable default args, no raw SQL against `genesis.db`, no `os.killpg`, no direct `task_states` inserts.
- Import hygiene clean: stdlib (`logging`, `time`, `dataclasses`) → third-party (`aiosqlite`, `networkx`) grouping preserved in `graph.py`; no star imports, no new circular imports introduced by dropping the `collections.deque` import.
- Test coverage: no new untested `.py` file — `_bfs_with_strength`'s behavior change in `src/genesis/memory/graph.py` is covered by two new tests in the same PR (`test_multi_parent_node_reports_the_strongest_reaching_edge`, `test_multi_parent_claim_does_not_follow_row_order`), plus updates to the existing tie-break test's docstring/scope note.
- Commit prefix: PR title carries `fix(memory):` — conventional prefix present.

One NOTE (non-blocking): `_bfs_with_strength`'s docstring (`src/genesis/memory/graph.py:118`, `"""BFS traversal with edge-attribute filtering."""`) wasn't updated to mention the new level-wide best-parent-wins / cross-run determinism guarantee — the inline comment block covers it in detail, but a reader skimming just the docstring wouldn't see it summarized there.

- Added lines scanned for private IPs (10.x/192.168.x/100.64-127.x), IPv6 ULA, API keys/tokens, SSH keys, DB connection strings: none found.
- No personal emails, home-directory paths, revealing timezones, cloud account IDs, or hardware identifiers in the diff.
- PR title/body and CHANGELOG addition describe a graph-traversal correctness bug (best-parent selection, row-order nondeterminism) with measured statistics — no mention of a leak, security incident, or specific private infrastructure. Reads as a standard bug-fix writeup, not exploitable/sensitive wording.

Structural review: no ERRORs, no WARNINGs (one NOTE above).

VERDICT: PASS



---

WingedGuardian:
@codex review

PR #1629, opened 2026-09-03
fix(health): a broken critical call site stays audible for weeks, not a day

WingedGuardian:
## What

Two-tier recency window for the dashboard's `callsite:down` detector: sites in
the curated `CRITICAL_CALL_SITES` set stay red for up to **35 days** after a
failed last run; every other site keeps the tight **24h** window.

## Why

The single 24h ceiling had an inversion for exactly the sites that matter
most. `call_site_last_run` holds one row per site (its latest run), so a
slow-cadence critical site — the weekly reflections — that failed its last
run left the alert set after a day and had no next attempt for ~a week:
**the longer something stayed broken, the quieter the dashboard got.** This
is the same architectural property that made a multi-day provider outage
inaudible, and the 24h constant's own comment anticipated it ("widen if
slow-cadence critical sites start under-surfacing" — that case arrived).

A blunt widen was deliberately rejected: the measured stale-noise rows that
motivated the original ceiling (one-off sites 600h+ stale) are all
NON-critical, so they keep the tight window and this widens nothing for them.

## Design

- 35 days mirrors the job-health never-succeeded floor
  (`crud/job_health.py` `recent_since` ~35d): wide enough for every cadence
  in the critical set (daily cycles, ~18h egos, weekly reflections), and
  still a **floor** — a critical site retired for over ~5 weeks stops
  alarming instead of alarming forever.
- The SQL fetches the wide window; the tight filter applies per row with the
  same string-compare semantics as the existing cutoff (both sides
  isoformat-with-offset — format equality verified against live rows).
- Dashboard-only surface by construction: `callsite:down` is never on the
  outreach escalation whitelist and is Sentinel-unmapped, so the blast
  radius of this change is the glanceable view only.

## Verification

Test-first: the real-DB expectation table was updated **before** the code
and observed RED (the stale critical failure absent from the alert set),
then green. It now pins all five cells: fresh-critical alerts,
fresh-non-critical alerts, **stale-critical stays audible** (the fix),
stale-non-critical excluded, ancient-critical floored out. A tier-inversion
mutant (the filter flipped to the wrong tier) dies on it, restore
hash-verified. Two mock fixtures with hardcoded stale timestamps —
previously passing only because the mock bypassed the SQL window — were
freshened to model what they claim. 107 tests green across the health
suites; ruff clean.

## Summary by CodeRabbit

- **New Features**
  - Critical call-site failures remain visible for up to 35 days after detection.
  - Non-critical call-site failures continue using the 24-hour detection window.
  - Critical failures remain marked as critical, while other failures retain warning status.

- **Documentation**
  - Added an Unreleased changelog entry describing the updated failure visibility windows.


WingedGuardian:
@codex review

WingedGuardian:

**Structural review: CLEAN**
- No pattern violations (nohup, raw task_states inserts, `/tmp/`, `git add -A`, force-push, bare `except:`, mutable defaults, etc.).
- Import hygiene fine — the two new imports (`CRITICAL_CALLSITE_DOWN_RECENCY_HOURS` in `errors.py`, and the constant's own module) follow the existing local-import-inside-try convention already used in that function.
- Test coverage: `tests/test_mcp/test_callsite_health.py` was updated to cover the new tier (fresh/stale/ancient × critical/non-critical), including a real-sqlite test proving the WHERE-clause + Python filter combination, not just mocks.
- Docstrings/comments: the new constant and the two-tier docstring in `critical_sites.py` are thorough and consistent with the code.
- Verified the "35 days mirrors the job-health floor" claim in the PR body against `src/genesis/mcp/health/errors.py:1034` (`job_health` never-succeeded query) — accurate, not just asserted.
- Verified the test fixtures' math against the actual filter (`errors.py:940-942`): `stale` (72h) sits inside the critical window (840h) but outside the tight window (24h) → correctly excluded for non-critical, retained for critical; `ancient` (864h) sits outside even the critical window → excluded by the SQL cutoff itself, not the Python filter. Matches the asserted expectations.
- Commit/PR title carries a conventional prefix (`fix(health): ...`).

- No credentials, tokens, private IPs, connection strings, or SSH key material in the diff.
- No personal emails, usernames-in-paths, or private hostnames introduced.
- PR title/body/CHANGELOG entry describe the behavior change in generic architectural terms (a dashboard-only alert's recency window) — no mention of a specific leak, incident details, or exploitable internal infrastructure.

No findings to report.

VERDICT: PASS




---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

PR #1350, opened 2026-08-09
docs(skill): spec-required-set discipline in genesis-development

WingedGuardian:
## What

Adds a discipline rule to the `genesis-development` skill: when a change must satisfy a canonical spec's **required-set** (a prompt's required JSON fields, a validator's mandatory keys, an allow-list), enforce the **whole set at once** and lock it with **one test** asserting the required block == the complete canonical set — not field-by-field.

Placed as the spec-facing corollary of the existing *"Read the reference implementation COMPLETELY before deriving from it"* bullet, plus a matching Anti-Rationalization row.

## Why

Origin: PR #1333 (deep-reflection salvage). `_SALVAGE_PROMPT`'s Required block was completed **one field at a time across three Codex review rounds** (cognitive_state_update → confidence → observations) for what `REFLECTION_DEEP.md` declares as one closed set. Deriving the full required-set upfront + a single locking test closes the axis in one round.

Docs-only change to the skill; no code paths touched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1200, opened 2026-07-22
docs(voice-master): anti-slop doc-level tells + generation guidance

WingedGuardian:
## What

Three doc-only additions to the `voice-master` skill, hardening its anti-slop coverage and generation guidance. No code touched.

### 1. SKILL.md — Generate step guidance (net-new)
Added the two missing workflow lines to the **Generate** step: first-person pronouns (I/my/me) are fine and read as honest, and exclamation marks are fine where they feel natural. This guidance already lived in `references/style-guide.md` (the human-writing reference) but was never surfaced as generation-workflow instruction, so a fast pass over SKILL.md missed it.

### 2. anti-slop.md — "deliverable shipped as raw .md" tell (net-new)
A finished deliverable handed back with visible markdown syntax (`#` headers, `**bold**`, `-` bullets) is a tell that a model generated it and nobody repackaged it into the format the context expects (.docx, PDF, deck, formatted email). Scoped to explicitly exclude cases where markdown *is* the medium (README, GitHub comment, chat).

### 3. anti-slop.md — document-level structural-tell checklist (generalized)
The existing "Structural tells" were sentence/paragraph-level, and the document-level ones existed only LinkedIn-scoped under "Banned Structural Patterns." Added a medium-agnostic checklist in the Universal section: meta-narration, exhaustive-coverage-as-virtue, equal-length sections, relentless parallelism, table overload, and absent point of view. Reconstructed from the concepts in Wikipedia "Signs of AI Writing" and the structural-fingerprint literature plus the existing banned list — no external guide purchased.

## Verification
- Doc-only; no `.py` changed, no tests to run.
- Diff scanned for private data (IPs, emails, real names, paths) — clean, synthetic examples only.
- Confirmed each addition was genuinely absent before adding (`grep` for "raw .md" = 0 hits; first-person/exclamation existed only in style-guide.md; doc-level tells were LinkedIn-scoped only).

Ledger: 93a4265b

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1651, opened 2026-09-03
fix(observability): tell a session its code moved, by observing HEAD

WingedGuardian:
## What this is

A Claude Code session's MCP subprocesses snapshot Genesis code at spawn and never
reload, and `genesis-server` holds imported code until restart. A deploy landing
mid-session therefore leaves recall — and its security read-exclusions — on the old
build. There is no auto-restart, so the remedy is a nudge, and one has shipped for a
while.

**It almost never fires.** It asks `commit_identity.is_stale`, which needs a deploy
recorded in `update_history` — written only by `scripts/update.sh`.

MEASURED on a live install, 30 days:

| | |
|---|---|
| HEAD movements | **46** |
| `update_history` success rows | **5** |
| coverage | **10.9%** |

Per-movement replay of both verdicts over the same history: the recorded-deploy
verdict fires **4/52 (7.7%)**, observing HEAD fires **52/52 (100%)**.

And the blind path is the sanctioned one — `scripts/lib/live_system_guard.sh` names
pull + reinstall + restart as *"the sanctioned path"* for a code-only change. It
writes no row.

Worse than sparse, it is **structurally** silent: `is_stale` also requires the
recorded deploy to POSTDATE the process, so after a pull-based deploy a newly spawned
session is invisible to it **until the next `update.sh` run** — days at a stretch in
practice, while the tree moves dozens of commits. Executed against the live spawn
plane, one slot sat **15 commits behind** with `is_stale` returning `False`; the
session that wrote this patch was **57 commits behind** and had been told nothing.

(Two figures appear above and they answer different questions: record **coverage**
5/46, and verdict **recall** 4/52 — lower, because a recorded deploy only helps a
session that started before it. 52 vs 46 because the replay counts every HEAD-moving
reflog entry. All three permanent records now state the distinction.)

## The fix

Ask the question that has no producer to fail: **does this process's commit differ
from what is checked out now?**

New `commit_identity.is_behind(spawn, head)`. Both sides resolve to the MAIN tree —
a spawn identity is captured from `env.repo_root()`, which is the installed package's
own location, and a worktree editable install is guard-blocked — so a worktree session
still compares on main's line. No time axis is needed: nothing can be *ahead* of live
HEAD, which is the only ambiguity `is_stale` needed timestamps to resolve.

The nudge now names what landed:

```
[⚠ Deploy: main moved 15 commits under this session (85f91765 → dd90bb39) —
 PRs #1573 #1560 #1577 #1563 #1588 +10 more — hooks/policy are already live;
 MCP + genesis-server need a session restart to load]
```

The trailing clause is the actionable half: CC hooks re-exec per invocation from the
main tree, so a pull is live for them immediately, while MCP and the server are not.
It speaks **once per deploy** (stamping the announced sha) rather than re-nudging on a
timer — a deploy is an event, and the old 3h cooldown would repeat for the life of a
session now that detection actually works.

## What deliberately did NOT change

`is_stale` has three consumers. Two are advisory (this hook, the dashboard badge) and
moved. The third — `observability/mcp_middleware.py` — **BLOCKS `procedure_store`**,
and `effective_mode()` is live-set to `block`.

Repointing it would widen a live block from ~never-firing to firing on most
long-lived sessions (~1.5 HEAD movements/day measured). That is a policy decision, not
a bug fix, so it is left alone and filed for an explicit call. The module docstring and
the subsystem-map entry both record the split as intentional so it is not later
"unified" by someone tidying up. `test_the_two_verdicts_*` pins the divergence.

## Verification

- **Acceptance bar — the real defect, replayed.** The pre-fix module is loaded from
  `origin/main` and run beside the new one on the measured live inputs (spawn
  `85f91765`, 15 commits, the `84c7259d` row that predates it). Old: silent. New:
  emits, naming 15 commits and the right PRs.
- **Measured rate, both directions** (above). False-positive side: 0 sessions at HEAD
  flagged; a non-fast-forward move (reset/force) is suppressed at count 0. Stated
  honestly — that diverged case has **no instance** in 30 days of real history, so it
  is proven by unit test against a real `git reset`, not by the corpus.
- **Verify-RED, 11 mutations** across two matrices, each reddening only its named
  test, anchors asserted to match exactly once, mutants compiled with the file's own
  parser, restores verified byte-identical: drop the stamp check · drop the zero-count
  guard · break the hex guard · unanchor the PR regex · make `is_behind` use `!=` ·
  revert the badge to the record-keyed verdict · drop the head guard · drop the argv
  boundary validation · restore `splitlines()` · narrow the stamp back to `OSError` ·
  stamp only on success.
  **Three initial survivors, all three faults in tests rather than code.** One targeted
  the wrong test; one revealed the PR-regex test's input never exercised the `$` anchor
  (`(supersedes #N)` has no `(#` to match, so it passed unanchored — now
  `revert: "feat: thing (#1200)" (#1350)`); one revealed two assertions on a filename
  git never writes. All now RED under mutation.
- **262 tests** across the consumer blast radius, including `mcp_middleware` and
  `mcp_staleness_guard` — those prove the blocking verdict did not move.
- **The text-pinning consumer class**, which a token-keyed grep misses: a session-id
  path-guard test named the removed function. Swept for every removed symbol
  repo-wide (code, tests, docs, config) — 0 remaining.
- **E2E on a live install**, not a fixture: the real hook, real spawn plane, real git,
  and no `GENESIS_REPO_ROOT` set (the production shape). First prompt emitted a real
  59-commit line naming real PRs; the second was silent; rolling the stamp back made it
  speak again; the stamp landed atomically with no temp files left; and
  `messages.jsonl` + `last_prompt_time` were both present, confirming the `main()`
  reorder protects the bookmark buffer.
- Git helpers are tested against a **real temporary repository** (including a
  `pack-refs` case), not a mocked subprocess — the git invocation is the part most
  likely to be wrong, and a fake proves nothing about it.
- All five locally-reproducible CI guards clean.

## Adversarial audit round (internal, pre-push)

A fresh-context `genesis-architect` pass over the full diff returned **0 blockers,
7 SHOULD-FIX, 9 NOTE**. Every finding was verified against ground truth before
acting, then fixed, then re-verified. The findings reduce to **five generators**,
each fixed at the class level rather than per instance:

| generator | findings | fix |
|---|---|---|
| a value crosses a boundary unvalidated | SF1, SF7, N8 | `_is_sha()` chokepoint, applied where the value **enters** |
| a renamed value's meaning didn't propagate | SF5 | renamed the payload key itself |
| fail-open narrowed to the wrong exception family | SF3 | `except (OSError, ValueError)` + atomic write |
| timeout not derived from the caller's budget | SF2, N9 | 5s→1.5s; dedicated request-path bound |
| cost paid on paths that produce no output | SF6 | stamp on any conclusive observation |

Three worth calling out:

- **SF1 — an unvalidated ref reached a git argv operand.** Reproduced with git
  2.43.0: `git log --oneline "--output=<p>..HEAD"` creates that file, exits 0,
  prints nothing — an arbitrary-write primitive that reads as a silent no-op. The
  module's own comment *claimed* hex-guarding, but the guard sat downstream of the
  git call. (The audit's example filename was slightly off — git writes the whole
  operand including the `..<ref>` suffix — corrected in the evidence.)
- **SF5 — the text-pinning class, third form.** Four HTML tooltips still said
  "pre-deploy code (deployed commit X)" for a value that is now live HEAD. My
  earlier sweep was `.py`-only and missed templates entirely. Fixed as a class:
  the key was renamed `deploy_commit` → `head_commit` so the name carries the
  meaning and the prose cannot drift from it again.
- **SF4 — independently found before the audit landed**, and confirmed by it: the
  hook fell back to `Path.home() / "genesis"` while the badge used `repo_root()`,
  and the launcher does not export `GENESIS_REPO_ROOT`, so the fallback *was* the
  production path. Both now use `repo_root()` — the same resolver the spawn commit
  is captured with, which makes the "both sides are on one tree" claim true rather
  than aspirational.

**The mutation matrix then caught two vacuous tests I had just written for those
fixes** — they asserted a filename git never writes (`{target}..HEAD` rather than
`{target}..<sha>`), so the silence they observed came from the zero-count rule, not
the guard under test. Both now glob the real path and go RED when the guards are
removed. Eleven mutations total across two matrices; three initial survivors, all
three faults in tests rather than code.

Full audit + per-finding dispositions: `~/.genesis/review_evidence/2e1e08854ce8.txt`.

## Notes for review

- `_current_head` shells out to `git rev-parse` rather than reading `.git/HEAD`
  by hand. A ref can be loose, packed, or both, so a loose-only reader works until the
  next `git gc` and then fails silently. Measured at 3.86 ms against a hook that
  already opens SQLite twice. Verified to run under the server's sandbox
  (`ProtectSystem=strict`, `NoNewPrivileges=yes`).
- The dashboard's head read is dispatched through `asyncio.to_thread`, and the offload
  test now asserts BOTH sync calls are — asserting only the first would let a later
  blocking call reach the loop unnoticed.
- The notice carries shas, a count and PR numbers only — no commit subjects, no paths,
  no branch names. That is a privacy floor and it also removes tag-forgery entirely,
  since no free text reaches the line.
- Pre-existing class NOT widened here: four call sites now hand-roll
  `git rev-parse HEAD` (`observability/mcp_spawn_identity.py`,
  `guardian/repo_bundle.py`, `contribution/version_gate.py`, and the new one). They
  have genuinely different fallback semantics; consolidating them is its own change,
  not scope for a bug fix.
- The `platform-data` subsystem-map stamp is deliberately NOT bumped: that entry spans
  12 modules and only the observability slice was re-verified, and the file is explicit
  that bumping on faith writes a false claim. Stale stamps warn, they do not fail.
- `scripts/genesis_urgent_alerts.py` is wired as a UserPromptSubmit hook, so treat this
  as hook-adjacent for merge purposes.

Ledger: c320c7908e1e410bbe05c8ab5ecf90cb

## Summary by CodeRabbit

* **New Features**
  * Sessions now notify once when the current code advances, including relevant commit and pull request details.
  * Dashboard indicators compare running sessions with the current checkout and show the relevant commit.

* **Bug Fixes**
  * Improved detection and messaging for changed, diverged, or reset session code.
  * Prevented duplicate marketing pitches and improved transcript cleanup and campaign-name handling.
  * Strengthened validation for invisible characters and campaign names.

* **Documentation**
  * Added guidance for responding to deploy notifications and restarting components when required.
  * Clarified provider availability, configuration, credentials, and code-awareness checks.


WingedGuardian:
@codex review

WingedGuardian:
## Automated first-pass review — PR #1651

Reviewed at head `d3b56a06c5cf668ba35e954473f3786b19884921` (14 files, +1193/-504 against the actual merge-base — note the PR API's reported `base.sha` had already diverged from `main`, so the diff was computed against `git merge-base` instead).

### Domain 1 — Structural review: CLEAN

No pattern violations found against the checklist:
- No `nohup`, no direct `task_states` writes, no `/tmp/` usage, no `git add -A`/`git add .`, no `--force`/`--hard`, no worktree-path `pip install -e`, no bare `except:`, no mutable default args, no `os.killpg`.
- Raw SQL: the old `update_history` query in `scripts/genesis_urgent_alerts.py` is *removed* (replaced by a git-based check), consistent with this hook's documented stdlib-only constraint — no new raw-SQL surface added.
- Imports: stdlib-only additions (`subprocess`, `tempfile` in the hook; `re` in `cc_sessions.py`) are correctly grouped; no circular/star imports.
- Exception handling: `except (OSError, ValueError)` around `_deploy_stamped`/atomic-write paths is deliberate and documented (`UnicodeDecodeError` is a `ValueError`, not `OSError` — verified against CPython behavior), not a bare-except violation.
- Docstrings: every new/changed public function (`is_behind`, `_current_head` ×2, `_deploy_span`, `_deploy_message`, `_deploy_stamped`, `_record_deploy_notice`, `_emit_deploy_nudge`, `_is_sha`) is documented, including rationale for non-obvious choices (timeout budget, `split("\n")` vs `splitlines()`, atomic tempfile+replace).
- Coverage: no new `src/` modules added. Every changed function has corresponding test updates (`test_commit_identity.py`, `test_cc_sessions_routes.py`, the new `test_urgent_alerts_deploy_nudge.py`, `test_session_id_path_guard.py`), including a dedicated test for the git-argv-injection guard itself (`test_span_refuses_an_option_shaped_ref_and_writes_no_file`).
- Commit message carries a conventional prefix (`fix(observability): ...`).

I verified one cross-file signature dependency directly rather than taking it on faith: `cc_sessions.py`'s new `_current_head()` calls `git_health._run_git(repo, "rev-parse", "HEAD", timeout=...)` — signature matches `git_health.py:75` exactly.

### Domain 2 — PII/Secrets/Wording: clean (soft warnings only)

- No API keys/tokens/passwords, private IPv4/IPv6 ranges, SSH keys, or DB connection strings in any added line.
- Email-shaped strings in tests (`t@example.invalid`, `user.email=t@example.invalid`) are synthetic git-identity fixtures, not real PII.
- No absolute paths with real usernames, no location-revealing timezone IDs, no cloud account IDs, no hardware identifiers, no private hostnames added.
- PR body/CHANGELOG: describes a code-freshness observability fix (sessions weren't told when a deploy landed). I specifically checked the "Notes for review" section, which documents a git-argv-injection defect (SF1: an unvalidated ref reaching a `git log` operand) found and fixed within this same commit, with a git-version-specific repro. This is a security-hardening writeup for a defect already patched in the diff — it doesn't name live/unpatched infrastructure, credentials, or a real data exposure, so I'm not treating it as a "vulnerability advertisement" hard block. Flagging it here only so a human reviewer's eyes land on that judgment call too.

No hard blocks, no ERRORs. Nothing above requires a change before merge from this pass.

VERDICT: PASS



---

WingedGuardian:
## Needs a dedicated architecture session — not a review pass

A review session screened this while working the open-PR queue and stopped before resolving it. The collision is **structural, not textual**.

Main's `79bb7e9c8` (#1556) made `scripts/hooks/hook_output.py` (`BoundedStdout`) the single chokepoint for model-facing stdout, and routed `scripts/genesis_urgent_alerts.py` through it. Measured at the two heads for that one file: `hook_output` appears 0 times on this branch versus 1 on main, `BoundedStdout` 0 versus 2, `_TAG_MAX_BYTES` 0 versus 3.

This branch deletes `_emit_staleness_nudge` (which main still references twice) along with the whole 312-line `test_urgent_alerts_staleness_nudge.py`, and adds `_emit_deploy_nudge` plus a 724-line test file that emits **outside** the chokepoint main just made mandatory. Its `test_urgent_alerts_charter_tag.py` conflict is against main's new `_TAG_MAX_BYTES` parity tests.

So this is not a rule being retired — the branch reintroduces per-site emission for a class main deliberately consolidated, and the emitter cap exists because a withheld hook is a silent context loss. Rebuilding the deploy nudge on top of the chokepoint is the promising direction, but it changes what this PR is, so it belongs in a design conversation rather than a conflict resolution.

No commits were pushed and nothing was merged; the branch is untouched. Flagging rather than forcing a resolution, because picking a side here decides policy rather than resolving text. Labelled `needs-architecture-session`.

WingedGuardian:
## Triage 2026-09-14 — back to a builder session (draft + needs-rework)

Batch triage of all open PRs ≤1700. Disposition: this branch emits model-facing stdout outside the chokepoint main made mandatory while the PR sat open — #1556 routed `genesis_urgent_alerts.py` through `hook_output.BoundedStdout` (re-verified today: 2 references at current main), and this branch deletes `_emit_staleness_nudge` (which main still references twice) plus its 312-line test file, adding a per-site emitter in their place.

Rebuild direction, per the rationale comment above: reimplement the deploy nudge ON TOP of the BoundedStdout chokepoint rather than beside it. The observability goal itself (tell a session its code moved) is not in dispute — only the emission path is.

PR #1664, opened 2026-09-03
fix(awareness): say which tick a resumed reflection actually runs against

WingedGuardian:
## The gap

When a user approves a reflection's gated CC fallback, the resume path dispatches it on a **later** tick, using `self._last_tick_result` — the current one. So what a person approved and what runs concern different moments.

The log said neither. It named only the approval id.

## What it says now

The approval, how long ago it was granted, the tick being run against, and — explicitly — that this is **not** the tick that requested it.

That last clause is the honest part rather than a hedge. The originating tick is genuinely unrecoverable: nothing writes it. The approval context is built from a fixed key set with no tick field, and the approval key deliberately excludes per-invocation identity so recurring dispatches reuse one pending row instead of minting a new approval every tick. Giving the line a real lineage would mean building a key-neutral carrier that does not exist today — so it states what is known, and what is not.

`CURRENT.md` records the same thing, including a warning against smuggling a tick id into `action_label` or `extra_context`, since both feed the approval key.

## Nothing under `autonomy/` is touched

`git diff origin/main -- src/genesis/autonomy/` is empty. Key derivation, the existing-approval lookup, both freshness checks and the consume are all unchanged. The new code only **reads** a column off an already-consumed row.

## Why a log line got this much care

It runs **after** `mark_consumed` and **before** `reflect()`, inside a broad `except Exception`. Anything raising there would consume a user-approved reflection, dispatch nothing, and leave only an error log. So the age helper is total over every value that column can hold — verified directly, not just through the log line.

## Review found two surviving mutants

Both now killed:

- **the entire age computation could be replaced by a constant** and the suite stayed green — the assertion matched the format literal `"(approved %s)"` rather than the value it renders
- **the tz-naive branch — the one that can actually raise** — was unreachable from every fixture, since all of them were aware, unparseable, or `None`

Seven mutants now, all killed. Also fixed: an `assert_awaited_once(), "msg"` that was a tuple expression and could never fail.

## Filed, not fixed here

- **The resume path's freshness filter admits approvals older than the 24 hours it documents.** It compares timestamp text whose two sides use different separators, so anything sharing the cutoff's calendar date compares greater regardless of time. Measured: 23h admitted (correct), 47h rejected (correct), **25h and 30h admitted** (not). Pre-existing, and this PR does not cause it — but it now *displays* the age, so the first operator to see `31h00m ago` would reasonably blame the log line. Bounded honestly: it widens a freshness window, it never admits an unapproved or already-consumed request.
- **This is the eleventh copy of the same relative-age helper**, and they disagree on timezone handling, fallback wording, and negative guarding — one would raise where the others degrade.

Neither belongs in a logging change.

### DIFF
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -1024,7 +1024,7 @@ The loops that make Genesis think between conversations.
 entry: ambient-cognition
 modules: [awareness, perception, reflection, attention, session_awareness,
           session_charter.py]
-verified: 50b79ffb 2026-09-01
+verified: 29a382e7 2026-09-03
 ```
 
 - **PR-watch inline surface (2026-07-21)**: a SessionStart hook
@@ -1151,6 +1151,21 @@ verified: 50b79ffb 2026-09-01
   `job.failed` is in the ego's `_REFLEX_OWNED_EVENT_TYPES` so it never spins a
   reactive ego cycle. Ingested by the reflex arc as of PR-2b (`ingest.py`
   consumes `job.failed`, guarded on `error_type` presence).
+- **A resumed reflection runs against a DIFFERENT tick than the one that asked.**
+  When a user approves a reflection's gated CC fallback, `_resume_approved_reflections`
+  dispatches it with `self._last_tick_result` — the CURRENT tick — because the
+  loop's own scoring may never reach that depth again. So what a person approved
+  and what runs are about different moments — and per that method's docstring
+  the reason is availability, not a preference for fresher state: the current
+  tick is the only one it has. The ORIGINATING tick is genuinely unrecoverable: the approval context is
+  built from a fixed key set with no tick field, and `_approval_key` excludes
+  per-invocation identity on purpose so recurring dispatches reuse one pending
+  row rather than minting a new approval every tick. The resume log therefore
+  names the approval, its age, and the tick being run against, and says outright
+  that this is not the requesting tick — rather than implying a lineage that
+  does not exist. Giving it one would mean a key-neutral carrier that does not
+  currently exist; do not smuggle a tick id into `action_label` or
+  `extra_context`, both of which feed the approval key.
 - **perception/**: the real-time reflection engine — MICRO (and LIGHT without
   a CC bridge) run in-process via the router; DEEP/STRATEGIC go to the CC
   reflection bridge. GROUNDWORK: user-model-synthesis, pre-execution-gate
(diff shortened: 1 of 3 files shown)


WingedGuardian:
@codex review

WingedGuardian (commented):

### Structural review

Files changed: `docs/architecture/CURRENT.md`, `src/genesis/awareness/loop.py`, `tests/test_awareness/test_loop_resilience.py`.

- No pattern violations (no `nohup`, no raw SQL, no bare `except:`, no mutable default args, no `git add -A`/force-push patterns, nothing matching the ERROR list).
- New `_approval_age()` helper (`src/genesis/awareness/loop.py:2716`) uses `datetime`/`UTC`/`timedelta`, all already imported at module top (`loop.py:20`) — no new imports, no import-hygiene issues.
- Verified `approved.get("resolved_at")` at `loop.py:3662` is safe: `gate.find_recently_approved` → `ar_crud.find_approved_unconsumed` returns `dict(row)` (`src/genesis/db/crud/approval_requests.py:282`), not a raw `sqlite3.Row`, so `.get()` works as intended and won't raise inside the post-`mark_consumed` block.
- Coverage: the changed function (`_resume_approved_reflections`) and the new helper (`_approval_age`) both get new/updated tests, including edge cases (naive timestamp, unparseable value, absent column, future timestamp/clock skew) — good WARNING-category coverage, nothing missing.
- Documentation: `_approval_age` has a docstring explaining its total/never-raise contract; behavior change is also reflected in `CURRENT.md`. No NOTE-level gaps.
- Commit/PR title uses the correct conventional prefix (`fix(awareness): ...`).

No ERROR, WARNING, or NOTE findings.

### PII/Secrets/Wording scan

- Added lines (diff `+`) in `loop.py` and the test file: no API keys/tokens/secrets, no private IPv4/IPv6 ranges, no SSH keys, no connection strings. Test fixture timestamps/IDs (`appr-deep-123456`, `appr-light`, `2026-03-24T03:00:00+00:00`) are synthetic test data only.
- `docs/architecture/CURRENT.md` changes are architecture narrative — no infra names, hostnames, or credentials.
- PR title/body: describes a logging/observability fix and a pre-existing (out-of-scope, explicitly flagged) freshness-filter bug in `find_approved_unconsumed`'s `resolved_at` comparison. No mention of a leak, security incident, or exploitable infra detail — this is a legitimate technical bug description, not a leak disclosure.
- No CHANGELOG.md changes in this PR.

PII/Secrets/Wording: clean (no soft warnings, no hard blocks).

VERDICT: PASS



---

PR #1667, opened 2026-09-03
docs(dev-skill): the gate description disagreed with the gate

WingedGuardian:
## The defect

`SKILL.md:1717` said the merge gate blocks on **"any P1"** and scoped the doc-path exclusion to **P2s**. Both halves were false — `git_push_guard.py` excludes doc-path P1s too, surfacing them as a non-blocking NOTE. That is the standing directive that Codex is a *code* reviewer (2026-08-10, PR #1362) working exactly as specified.

The divergence was not cosmetic. A session read that paragraph, observed a P1 merge on #1606, concluded the gate had a hole, and built a reversal of that directive before checking whether one existed. It was reverted unshipped. The skill's own Reference Router says *"docs describe intent; code describes reality"* — this is that rule failing on the skill itself, so the correction states it in place rather than quietly changing the wording.

## What makes it a discipline problem, not a gate problem

Two correct non-blocking findings merged **unfixed** on 2026-09-03 because nobody read them: #1620's P2 (invalid `wing` → `ValueError` → `tool_api.py:200-202` maps every exception to HTTP 500, so a caller input error reads as retryable) and #1606's doc-path P1.

**Neither cost anything to fix**, and that is the part the skill never said. A docs-only follow-up commit is review-trivial: `_classify_post_review_delta` tests the **delta's** files, not the PR's, so a CHANGELOG-only fix stays trivial even on a hook-surface PR. Zero blocks, zero review rounds, one commit. There was no tradeoff to weigh.

## Measured, re-verified against live code before writing

| check | result |
|---|---|
| `_is_doc_path("CHANGELOG.md" / "guide.rst" / "docs/x.md")` | True, True, True |
| `classify_compare_substantiality` — CHANGELOG, 500 lines | `inline` |
| `_classify_post_review_delta` — CHANGELOG-only | `inline` |
| `_classify_post_review_delta` — 1-line guard change | `substantial` |
| `--check-pr 1606` on the unmodified gate | prints the P1 as "NOT blocking", then `inline-findings: ok` — no override was used |

## Both directions

The edit says *read and fix non-blocking findings* **and** *do not escalate a cheap finding into a mechanism change*. The second is there because the same session committed that error while correcting the first — diagnosing its own inattention as a gate defect and proposing weighting that would have taxed every PR. A one-sided rule invites the opposite failure.

Also corrects `:1782`, which credited the weighted score with closing the unread-P2 gap. True for pairs; a lone P2 still passes unread, which is how #1620's merged.

Docs-only. No runtime path can regress.

## Summary by CodeRabbit

- **Documentation**
  - Clarified code review guidance for non-blocking findings and encouraged applying inexpensive fixes before merging.
  - Documented how documentation-related findings are handled during review scoring.
  - Added guidance that documentation references may still reveal real code defects.
  - Clarified that review reports must be read fully, including cases where individual findings may not block approval.

### DIFF
--- a/.claude/skills/genesis-development/SKILL.md
+++ b/.claude/skills/genesis-development/SKILL.md
@@ -1313,7 +1313,14 @@ above (full definitions in `.claude/agents/genesis-architect.md`):
   paths at ANY size, and `_classify_post_review_delta` tests the DELTA's files
   rather than the PR's — so a CHANGELOG-only fix stays trivial even on a
   hook-surface PR (verified: CHANGELOG-only → `inline`; a 1-line guard change →
-  `substantial`). Zero blocks, zero review rounds, one commit.
+  `substantial`). So it costs zero CODEX rounds — but not zero blocks, and the
+  difference matters when you are budgeting the follow-up. On the canonical
+  public repo the SCHEDULED-review gate is head-pinned per kind, and only
+  `leaks` has ancestor relief (`_MECHANICAL_RESCAN_BY_KIND`); `code-review` has
+  none. So the push moves the head, the earlier `code-review` marker stops
+  counting, and that gate blocks until a fresh scheduled review lands at the new
+  head. Budget the follow-up as: one commit, no Codex round, one scheduled
+  `code-review` at the new head.
   So: read every finding the report prints, fix the cheap ones, then merge.
   The read is what the standing merge-when-green policy is buying — a gate
   verdict of `ok` is not a report that there is nothing there.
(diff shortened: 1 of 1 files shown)


WingedGuardian:
@codex review

WingedGuardian:
**First-pass review — docs-only PR, PII/wording scan only**

Scope: `.claude/skills/genesis-development/SKILL.md` is the only changed file (markdown, no code) — per process, structural review is skipped for documentation-only PRs.

PII/Secrets (code diff): clean — no API keys, tokens, secrets, private IPs (10.x/192.168.x/100.64–127.x), IPv6 ULA, SSH keys, or DB connection strings in the added lines. No soft-warning items either (no personal emails, no usernamed absolute paths, no location-revealing timezones, no cloud account IDs, no hardware/host identifiers).

Wording (PR title/body + diff prose): clean — content is internal review-gate documentation (references to `git_push_guard.py`, `_is_doc_path`, `tool_api.py`, PR numbers #1606/#1620/#1439/#1362) correcting a stale description of existing gate logic. No mention of a PII/secrets leak, security incident, data exposure, or exploitable infrastructure detail.

PII/Secrets/Wording: CLEAN

VERDICT: PASS



---

PR #1668, opened 2026-09-03
fix(hooks): guards must read the command, not the text it contains

WingedGuardian:
## What this is

Four PreToolUse checks decided whether to block by matching **raw command text**. So a
phrase inside a quoted grep pattern, a heredoc body, a docstring or a commit message read
as the operation itself.

That is not friction. **A block discards the whole Bash call**, so a false match on step 3
silently threw away the `cd`, the heredoc and the file write in steps 1–2, while the error
named only the rule that fired. It was reproduced **four times while building this PR** —
twice on heredocs writing these very files, once on a read-only `grep` whose *search
pattern* contained the phrase, and once by a subagent that tripped it while reading the
guard's own source.

## The fix

The worktree-removal check now keys on **parsed structure** — resolved executable,
subcommand, real argv operands. The three shell checks are **anchored**: the operation must
sit at command position (optionally behind env assignments) with its flags as real tokens
rather than substrings. They stay shell-side deliberately — the test suite records that
they have no tracked Python backstop and must remain inline.

## A false NEGATIVE closed at the same time

The removal check's regex required `git` to be followed *immediately* by the subcommand, so

```
git -C /path/to/repo worktree remove <target>
```

bypassed it **entirely** — six real instances in the corpus. Nobody knew; only the replay
surfaced it.

So that guard now blocks **more** commands than before, and that is the correct outcome:

| | |
|---|---|
| before | 150 blocked |
| after | 152 blocked |
| composition | **4 mentions freed, 6 genuine removals newly caught** |

## Measured

New: **`scripts/replay_guard_corpus.py`**, which Rule 5 of the hook contract requires and
which did not exist — the previous 11,488-command figure survived only as a comment. It
rebuilds a corpus from this install's own transcripts (**48,363 unique commands**), replays
it through a named guard, and reports `k/N`.

| guard | before | after |
|---|---|---|
| inline wrapper arm | 12/6,000 (0.20%) | **7/6,000 (0.12%)** |
| worktree removal | 150/48,363 | 152/48,363 (composition above) |

The corpus is **never checked in** and is written `0600`: it is verbatim command lines and
demonstrably contains secrets passed in argv. `--show` prints nothing by default for the
same reason.

## Both directions, and the anti-vacuity proof

A benign-block rate of zero reads identically for a correct guard and an inert one, so
every allow-case is paired with a true-positive control.

More than that: **every new allow-case was run through `origin/main`'s hook to prove it
discriminates.** 8 of 9 return old=BLOCK / new=ALLOW. The ninth does not discriminate, and
its docstring now says so plainly rather than posing as a regression pin. A first version
of these tests had 12 of 14 passing unchanged against the old hook — they pinned nothing,
and the audit caught it.

Verify-RED, 6 mutations. Two initially survived; diagnosed as **behaviourally null** (I had
mutated the gate, while the load-bearing change is the target extraction) rather than as
vacuous tests. Mutating the actual mechanism gave the right signature: FP test FAIL,
true-positive control PASS, fail-open pin FAIL.

**897 passed / 24 skipped** across the consumer blast radius; lint and all four local CI
guards clean.

## Also here

- **A `bash -n` test for the global Bash hook.** It is wired at *user* level for every
  session on the machine, and bash exits 2 on a syntax error — which the harness reads as
  BLOCK. A typo there would refuse every subsequent command, including the one needed to
  repair it. Nothing checked for that.

## Notes for review

- The adversarial audit returned 2 blockers and 11 should-fixes, and its three summary
  criticisms were all correct and all self-inflicted: a comment asserted a fail-closed
  property the code lacked, the CHANGELOG overstated the fix, and the tests did not fail
  against the old code. All are fixed; full dispositions in the review evidence.
- **SHOULD-FIX 11 partially rejected with reasoning**: CI runs `ruff check`, not
  `ruff format`, and 702 of 1,522 files would reformat — blanket formatting is not this
  repo's convention. But the narrower claim held: two of my three files were clean on main
  and I drifted them, so exactly those two were formatted and the pre-existing one left
  alone.
- **Known limit, stated rather than closed**: wrapper spellings (`sudo`, `timeout`, `uv`)
  remain unrecovered in the two shell arms, because those arms stay shell-side and cannot
  reach `shell_parse._WRAPPER_SPEC`. Measured real-traffic loss: zero — all corpus
  differences are mentions.
- **Deliberately NOT in scope**, each measured and recorded rather than quietly folded in:
  `protected_paths`' untokenizable branch (6 hits in 48,363; scoping it risks a real
  deletion bypass — a bad trade), the `genesis serve` arm's residual 7 FPs, and the `rm`
  delegation, which is the single largest source of guard blocks on ordinary work at
  372/48,363 and is tracked separately.

Ledger: cb57b0facbaf4b0685fd63a88bfe2d06

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved safety checks to distinguish executable risky commands from mentions in text, paths, documentation, or unrelated command arguments.
  * Continued blocking genuine launches, editable installs, and forced worktree removals, including commands using global flags, environment assignments, and nested shell commands.
  * Reduced false positives while preserving fail-closed behavior when command parsing is inconclusive.

* **New Features**
  * Added a command replay utility for evaluating safety guards against collected shell-command histories, with filtering, caching, reporting, and parallel execution options.

* **Documentation**
  * Documented the updated command-detection behavior and validation coverage.


WingedGuardian:
@codex review

WingedGuardian:

### Structural review

No pattern-violation errors, circular imports, star imports, or misordered imports found across the changed files (`.claude/settings.json`, `scripts/bash_safety_hook.sh`, `scripts/hooks/worktree_cwd_guard.py`, `scripts/replay_guard_corpus.py`, three test files, `CHANGELOG.md`). Specifically checked and clear: no `nohup`-started services, no raw `task_states` inserts, no `/tmp/` temp files (the new corpus cache correctly uses `~/.genesis/output/`, matching the output-files convention), no `git add .`/`-A`, no `git push --force`/`git reset --hard` invocations (only guard *detection* code for these), no unanchored/bare `except:` clauses (all `except Exception:` are typed), no mutable default arguments, no raw SQL against `genesis.db`. `scripts/hooks/worktree_cwd_guard.py`'s new `from shell_parse import ...` is a pre-existing, stdlib-only local module (`re`/`shlex`/`dataclasses`/`typing`) — not a new heavy dependency in a hook script. Commit/PR title carries a conventional `fix(hooks):` prefix.

Two NOTE-level (non-blocking) observations:
- `scripts/replay_guard_corpus.py` (new, 327 lines) has no dedicated test file under `tests/`. It's in `scripts/`, not `src/`, so it's outside this checklist's strict scope, but it's substantial new tooling with security-sensitive corpus handling (0600 perms, exclusion logic for `git_discard_guard`) that would benefit from direct unit coverage independent of the shell-guard replays it drives.
- `tests/test_scripts/test_bash_safety_hook.py`: the new top-level comment block/`TestPipEditableArm` class is appended immediately after the previous test method with no blank-line separation (PEP8 E302-style). Low severity — ruff/CI formatting should catch this if it's enforced.

Both changed production files (`bash_safety_hook.sh`, `worktree_cwd_guard.py`) have corresponding test updates in the same PR, so no missing-coverage warning there.

### PII/Secrets/Wording: clean (soft warnings only)

- Scanned all added (`+`) lines for API keys/tokens, private IPs (10.x/192.168.x/100.64–127.x), IPv6 ULA, SSH keys, DB connection strings, and secret-valued env vars — none found. Test fixtures use only placeholder values (`VIRTUAL_ENV=/x`, `PYTHONPATH=/x`, `https://example.invalid/s`).
- Soft warning: hardcoded absolute paths containing a username (`/home/ubuntu/genesis/...`) appear throughout the new test parametrizations in `test_inline_settings_guard.py` and `test_bash_safety_hook.py`. Per this project's known false positives, paths in test fixtures are ignored — noting only for completeness, no action needed.
- Soft warning: a comment in `scripts/replay_guard_corpus.py` names a specific real credential variable found in local session transcripts during development (`an inline SSHPASS=… was found in it`). The value itself is correctly redacted with `…`, so this isn't a hard block, but consider genericizing to "a shell credential variable" rather than naming the specific one encountered, to minimize any hint about real local data shape.
- PR title, body, and CHANGELOG.md additions describe fixing false-positive/false-negative guard predicates (quote-blind text matching → anchored parsing). None of it discloses an actual PII/secrets leak, security incident, or specific private infrastructure (no real IPs/hostnames/account IDs named) — this is guard-logic engineering narrative, not an incident disclosure. The one place secrets are discussed (`replay_guard_corpus.py`'s docstring/comments about the corpus containing "secrets passed in argv") is process/design rationale for why the corpus is never checked in and is written 0600 — not a report of a leak that occurred in this repo.

No hard blocks, no wording issues requiring changes.

VERDICT: PASS



---

WingedGuardian:
Measurement note, for completeness.

The PR body reports the two post-fix numbers that finished: the inline wrapper arm (12/6,000 → 7/6,000) and the worktree guard composition (150 → 152, being 4 mentions freed and 6 genuine removals newly caught).

A full-corpus re-measure of `bash_safety_hook.sh` across all 48,363 commands is still running (it spawns a process per command, so it takes hours). Its **baseline** is 372/48,363 (0.77%), and that figure is dominated by the `rm` delegation to `destructive_command_guard` — an arm this PR does **not** touch, tracked separately. The two arms this PR does change (pip-editable, worktree-force-removal) measured **0 hits in a 1,200-command random sample** before the fix, so the expected movement in the overall number is small and the honest case for those two arms rests on severity, not frequency: when they fire they discard an entire compound command and report only the rule that matched.

Both arms are nonetheless covered by discriminating tests — each allow-case was verified to return BLOCK from `origin/main`'s hook and ALLOW from this one.

WingedGuardian:
The full-corpus re-measure finished. Completing the record, and correcting my own note above.

`bash_safety_hook.sh`, all 48,363 commands: **372 → 353 blocked (0.77% → 0.73%)** — **19 commands freed**. Both arms this PR changes live in that file and are the only edits to it, so the movement is attributable to them.

That corrects the expectation I set in the previous comment. I had said the movement would be small based on a 1,200-command random sample showing 0 hits on those arms — but 19/48,363 is 0.039%, so a sample that size would expect roughly half a hit. **Seeing zero was the expected outcome of too small a sample, not evidence of no effect.** The full run was the right instrument and it was already running when I wrote that; I should have waited for it rather than generalising from the sample.

The substantive point in that comment still stands unchanged: the case for these two arms rests on severity rather than frequency. 19 commands is not a lot, but each one was an entire compound command discarded — heredocs, file writes and all — with an error naming only the rule that matched.

Full set of post-fix numbers now:

| guard | before | after |
|---|---|---|
| inline wrapper arm | 12/6,000 | 7/6,000 |
| bash_safety (both arms) | 372/48,363 | **353/48,363** |
| worktree removal | 150/48,363 | 152/48,363 — 4 mentions freed, 6 genuine removals newly caught |

WingedGuardian:
Closing in favour of #1675, which carries the identical change as one clean commit.

This PR's `leak-detector` check could not be made to pass. Its first commit's patch
added test-fixture lines that match the repository's private-fingerprint gate, and
that gate scans the added lines of *every* commit in the range by design — so that a
value introduced and then removed within a single PR is still caught. A follow-up
commit therefore could not clear it, and rewriting the branch in place is not
available: force-pushing this repository is a hard invariant, not a preference.

The replacement rebuilds the branch on current `main` with the fixtures using an
install-generic path. It also carries four fixes from the adversarial review of that
change, and a correction to this PR's own allow-case census: the body here claimed
8 of 9 allow-cases discriminate against the pre-change hook; re-measuring all of them
gives 11 of 13, with two non-discriminating cases now labelled as forward-looking
rather than left looking like regression pins.

No review comments here need carrying over — there were none.

PR #1683, opened 2026-09-04
fix(hooks): the wrong-repo check now says when it did not run

WingedGuardian:
## What was wrong

`repo_routing_guard` catches a `git add`/`commit` that stages files belonging to a
*different* repository. It identifies the current repo by reading the command text to
work out which directory the command runs in.

When the text did not determine one — `cd $WORKTREE`, `cd $(…)`, a glob — it joined the
unexpanded token onto the current path anyway, producing something like `<cwd>/$W`. That
path cannot exist, so every `git` lookup against it failed and the invocation was dropped
at `if not origin: continue`.

**That is the same branch that means "this repository is not in the topology"** — a case
the module docstring documents as a deliberate non-guard. So a command the guard *could
not inspect* was indistinguishable from one it *chose to ignore*, and nothing said so.

## Why this is an advisory and not a refusal

This is the number that decided the shape of the fix, and it inverts the obvious one.

| | |
|---|---|
| real commands carrying a `git add`/`commit` | 2,264 |
| …whose directory the guard cannot derive | **119 (5.3%)** |
| …that the **pre-change** guard blocked | **0 / 119** |

The failure is **open**: the check is silently *skipped*, never mis-fired. Nobody has ever
been wrongly stopped by this. So the defect is the **silence**, not the verdict — and
making it fail closed would newly refuse 119 ordinary commands to fix a problem nobody
has. Exit codes are unchanged for every command in the corpus.

The two skips that shared one `continue` are now separated: a real directory whose repo
is not in the topology stays silent (documented as unguarded), and conflating them would
have fired the notice on every unrelated checkout on the machine.

## An adversarial audit found this PR turning refusals into allows

Recorded rather than quietly fixed, because it is the more useful half of the story.

The first version's resolver consulted its **base** directory before checking whether the
target was **absolute**. So one unresolvable `cd` blinded every later leg of the command —
including legs naming their directory absolutely. Measured with both binaries over real
scratch repositories holding a wrong-repo file:

| command | pre-PR | first version | fixed |
|---|---|---|---|
| `cd $W && git -C <repo> add firmware/boot.py` | BLOCK | **ALLOW** | BLOCK |
| `cd $W && cd <repo> && git add firmware/boot.py` | BLOCK | **ALLOW** | BLOCK |
| `cd '<repo (with parens)>' && git add firmware/boot.py` | BLOCK | **ALLOW** | BLOCK |
| plain wrong-repo add *(control)* | BLOCK | BLOCK | BLOCK |

Three real wrong-repo commits, waved through with a "check did not run" note.

The correct rule is written in the sibling resolvers' own docstrings — *an absolute target
is normalized and returned; it recovers even from a prior unknown*. **I had implemented it
correctly in `pre_push_privacy_review._resolve` earlier the same day, with this exact
`cd $W && git -C /wt` case spelled out in its docstring, and then wrote it the other way
here.** Two copies of one rule, drifting inside a day, which is the argument for the
follow-up below.

## The metacharacter set was over-broad, for the same underlying reason

`shlex` erases quoting before the token arrives, so a wide set cannot tell an expansion
from a quoted **literal** containing the same character. A directory whose name contains
parentheses exists, is a git repo, is in the topology and held a strong wrong-repo hit —
and was being skipped as unresolvable (row 3 above).

It is narrowed to `` "$`*?" `` (what `git_push_guard` actually uses), and the positive
check that the answer **is a real directory** closes the rest without enumerating
anything: a glob that expands to nothing, a brace expression, a literal that is not there
all fail that test and are reported. Enumerating what a shell expands only ever catches
the constructs somebody thought of.

The set was an open set in **both** directions — under-covering constructs nobody listed,
over-covering literals that merely look like them. Re-measured after narrowing: the notice
rate is unchanged at 119/2,264.

## A parity claim that was false

A comment asserted this set was "the same class the sibling cwd resolvers use". It is not,
and they differ from each other too: `git_push_guard` uses `` "$`*?" ``;
`review_enforcement_commit`'s copies are wider and additionally handle `cd -`, a bare `cd`
and `CDPATH`. The comment now states that, and states that this is a **fifth** copy sitting
outside the existing drift trip-wire (`test_value_flag_consistency.py` locks the
*value-flag* sets; `_UNRESOLVABLE` appears nowhere in it).

Unifying the copies belongs in `shell_parse.py`, which another open branch is editing in
the same session — doing it here would conflict two of my own PRs — so it is **filed as a
follow-up**, along with the bare `cd` / `cd -` cases this copy still handles silently.

## Two numbers that were presented as one

An earlier draft quoted "0 blocks across 70 real commands" beside "119 of 2,264" as though
they described one measurement. They were two populations from two corpora with different
predicates. Both are re-derived on a single population above, and all three sites (the
guard comment, the test header, the CHANGELOG) now quote that reconciled pair.

## Verification

**Verify-RED:** 8 of the 10 new tests fail against the pre-change guard. The 2 that pass
both ways are the true-negative controls and say so in their docstrings — an ordinary
scopeable command stays silent, and a repo with no origin stays silent. Without those two,
a guard that emitted the notice unconditionally would satisfy every other test.

Tests pin **both** orderings. The allow-regression hid behind a docstring that claimed the
general property ("the verdict is unchanged") while the body checked only the safe
direction — hit first, unscoped leg second. That docstring is narrowed to what it pins,
and the reversed order is now its own test.

Output contract re-checked: exactly one `json.dumps` site, always paired with `return 0`;
the block path writes stderr only; no `permissionDecision` anywhere; `main()` still wrapped
in `except Exception: return 0`. A second `print` would have emitted two concatenated JSON
objects and the harness would have parsed neither — silently destroying the pre-existing
weak-hit advisory as well.

**50 tests pass** (`test_repo_routing_guard`, `test_value_flag_consistency`,
`test_hook_versions_complete`). Lint and format clean.

Deploy path: hook changes take effect at the next Claude Code session start, not on merge.

## Summary by CodeRabbit

- **Bug Fixes**
  - The wrong-repository commit check now reports when it cannot determine the target directory for certain shell commands.
  - These cases remain advisory and fail open, so valid commands are not newly blocked.
  - Improved handling covers quoted directories, absolute paths, literal shell-like directory names, and commands that return to a resolvable directory.
  - Existing repository-routing checks continue to report strong and weak matches as before.

- **Documentation**
  - Added an Unreleased changelog entry describing the advisory reporting and its impact.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,19 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **The wrong-repo commit check now says when it did not run.** It works out which
+  repository a `git add`/`commit` targets by reading the command text, and when that
+  text did not determine a directory — a shell variable, a command substitution, a
+  glob — it joined the unexpanded token onto the current path anyway. The result
+  cannot exist, so every lookup against it failed and the check was skipped through
+  the same branch that means "this repository is not covered". A command it could
+  not inspect was therefore indistinguishable from one it deliberately ignored.
+  It now reports that the check did not run, on **119 of 2,264 (5.3%)** real
+  `add`/`commit` commands. Deliberately an advisory and **not** a new refusal:
+  replaying those same 119 through the old behaviour, it blocked **0** of them —
+  it was failing open, so nobody has ever been wrongly stopped by this, and making
+  it refuse would newly stop 119 ordinary commands to fix a silence.
+
 - **The cold-marketing campaign no longer re-pitches the same person.** Once a
   marketing pitch is delivered to a prospect, that prospect is marked contacted and
   drops out of the campaign's target list — previously nothing recorded the contact,
--- a/scripts/hooks/repo_routing_guard.py
+++ b/scripts/hooks/repo_routing_guard.py
@@ -87,9 +87,74 @@ def _load_topology() -> dict | None:
 
 _SHELL_OPS = {"&&", "||", ";", "|", "&"}
 
+# Characters bash would ACT on, so reading the token literally would name a
+# DIFFERENT directory than the one bash enters: parameter expansion, command
+# substitution, and globs.
+#
+# DELIBERATELY NARROW, and the narrowness is the point. shlex has already erased
+# the quoting by the time a token reaches here, so a wider class cannot tell an
+# expansion from a quoted LITERAL that merely contains the same character. A
+# first version added `[{()<>` and MEASURED a real cost: `cd '/srv/a (wt)'`
+# names a directory that exists, is a git repo and held a wrong-repo hit, and
+# the guard stopped checking it. The class was over-covering as well as
+# under-covering — an open set in both directions.
+#
+# The positive validation in main() (`not os.path.isdir(git_dir)`) closes the
+# rest of the class without enumerating anything: a bracket-glob that expands to
+# nothing, a brace expression, a literal that does not exist — all fail the
+# is-it-a-real-directory test and are reported. Enumerating what bash expands
+# only ever catches the constructs somebody thought of; asking whether the
+# answer exists catches the ones nobody did.
+#
+# NOT a parity claim: the sibling resolvers (git_push_guard,
+# review_enforcement_commit, pre_push_privacy_review) each carry their OWN
+# class and they are NOT identical — review_enforcement_commit's is wider and
+# also handles `cd -`, a bare `cd` and CDPATH. An earlier comment here asserted
+# they matched; they do not. Canonicalizing the four copies is filed as a
+# follow-up rather than done here, because the shared module it belongs in is
+# being edited by another open branch.
+_UNRESOLVABLE_DIR_CHARS = "$`*?"
 
-def _parse_git_invocations(cmd: str) -> list[tuple[str, list[str], str]]:
+
+def _resolve_dir(base: str | None, token: str) -> str | None:
+    """Absolute directory a ``cd``/``-C`` lands in, or None if unknowable.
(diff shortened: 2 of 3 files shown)


WingedGuardian:

Files: `CHANGELOG.md`, `scripts/hooks/repo_routing_guard.py`, `tests/test_hooks/test_repo_routing_guard.py`

- No new heavy imports in the hook (still stdlib + local `hook_input` + lazily-guarded `yaml`, unchanged by this diff).
- New/changed function (`_resolve_dir`, updated `_parse_git_invocations`) carry full docstrings covering the WHY (why None must only be returned for the right tokens, why order of the absolute/relative check matters).
- No bare `except:`, no mutable default args, no `nohup`, no `/tmp/` usage, no `git add -A`/`.`, no force-push/reset --hard, no raw SQL, no `os.killpg`.
- Test coverage present and proportionate: 10 new regression/control tests in `tests/test_hooks/test_repo_routing_guard.py` covering the fix, the allow-regression it caught, and both silent-skip true-negative controls.
- Commit message and PR title both carry the `fix(hooks):` conventional prefix.

- No private IPs, credentials, connection strings, or SSH key material in the diff.
- No personal emails or usernamed absolute paths — test fixtures use `tmp_path`/synthetic paths (`/srv/a (wt)`, `$WORKTREE`, etc.).
- PR title/body/changelog describe a hook logic fix (silent skip → advisory notice) with internal measurement numbers (119/2,264); no mention of a PII/secrets leak, security incident, or exploitable infra detail. The `Install:`/`Genesis-Session:` commit trailer and `genesis@genesis` author are opaque system identifiers, not PII.

VERDICT: PASS



---

WingedGuardian:
@codex review

PR #1688, opened 2026-09-04
feat(outreach): route marketing campaign updates to their own Telegram topic

WingedGuardian:
## What

Marketing campaign updates (its tick digest / reply pings) previously routed to the shared **Morning Reports** Telegram topic. This adds a dedicated `marketing` outreach category so they land in their own **"Marketing"** forum topic, cleanly separated from the morning report and other digests.

Root cause: the campaign called `outreach_send(category='digest')` → `resolve_outreach_category('digest')` → `'morning_report'`. `outreach_send` stamps `signal_type=category` with no per-campaign discriminator, so a dedicated category is the routing mechanism.

## Changes (6 coordinated spots)

- `OutreachCategory.MARKETING` enum — the validation gate `outreach_send` checks — plus an enum-subset-of-DB-CHECK lock test (catches "added the enum, forgot the CHECK" for any future category).
- `resolve_outreach_category` `"marketing" → "marketing"` + `DEFAULT_CATEGORIES` `"Marketing"` display name.
- `outreach_history.category` CHECK: canonical `_tables.py` DDL + a guarded table-rebuild block (the inline fragment-probe chain, block #5) that carries the enforcing `engagement_outcome` CHECK and self-heals via a CASE-normalized copy — so it never undoes the earlier engagement-CHECK fix and never re-fires.
- Governance threshold `marketing: 0.0` (`config.py` defaults + `config/outreach.yaml`) so an owner-facing digest is not dropped by the default salience gate. It remains subject to quiet-hours + the shared daily cap, exactly like `notification`/`content`.

## Deploy

Additive migration — existing installs pick up the category automatically on `git pull` + server restart. The morning report and all other topics are unaffected. The behavioral rewire (the campaign sending `category='marketing'`, plus a material-only "only ping when something material happened" rule) lives in the install-local campaign strategy doc, paired with this infra.

## Verification

- Targeted suites green: telegram-topics 12, marketing-rebuild 5, engagement-rebuild 5, governance 27, types 9, schema 34; ruff clean.
- Real-data acceptance bar: ran the migration against a consistent online-backup snapshot of the live `outreach_history` (1670 rows) — all rows preserved, `marketing` enabled, garbage rejected, engagement enforcement intact.
- Adversarial review (genesis-architect): no BLOCKER; one SHOULD-FIX (a comment that overclaimed "always deliver") corrected, one NOTE (rebuild independence) fixed with the CASE copy.

Ledger: 23f4fa665204438290bd6956eb0ecfbf

## Summary by CodeRabbit

- **New Features**
  - Marketing campaign updates now appear in a dedicated Telegram **Marketing** topic instead of the shared Morning Reports topic.
  - Marketing updates bypass the salience threshold while remaining subject to quiet hours and the shared daily sending limit.
  - Marketing reply notifications are routed to the dedicated Marketing topic.
- **Migration**
  - Existing installations automatically receive the new marketing category when restarted.
  - Other Telegram topic routing remains unchanged.


WingedGuardian:
@codex review

WingedGuardian:
**Structural review:** CLEAN — no ERROR-level pattern violations found.

One NOTE:
- `src/genesis/channels/telegram/topics.py` (`resolve_outreach_category` docstring, ~line 311) — the docstring's example list ("E.g., `\"blocker\"` → `\"alert\"`, `\"surplus\"` → `\"surplus\"`") wasn't updated to mention the new `marketing` → `marketing` mapping this PR adds. Non-blocking (docstring examples are illustrative, not exhaustive), but the method's behavior changed and the doc didn't follow.

**PII/Secrets/Wording scan:** CLEAN — no secrets, tokens, private IPs/hostnames, SSH keys, or personal emails in the added lines; PR title/body/CHANGELOG describe a routing/UX fix (marketing campaign digests moving to their own Telegram forum topic) with no security-incident, data-exposure, or private-infrastructure language.

Spot-checks performed:
- Raw SQL in `_migrations.py` / `_tables.py` is confined to schema/migration code (exempt from the raw-SQL rule by policy); the one f-string-interpolated fragment in the new rebuild block (`_CANONICAL_OUTCOMES_SQL`) is a module-local literal constant defined earlier in the same function, correctly in scope, with a justified `noqa: S608`.
- `except Exception:` blocks in the new migration code name an exception type — not bare excepts.
- The new `OutreachCategory.MARKETING` enum value is covered by the DB CHECK (enforced by the new `test_every_outreach_category_is_in_the_db_check` lock test) and correctly left out of `governance._BYPASS_CATEGORIES`, matching the PR's own inline comment claim (verified against `src/genesis/outreach/governance.py:22`).
- No new files were added under `src/`, so no missing-test-file WARNING; both touched behaviors (`resolve_outreach_category`, the migration chain) have new/updated tests, and the migration's rebuild-#5 ordering/probe logic checks out against the surrounding rebuild #4 code.

VERDICT: PASS



---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
## Codex round-1 dispositions (head `6f0a22c4`)

**P1#1 — extra-column drift → FIXED.** Rebuild #5 now copies via the canonical `_intersection_copy` (raises on a live-only column instead of dropping it; try/except keeps the original table). Verified drift-safe: the live `outreach_history` has exactly the 20 canonical columns.

**P2#1 — reply-ping category → FIXED.** `make_marketing_reply_notifier` now uses `MARKETING` (own topic) instead of `NOTIFICATION` (DM). Verified `submit_raw → _deliver` reaches topic resolution via `delivery_routing.marketing=supergroup`.

**P2#2 — pin delivery → FIXED.** Added explicit `delivery_routing.marketing: supergroup` (yaml + defaults), matching approval/content.

**P2#3 — merge defaults into saved configs → FIXED.** `load_outreach_config` now merges shipped defaults **under** the user's `channel_preferences`/`thresholds`/`delivery_routing` maps (user overrides still win), so a config saved before the marketing category existed still resolves marketing's channel/threshold/routing. New test locks it.

**P1#2 — "fail startup on migration failure" → DISPUTED (recommend override).** The proposed fix is inappropriate for this codebase:
- Its premise ("another process holds the SQLite write lock") is already handled — the connection layer's `_retry_locked` retries `database is locked` with jittered backoff on every execute, so a transient lock retries rather than fails.
- `log-and-continue` matches the four sibling `outreach_history` rebuild blocks and logs at **ERROR** (surfaced in health + dashboard) — not silent.
- A genuine non-lock DDL failure fails on **every** boot → caught in CI, never prod; refusing to boot the whole server (denying every other table) over one category is a worse, inconsistent failure mode.
- The dup-send it describes needs a marketing campaign actively sending on an install where the migration genuinely failed and stayed failed; campaigns ship as zero data, and this install keeps the campaign paused until the migration is verified applied at deploy.

CodeRabbit's atomicity note: the `_intersection_copy` change covers the data-loss half; the CancelledError-after-DROP half is a pre-existing pattern shared by all four sibling rebuilds (separate follow-up, not this PR's scope).

PR #1689, opened 2026-09-04
fix(hooks): documentation is prose wherever it lives, and the treatment is a lever

WingedGuardian:
## What

A reviewer's finding on documentation no longer blocks a merge, and how documentation is treated is now a config lever rather than a hardcoded rule.

Owner-directed, taken today. **Scoped to the merge gate only — the commit gate is deliberately untouched.**

## Why the old behaviour was wrong for us

The gate already declined to score findings on prose, but its definition of prose was narrow: markdown counted only underneath a top-level `docs/`. Its own docstring stated the consequence — *"a random top-level `*.md` … blocks"*. So a finding on `AGENTS.md`, `CLAUDE.md`, or any `.claude/skills/**/SKILL.md` was treated as a finding on **code** and held the merge.

`_is_doc_path` rule 3 now accepts an unambiguous documentation extension **at any depth**. `.txt` is deliberately unchanged and stays prose only on a known doc stem, because a build or dependency manifest carries it too.

## This reverses a deliberate design, on one specific ground

The narrowness was not an oversight — a `.md` in this repo is frequently an executable **prompt surface** rather than prose, and treating one as documentation means a reviewer's finding on it cannot block.

The ground for relaxing it, verified by reading rather than assumed: `review_enforcement_commit.py` classifies those files independently via `_is_prompt_surface()`, and at line 127 says *"behavior surface — reviewable, never docs-skipped"*. **That gate is not touched by this PR.** Editing a skill or agent file still requires an adversarial review at commit time. What moves is only whether a reviewer's finding can block a **merge** on prose — not whether prose gets reviewed.

## The lever

```yaml
# ~/.genesis/config/genesis.yaml
merge_gate:
  doc_findings: skip   # skip (default) | p1_only | score
```

- `skip` — a doc-path finding never scores, at any severity. The default and current policy.
- `p1_only` — a doc-path P1 blocks; lower severities only surface.
- `score` — no special treatment; restores the previous behaviour exactly.

Findings are **surfaced in every mode**. The lever decides whether they count toward the blocking score, never whether a human sees them.

One deliberate asymmetry, stated rather than left to be discovered: this reader does **not** fail toward more review, unlike `_required_scheduled_review_kinds`. The shipped default is the permissive end by decision, so there is no stricter default to fall back to; a malformed value lands on the documented default rather than silently tightening a policy nobody asked to change.

## Blast radius — enumerated, not sampled

`grep -rn "_is_doc_path" --include=*.py` over the tree: the definition, **exactly two call sites** (the P1 and P2 branches of `_check_inline_review_findings`), and one docstring mention. Zero test files reference it by name. The blast radius is precisely the merge gate's inline-findings scan.

## Tests

- **8 paths asserted prose**: `AGENTS.md`, `CLAUDE.md`, a `SKILL.md`, an agent `.md`, a nested `docs/**/*.md`, a source-tree `NOTES.md`, `README.rst`, `LICENSE`.
- **8 asserted NOT prose**, so the allowlist stays fail-closed: `docs/requirements.txt`, `docs/conf.py`, `docs/CMakeLists.txt`, two source files, a yaml, the empty path, and a path carrying a literal NEL (0x85) — the control character `gh --jq` actually emits.
- **Mode matrix**: default is `skip`; six malformed values fall back to `skip`; `skip` → a doc P1 does not block *and* is still surfaced; `p1_only` → a doc P1 blocks while a doc P2 does not; `score` → two doc P2s block exactly as two code P2s would.
- **The control**: a *code* finding blocks in all three modes. Without it the lever could leak into code paths and every other test would still pass.

An existing test was **flipped, not deleted**. `test_inline_p1_on_random_md_still_blocks` locked the old behaviour; it is now `test_inline_p1_on_a_markdown_file_no_longer_blocks`, carrying a docstring naming the reversal, the decision behind it, and the `score` setting that restores what it used to assert. A security-relevant test must never silently change meaning.

Stale report text also corrected — the doc-path NOTE still advertised `(CHANGELOG/README/LICENSE/NOTICE/docs/**/*.rst)`, which no longer describes what the predicate accepts.

## Verification

326 tests green in `tests/test_hooks/test_merge_review_gate.py`; `ruff --no-cache` clean; privacy grep clean over the whole diff. `git diff --stat` confirms the commit gate is not in this change: CHANGELOG, `git_push_guard.py`, and one test file.

## Merge posture

Directed as an emergency change to merge without a cross-model review round. Recorded so the disposition is legible later rather than looking like a skipped step.

## Summary by CodeRabbit

- **New Features**
  - Documentation files are now recognized across all directory levels, including Markdown, reStructuredText, and AsciiDoc formats.
  - Added configurable documentation-review modes: `skip`, `p1_only`, and `score`.
  - Documentation findings continue to be reported while merge blocking follows the selected mode; code findings remain blocking.

- **Documentation**
  - Updated the unreleased changelog with the new documentation-review behavior and configuration options.


WingedGuardian:
**Structural review: CLEAN** — no pattern violations found in `scripts/hooks/git_push_guard.py` or `tests/test_hooks/test_merge_review_gate.py`.

Checked: no `nohup`, no direct `task_states` inserts, no `/tmp/` usage, no `git add -A`/`git add .`, no force-push/`reset --hard`, no bare `except:` (the new `except Exception:` in `_doc_findings_mode()` names its exception type), no mutable default args, no raw SQL, no `os.killpg`. Import hygiene is fine — the new `import yaml` inside `_doc_findings_mode()` is a lazy, function-scoped import that matches the three existing sibling readers already in this file (~lines 1097/2671/2759/2816), not a new heavy top-level dependency in a hook. Docstrings on `_is_doc_path` and the new `_doc_findings_mode()` are thorough and reflect the behavior change. Test coverage for the widened `_is_doc_path` allowlist and the new `_doc_findings_mode` lever is extensive — parametrized prose/non-prose cases, all three modes, and a code-path control test. PR title carries the `fix(hooks):` conventional commit prefix.

**PII/Secrets/Wording scan: clean** (soft warnings only: none found) — no API keys/tokens/credentials, no private IPs, no SSH keys, no connection strings in the added lines. The `CHANGELOG.md` addition and the PR title/body describe an internal merge-gate policy change (how doc-path findings are scored) without naming private infrastructure, referencing a security incident, or giving exploit-level detail on a vulnerability.

Note (non-blocking, outside this pass's scope): the diff modifies the inline-findings scan (`_check_inline_review_findings` / `_is_doc_path`) that decides whether a reviewer's own finding on this very PR would count toward its score — worth a second pair of eyes on the decision logic itself, since this review covers structural/PII patterns only, not gate logic.

VERDICT: PASS



---

PR #1703, opened 2026-09-04
fix(deploy): restart resident daemons running pre-pull code, and alert between updates

WingedGuardian:
## What

Merged changes to **resident** systemd user daemons never deployed: nothing restarts them, so they keep running their pre-merge script indefinitely. Verified instance: the temp-protection watchgod ran a 7-week-old script for two weeks past its OOM-capture feature merge — the one instrument built to explain silent CC session deaths was inert through the incident it existed for. Timer oneshots are immune (re-exec each tick); `genesis-server`/`bridge` already have restart handling; the gap is the resident repo-script class.

## Fix — two halves, one verdict

**Heal (update.sh):** `_sync_deploy_targets` gains a `RESIDENT_UNIT_SCRIPTS` table + a pure `_unit_restart_reason` verdict (mirroring `_guardian_redeploy_reason`): restart a RUNNING unit whose `ExecMainStartTimestamp` predates its script's **mtime**. Inside that function deliberately — the only site reached on both the post-merge and "Already up to date" paths, and a stale daemon survives no-op runs by definition (the phase-order suite pins exactly two bare call sites). mtime over commit time: mtime is stamped at checkout — when the code arrived on *this* install — while a commit timestamp is authored upstream and misses a daemon started between commit and pull. Every step guarded (`set -e` live at both call sites); every unreadable fact fails open to no-restart; an inactive unit is never started (enablement stays bootstrap's decision).

**Alert (deploy_health → awareness):** `collect_stale_units` + a `stale_units:` finding class, sharing the verdict and a **lockstep-tested** unit table with the heal. A daemon running pre-update code now raises the standing deploy-drift observation between updates — and that alert's existing recovery line ("run update.sh") becomes true only because of the restart half. Broken probe → `None`, never a clean empty (fail-closed data-access doctrine).

**Cross-install correctness (measured):** `TZ=UTC` pins both halves' systemctl render *and* parse. systemctl formats timestamps client-side, and GNU date resolves ambiguous zone abbreviations to its own guess — measured 50400s error for a UTC+8 install, which would have read a stale daemon as fresh for 14h and silently under-delivered the fix on non-US installs.

**Installer:** step 7 already renders the watchgod unit from the template; the later legacy `cp` from `config/` (hardcoded repo path) silently overwrote that render on any non-default clone location. The cp is gone; the orphaned legacy copy is deleted (the protection registry classifies the `*.service` pattern, not file existence — suite green).

## Review

Fresh-context adversarial architect audit with measured findings; all 3 SHOULD-FIXes (TZ ambiguity, table lockstep test, RED-proof for the set-e guards) and 3 NOTEs (DST fold, stale trap comment, orphaned unit file) resolved in-diff.

## Tests

110 green: 14 new (extraction locks incl. block-inside-function and non-fatality; pure-verdict matrix; recording-stub loop tests incl. probe-failure and missing-script fail-open; table lockstep) + phase-order/traps/settings-transition locks + deploy_health (20) + awareness (12) + protection.

## Deploy

update.sh change deploys itself on its next run (tier-2 self-activation); deploy_health/awareness land with server restart. On existing installs the first update.sh run after merge heals any currently-stale daemon — including on a no-op run.

## Summary by CodeRabbit

* **New Features**
  * Deployment health snapshots now detect resident services running pre-update code using startup files and rendered service configurations.
  * Deployment alerts identify affected services with stale resident code.
  * Stale active services are refreshed and checked for stabilization afterward.
* **Bug Fixes**
  * Invalid, unreadable, missing, or future-dated timestamps no longer trigger unnecessary restarts.
  * Restart failures and services that remain inactive are recorded as degraded deployment health.
  * Unavailable health data is reported as unknown without incorrectly clearing active alerts.

Ledger: 2e85e1df8aa94ea3be3e35ea49f08e6c

---

<a href="https://app.devin.ai/review/wingedguardian/genesis-agi/pull/1703" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


WingedGuardian:
@codex review

WingedGuardian:

Reviewed files: `CHANGELOG.md`, `config/genesis-tmp-watchgod.service` (removed), `scripts/install.sh`, `scripts/update.sh`, `src/genesis/awareness/loop.py`, `src/genesis/observability/snapshots/deploy_health.py`, `tests/test_awareness/test_deploy_staleness_check.py`, `tests/test_observability/test_deploy_health.py`, `tests/test_scripts/test_update_unit_restart.py` (new).

Notes (non-blocking, no action required):
- Domain 1 (structural): no ERROR/WARNING/NOTE findings. Import additions in `deploy_health.py` (`os`, `subprocess`) are grouped and ordered correctly. No bare `except:`, no mutable default args, no raw SQL, no `nohup`/`/tmp/`/`git add -A` patterns. The new restart-heal (`RESIDENT_UNIT_SCRIPTS` + `_unit_restart_reason` in `update.sh`) and the matching `collect_stale_units`/`stale_units` finding in `deploy_health.py` are both covered by new/updated tests, including a lockstep test asserting the two unit tables stay in sync. Commit uses the conventional `fix(deploy):` prefix. Spot-checked the install.sh claim: `scripts/systemd/genesis-tmp-watchgod.service.template` does exist, so the enable-only change (dropping the legacy `config/` copy) is backed by a real template render at step 7, not a dangling assumption.
- Domain 2 (PII/secrets/wording): no API keys, tokens, private IPs, SSH keys, or credential strings in the diff. Commit author address is a `users.noreply.github.com` address (known false positive, not personal PII). PR title/body/CHANGELOG describe a generic deploy-drift bug fix without naming private infrastructure or exposing an exploitable vulnerability.

VERDICT: PASS



---

WingedGuardian:
@codex review

WingedGuardian:

Pushed 776d6f8587b94b0ed46621891c5e70360fa61553 for the owner-requested shared audit correction. The dependency-audit step now reports failures without failing the job, matching the verified #1960 reporting policy. Setup/install failures remain blocking. All existing audit commands and ignore arguments are preserved; no dependencies or vulnerability suppressions changed. SECURITY/Dependabot descriptions now match the policy.

WingedGuardian:
## Moving this to draft and labelling it `needs-rework` — the rounds are not converging, and the reason is structural

Five review rounds (`bdc407c5` → `f59da71a` → `5daac45a` → `2c563b04` → `34dd0a24`), eight commits, and **13 findings still unresolved**. That is not a PR that needs a sixth round of fixes; the repo's own stop condition says to tabulate findings against root cause once a third round lands, and this is two rounds past it.

So here is the tabulation. **13 findings, 5 roots, and three of those roots own 10 of the 13.**

### Root A — the "startup-loaded inputs" inventory is hand-maintained, so it keeps being incomplete (4)

| finding | what the list was missing |
|---|---|
| Track startup-loaded dependencies in the freshness check | `scripts/lib/alert_queue.sh`, which `tmp_watchgod.sh:42-44` sources at startup — a pull that changes only it leaves the daemon on the old emergency-paging code while both the healer and the detector call it fresh |
| Track unit templates in the resident restart inventory | `scripts/systemd/genesis-tmp-watchgod.service.template` — bootstrap rewrites the installed unit and calls only `daemon-reload`, so the running daemon keeps the old directives |
| Include the code-intelligence freeze daemon | `genesis-code-intel-freeze.service` — a second resident service running a repo script, absent from the table entirely |
| Document all startup-loaded files | the changelog describes only the daemon's own script, implying sourced-library changes do not trigger a restart |

Each round adds one more entry to a typed list, and the next round finds the next entry. That is a **denylist wearing an inventory's clothes**: the design guarantees a missing item exists, so reviewers will keep producing them indefinitely. Entry #5 is not the fix.

The fix is to stop typing the set and start deriving it. For each resident unit, its startup-loaded inputs are computable: the `ExecStart` script, everything that script `source`s (transitively), and the unit template it was rendered from. Compare the daemon's start time against the newest mtime of *that derived set*. A new sourced library or a new resident service then joins the comparison on its own, and the class of finding disappears rather than shrinking by one per round.

### Root B — a failed step converts into a silent success (4)

All four are the same fail-open, in the component whose entire job is to not lie about deploy state:

- **Record failed resident restarts as degraded** — `systemctl restart` returns nonzero, the branch only warns, `HOST_CC_DEGRADED` stays empty, and the normal path at `update.sh:1920-1934` then records an **undegraded success**.
- **Failed rendering masks unit drift** / **Skip `try-restart` when unit preparation fails** — a render, target-write, or `daemon-reload` failure still reaches `try-restart`, which refreshes `ExecMainStartTimestamp` while the installed unit keeps its old directives. The next run compares the new start time against the template mtime and calls it fresh. **The failure erases its own evidence.**
- **Propagate unreadable timestamps as an unknown snapshot** — `show` returns nothing parsable, `_probe_unit` returns `(True, None)`, the unit is silently skipped, `collect_stale_units` returns `[]`, and `deploy_health` reports **`healthy`**.

The last one is the sharp end: the detector reports healthy *because* it could not read the evidence. #1790 just spent six rounds removing exactly this shape from the OOM monitor, and the rule it landed on applies here unchanged — **missing evidence must never read as exculpatory**. There is no `unknown` state in this model, only ok and not-ok, and every error path defaults to ok.

The fix is one state, not four patches: a probe returns `ok | degraded | unknown`, `unknown` propagates to the snapshot, and no step whose precondition failed is allowed to write a fresher timestamp.

### Root E — the fix-on-fix, which is why the label is `needs-rework` specifically (2)

An earlier round in this PR changed an unconditional `restart` to `try-restart`, to stop clobbering a unit an operator had deliberately stopped. The next round then found what that fix introduced:

- **Use try-restart to preserve an operator-stopped unit** (the original)
- **Do not mark an operator-stopped unit as a failed restart** — `try-restart` returns *success without starting the unit*; the success branch then checks `is-active`, finds it inactive, and appends `unit_restart_*`. An intentional operator stop is now recorded as a degraded deployment, with no restart having occurred.

Round *n* fixed the symptom by swapping the verb; round *n+1* found that the new verb's no-op is indistinguishable from a failure under the same liveness check. Both come from the same missing distinction: the code never asks whether a **new unit invocation actually happened**, only whether the unit is active afterwards. That question has a direct answer (`InvocationID` / `NRestarts` before and after), and asking it retires both findings and the `is-active`-after-the-fact race underneath them.

### The small two (3)

Genuinely independent and genuinely small — these are fine as ordinary fixes:

- **Build the systemctl environment before probing** — `deploy_health.py` copies raw `os.environ` instead of using this repo's own `genesis.util.systemd.systemctl_env()`, so `systemctl --user` cannot reach the user manager in the supported systemd-service case (no `XDG_RUNTIME_DIR` / `DBUS_SESSION_BUS_ADDRESS`).
- **Compare timestamps at a consistent precision** — `_probe_unit` truncates to whole seconds via `strptime` while `st_mtime` keeps fractions, so a script at `1000.8` and a start at `1000.9` compare as `1000.0 < 1000.8` and a fresh daemon reads stale.
- **Special paths corrupt rendered units** — `&`, `|` and backslashes break `render_systemd_template` substitution, and the stale-unit repair can overwrite a valid unit with a corrupted one.

### What I am asking for

Not a sixth fix round. Two design changes that retire 10 of the 13 together:

1. **Derive the startup-loaded input set** instead of enumerating it (Root A, 4 findings).
2. **Give the probe an explicit `unknown`, and gate every timestamp-refreshing step on its precondition** (Root B, 4 findings) — and while in there, decide restart success from *whether a new invocation occurred*, not from liveness after the fact (Root E, 2 findings).

Then C and D land as ordinary small fixes on top.

Draft + `needs-rework` rather than closed: the problem this PR is solving is real and the diagnosis in it is good — resident daemons genuinely do run pre-pull code between updates, and nothing else catches it. It is the enumeration-and-patch shape that needs replacing, not the goal.

I have not touched the branch. Happy to take the rework myself if that is useful — say the word and I will, otherwise it is yours.

PR #1704, opened 2026-09-05
docs(process): the zero-drop rule — finish before you stop, and review as you build

WingedGuardian:
## What

Two instruction-surface additions from a measured failure: finished, tested code sat unpushed on a local branch for **1.5 days**, invisible to every session, because it was recorded only in a plan file nothing reads back.

**CLAUDE.md — the zero-drop rule.** A turn runs as long as the work requires; never yield with steps you could still complete *and are cleared to complete* — stopping for an approval, a blocking question, plan approval, or a designed hard stop is finishing correctly, not dropping work. What genuinely can't finish becomes a tracked row before stopping (and an issue keeps its per-instance approval gate — the rule waives nothing). Next prompt: reconcile your own stranded artifacts early, ordered behind whatever the user made urgent. "What fell through the cracks?" is *made* zero — enumerate, then fix or file what the enumeration finds.

**genesis-development — review as you build.** The built-in `/code-review` at **low** effort after each meaningful chunk of code: cheap, not a gate, catches drift while the fix is one edit instead of a review round. For a **substantial** change (the commit gate's own definition), the end-of-build review is canonical, not the builder's choice: `/deep-review`. Below substantial, the existing table governs unchanged.

## Reviewed as instructions, and the review earned its keep

An adversarial pass hunted for readings where a literal agent obeys the text into the wrong behavior. It found a **blocker in the rule itself**: the first draft listed "issue" among the row types to create *before stopping*, with no approval caveat — thirty lines above the same file's rule that a public post needs explicit approval every time and that a channel-driven session cannot confirm anything. A gate bypass written in the grammar of hygiene, inside the rule built to prevent drops.

Also fixed before commit: "never yield with steps you could still complete" tested *capability* where the legitimate stops are *permission* (the carve-out now names them); "reconcile FIRST" hardened the deliberately-soft ≈51/49 close-bias and inverted priority for urgent prompts (now "early, ordered behind the user's ask"); the canonical-review claim overclaimed below the gate's substantiality line (verified against the gate source — it now uses the gate's definition verbatim); a citation pointed at a table that didn't contain the claim; and "the answer is always zero" pre-committed a conclusion a lazy agent could satisfy by asserting — it is now a process.

The mechanical half of this rule — the stranded-work detector and session-exit gate that *enforce* what this prose asks — is specced separately (session-roles design §8.11) and handed off for build; this PR is the floor, not the mechanism.

## Summary by CodeRabbit

- **Documentation**
  - Clarified that continuously committed code changes are the standard unit for routine reviews.
  - Established lightweight reviews as the default for ordinary code commits, with deeper review required for substantial branch changes.
  - Documented exemptions for documentation-only and configuration-only commits.
  - Added a requirement to complete cleared work before stopping or explicitly identify every unfinished item in the final handoff.

### DIFF
--- a/.claude/skills/genesis-development/SKILL.md
+++ b/.claude/skills/genesis-development/SKILL.md
@@ -929,22 +929,36 @@ interactive one does). What this paragraph adds is the named DEFAULT below the
 substantial line: for an ordinary code commit, run the built-in `/code-review`
 at **low** effort and let its findings feed the evidence you mark. Cheap,
 catches drift while the fix is one edit instead of a review round, and it is
-NOT a clearance — the marker flow is unchanged. (Commits the table below rates
-"None" — docs, mechanical renames — stay exempt as ever.) Every defect that
+NOT a clearance — the marker flow is unchanged. (The gate auto-allows a
+docs/config-only commit; any code change goes through the marker flow, and the
+gate itself sets the depth — an ordinary edit takes a plain marker, a
+substantial or prompt-surface one takes the adversarial pass.)
+Every defect that
 survives to the end-of-build review costs an adversarial cycle there, and
 every one that survives THAT costs an external Codex round against the
 escalation cap.
 
-**For a SUBSTANTIAL change, the end-of-build review is canonical, not the
-builder's choice: run `/deep-review`.** Substantial is the commit gate's own
-definition — ≥50 reviewable lines, more than one code file, or any auth /
-API / migration / prompt / agent / skill surface. `/deep-review` dispatches
-the adversarial pass in the required shape (genesis-architect, plus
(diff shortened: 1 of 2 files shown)


WingedGuardian:
Automated first-pass review (documentation-only PR — two `.md` files, no code changes — so structural review is skipped per protocol; PII/wording scan only).

- No secrets, tokens, credentials, private IPs, SSH keys, or connection strings in the diff.
- No personal emails, username-bearing paths, location-revealing timezones, cloud account IDs, or hardware/hostname identifiers.
- PR title/body describe a process/instruction change (a "finish before you stop" rule and an in-build review cadence) — no reference to a PII/secrets leak, security incident, private infrastructure, or exploitable vulnerability detail.

VERDICT: PASS



---

PR #1711, opened 2026-09-05
fix(hooks): the tokenizability probe now reports what the segmenter cannot represent

WingedGuardian:
`untokenizable()` began as "shlex raised", and that is strictly narrower than the question every caller actually asks: **can this parse be trusted?**

A construct can tokenize perfectly and still leave `analyze()` with no segment for the command inside it, while the probe reports a clean parse. *"Nothing gated here"* and *"I could not see this"* then return the same value — the exact confusion the probe exists to remove.

## The class is seven, not the two that were reported

Two were raised in review. Rather than fix the named cases, every bash reserved word and grouping construct was run with a **neutral inner command** and checked for whether a segment survived:

| construct | segments | inner command |
|---|---|---|
| `case … esac` | `['case','esac']` | absent |
| `name() { … }` | `['f()','}','f']` | absent |
| `function name { … }` | `['function','}','f']` | absent |
| `coproc …` | `['coproc']` | absent |
| `bash -ce` / `-cl`, `sh -ce` | `['bash','e']` | absent |

`untokenizable()` returned `False` for all of them, so nothing reported the gap.

`if` / `while` / `until` / `for` / `select`, brace groups, subshells, negation and `time` all resolve correctly and are **deliberately absent** — every entry costs a fallback to coarse matching, so the set is only what was measured to fail. `bash -ec` resolves too, which is why the rule is *"c present and not last"* rather than *"contains c"*.

## Why a list is acceptable here, and is not for carriers

This repo has a recorded lesson that enumerating command **carriers** is a losing race, and it's right — the set of programs taking a command as an argument grows forever. Shell **reserved words are a closed set**, fixed by the grammar. So *"the parser reported a reserved word as an executable, therefore the parse is wrong"* converges, while *"the parser did not report a carrier"* never can.

That's the line, and it's why this belongs in the probe rather than in any guard's carrier list.

## A consumer whose fallback was not a superset

`protected_paths_guard.py:209` ran the legacy substring check and **returned**. Sound while that check was strictly broader; it stopped being sound when brace expansion was added to the parsed path only, because an expanded operand is caught there and is invisible to a substring scan.

The consequence inverts the intent of this change: **making the probe more sensitive made that guard more permissive.** Measured before the fix, three verdicts changed and one flipped from block to allow. The branch falls through now, so both checks run and the probe firing can only ever *add* a reason to block.

Pre-existing and independent of the probe — fixed here because the probe change is what makes it reachable often enough to matter.

## Measured rate, both directions — 51,052 real `(command, cwd)` pairs

Over-reporting is the safe direction but it is **not free**: every consumer treats this signal as "use the coarse path", so a probe that fires on ordinary syntax quietly converts the guard layer back to substring matching — the very false-positive class the parser migration exists to reduce.

| iteration | newly reported | verdict changes |
|---|---|---|
| as first written | 847 (1.66%) | 3, **one fail-open** |
| function definition requires `{` | 381 (0.75%) | — |
| `case` requires `esac`; consumer fixed | 269 (0.53%) | 2, both fail-closed |
| glued operators stripped | **320 (0.63%)** | **2, both fail-closed** |

The first tightening removed 466 hits that were a zero-argument call inside a **Python heredoc body**, which this probe reads as ordinary text by design. The second removed a commit message whose body used the word "case" in a sentence — that alone had been enough to divert a guard off its parsed path, and it is the command that produced the fail-open.

Of the final 320, only **2** change any verdict — both `ALLOW→BLOCK`, both carrying a real shell function definition the guard genuinely cannot see inside. `worktree_cwd_guard`: **0 changes out of 320**. Nothing that previously reported stopped reporting.

## Acceptance bar, through the real consumer

The reported constructs move from silent to a **native approval dialog** interactively, and to a **refusal** in a dispatched session where no human is present to answer. Controls unchanged in both directions — including a command that merely *mentions* the operation, which must stay allowed and does.

The consumer wiring for this already existed at `git_push_guard.py:5439`. It was simply never reached, because the signal it depends on never fired.

## Two things found by testing, not by review

**Reserved words arrive glued to shell operators.** `shlex` doesn't treat `;` `&` `|` `(` `)` as separators, so `esac;` is one token. Comparing raw tokens only ever matched a keyword that happened to sit at the *end* of a command — the shape a hand-written fixture takes, not the shape real commands take. Found by a test precondition failing. The keyword comparison strips those characters; the function-definition pattern deliberately does not, because `()` **is** its shape.

**Verify-RED: 7 mutations, 7 killed** — and one survived on the first sweep, which is the most useful result here. The consumer test passed against the unfixed code because its fixture used an operand whose *parent* is itself protected, so the substring path caught it and the parsed path was never needed: green while pinning nothing. The first draft had a precondition asserting the substring path does **not** catch the fixture; it was dropped during a rewrite, and dropping it is what made the test vacuous. It's back, with a comment saying it is load-bearing.

Also recorded: the first run of the acceptance bar read only the process exit code and reported that the fix did nothing. An `ask` decision is exit 0 with JSON on stdout — indistinguishable from an allow by return code alone.

## Disclosure

Test fixtures use a **neutral inner command** rather than a gated verb, so they carry no working shape at all. Comments describe what the segmenter cannot represent without pairing a construct with a claim about any specific gate. The repo's own prose-recipe test passes with them in place.

---

`ruff` clean · **501 passed** across eight guard suites · **183** on the parser and probe · **432** on the commit and merge gates

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved protected-path safeguards for complex shell commands, including commands using brace expansion and unsupported shell constructs.
  - Prevented unsafe commands from bypassing protected-directory checks when command parsing is incomplete or ambiguous.
  - Reduced false positives for quoted text and function-like commands that do not define shell functions.

- **Tests**
  - Added regression coverage for complex command parsing and protected-path enforcement.

### DIFF
--- a/scripts/hooks/protected_paths_guard.py
+++ b/scripts/hooks/protected_paths_guard.py
@@ -208,7 +208,16 @@ def main() -> int:
     # no observed behaviour change.
     if untokenizable(cmd):
         reason = _legacy_substring_block(cmd, dirs)
-        return _block(reason) if reason else 0
+        if reason:
+            return _block(reason)
+        # FALL THROUGH rather than returning 0. This branch used to end here, on
+        # the assumption that the substring check is a superset of the parsed one
+        # below. It is not, and stopped being so the moment brace expansion was
+        # added to the parsed path only: an expanded operand is caught below and
+        # is invisible to a substring scan, so a command that merely TRIPPED the
+        # probe skipped the very check that would have blocked it. Running both
+        # is strictly fail-closed — the probe firing can now only ever add a
+        # reason to block, never remove one.
 
     cwd = payload.get("cwd") if isinstance(payload, dict) else None
     for seg in analyze(cmd):
--- a/scripts/hooks/shell_parse.py
+++ b/scripts/hooks/shell_parse.py
@@ -667,10 +667,75 @@ def untokenizable(command: str) -> bool:
     normalization bought nothing and is removed rather than documented.
     """
     try:
-        shlex.split(command)
-        return False
+        tokens = shlex.split(command)
     except ValueError:
         return True
+    return _has_opaque_construct(tokens)
+
+
+# Reserved words the segmenter does not model. CLOSED SET, and that is the whole
+# reason this list is safe where a list of command CARRIERS would not be: the
+# shell grammar fixes its reserved words, while the set of programs that take a
+# command as an argument grows forever. Enumerating the first converges;
+# enumerating the second is a race. Only words MEASURED to leave `analyze()`
+# without the inner command are here — `if`/`while`/`for`/`select` and the
+# grouping operators all resolve correctly and are deliberately absent, because
+# every entry costs a fallback to coarse matching.
+_OPAQUE_WORDS = frozenset({"coproc", "function"})
+
+# `name()` — a function definition, whose body the segmenter reads as operands.
+_FUNCTION_DEF = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*\(\)$")
+
+# An option bundle carrying `c` somewhere other than the END. `-c` and `-ec`
+# resolve, because the script is the next argument either way; `-ce` and `-cl`
+# do not, because the letters after the `c` are read as the thing to run.
+_INTERPRETER_C_BUNDLE = re.compile(r"^-[A-Za-z]*c[A-Za-z]+$")
+_INTERPRETERS = frozenset({"bash", "sh", "zsh", "ksh", "dash"})
+
+
+def _has_opaque_construct(tokens: list[str]) -> bool:
+    """True when a TOKEN marks a construct `analyze()` cannot represent.
+
+    Tokens, never raw text, and that is what makes this quote-safe by
+    construction rather than by care: shlex keeps a quoted argument as ONE
+    token, so a reserved word inside `grep -r '...'` can never appear as a bare
+    token here. A raw-text scan would fire on every command that merely mentions
+    one, which is the false-positive class the parser exists to release.
+    """
+    # Shell operators are NOT token separators for shlex, so a reserved word
+    # adjacent to one arrives glued (`esac;`, `(git`, `done&&`). Comparing raw
+    # tokens therefore only ever matched a keyword that happened to sit at the
+    # end of the command — which is exactly the shape a hand-written fixture
+    # takes, and not the shape real commands take.
+    words = {t.strip(";&|()<>") for t in tokens}
+    # `case` is required to CLOSE. A real case statement always ends in `esac`,
+    # while prose that happens to contain the word does not — and prose reaches
+    # here routinely, because this probe reads heredoc bodies as ordinary text
+    # by design. MEASURED: requiring the pair removes a commit message whose
+    # body used "case" in an ordinary sentence, which had been enough on its own
+    # to divert a guard off its parsed path.
+    if "case" in words and "esac" in words:
+        return True
+    for i, token in enumerate(tokens):
+        bare = token.strip(";&|()<>")
+        if bare in _OPAQUE_WORDS:
+            return True
+        # `name()` alone is not enough. A shell function definition is
+        # `name() {` — the brace is part of the grammar. Without requiring it,
+        # MEASURED over 51,052 real commands, 574 fired on a zero-argument call
+        # inside a PYTHON HEREDOC body, which this probe reads as ordinary text
+        # by design. Every one of those would push a guard back to substring
+        # matching, which is the false-positive class the parser exists to
+        # reduce — so the probe would have undone at the parser layer what the
+        # guards gained.
+        # Against the RAW token, never the stripped one: `()` IS the shape here,
+        # so stripping the operator characters that matter for a reserved word
+        # erases the thing this pattern matches on.
+        if _FUNCTION_DEF.match(token) and i + 1 < len(tokens) and tokens[i + 1] == "{":
+            return True
+        if _INTERPRETER_C_BUNDLE.match(token) and i and _basename(tokens[i - 1]) in _INTERPRETERS:
+            return True
+    return False
 
 
 def _basename(token: str) -> str:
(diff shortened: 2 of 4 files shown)


WingedGuardian:
@codex review

WingedGuardian:

Files reviewed: `scripts/hooks/protected_paths_guard.py`, `scripts/hooks/shell_parse.py`, `tests/test_hooks/test_protected_paths_guard.py`, `tests/test_hooks/test_untokenizable_probe.py`

### Structural review

No pattern violations found.

- Fall-through fix in `protected_paths_guard.py` (removing the early `return 0` after the substring check) is minimal and well-commented; no `nohup`, no raw SQL, no bare `except:`, no mutable default args, no `os.killpg`, no `/tmp/` usage.
- New helpers in `shell_parse.py` (`_has_opaque_construct`, `_FUNCTION_DEF`, `_INTERPRETER_C_BUNDLE`, etc.) are all private (`_`-prefixed) — no missing-docstring flag, and each carries a docstring/comment anyway.
- Imports stay stdlib-only (`re`, `shlex`, `os`) — consistent with the hook-script constraint.
- No new `.py` files added under `src/` or `scripts/`; both modified files already had corresponding test files, and this PR extends them (`test_protected_paths_guard.py`, `test_untokenizable_probe.py`) rather than leaving new behavior uncovered.
- Commit/PR title carries a conventional prefix (`fix(hooks): ...`).
- Minor style nit only (not a listed rule, not blocking): the four new test methods in `test_untokenizable_probe.py` do a per-test `import shell_parse as sp`, redundant since the module already imports it at top (line 33) — harmless, no action needed.

### PII / Secrets / Wording

- No API keys, tokens, credentials, private IPs (10.x/192.168.x/100.64-127.x), IPv6 ULA, SSH keys, or connection strings in the added lines.
- No personal emails, username-bearing paths, location-revealing timezones, cloud account IDs, or hardware identifiers. The `H = "$HOME"` test shorthand is a literal placeholder string, not a real path.
- PR description is a detailed technical write-up of a fail-open bug in the shell-command approval guard (`protected_paths_guard.py`) and its fix — this is disclosure of a bug that is *fixed in the same PR* (standard security-patch changelog practice), not an incident report, not a leak of live credentials/infrastructure, and it doesn't name any private hosts/IPs/accounts. Not treated as a wording hard block.

PII/Secrets/Wording: CLEAN

VERDICT: PASS



---

WingedGuardian:
Follow-on filed as #1713 — this PR makes the probe *report* the constructs `analyze()` cannot represent, which is the floor. #1713 is teaching the parser to actually read them, at which point each construct's entry here is removed so the probe and the parser cannot disagree about what is representable.

WingedGuardian:
## Measured: the blind-spot net never fires for the class this PR adds

Reviewed the five P1s together rather than one at a time, because four of them
land in one file and all five end the same way — a guarded operation exits the
gate with 0. They are not five bugs; they are two.

**Three are detector-matching bugs** (whitespace `f () { }`; `bash --noprofile -ce`
with options before the `-c`; control operators glued between words, `true;case`).
Codex is right on each, and the last one names the mechanism: the detector trims
operator characters off token EDGES where it needs quote-aware tokenization.

**Two are caller bugs**, and they survive any detector improvement. That is the
part worth the corpus time.

### The consumer only consults opacity when nothing else parsed

`git_push_guard.py`:

```python
if (
    not (push_segs or merge_pr_segs or merge_git_segs or create_segs)
    and untokenizable(cmd)
    and _mentions_gated_op(cmd)
):
```

`review_enforcement_commit.py` has the same shape (`if not any(... commit ...)`
guarding its net). So an opaque construct hiding a guarded operation, alongside
any visible gated segment, is invisible.

MEASURED over **43,427 unique real commands** extracted from local session
transcripts:

| rule | fires | closes the reported scenario? |
|---|---|---|
| **A** — today: only when NO gated segment parsed | **0** (0.000%) | — |
| **B** — whenever opaque + a gated op is mentioned | 52 (0.120%) | yes |
| **C** — fire unless a visible `git`/`gh` executor parsed | 24 (0.055%) | **no** |
| **D** — fire when a MENTIONED op has no parsed executor | 47 (0.108%) | yes |

Two results decide the design.

**A fires zero times.** 318 commands (0.73%) contain an opaque construct, and
every one of them also has a parsed segment — so the precondition suppresses the
net on 100% of real opaque commands. This PR adds a detector for a class its
main consumer structurally never asks about. That is why no test caught it: the
feature is inert rather than wrong.

**C is the intuitive fix and it does not work.** `git merge --abort` is a visible
`git` executor, so it suppresses the net exactly as Codex describes. Worth
recording because C is what "only fire if nothing gated is visible" naturally
becomes.

**D buys almost nothing over B** — 5 commands in 43,427 — while requiring a
per-operation table that has to track the gate's vocabulary as it grows. Most of
the residual firing in both is heredoc bodies whose PROSE mentions merge/push
with nothing executing, which the probe reads as text by design and documents.

### Direction agreed with the repo owner

- **Interactive: rule B.** Honour whole-command opacity regardless of what else
  parsed, and ask. ~1 prompt per 835 commands.
- **Dispatched: rule D.** Keep the harder outcome narrow — block only when the
  operation the text names has no parsed executor at all — so a background
  session is not denied on heredoc prose. The standing axiom is that background
  stays as capable as foreground.

That split is the reason to carry both predicates rather than pick one.

### A third gap, from the same root

The `protected_paths_guard` P1 is not the same as the other two. For the newly
opaque class, `analyze()` is blind by definition AND `_legacy_substring_block()`
cannot see a target assembled by brace expansion — so falling through runs two
checks that both structurally miss it. Widening what `untokenizable()` reports
without giving that caller a check that works for the new class leaves it with
no line of defence, where before the probe simply never fired there.

### Not implemented here

Handing this back rather than reworking the approach: the closed-set argument in
the PR description is sound and worth keeping, and the author is better placed to
decide how the two predicates should live. The measurement is the contribution —
the scripts read transcripts only and can be re-run.

WingedGuardian:
## Converting to draft

`origin/main` has since replaced the whole `untokenizable(cmd)` region this PR rewrites with the `analyze_checked`/`blind` machinery — bounds-induced vs non-bounds blind spots, and a different `_legacy_substring_block` signature carrying a cause/hint argument.

The load-bearing question is whether main's version **already subsumes** this PR's fall-through fix. If it does, the right outcome may be closing this rather than merging it — and that is a judgement about intent, not a merge resolution.

PR #1730, opened 2026-09-05
fix(cc-tmp): watchgod RED sweep spares unix sockets, the control plane it severed

WingedGuardian:
## What

The temp-space watchdog's RED (nuclear) tier deleted every depth-1 directory of Claude Code's working temp except the newest session's ancestor. That included the directory where the CC binary binds one unix socket per live session for cross-session messaging. The sockets are 0 bytes — deleting them reclaimed nothing — while every running session silently became unreachable: still listening on a bound-but-unlinked socket, inbound connects failing ENOENT, nothing logged on either side. Measured on a live install after a real RED event: 3 of 4 running sessions severed until restart.

## Fix

- New `reap_dir_sparing_sockets()` helper: `find "$dir" -depth -not -type s -delete` — object-level deletion that spares socket inodes only. All reclaimable bytes are still deleted; a socket's ancestor dirs survive as empty shells; a socket-free dir is removed exactly as `rm -rf` did.
- The RED depth-1 loop uses the helper; the newest-session guard is untouched.
- After all sweeps, RED logs `preserved N unix socket(s)` — counted post-sweep, so the line is true by construction whatever any sweep did.
- The sweep pipeline gains the file's standard `|| true` idiom so a transient `find` error can't abort the daemon mid-RED under `set -euo pipefail`.
- Yellow/orange tiers, file sweeps, and cache sweeps deliberately unchanged; a pin test locks them out of socket deletion for the layout CC creates.

## Verification

- Verify-RED: 3 of the 4 new tests fail against the pre-fix script on the socket-survival assertions; the pin test was seen failing against a behaviorally-real mutant before trusting its green.
- 18/18 green post-fix (4 new + all 14 existing watchgod tests).
- Acceptance replay of the live incident's depth-1 shape through both scripts: pre-fix 0/4 sockets survive and the socket dir is deleted (incident reproduced); post-fix 4/4 preserved, newest-session dir spared, identical 7/7 junk-dir reclaim, INFO line present.
- Helper semantics measured against GNU findutils 4.9.0 (the binary the systemd unit resolves). shellcheck at baseline parity (no findings on new code).
- Adversarial review: architect + security passes, sequential; all findings fixed or dispositioned in-diff. Fail direction of every change is under-deleting only.

## Accepted limitation

Sockets orphaned by crashed sessions are now spared indefinitely — zero-byte inodes, no disk cost. Liveness-aware reaping belongs to session-roster tooling, not a byte-reclaiming daemon, and is tracked separately.

## Deploy

Container-local systemd user unit (`genesis-tmp-watchgod.service` runs the checkout directly). Post-merge: `systemctl --user restart genesis-tmp-watchgod`; verify via `is-active` + a fresh "Watchgod starting" line in the watchgod log. No host-side deploy path involved.

## Summary by CodeRabbit

- **Bug Fixes**
  - Cleanup now preserves Unix sockets and their required parent directories.
  - Unrelated temporary files and directories continue to be removed during cleanup.
  - Cleanup works correctly even when no active session is present.
  - Improved cleanup reporting includes the number of sockets preserved.

- **Tests**
  - Added coverage for socket preservation across cleanup scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,15 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Fixed
 
+- **The temp-space watchdog no longer severs cross-session messaging when it
+  goes nuclear.** At its most aggressive cleanup tier the watchdog deleted every
+  top-level directory of Claude Code's working temp — including the directory
+  holding each live session's messaging socket. The sockets are zero bytes, so
+  deleting them reclaimed nothing, while every running session silently became
+  unreachable to its peers until restarted. The nuclear sweep now spares unix
+  sockets (and only them — all reclaimable bytes are still deleted) and logs
+  how many it preserved.
+
 - **The pre-merge check now reads the second review bot it was already fetching.**
   Two automated reviewers comment on every pull request, and the gate that decides
   whether a change is ready to merge only understood the format of one of them. The
--- a/scripts/tmp_watchgod.sh
+++ b/scripts/tmp_watchgod.sh
@@ -114,6 +114,19 @@ fs_total_mb() {
     echo "${result:-0}"
 }
 
+reap_dir_sparing_sockets() {
+    # Object-level deletion that NEVER removes unix sockets. CC binds one
+    # socket per live session under cc-tmp (cross-session messaging); they are
+    # 0 bytes, so deleting them reclaims nothing and silently severs the local
+    # coordination plane — sessions keep listening on bound-but-unlinked
+    # sockets and inbound connects fail ENOENT (measured, 2026-09-05 RED
+    # incident). Deletes everything else depth-first; a socket's ancestor dirs
+    # stay non-empty so they survive; a dir holding no sockets is removed
+    # entirely, exactly like rm -rf. -delete failures on non-empty dirs are
+    # expected and suppressed; GNU find continues past them.
+    find "$1" -depth -not -type s -delete 2>/dev/null || true
+}
+
 write_state() {
     local cc_tier="$1" cc_used="$2" sys_tier="$3" sys_pct="$4"
     local is_tmpfs="false"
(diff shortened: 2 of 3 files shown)


WingedGuardian:
@codex review

WingedGuardian:
**First-pass review (structural + PII/secrets/wording)**

Files changed: `CHANGELOG.md`, `scripts/tmp_watchgod.sh`, `tests/test_scripts/test_watchgod_socket_sparing.py` (209 additions / 3 deletions across 3 files).

Checked `scripts/tmp_watchgod.sh` and the new test file against the pattern list:

- No `nohup`, no direct `task_states` writes, no `/tmp/` literal usage (all paths route through `$CC_TMP_DIR` / pytest's `tmp_path`), no `git add .`/`-A`, no force-push/reset, no bare `except:` (n/a — bash), no mutable default args (n/a — bash), no raw SQL, no `os.killpg` misuse.
- Import hygiene: the new test file's imports (`os`, `stat`, `subprocess`, `time`, `pathlib.Path`) are stdlib-only and correctly ordered/grouped. No star imports, no circularity.
- Coverage: the changed function (`clean_cc_red`, plus the new `reap_dir_sparing_sockets` helper) gets a dedicated new test file with 4 cases (incident replay, empty-`newest_session` edge case, object-level-vs-wholesale distinction, and an orange-tier pin test). Checked the two pre-existing watchgod test files (`test_watchgod_instrumentation.py`, `test_watchgod_loopbreak.py`) for assumptions the behavior change could break (e.g. dirs previously fully `rm -rf`'d now surviving as empty shells when they contain a socket) — neither references `clean_cc_red` or the depth-1 sweep, so no conflict.
- Documentation: new bash function and test module/functions are all commented/docstringed appropriately for the non-obvious "why" (object-level vs wholesale deletion, the incident that motivated it).
- Genesis conventions: commit/PR title carries the `fix(cc-tmp):` prefix. Not a hook script (no stdlib-only constraint applies). No `IntervalTrigger`/`$CLAUDE_PROJECT_DIR` misuse (n/a to this diff).

No structural findings.

- Added lines (`+`) contain no API keys/tokens/secrets, no private IPv4 (10.x/192.168.x/100.64-127.x) or IPv6 ULA addresses, no SSH keys, no DB connection strings.
- No personal emails, no usernamed absolute paths (script/tests consistently use `$HOME`/`$CC_TMP_DIR`/pytest `tmp_path`), no location-revealing tz identifiers, no cloud account IDs, no hardware identifiers, no private hostnames.
- PR title/body/CHANGELOG describe an operational bug (RED-tier temp cleanup deleting cross-session unix sockets) in plain engineering terms — no infra names, no exploit-enabling detail, and this reads as a bug-fix narrative rather than a security-incident/PII-leak disclosure.

VERDICT: PASS



---

PR #1733, opened 2026-09-05
fix(routing): stop retrying provider refusals that cannot change

WingedGuardian:
## What was wrong

A 403 meaning *"your account tier may not use this model"* was classified as an
exhausted quota. The vendor message names the plan or subscription you would
need — and `subscription`, `plan` and `limit` are all members of
`_QUOTA_KEYWORDS`, so the quota arm won.

That matters because `QUOTA_EXHAUSTED` was not in the router's fail-fast set.
The router waited and re-asked, with backoff, on a refusal whose answer could
not change — spending the chain's aggregate `max_total_s` on a provider that
could never answer, so a walk could run out of budget before reaching a
provider that works.

`circuit_breaker.record_probe_success`'s docstring has named this exact
provider and error since 2026-08-27. #1563 fixed the probe-healing half; this
is the classification half.

## What changed

`ErrorCategory.NOT_ENTITLED`, matched ahead of the quota keywords. It takes one
behaviour from each neighbour, because neither alone is right:

| axis | `QUOTA_EXHAUSTED` | `PERMANENT` | `NOT_ENTITLED` |
|---|---|---|---|
| router fail-fast | no — retries | yes | **yes** |
| breaker open cap | 4h | 30m | **4h** |

**The class, not the instance.** Review found the first draft fixed one member
of a defect class and left the other live: `QUOTA_EXHAUSTED` had the identical
problem for the identical reason, so it now fails fast too. The old split was
inverted — `RATE_LIMITED`, the one 4xx that genuinely *might* clear inside a
backoff, already failed fast, while a spent allowance, which cannot, was
retried. It matters more than one provider, because such a limit is usually
account-global: a single walk could pay the wait once per chain member.

The two categories stay distinct on every reporting surface. Only the retry
behaviour is shared.

## Measured

Before the change, from `activity_log.latency_ms` (which times the whole
`_call_with_retry`, so it *is* the retry burn):

| refusal | n | avg | max |
|---|---|---|---|
| entitlement 403 | 15 | 11.9s | 19.2s |
| quota 403 | 22 | 4.1–6.8s | 16.9s |

…against a 180–600s aggregate budget.

Classification measured **both directions** over 1,093 real error messages:
15 fire, all genuine entitlement denials (zero false positives), and the one
403 that does *not* fire is a real quota refusal that correctly keeps its own
category. That corpus has a retention window, so the denominator is not
reproducible later.

## Verification

- 235 targeted tests green; ruff and all four CI-reproducible guards clean.
- **Four mutation controls**, one per guard, each restored by hash. One of them
  replays a mutation the reviewer used to prove a test was claiming a guard it
  did not provide — it passed silently then, and fails now.
- **Live E2E** against the real API.
- **Runtime-context check** — the assumption that could have made this inert:
  verified that the client library surfaces this as an `APIError` carrying
  `status_code == 403`, which the delegate's generic handler passes through, so
  the 403-gated branch is genuinely reachable in production rather than only on
  a raw HTTP body.

## Regression marker

Checkable one day after deploy: `activity_log.latency_ms` for the affected
providers should fall from seconds to sub-second, and their breaker
`trip_count` should stop climbing. If those latencies persist, the fail-fast
wiring is not being reached and the diagnosis is wrong.

## Not in scope

Terminal-state semantics (chain removal, a category-aware operator alert)
are not here, so that follow-up stays open. Operator-facing surfaces still
branch on the literal `quota_exhausted` and will render the generic
"provider down" text for a `NOT_ENTITLED` breaker — nothing breaks, both are
fall-through chains, but the distinction does not yet reach the dashboard.

Ledger: 998c59178aa843528c81e32f748aa829

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved provider failover for entitlement-related and exhausted-quota errors by avoiding unnecessary retries.
  - Distinguished entitlement-related access denials from quota exhaustion and other permanent failures.
  - Extended circuit-breaker protection for entitlement and quota failures.
  - Improved exhaustion reporting with clearer provider status, attempt counts, and routing-chain details.
- **Tests**
  - Added coverage for failover, error classification, circuit-breaker behavior, and exhaustion reporting.

---

E2E: after deploy, exercise a call site whose chain carries `mistral-large-free` and confirm in `activity_log` that the 403 is recorded as `not_entitled` after a single attempt, rather than as a retried `quota_exhausted` — the whole point of the reclassification is the attempt count, not the label.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,29 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ## [Unreleased]
 
+### Fixed
+
+- **A model your account tier cannot use no longer stalls the fallback chain.**
+  When a provider refuses a call because the plan does not include that model,
+  the refusal arrives as an HTTP 403 whose message names the plan or subscription
+  you would need. Those are the same words a genuine "you have used up your
+  allowance" message uses, so the router read the refusal as an exhausted quota
+  and did what that calls for: waited, and tried the same provider again. Since
+  the answer can never change, that wait was spent against the chain's overall
+  time budget for nothing — measured at several seconds on each attempt.
+
+  Entitlement refusals are now recognised on their own terms. The router gives up
+  on that provider immediately and moves to the next one in the chain, and the
+  provider is held out of rotation on the long window rather than being re-tried
+  every half hour.
+
+  The same correction is applied to an exhausted allowance, which had the same
+  problem for the same reason: a spent quota is a billing state, so waiting a
+  few seconds cannot change it either, and the limit usually applies to the
+  whole account rather than one model — so a single walk could pay that wait
+  once per provider it tried. Both now behave the way a rate-limit already did,
+  which is to stop asking and move on.
+
 ### Changed
 
 - **A review comment on documentation no longer blocks a merge.** The pre-merge
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,9 +21,12 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
   time budget for nothing — measured at several seconds on each attempt.
 
   Entitlement refusals are now recognised on their own terms. The router gives up
-  on that provider immediately and moves to the next one in the chain, and the
-  provider is held out of rotation on the long window rather than being re-tried
-  every half hour.
+  on that provider immediately and moves to the next one in the chain, and a
+  provider that keeps refusing escalates onto the longer hold-out window instead
+  of levelling off at half an hour. To be exact about what that is worth: the
+  hold-out is identical to the existing one for the first four trips and only
+  pulls ahead during a sustained outage, so the gain is in not re-probing a
+  provider that has been dead for hours — not in the first few minutes.
 
   The same correction is applied to an exhausted allowance, which had the same
   problem for the same reason: a spent quota is a billing state, so waiting a
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -1707,7 +1707,7 @@ verified: ee9ebf85c 2026-09-05
   like from outside. The 403-on-use half of that is now classified in its own
   right: `NOT_ENTITLED` (`retry.py` `_ENTITLEMENT_MARKERS`, checked BEFORE
   `_QUOTA_KEYWORDS` because an entitlement message names the plan/tier/
-  subscription it lacks and three of those words are quota keywords). It takes
+  subscription it lacks, and two of those words are themselves quota keywords). It takes
   one behaviour from each neighbour — fail-fast like PERMANENT, so a dead
   provider cannot spend the chain's aggregate `max_total_s` before the walk
   reaches a working one; the 4h cap like QUOTA_EXHAUSTED, since an entitlement
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -1649,21 +1649,37 @@ How every LLM call picks a provider, and the registry for non-LLM tools.
 ```yaml subsystem-map
 entry: routing-providers
 modules: [routing, providers]
-verified: b8232425 2026-09-02
+verified: df6605eda 2026-09-05
 ```
 
 - **routing/**: `config/model_routing.yaml` defines ~54 numbered call sites,
   each a free-first → paid-last chain; `never_pays` sites are filtered to
   free-only. Per-provider circuit breaker (3 failures, exponential backoff
-  capped 30 min — 4h for QUOTA_EXHAUSTED; 429 = backpressure, NOT a breaker
-  failure; state persisted cross-process to
+  capped 30 min — 4h for QUOTA_EXHAUSTED and NOT_ENTITLED; 429 = backpressure,
+  NOT a breaker failure; state persisted cross-process to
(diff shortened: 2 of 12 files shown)


WingedGuardian:
@codex review

WingedGuardian:

Notes (non-blocking):
- All 4 modified `src/genesis/routing/*.py` files have corresponding test additions in this same PR (test_circuit_breaker.py, test_retry.py, test_router.py) — good coverage.
- Commit/PR title carries the correct `fix(routing):` conventional prefix.
- No secrets, private IPs, credentials, or PII-related wording found in code or changelog/PR text; vendor error bodies (Mistral/OpenRouter) and model names in test fixtures are not sensitive.

VERDICT: PASS



---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
## Cross-model review (GLM-5.3) — findings and dispositions

GitHub Codex is quota-capped (it answered `@codex review` with an explicit usage-limits
message), so this went to the install's approved fallback reviewer with per-use
authorization. Two independent runs landed against head `612ea18`; both are recorded.
They agree on the core findings, and each was re-derived from source before being acted
on — a review finding is a hypothesis, not a fact.

### P2 — `NOT_ENTITLED` rendered as a provider outage · **FIXED (as a class)**

The alert builder special-cased one category. A `NOT_ENTITLED` breaker fell to the
generic arm, so the attention strip said **"Mistral down (circuit breaker open)"** and
the overview said **"API down"**.

The sharp part: the identical 403 rendered **"credits depleted"** *before* this branch,
because its body matched `_QUOTA_KEYWORDS`. So the change made operator diagnosis worse
on its own motivating scenario — sending the operator to check a provider that is
perfectly healthy, for an account-tier fact only they can fix.

Fixed as a class, because "add a label for `not_entitled`" leaves the next category to
repeat it:

- `_CB_ALERT_BY_CATEGORY` + `_CB_GENERIC` are now the single source of truth and are
  **exhaustive over `ErrorCategory` by contract**. A category that genuinely reads as an
  outage is *listed* as generic rather than omitted, so the set is auditable.
- `overview.html` carried its **own copy** of the mapping (`cb_reason === 'quota_exhausted'
  ? … : 'API down'`) — a second decision no Python test could reach. The backend now emits
  `cb_label` and the template renders it. The duplicate is deleted, not corrected.
- `test_every_error_category_has_an_operator_label` fails when a member is in **neither**
  table. **Mutation-verified:** adding a `REGION_BLOCKED` member and touching nothing else
  turns it RED with the fix named in the failure message.

### P3 — the long-window claim was overstated · **CLAIM FIXED, mechanism untouched**

`_effective_open_duration()` is `min(open_duration_s * 2**(trip-1), cap)`, and every
provider sets `open_duration_s: 120`:

```
long cap   [120, 240, 480, 960, 1920, 3840, 7680, 14400]
short cap  [120, 240, 480, 960, 1800, 1800, 1800, 1800]
```

**Identical for trips 1-4.** They diverge at trip 5 and reach the nominal 8× only at
trip 8 — roughly 4.2h into a continuous outage. `load_state` also caps `trip_count` at 3
on restart, so any deploy resets to the shared part of the curve.

The long cap is a *saturation* property, not a first-failure one. `types.py` and the
CHANGELOG now carry both curves and say so, instead of implying a 4h hold from the first
refusal. The restart cap is pre-existing (`QUOTA_EXHAUSTED` had it) and changing breaker
escalation is not a safe trade hours before a release cut.

### P2 — `_QUOTA_KEYWORDS` contains the bare word `"limit"` · **DEFERRED, deliberately**

A resettable 403 (`"Usage limit exceeded — resets in 60s"`) classifies `QUOTA_EXHAUSTED`,
so it now fails fast with no retry and accrues breaker failures toward the long cap for
something that clears in a minute. The measured `n=22` justification covers OpenRouter's
hard `"Key limit exceeded (total limit)"`; there is **no equivalent measurement for the
quota vocabulary itself**, which is what now gates the no-retry decision. `CallResult.retry_after_s`
already exists and would disambiguate — and is ignored.

Not fixed here: consuming `retry_after_s` changes routing behaviour for every provider,
and the triggering vendor behaviour is inferred rather than observed (the reviewer rated
it MEDIUM and could not settle from the tree whether any configured provider returns 403
for a short window). Shipping a behaviour change on an unmeasured premise before a
release cut is the wrong direction. Tracked for post-release.

### P3 — inferred entitlement markers · **DOC-ACCEPT**

`"not authorized"` / `"permission_denied"` move previously-`PERMANENT` 403s onto the long
cap. Already labelled INFERRED in-code with the over-capture called out. Per the corrected
P3 above, the practical difference across the first four trips is nil, which materially
shrinks this finding's cost.

---

**Verified clean by the reviewer:** no exhaustive dispatch on `ErrorCategory` anywhere in
`src/`; the `load_state` `ValueError` guard handles a rollback reading `not_entitled`;
exactly one fail-fast set exists; the new router tests bite (with `max_retries=1` the
pre-change code produced 2 attempts where the assertions pin 1).

**Tests:** 141 passed across `test_api_keys` (24), `test_circuit_breaker`, `test_retry`,
`test_router`, plus 11 dashboard tests touching the template. `ruff` clean.

**Residual:** the Alpine change is verified by the Python-side label contract and existing
dashboard tests; the overview tab was not rendered in a browser to confirm the `cb_label`
binding visually.

PR #1774, opened 2026-09-05
feat(sessions): roster spine — identity columns, observed liveness, honest peer line

WingedGuardian:
## What

First batch of the session roster: make concurrent-session identity legible. Today, nine partial sources of session identity exist on an install (process table, transcripts, heartbeat rows, state dirs, socket dir, model cache, …) and none joins to another — the one connecting edge, *which OS process a session id belongs to*, was never recorded. Measured costs on a live install: a peer session working in another's region with nothing surfacing it; the per-prompt concurrent-sessions line listing an 8.5h-idle live session and a crashed one identically; cross-session messages meant for local peers routing through remote machines.

## How

**Identity is stored where it's cheapest** — in the heartbeat row the in-session hooks already write:
- `pid` + `pid_started_at` as an atomic pair (ppid-walk to `comm == "claude"`; the start time exists so a recycled pid is rejected at read time instead of impersonating a session)
- `cwd` (from the hook payload) and `git_branch` (pure bounded file reads, linked-worktree `.git`-file indirection included; three-valued: name / known-none / unknown)
- `slot` (advisory launcher label; joins trust the pid)

All resolution is fail-open; the upsert's COALESCE contract means unknown never clears known. Everything runs behind the existing 60s throttle / per-prompt writes — the observer hook's <50ms common path is untouched.

**Liveness is never stored** — a new stdlib-only observability leaf derives it at read time: ALIVE (`/proc` + comm + start-time match), REACHABLE (alive AND the session's messaging socket path exists on disk), IDLE-FOR (a duration, never a judgement). Idle is not dead: an idle session in an open terminal stays visible indefinitely; the GC now deletes only observed-gone rows (24h backstop), and successive conversations in one terminal dedupe to the newest per pid.

**The `[Concurrent]` line renders it**: `[Concurrent | live | <model> | <id8> | <branch> @<dir> | idle 4m] topic · digest` — live peers outrank recently-gone ones for the 8 capped slots, overflow is counted, and the section is budget-bounded (it was previously unbounded in peer count). Every peer-authored string keeps the sanitize-before-render contract, extended to the new fields; the liveness token is closed-set validated.

**Deploy-window honesty**: reads against a not-yet-migrated DB (the state every install passes through between session start and server restart) degrade to the legacy column set instead of rendering an empty roster — that failure was measured live during this build and is pinned by test.

## Verification

- TDD throughout: every behavior change seen RED first (23 new tests; the green-from-birth pin tests were verified RED against mutants; one fixture off-by-one was itself caught by its RED run).
- Full battery 196/196 (collected == passed), ruff clean, schema three-place parity guard green.
- E2E with real binaries: the actual observer-hook subprocess wrote a complete identity row (pid = the real claude ancestor, start time matching `ps` to the second, worktree branch, slot), and the real renderer emitted the target line for a peer. Read-only E2E against the live pre-migration DB: 7 legacy rows render honestly as `unknown` with true idle durations.
- Two sequential adversarial reviews (architecture, then security on the fixed code). All 4 architecture SHOULD-FIXes and all 3 security should-address findings fixed in-diff with pin tests: hostile comm bytes (probe-confirmed raise → bytes compare), phantom peers after `/clear` (pid-dedupe), GC TOCTOU (re-checked delete), untested recycled-pid mechanism (now pinned), bounded reads of attacker-influenced `.git`/`HEAD` files, absurd branch names store as unknown, `observe()` honors its never-raise contract without a caller net.

## Known limitations (deliberate, documented)

- Subagent tool calls merge into the parent's "recent tools" digest (pre-existing; probe queued).
- Only the roster section of the proactive hook is budget-bounded; its other emitters are a pre-existing residual.
- `get_roster` (async), `slot`, and `alive_verified` have no consumers yet — GROUNDWORK-tagged for the next batch (MCP roster tool + dashboard join + dispatched-session visibility).

## Deploy

Runtime code + additive idempotent migration: `git pull` + server restart applies the migration; hooks pick up at next CC session start. Mid-window combinations are covered by the degradation paths (pinned by test). No host-side deploy.

## Summary by CodeRabbit

- **New Features**
  - Session rosters now show process identity, working directory, Git branch, and workspace slot.
  - Concurrent-session information includes liveness, idle duration, session IDs, branch, and sanitized summaries.
  - Session status distinguishes active, unreachable, ended, and unknown sessions.

- **Bug Fixes**
  - Improved detection of stale, ended, and recycled sessions.
  - Preserved live sessions during cleanup while removing obsolete heartbeat records.
  - Added compatibility for existing databases and safer handling of malformed session data.


WingedGuardian:
@codex review

WingedGuardian:
## First-pass code review — PR #1774

Reviewed: full diff (15 files, 1641/-107), PR title/body, CHANGELOG addition. This is a code change (hooks + CRUD + migration + observability module), so both passes ran.

### Domain 1 — Structural Review

No pattern violations (nohup, raw SQL outside the CRUD/migration layer, bare `except:`, mutable defaults, `git add -A`/`--force`/`--hard`, `os.killpg` misuse, `$CLAUDE_PROJECT_DIR` misuse, oversized `IntervalTrigger`) found in the diff.

Checked and clean:
- Both new `src/` files (`src/genesis/observability/session_roster.py`, `src/genesis/db/migrations/20260905194140_roster_identity_columns.py`) have matching test files (`tests/test_observability/test_session_roster.py`, `tests/test_db/test_migration_roster_identity.py`).
- New public functions (`claude_ancestor_pid`, `proc_start_iso`, `resolve_git_branch`, `get_roster`, `get_roster_sync`, `cleanup_stale`, `observe`, `observe_all`) all carry docstrings.
- Hook-side additions (`scripts/hooks/session_heartbeat.py`, the ancestor-walk/branch-resolution helpers) stay stdlib-only, as the hook-import convention requires.
- Import grouping is correct throughout (stdlib / third-party / local); no circular imports or star imports.
- PR title carries a conventional `feat(...)` prefix.

**NOTE (doc lag, cosmetic):** `scripts/proactive_memory_hook.py` — `_heartbeat_write`'s docstring ("Write session heartbeat. Returns elapsed ms. Best-effort.") wasn't updated to mention the new roster identity ride-along (pid, pid_started_at, cwd, git_branch, slot) it now writes on every call. An inline comment directly above the write already documents it, so this is a docstring-only gap, not a behavior gap.

### Domain 2 — PII, Secrets, and Wording

Scanned all added (`+`) lines plus the PR title, body, and CHANGELOG.md addition.

- No API keys/tokens/secrets, private IPs (10.x/192.168.x/100.64–127.x), IPv6 ULA, SSH keys, or credentialed connection strings anywhere in the diff.
- No personal email addresses, username-bearing absolute paths, location-revealing timezones, cloud account IDs, hardware identifiers, or private hostnames. Test-fixture paths such as `/home/u/genesis/wt` and `/some/dir` are generic placeholders, not real usernames.
- The PR body's reference to a "2026-09-05 incident" (an unlinked-socket edge case) and to a completed adversarial security review (hostile `comm` bytes, GC TOCTOU, recycled-pid rejection, bounded reads of attacker-influenced `.git`/HEAD files) describes hardening applied to this new code — not a leak, an exploitable disclosure about a live system, or named private infrastructure.

PII/Secrets/Wording: clean (no soft warnings found either)

VERDICT: PASS



---

WingedGuardian:

Pushed b2427dd6514cfd963549199127d4ab93fc51eaa5 for the owner-requested shared audit correction. The dependency-audit step now reports failures without failing the job, matching the verified #1960 reporting policy. Setup/install failures remain blocking. All existing audit commands and ignore arguments are preserved; no dependencies or vulnerability suppressions changed. SECURITY/Dependabot descriptions now match the policy.

WingedGuardian:
Closing this unmerged. The work it does has been overtaken from two directions, and the evidence is worth recording so nobody re-derives it.

## The renderer half is superseded

This PR rewrites the `[Concurrent]` peer renderer in `proactive_memory_hook.py` with its own bounded writer, row cap and overflow line. The default branch has since rewritten the same area independently — `_active_peers`, `_peer_closing_line`, a documented peer limit that surfaces what the limit hid, and bounded output that protects the injection directive. That landed reviewed and is live.

Merging this branch would mean hand-reconciling two competing bounded renderers, and resolving it carelessly regresses the one that shipped.

## The spine half is mostly redundant

Measured against the current default branch:

- `observability/cc_slots.enumerate_cc_slots()` already walks `/proc` for interactive sessions (rejecting headless `-p` invocations), already sources the slot label from the environ **or** the spawn file plane — the only source that survives the server's sandbox, where `/proc/<pid>/environ` returns EACCES — and already rejects a recycled pid by start time. Run live, it returns one row per live session with slot, pid, status and start time.
- `cc_slots.read_proc_start_iso(pid)` already computes the ISO start time, verified against `ps lstart`, and already catches the `ValueError` that this PR's own roster module misses in its `except OSError` (the reviewer finding on `session_roster.py`).
- This PR would therefore be a third implementation of the `/proc` start-time arithmetic.

`pid_started_at` also turns out to be unnecessary for its stated job. A pid can only be recycled after the process holding it exits, which is after that session's last heartbeat write — so for a recycled pid the observed start time is always later than the row's `updated_at`, and the existing `spawn_record_is_current` predicate answers it with no new column.

## The one thing that was genuinely missing

`cc_session_id -> pid` for interactive sessions. Measured on a live database: the heartbeat table held only interactive foreground sessions while the session table's rows carrying a pid were exclusively background reflection runs — 22 of 837 id-bearing rows, disjoint populations, no overlap.

That edge is already built in #1706, whose SessionStart hook registers the interactive terminal session with the claude process's pid under the terminal convention `id == cc_session_id`, skipping headless invocations so it cannot race managed rows. Landing that gives the join, and the roster then becomes a read across three existing stores rather than a schema change.

## Also worth carrying forward

- `cc_slots.enumerate_cc_slots` reads `comm` in text mode, so a hostile process name raises `UnicodeDecodeError` past its `except OSError` — the same class as the reviewer finding on this branch's roster module, already present on the default branch. Filing separately.
- No hook may import `genesis.observability`: measured at roughly half a second of import cost, because the package eagerly pulls in the health module and its HTTP stack. Any liveness helper is server-side only.

The branch is left in place if anyone wants the renderer history or the test fixtures. Reopening is fine if the reasoning above turns out to be wrong.

WingedGuardian:
Superseded — see the evidence comment above. The pid edge this branch was for is already built in #1706; the renderer half is superseded by the bounded-output rewrite on main.

PR #1778, opened 2026-09-05
fix(guardian): do not resume an escalation episode across a host reboot

WingedGuardian:
## What

The Guardian's state file is its only continuity — each timer tick is a fresh `Type=oneshot`
process — and `load_state` restored it verbatim, starting fresh only on corruption. The
300s bootstrap grace is consulted from **exactly one transition of nine**
(`CONFIRMING → SURVEYING`).

So after a host reboot the escalation ladder resumed wherever it left off:

- **past CONFIRMING** — the grace was never reached at all;
- **inside CONFIRMING** — it was measured from the persisted *pre-reboot*
  `first_failure_at` and had already expired on arrival.

Against a container that was legitimately still bootstrapping, and a ladder whose upper
rungs kill processes, drop page cache, revert commits and restore snapshots.

## How

Reset to HEALTHY so the normal `HEALTHY → SIGNAL_DROPPED → CONFIRMING` path stamps a fresh
onset — re-arming the grace that already works rather than adding a second mechanism.
It deliberately does **not** call `_reset_to_healthy`, which zeroes three values that must
survive.

**Preserved, all bounded behaviours:** `recovery_attempts`, `io_triage_attempts`,
`auto_reset_count` — a reboot is not evidence the incident is over, and refunding these
turns a bounded ladder into an unbounded loop on a box that is reboot-cycling, which is the
box where reboots get observed. `down_alert_sent` (clearing it sends a fresh CRITICAL every
30s). `snapshot_size_history` — the disk-space baseline, aliased **by reference** from the
snapshot manager, so the same list object survives. `PAUSED` is excluded from the reset
entirely: forcing it to HEALTHY made the next tick read as a *new* pause and restarted the
long-pause reminder clock.

**Boot identity is `/proc/sys/kernel/random/boot_id`**, a per-boot UUID — not `/proc/stat`
btime. btime is wall-clock minus uptime, so it moves by the full amount of any NTP step or
a suspend/resume, which would read as a reboot and discard a live episode: the one false
positive this must avoid. A UUID needs no tolerance, no direction reasoning, no drift anchor.

## The import contract — and the bug an earlier revision had

The primitive lives in `genesis/util/host_boot.py` and is **stdlib-only on purpose**. The
host Guardian venv installs `pyyaml` and nothing else (`install_guardian.sh:447-448`), and
importing anything under `genesis.observability` executes that package's `__init__`, which
pulls `aiohttp`/`aiosqlite`.

An earlier revision of this change did exactly that. The import sat *before* the tick's
`try:`, so on the host it would have raised, skipped the heartbeat and the
`finally: save_state`, and been restart-looped by the container watchdog — **the recovery
brain silently ceasing to watch, which is worse than the bug being fixed.** Verifying that
the file ships (`GUARDIAN_PATHS`) was not enough; the constraint is the import chain.

So the **class** is guarded, not the instance: `test_host_import_contract.py` imports every
host-side Guardian module in a subprocess with everything outside the stdlib and `pyyaml`
blocked — expressed as an **allowlist**, because a denylist only catches the dependency
somebody already thought of. (Measured: `requests`, `psutil`, `pydantic`, `jinja2` and six
others were importable and invisible to the denylist first written.) It carries a control
asserting the blocker fires, and asserts the worktree copy is the one imported.

**Acceptance-replayed:** restoring the original bad import turns that guard RED.

## Also fixed

The "Genesis recovered" ping was gated on the previous state being `CONFIRMED_DEAD`.
Anything that reset the state first swallowed it, leaving `down_alert_sent` stuck true and
**permanently muting every later down-alert**. Pre-existing (`AWAITING_SELF_HEAL → HEALTHY`
and the auto-reset both hit it), but this change makes it guaranteed rather than occasional,
so it is fixed here. The flag alone is the right predicate.

## Verification

- 134 passed across `test_check` / `test_state_machine` / `test_host_import_contract` /
  `test_util/test_host_boot`; `ruff` clean.
- **Verify-RED, per mechanism**, M0 control green, hash-verified restore. Three mutations
  that were *green before this round* are now RED: re-adding the `old_state` gate,
  `read_boot_id` always falsy, and deleting the `run_check` call site entirely. Plus:
  adopt-legacy removed, fail-open removed, same-boot check removed,
  uses-`_reset_to_healthy`, PAUSED exclusion removed, `signal_history` clear removed,
  `cc_unavailable` clear removed, `boot_id` coercion removed.
- Live: the import contract measured against a simulated minimal venv **with controls both
  ways**, and `read_boot_id` against the real kernel path.

## Caveats

- The reset only fires on an actual **host** reboot — `boot_id` does not change on a bare
  `incus restart`. Correct (the Guardian and the state it protects are both host-side), and
  stated in the docstring so nobody assumes wider coverage.
- The ladder has never executed on this install, so the reboot path cannot be exercised
  before deploy. Guardian code reaches the host only via `scripts/update.sh`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Guardian episodes now reset correctly after a host reboot while preserving recovery attempts and paused state.
  * Recovery notifications are sent consistently after successful recovery, including post-reboot and self-healing scenarios.
  * Reboot detection safely handles unavailable or malformed boot information.

* **Tests**
  * Added coverage for reboot recovery, state persistence, notification behavior, and host-environment compatibility.

### DIFF
--- a/src/genesis/guardian/check.py
+++ b/src/genesis/guardian/check.py
@@ -50,6 +50,11 @@ from genesis.guardian.state_machine import (
     GuardianState,
 )
 
+# genesis.util is import-safe on the host (stdlib-only chain); genesis.observability
+# is NOT — its package __init__ pulls aiohttp/aiosqlite, which the Guardian venv
+# does not carry. See genesis/util/host_boot.py.
+from genesis.util.host_boot import read_boot_id
+
 logger = logging.getLogger(__name__)
 
 _STARTED_AT = datetime.now(UTC)
@@ -439,6 +444,23 @@ async def run_check(config: GuardianConfig | None = None) -> None:
     sm = ConfirmationStateMachine(config)
     sm.load_state(state_path)
 
+    # Each timer tick is a fresh oneshot process, so the state on disk is the
+    # only continuity — including across a host reboot, where an escalation
+    # episode would otherwise resume mid-climb against a container that is
+    # legitimately still bootstrapping. Recovery budgets are preserved; see
+    # ConfirmationStateMachine.reset_if_rebooted.
+    #
+    # Contained like every other host-side helper in this tick: this runs BEFORE
+    # the heartbeat write and the `finally: save_state`, so an exception here
+    # would leave the Guardian silent and get it restart-looped by the container
+    # watchdog — a worse outcome than carrying stale state for one tick.
+    try:
+        sm.reset_if_rebooted(read_boot_id())
+    except Exception:
+        logger.warning(
+            "boot-aware state reset failed — carrying state forward", exc_info=True,
+        )
+
     dispatcher = _build_dispatcher(config)
     snapshots = SnapshotManager(config)
     diagnosis_engine = DiagnosisEngine(config)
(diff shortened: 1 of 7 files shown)


WingedGuardian:

### Structural review
CLEAN — no pattern violations found.

- No `nohup`, no direct `task_states` inserts, no `/tmp/` usage, no `git add -A`/`--force`/`--hard`, no worktree-path `pip install -e`, no bare `except:`, no mutable default args (`StateData.snapshot_size_history` correctly uses `field(default_factory=list)`), no unvalidated `os.killpg`, no raw SQL against `genesis.db`.
- Import hygiene: no circular imports, no star imports; `src/genesis/guardian/check.py`'s new `from genesis.util.host_boot import read_boot_id` and `tests/test_guardian/test_state_machine.py`'s new `datetime` import are both correctly grouped (stdlib → third-party → local).
- Coverage: the new `src/genesis/util/host_boot.py` has a corresponding `tests/test_util/test_host_boot.py`; the changed `check.py`/`state_machine.py` behavior is covered by new tests in `test_check.py` and `test_state_machine.py` (including negative/edge cases: same-boot no-op, legacy state adoption, unreadable boot id fail-open, non-string boot id coercion).
- Documentation: new public surface (`read_boot_id`, `reset_if_rebooted`) has thorough docstrings explaining rationale and failure-mode contracts.
- Commit/PR prefix: `fix(guardian): ...` — conventional.
- Genesis conventions: no `$CLAUDE_PROJECT_DIR` misuse, no `IntervalTrigger` for >1h intervals, no hook scripts touched.

### PII/Secrets/Wording scan

- No API keys/tokens/passwords/secrets, no private IPv4 ranges (10.x/192.168.x/100.64–127.x), no IPv6 ULA, no SSH keys, no DB connection strings with credentials.
- No personal emails, no username-bearing absolute paths, no location-revealing timezone identifiers, no cloud account IDs, no hardware identifiers, no private hostnames. The UUID-shaped values in test fixtures (e.g. `1a2b3c4d-0000-4111-8000-...`) are clearly synthetic boot-id test data, not real identifiers.
- PR title/body describe the fix mechanically (state machine transitions, host-venv import contract) without naming real private infrastructure or advertising an exploitable vulnerability with exploit-enabling detail.

VERDICT: PASS



---

WingedGuardian:
@codex review

WingedGuardian:
## ⚠️ POST-MERGE E2E IS REQUIRED — it cannot be done before merge

**Merged ≠ deployed.** Guardian code reaches the host ONLY via `scripts/update.sh` (Host-Deploy Gate, `docs/architecture/CURRENT.md`). Merging and walking away leaves the host running the old code.

The recovery ladder has **never executed on this install** (`recovery_attempts=0`, `last_recovery_at=None`, read live from the host guardian 2026-09-05 05:27Z), so there is no production path to exercise, and inducing an outage to watch it fire would be the destructive act these changes guard against.

### Checklist for whoever merges
- [ ] `scripts/update.sh` from `~/genesis`, **backgrounded** (exceeds the 600s Bash ceiling; a SIGTERM mid-run rolls the deploy back)
- [ ] guardian gateway `version` → `deployed_commit` matches the merged head
- [ ] guardian gateway `status` → `current_state` healthy, `recovery_attempts` unchanged
- [ ] guardian journal → clean tick, no unexpected reset warning
- [ ] **Do NOT** induce an outage to exercise the ladder

Tracked locally as follow-up `871a9e5e1d0b467b92c01e620d02e519` — bare id on purpose, so the repo-pulse worker does not auto-close the row at merge; it must stay open until the deploy actually runs.

PR #1787, opened 2026-09-05
fix: correct the bypass-mode finding, and the guard that rejected it

WingedGuardian:
Started as docs-only; it is not any more. The repo's own guard rejected the
documentation, and fixing that meant changing the guard.

## The bypass-mode "edit via Bash" finding

In bypass/auto permission mode the CC binary injects a meta message telling the
session to prefer Bash (`sed`, heredocs, scripts) over Read/Edit/Write. Measured
(2.1.246): it is **compiled into the binary** and is **not** install-specific
(the sibling `AgentTool`/workflows lines are in the binary too; neither is config).

**It is not gated only on permission mode.** The selector has a third branch on
a cohort assignment, so a **default-mode session can receive it**. An earlier
revision of this PR claimed it was emitted purely on permission mode; that was
read off a truncated transcription and is corrected here. It is still an
efficiency heuristic making no correctness claim, but that now follows from what
the message SAYS — it carries its own fall-back-to-a-dedicated-tool escape hatch
— rather than from where it is gated. That escape hatch was also the last
sentence of the quoted message, which the blockquote had cut; it is restored.

For reads/search the advice is right. For **writes to source** it is not: bare
`sed -i` returns exit 0 on an absent anchor (silent no-op), rewrites all matches
of a non-unique one, and hits the wrong line on a regex-metacharacter anchor —
where Edit fails loudly. A Bash write also fires none of the five edit-specific
`Edit|Write` PostToolUse hooks (the repo's own post-edit verification plane) and
heredocs sit in blocked-compound blast radius.

**Measured, with the dialect stated, because without it the number misleads:**
`sed` uses BRE by default, where `+ ? | ( ) { }` are literal. Over the last 200
non-merge commits on `src/` + `scripts/`, removed lines ≥8 chars: **2,750 /
5,013 = 54.9%** carry a BRE special, against 3,708 / 5,013 = 74.0% for an ERE
set. An earlier revision published the ERE figure as "75.2%" — a real number for
the wrong dialect. The claim's direction survives; its precision did not.

## Why this is no longer docs-only

`tests/test_hooks/test_no_published_bypass_recipes.py` forbids prose that names
a shell construct AND, nearby, claims a gate was defeated by it. It flagged this
branch six times. Two were real and are reworded. **Four were the word "bypass"
naming the permission mode these docs exist to describe** — not a claim, and
unfixable by rewording.

So the detector learns the naming, with the trailing "mode" REQUIRED. That
requirement is the whole safety of it: the first version of this carve-out
matched the namings as prefixes while its comment claimed it was narrow, which
exempted every continuation. Measured against the pre-change detector, **9
constructed cases went from caught to missed** — an exemption broadened from
anchored to unanchored, which loosens a gate while reading as a tightening. That
is the failure this file exists to prevent, committed inside this file, and it
was caught by an adversarial audit rather than by the green suite.

A second carve-out for a quoted attribute access was removed entirely — it
applied to all ten outcomes and to markdown paths. The prose it protected now
describes the selector instead of pasting the identifier, which is what the
guard's own failure message tells authors to do.

**Separately, a fail-open hole older than this branch:** `_recipe_hits` located
each outcome with `find` — the first occurrence only — so a window opening with
a denied mention took the negation `continue` and abandoned that outcome
entirely; a genuine claim later in the same window was never examined. It now
walks every occurrence, and the control class pins the traversal in both
directions, since a last-occurrence-only implementation was passing every
earlier case.

## Injection-map refresh (CURRENT.md)

- Document the **prune↔watcher cross-store coupling**: the injection watcher's
  blind-scan guard relies on a non-empty `~/.genesis/sessions/`, which
  `disk_hygiene.sh` prunes at 60 days — so that floor is load-bearing for the
  guard too. This lived only in the script's comment.
- **Re-verify + re-stamp** `ambient-cognition` (delta trivial) and `platform-data`
  (14-commit delta checked). The re-verification found a stale count — the entry
  claimed 118 `CREATE TABLE`, HEAD has 117 — corrected and reframed as
  drift-prone per the map's own rate-of-change rule.
- The `routing-providers` orphan stamp is left untouched (owned by another
  in-flight branch; the subsystem-map guard already warns on it there).

## Verification

- **26 passed** on the guard suite, up from 14 on main; ruff clean.
- **verify-RED 7/7**, each mutation proved applied against a hash-checked
  baseline: restoring the first-occurrence scan, walking only the last
  occurrence, reverting the naming match to a prefix, dropping the required
  trailing "mode", restoring the blanket dotted carve-out, and restoring each of
  the two reworded sentences.
- **237 tracked `.md` and 2,815 `.py` rescanned** under the stricter walk: zero
  offenders.
- subsystem-map guard coverage CLEAN; `test_subsystem_traps_hook` (consumes
  CURRENT.md) passes.
- Every documentation correction above was re-derived against the source — the
  binary for the selector, `.claude/settings.json` for the hook count, git
  history for the metacharacter rate — not reasoned about.

E2E: none — documentation plus a test-only guard. The guard's own effect is
exercised by CI on this PR: it must reject the pre-rewording prose and accept
the corrected prose, which is what the 7/7 mutation set asserts.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recipe validation to inspect all relevant claims, including later occurrences.
  - Improved detection of embedded or dotted identifiers containing bypass/auto references.
  - Prevented valid permission-mode names and negated claims from being incorrectly flagged.
  - Tightened matching to distinguish complete permission modes from similarly named values.

### DIFF
--- a/.claude/skills/genesis-development/SKILL.md
+++ b/.claude/skills/genesis-development/SKILL.md
@@ -456,18 +456,23 @@ bodies, so legitimate commands get blocked too.
 ### Editing source in bypass/auto mode: reads via Bash, WRITES via Edit/Write
 
 In bypass or auto permission mode the CC binary injects a meta message telling
-the session to prefer Bash (`sed`, heredocs, scripts) over Read/Edit/Write. It
-is gated on permission mode, so it is an EFFICIENCY heuristic and makes no
-correctness claim (provenance + full measurement:
+the session to prefer Bash (`sed`, heredocs, scripts) over Read/Edit/Write —
+and NOT only there: a third selector branch is a cohort assignment, so a
+default-mode session can receive it too. It is an EFFICIENCY heuristic making no
+correctness claim, which follows from how the message is phrased (a preference,
+with its own escape hatch back to the dedicated tools) rather than from where it
+is gated (provenance + full measurement:
 `docs/reference/cc-compatibility.md` "Bypass/auto mode tells the agent to edit
 via Bash"). For reads and search it is simply right — use `cat`/`sed -n`/`grep`/
 `find` freely. For WRITES TO SOURCE it is not: a bare `sed -i` returns exit 0 on
 an absent anchor (silent no-op), rewrites ALL matches of a non-unique one, and
 silently hits the wrong line on a regex-metacharacter anchor where a literal was
-meant — where Edit fails LOUDLY on each. A Bash write also bypasses the five
+meant — where Edit fails LOUDLY on each. A Bash write also fires none of the five
 `Edit|Write` PostToolUse hooks (the repo's own post-edit verification plane) and
-a heredoc sits in the blocked-compound blast radius above. MEASURED: 75% of real
-edit anchors in this repo carry a `sed` metacharacter, so the unsafe case is the
+a heredoc sits in the blocked-compound blast radius above. MEASURED: 54.9% of
+real edit anchors in this repo carry a character `sed` treats as special in BRE,
+its default dialect (2,750 / 5,013; an ERE set scores 74.0% over the same
+corpus, so quote the dialect or the number misleads) — the unsafe case is the
 common one. Use Edit/Write for source; a script that does a literal replace AND
 asserts its occurrence count is acceptable (it re-implements Edit's checks). This
 is universal — the machinery is tracked, so every clone inherits it.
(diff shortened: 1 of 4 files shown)


WingedGuardian:
@codex review

WingedGuardian:
Docs-only PR (`.claude/skills/genesis-development/SKILL.md`, `docs/architecture/CURRENT.md`, `docs/reference/cc-compatibility.md`) — structural review skipped per process, PII/Secrets/Wording scan run.

PII/Secrets/Wording scan: CLEAN — no secrets, private IPs/hostnames, personal emails, or username-bearing paths in the added lines; commit-hash references (`9730efe9`, `29a382e7`, `50b79ffb`) are doc-provenance SHAs, not credentials. Title/body describe a CC binary editing-heuristic finding and internal doc-accuracy corrections — no leak/incident language, no specific private infra named, nothing exploitable disclosed.

VERDICT: PASS



---

WingedGuardian:
@codex review

PR #1795, opened 2026-09-06
docs: say when to store a memory, not just which store to put it in

WingedGuardian:
## What

`CLAUDE.md` already says which memory store to use — a fact goes to Genesis memory, a behavioural rule to the session's own file memory. That is a **routing** rule, and routing only helps once you have already decided to store something. Nothing said *when that moment is*.

The one concrete trigger it offered — *"if you synthesize an answer from multiple recalled memories"* — covers synthesis **from recall**. Work that produced something new and expensive did not qualify.

## The change

The trigger is now **re-derivation cost**: store it when a future session would have to redo the work to know it. That is a question about work already done, which can be answered honestly. *"Is this important?"* is a guess about the future, and it tends to resolve as no.

Three concrete firings:

1. **A measurement** that took real effort — kept with its denominator *and its method*, because the method is usually what the next session gets wrong.
2. **A durable architectural fact** — especially one found while expecting the opposite.
3. **A correction** to a previously-held belief, linked with `supersedes`. Highest value, easiest to skip, because being wrong does not feel like a finding.

It also names a distinction the routing text left implicit: **the two stores differ in lifetime, not only audience.** File memory is compacted away within months, so a durable fact parked there is a fact that will quietly go missing. A lesson that is both a rule and a fact belongs in both.

## Why this is worth a rule rather than better intentions

The origin is recorded in the text. A session measured an embedding regression, enumerated an init-path invariant across 33 modules, and corrected a wrong architectural claim of its own — then wrote only the behavioural lessons to file memory and stored nothing durable until asked why. The routing rule was clear and correctly followed; nothing told it *when*.

## Honest limitation

This is a behavioural claim about future conduct, and it is unproven. The most that can be said is that the trigger is **concrete and checkable** — each of the three cases is a fact about work just completed, not a judgement call — where the previous phrasing was neither.

The failure mode to watch for is the opposite one: over-firing, and a store filling with restatements of what the repo already records. The text guards against that explicitly ("don't store routine answers, restatements of what a file already says, or anything the repo records on its own"), but whether that guard holds is a question for observation, not for this PR.

No code paths touched; documentation only.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -272,12 +272,42 @@ the question. Query SQLite `cc_sessions` for structured session data. Use
 `db_schema` MCP to discover table schemas before any SQLite query (60+ tables).
 **Grep transcripts is LAST RESORT** — only after all above fail.
 
-**When to store back:**
-If you synthesize an answer from multiple recalled memories — something that
-connects information in a new way — store it via `memory_store` with
-`tags: ["synthesis"]` and appropriate wing/room tags. This is how the memory
-system compounds over time. Don't store routine answers; store genuine syntheses
-that would be expensive to re-derive.
+**When to store back — the test is RE-DERIVATION COST, not importance.**
+Store to Genesis memory when a future session would have to REDO WORK to know
+this. That is a question about the work you just did, which you can answer; "is
+this important?" is a question about the future, which you cannot, and it is why
+the rule below used to fire so rarely. Three concrete triggers:
+
+1. **You MEASURED something** that took real effort to obtain — a probe you had
+   to design, an enumeration across a directory, a live-state check. Store the
+   number WITH its denominator and the method, because the method is usually what
+   the next session gets wrong.
+2. **You established a durable ARCHITECTURAL fact** — X exists, X works this way,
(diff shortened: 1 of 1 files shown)


WingedGuardian:
Automated first-pass review (scheduled).

This PR is documentation-only (`CLAUDE.md` only, 1 file changed) — per process, structural review is skipped and only the PII/Secrets/Wording scan ran.

**PII/Secrets/Wording scan:** clean (no hard blocks, no soft warnings).
- Added lines contain no API keys/tokens/secrets, no private IPv4 ranges, no IPv6 ULA addresses, no SSH keys, and no DB connection strings.
- No personal email addresses, usernamed absolute paths, location-revealing timezone identifiers, cloud account/project IDs, or hardware identifiers.
- No `CHANGELOG.md` changes to scan.
- PR title/body describe internal Genesis architecture (memory-store trigger rules, an illustrative "embedding regression" measurement and a "33 modules" enumeration used as a worked example) as rationale for a documentation change — not a leak disclosure, security-incident reference, named private infrastructure, or exploitable vulnerability detail.

VERDICT: PASS



---

PR #1796, opened 2026-09-06
fix: report full deterministic stream size

### LINKED ISSUE #1707: executor: _read_limited reports a clipped stream's retained length as the total
## What happens

`autonomy/executor/deterministic.py` bounds each subprocess output stream at
`_MAX_STREAM_BYTES` (2 MiB) via `_read_limited`, which reads to the limit and
discards the remainder — deliberately, to prevent memory exhaustion from
commands like `yes` or verbose test runs. That cap is correct and should stay.

The gap is that the cut is silent end to end:

- `_read_limited` returns only the retained bytes, with no signal that anything
  was discarded.
- Its caller then reports the retained buffer's length, so a 3 MiB stream is
  described as `2097152 bytes total` — a false total, not merely a missing one.

A reader of the step output (human or model) sees a byte count that looks
authoritative and a log that looks complete, with no way to tell it was
clipped.

## Suggested shape

- `_read_limited` returns `(bytes, truncated: bool)` (or an equivalent
  signal).
- The caller reports honestly when clipped, e.g.
  `>= 2097152 bytes (output bounded; remainder discarded)` — a lower bound,
  since counting the discarded remainder would require reading it, which is
  the cost the bound exists to avoid.

## Provenance

Surfaced during review of PR #1676 (the truncation-principle docs), where the
guidance text initially described this cap as satisfying the
declare-what-you-cut rule and had to be corrected: the cap is right, its
silence is the improvable half. Verified against the source on 2026-09-04.

be-student:
Closes #1707

## Summary
- count every drained stdout/stderr byte while retaining the existing 2 MiB cap
- report the exact total in truncation messages

## Tests
- `pytest tests/test_autonomy/test_executor/test_deterministic.py -q`
- `ruff check src/genesis/autonomy/executor/deterministic.py tests/test_autonomy/test_executor/test_deterministic.py`

Existing installs receive the runtime fix after pulling and restarting Genesis; fresh clones include it directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved output truncation messages to report the complete amount of data produced, rather than only the portion retained.
  * Standardized accurate size reporting for both standard output and standard error during execution.

* **Tests**
  * Added coverage for truncated output and full stream-size reporting.

### DIFF
--- a/src/genesis/autonomy/executor/deterministic.py
+++ b/src/genesis/autonomy/executor/deterministic.py
@@ -108,24 +108,26 @@ def validate_command(command: str) -> str | None:
 async def _read_limited(
     stream: asyncio.StreamReader,
     limit: int = _MAX_STREAM_BYTES,
-) -> bytes:
-    """Read up to *limit* bytes from *stream*, discarding the rest."""
+) -> tuple[bytes, int]:
+    """Retain up to *limit* bytes while counting and draining the full stream."""
     chunks: list[bytes] = []
-    total = 0
+    retained_size = 0
+    total_size = 0
     while True:
         chunk = await stream.read(8192)
         if not chunk:
             break
-        remaining = limit - total
+        total_size += len(chunk)
+        remaining = limit - retained_size
         if remaining <= 0:
             continue
         if len(chunk) > remaining:
             chunks.append(chunk[:remaining])
-            total = limit
+            retained_size = limit
         else:
             chunks.append(chunk)
-            total += len(chunk)
-    return b"".join(chunks)
+            retained_size += len(chunk)
+    return b"".join(chunks), total_size
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


WingedGuardian (approved):
Reviewed, and this is a good fix. Approving.

The defect is real and this repo had it documented as a known counterexample — a stream cap that discarded bytes silently while the caller reported the retained length as the total, so a 3 MiB stream read as `2097152 bytes total`. A clipped log that looks complete is worse than one that says it was clipped.

What I verified, rather than took on faith:

**The count is correct at the boundary.** `total_size += len(chunk)` runs before the `remaining <= 0` check, so chunks that are discarded entirely are still counted. That ordering is the whole fix and it is right.

**The drain property survives.** Past the cap the loop `continue`s rather than breaking, so the pipe keeps being read and the child never blocks on a full buffer. `chunks` stops growing at the same moment, so the reader stays constant-space. Both properties that made the original cap correct are preserved.

**The signature change is swept.** `_read_limited` has exactly two production call sites, both in the `asyncio.gather` this PR updates, and the tuple unpack covers them. Grepped the tree to confirm there is no third.

**Both tests fail against the pre-change code.** I re-ran them as mutations rather than trusting that they pass: reverting the message to `len(stdout_bytes)` goes red, and removing `total_size += len(chunk)` goes red. A narrowing test that passes both ways pins nothing, so this matters more than the green run.

**`seq 1 400000` is a well-chosen fixture** — 2,688,895 bytes, which genuinely exceeds the 2 MiB retention cap, so the test exercises the discard path rather than approximating it. Easy to get wrong by picking a number that stays under the cap and asserting on a path that never runs.

46 tests pass, ruff clean.

One note on timing, not on the change: merging is currently gated on a review bot that is rate-limited, so this may sit for a while at no fault of the PR. It is queued and nothing further is needed from you.

Thanks for picking this one up, and for adding the direct unit test on `_read_limited` alongside the end-to-end one — the pair is what makes the fix legible.

PR #1797, opened 2026-09-06
feat(awareness): alert when the code index is dead instead of logging it

WingedGuardian:
## The problem

The code indexer can give up on a repository **permanently** and report it only into a
log file. Measured on a live install:

- the main repo's index request was euthanized after 5 failed attempts, to a marker
  `scripts/lib/index_marker.py` itself describes as **"never retried"**;
- its database sat as a 164 MB `.db.corrupt` for **two weeks**;
- `code-intelligence-runner.log` carried **35** `index failed` lines (the last three
  `rc=143` — killed under the memory cap);
- the live server reported two indexed projects: a worktree whose root no longer exists,
  and a 26-node scratch dir. The main repo was **not indexed at all**.

Nothing surfaced any of it: no awareness check covered code-intel, and the only `src/`
consumer of the marker system is a *writer*. It did not fail silently — it failed loudly
into a log, which is operationally identical. Same generator as the SessionStart-injection
silence (#1556), one subsystem over.

## The fix

An hourly awareness check reading the indexer's **own on-disk artifacts** — euthanized
request markers, and whether an index exists for the configured target. It never asks the
code-intel tool how it is doing: a crashed indexer cannot report its own health, and its
daemon may not be running.

Scope is deliberately narrow (the configured target only). Worktree indexes churn
constantly here — a live example is a 31,431-node index whose root is gone — and alerting
on those is how an alarm gets muted, taking the real signal with it.

Priority `high` (morning report), **not** `critical`: a dead index degrades tool quality;
it does not lose the user's data or session context.

## Two traps found in adversarial review

Both the same shape — the *watcher's* contract verified, the *producer's* asserted:

1. **`.db.corrupt` is a retained BACKUP, not a live-state flag.** The indexer's own binary
   says `backing up corrupt db to .corrupt`, rebuilds the db in place, and never unlinks
   the backup. Keying on its existence fired the alarm **forever after a successful
   rebuild** — a permanently-wrong alarm, which is precisely what this check exists to
   prevent. Now: corrupt only when the backup is *newer* than the live db. Regression
   cells in both directions, because a blanket "ignore `.corrupt`" would blind the check.
2. **`indexed_path` steers only this check.** Every index-request writer (post-commit
   hook, `disk_reclaim`, the gitnexus surplus job) hardcodes the repo root, and nothing in
   the indexing path reads this config — so the alert's original remedy ("narrow
   `indexed_path`") would have made the check watch an index nothing builds. The remedy
   and all doc sites now say narrowing means re-pointing the *writers* and matching the
   knob; setting it alone reports ABSENT forever.

Also fixed: an `any(*.json)` fresh-install guard that accepted *another* repo's pending
request (marker filenames are opaque hashes); a non-object marker raising `AttributeError`
straight out of the collector into a bare `except Exception: logger.warning` — this
module's own failure mode applied to itself; two unescaped values reaching a first-party
observation; and a vacuous escaping test whose injected value was dropped by an upstream
scope filter before any escaping happened.

One consciously accepted limit, stated rather than inherited silently: a raise *outside*
the collector is still log-only. The sibling `_check_context_injection_health` has the same
hole, and fixing one while leaving the other is worse than fixing the class deliberately —
that belongs in its own change. The concrete trigger is fixed here.

## Verification

198 targeted tests green (20 in the new suite); ruff clean; 4 local CI guards OK.
**Acceptance bar: run against real live state, it fires on the actual two-week-old
defect.** Live E2E through the real awareness path: exactly one `high` alert after two
ticks (no re-paging), resolving on recovery and on disable.

## Summary by CodeRabbit

- **New Features**
  - Added hourly monitoring for code-intelligence index health.
  - Added configurable monitoring controls for enablement, indexed path, and alert priority.
  - Added detection of missing, stale, corrupted, unreadable, or incomplete index data.
  - Added clear health findings based on available index artifacts.
  - Added high-priority alerts with duplicate prevention and automatic resolution when issues recover or monitoring is disabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,19 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ### Added
 
+- **A dead code index now says so.** The code indexer could give up on a
+  repository permanently and report it only into a log file: on one install the
+  main repo's index request was abandoned after five failed attempts, its
+  database sat corrupt for two weeks, and code-intelligence tools kept answering
+  from nothing — with no alert anywhere, because nothing read the indexer's
+  failure records. An hourly check now reads those records directly (abandoned
+  index requests, and whether an index actually exists for the path being
+  indexed) and raises one self-resolving alert that clears when the index comes
+  back. It reads the indexer's own files rather than asking the tool how it is
+  doing, since a crashed indexer cannot answer. Configurable via
+  `code_intel_health` (`enabled`, `indexed_path`, `alert_priority`); set
+  `indexed_path` to whatever the indexer is actually pointed at.
+
 - **A `tmux kill-server` with no socket binding now draws an advisory.** tmux
   resolves its target server from the inherited `$TMUX` variable before
   `TMUX_TMPDIR`, so a cleanup aimed at a scratch or probe server can address
--- a/config/code_intel_health.yaml
+++ b/config/code_intel_health.yaml
@@ -1,0 +1,38 @@
+# Code-intel index health (awareness hourly band).
+#
+# The code indexer can fail permanently and say nothing. MEASURED on a live
+# install 2026-09-05: the main repo's index request was euthanized to a
+# <hash>.failed.json marker after 5 genuine failures (a state index_marker.py
+# describes as "never retried"), its database sat as a 164 MB .db.corrupt for
+# two weeks, and the runner logged "index failed" 35 times — while no awareness
+# check covered code-intel and nothing in src/ read either signal. It did not
+# fail silently; it failed loudly into a log file, which is the same thing.
+#
+# This check reads the indexer's OWN artifacts on disk (euthanized request
+# markers + the index database for the configured target). It never asks the
+# code-intel tool how it is doing: a tool that has crashed cannot report its own
+# health, and its daemon may not be running at all.
+#
+# Values here merge over built-in defaults; a .local.yaml overlay merges over
+# this file. Env kill switch: GENESIS_CODE_INTEL_HEALTH_DISABLED=1.
+
+enabled: true
+
+# Which path code-intel is expected to have an index for. Empty = repo root.
+#
+# LEAVE THIS EMPTY unless you have also re-pointed the indexer. Every writer of
+# an index request hardcodes the repo root today (the post-commit hook,
+# disk_reclaim, the gitnexus surplus job), and the runner indexes whatever the
+# request names — nothing in that path reads this file.
+#
+# The knob exists because the index's on-disk name derives from the FULL path,
+# so an install that genuinely indexes a subdirectory needs this check to
+# follow. But it steers ONLY this check: set it on its own and the check will
+# look for an index nothing ever builds, reporting ABSENT forever. Narrowing the
+# indexer means changing the writers AND setting this to match.
+indexed_path: ""
+
+# high = the morning report. A dead code index degrades tool quality; it does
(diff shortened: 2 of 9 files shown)


WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:

### Structural review

**Warnings**

- `src/genesis/awareness/code_intel_health_config.py` (new file) has no direct unit-test coverage of its own logic. `tests/test_awareness/test_code_intel_health.py` monkeypatches `load_config`/`indexed_path` for the awareness-wiring tests rather than calling the config module's functions directly, so these paths are untested: `load_config()`'s layer-by-layer degrade-to-DEFAULTS behavior on a missing/corrupt base file or overlay; `is_enabled()`'s non-bool fail-open guard (`enabled: null`/`0`/`[]`); `indexed_path()`'s `OSError` fallback when `.resolve()` fails; `alert_priority()`'s fallback when the configured value isn't in `_VALID_ALERT_PRIORITY`. Given the module's own docstring stresses "config damage degrades toward WATCHING," a direct test that a damaged config actually keeps the check enabled (rather than just asserting the intent in a comment) would close the gap.

**Notes**

- `default_cache_dir()` in `src/genesis/observability/snapshots/code_intel_health.py` has no docstring, unlike its siblings (`default_marker_dir()`, `index_slug()`, `collect()`, `derive_findings()`, `alert_identity()`) in the same file.

No structural pattern violations found otherwise: no `nohup`, no raw SQL against `genesis.db` in production code (uses `observations.resolve_by_source_and_type` / `supersede_except_hash` / `create`), no bare `except:`, no mutable default arguments, import grouping (stdlib → third-party → local) is correct in both new modules, no star imports, no circular-import risk (lazy imports inside functions match the existing sibling pattern in `awareness/loop.py` and `mcp/health/settings.py`), and the PR title carries a conventional `feat(...)` prefix.

### PII/Secrets/Wording scan

PII/Secrets: clean (soft warnings only) — no API keys, tokens, private IPs, SSH keys, or connection strings in the diff.

**Soft warning**: a hardcoded absolute path containing a username-style directory, `/home/ubuntu/tmp/kimi_review`, appears twice as an illustrative example — in the `index_slug()` docstring in `src/genesis/observability/snapshots/code_intel_health.py`, and in the `_index()` helper docstring in `tests/test_awareness/test_code_intel_health.py`. This is a documentation example (verifying the slug-derivation shape against a live observation), not a live secret, but it matches the checklist's named pattern for hardcoded `/home/ubuntu/...` paths, so flagging for awareness — consider genericizing the example path if this file is ever surfaced outside the repo.

PR title/body and the CHANGELOG.md addition were scanned for leak/incident wording: none found. The write-up describes a health-monitoring gap and its measured symptoms (abandoned index markers, a `.db.corrupt` file, log line counts) — this is a bug-fix narrative about an internal observability defect, not a disclosure of a security incident, exploit detail, or specific private infrastructure (no IPs/hostnames named).

No hard blocks, no structural errors, no wording issues.

VERDICT: PASS



---

WingedGuardian:
@codex review

WingedGuardian:
Superseded by #1802 — identical content on a clean branch.

The private-pattern hard gate sweeps `base..HEAD` **per commit**, so it re-flagged two `/home/<user>` docstring examples that the first commit added and the second removed (net diff: 0 matches; commit `8e015167`: 2). Force-push to this repo is banned, so recreating the branch is the remedy rather than amending. All review findings and fixes carry over unchanged.

PR #1802, opened 2026-09-06
feat(awareness): alert when the code index is dead instead of logging it

WingedGuardian:
## The problem

The code indexer can give up on a repository **permanently** and report it only into a
log file. Measured on a live install:

- the main repo's index request was euthanized after 5 failed attempts, to a marker
  `scripts/lib/index_marker.py` itself describes as **"never retried"**;
- its database sat as a 164 MB `.db.corrupt` for **two weeks**;
- `code-intelligence-runner.log` carried **35** `index failed` lines (the last three
  `rc=143` — killed under the memory cap);
- the live server reported two indexed projects: a worktree whose root no longer exists,
  and a 26-node scratch dir. The main repo was **not indexed at all**.

Nothing surfaced any of it: no awareness check covered code-intel, and the only `src/`
consumer of the marker system is a *writer*. It did not fail silently — it failed loudly
into a log, which is operationally identical. Same generator as the SessionStart-injection
silence (#1556), one subsystem over.

## The fix

An hourly awareness check reading the indexer's **own on-disk artifacts** — euthanized
request markers, and whether an index exists for the configured target. It never asks the
code-intel tool how it is doing: a crashed indexer cannot report its own health, and its
daemon may not be running.

Scope is deliberately narrow (the configured target only). Worktree indexes churn
constantly here — a live example is a 31,431-node index whose root is gone — and alerting
on those is how an alarm gets muted, taking the real signal with it.

Priority `high` (morning report), **not** `critical`: a dead index degrades tool quality;
it does not lose the user's data or session context.

## Two traps found in adversarial review

Both the same shape — the *watcher's* contract verified, the *producer's* asserted:

1. **`.db.corrupt` is a retained BACKUP, not a live-state flag.** The indexer's own binary
   says `backing up corrupt db to .corrupt`, rebuilds the db in place, and never unlinks
   the backup. Keying on its existence fired the alarm **forever after a successful
   rebuild** — a permanently-wrong alarm, which is precisely what this check exists to
   prevent. Now: corrupt only when the backup is *newer* than the live db. Regression
   cells in both directions, because a blanket "ignore `.corrupt`" would blind the check.
2. **`indexed_path` steers only this check.** Every index-request writer (post-commit
   hook, `disk_reclaim`, the gitnexus surplus job) hardcodes the repo root, and nothing in
   the indexing path reads this config — so the alert's original remedy ("narrow
   `indexed_path`") would have made the check watch an index nothing builds. The remedy
   and all doc sites now say narrowing means re-pointing the *writers* and matching the
   knob; setting it alone reports ABSENT forever.

Also fixed: an `any(*.json)` fresh-install guard that accepted *another* repo's pending
request (marker filenames are opaque hashes); a non-object marker raising `AttributeError`
straight out of the collector into a bare `except Exception: logger.warning` — this
module's own failure mode applied to itself; two unescaped values reaching a first-party
observation; and a vacuous escaping test whose injected value was dropped by an upstream
scope filter before any escaping happened.

One consciously accepted limit, stated rather than inherited silently: a raise *outside*
the collector is still log-only. The sibling `_check_context_injection_health` has the same
hole, and fixing one while leaving the other is worse than fixing the class deliberately —
that belongs in its own change. The concrete trigger is fixed here.

## Verification

198 targeted tests green (20 in the new suite); ruff clean; 4 local CI guards OK.
**Acceptance bar: run against real live state, it fires on the actual two-week-old
defect.** Live E2E through the real awareness path: exactly one `high` alert after two
ticks (no re-paging), resolving on recovery and on disable.

## Summary by CodeRabbit

* **New Features**
  * Added configurable, hourly monitoring for code-intelligence index health.
  * Added detection of missing, corrupt, unreadable, stale, or failed indexes and abandoned requests.
  * Added configurable indexed paths, alert priority, enablement controls, and live settings reload.
  * Added deduplicated infrastructure alerts with recovery resolution and remediation guidance.
* **Tests**
  * Added comprehensive coverage for healthy, degraded, failed, disabled, and recovered index states.


WingedGuardian:
## Structural review

No ERROR-level pattern violations found (no nohup, no direct `task_states` writes, no `/tmp/` usage, no `git add -A`/force-push patterns, no bare-except, no mutable default args, no raw SQL in production code paths — `loop.py` correctly routes through the `observations` crud module for all reads/writes). Import grouping (stdlib → third-party → local) is correct in all three touched/new Python modules. Commit/PR title carries `feat(awareness):` prefix.

### WARNING — test coverage gaps (not blocking)

- `src/genesis/awareness/code_intel_health_config.py` (new file): `tests/test_awareness/test_code_intel_health.py`'s `wired` fixture monkeypatches `cfg_mod.load_config` and `cfg_mod.indexed_path` directly, so this module's own implementation is never exercised end-to-end except for the env kill switch (covered via `test_disabling_the_check_resolves_a_standing_alert`). Untested: YAML merge/overlay behavior in `load_config()` (corrupt base file, `.local.yaml` overlay), the non-bool `enabled` fallback in `is_enabled()`, the invalid-`alert_priority` fallback in `alert_priority()`, and the symlink/`..`-resolution behavior documented in `indexed_path()`'s docstring. Consider a dedicated `test_code_intel_health_config.py`.
- `src/genesis/mcp/health/settings.py:_validate_code_intel_health` (new function): no test in this diff exercises its validation branches (unknown key, non-bool `enabled`, invalid `alert_priority`, non-string `indexed_path`).
- `src/genesis/memory/provenance.py`: one-line addition of `code_intel_health_monitor` to the first-party source set — no test in this diff directly asserts this new entry's classification (may be covered by an existing generic test over the set; not visible in this diff).

### NOTE — minor

- `tests/test_awareness/test_code_intel_health.py`'s `_alerts()` helper runs a raw `SELECT ... FROM observations` against the test's in-memory sqlite connection rather than a crud accessor. This is a test-only assertion against a `:memory:` fixture DB (not `genesis.db`), so it's very likely fine as written — flagging only for awareness of the raw-SQL convention.

No circular-import, star-import, hook-dependency, `$CLAUDE_PROJECT_DIR`, or `IntervalTrigger`/`CronTrigger` issues found. All new public functions/classes carry docstrings.

## PII/Secrets/Wording scan

Scanned all added (`+`) lines across the diff, plus PR title, body, and the `CHANGELOG.md` addition.

PII/Secrets/Wording: clean (no soft warnings either) — no API keys/tokens/secrets, no private IPv4/IPv6 ranges, no SSH keys or connection strings with credentials, no personal emails, no hardcoded paths with real usernames (docstring examples use a `<user>` placeholder, not an actual account), no timezone/AWS/GCP/hardware identifiers, no private hostnames. The PR body's "MEASURED" details (164 MB `.db.corrupt`, 35 log lines, node counts) describe an internal indexer-reliability defect in generic terms — not a PII/security-incident disclosure and not specific enough to be exploitable.

VERDICT: PASS



---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
Review work claimed by `codex-01a091e8-d482-78a3-bc7f-4daf10bd0bdd`.

Pushed b192b3549: merged current main to refresh parser integration and CI. Independent integration review verified awareness-loop and settings-registry parent preservation. Local parser/chokepoint suite: 167 passed. Atomic-write check: 60 sites, 30 clean, 30 known-dirty, zero NEW.

This is an integration refresh, not a resolution of the original health-state, marker-identity, recovery, priority, or TTL review findings. No finding is cleared by this comment.

Ready to merge: No — original review findings remain; fresh CI required.

WingedGuardian:
@codex review

WingedGuardian:

Pushed 1cd8305b1ea976ad2f7d3fb248cbfb0d50f59f8c for the owner-requested shared audit correction. The dependency-audit step now reports failures without failing the job, matching the verified #1960 reporting policy. Setup/install failures remain blocking. All existing audit commands and ignore arguments are preserved; no dependencies or vulnerability suppressions changed. SECURITY/Dependabot descriptions now match the policy.

WingedGuardian:
@codex review

Head moved to 03c7cd882. Merged origin/main (the branch was 22 behind and CONFLICTING) and fixed two blockers an internal adversarial audit found — both of them the alarm committing the failure it exists to prevent.

1. It fired high on every healthy fresh install. install.sh and setup_claude_config.py both queue an index request at setup, so a new clone has a pending marker within seconds, while the first index legitimately takes hours — the runner is idle-gated and only relaxes after RELAX_AFTER_S. The rule was absent AND anything-ever-asked, so it accused the runner of failing while the runner was correctly waiting. The remedy compounded it: it told the operator to clear the euthanized marker, and in that state the only marker present is the pending one that would eventually build the index. An absent index now speaks only once the request is euthanized or has waited past the runner's own relax window, and the recovery text is derived from the state.

2. A .failed.json tombstone is written once and deleted by nothing, while the request is recreated routinely by the post-commit hook, the twice-weekly gitnexus reindex and disk_reclaim. So the GAVE UP alarm outlived the rebuild that answered it and could never self-clear — the same defect already fixed for .db.corrupt forty lines below, not carried across. A tombstone older than the live index is now history. The producer-side half (index_marker.consume should unlink it on success) belongs to that module's own change and is deliberately not done here.

Also: the headline no longer claims an outage when the only finding is an unreadable marker; the error list is bounded in both the text and the alert identity, so a new unreadable file cannot mint a fresh identity and re-page.

The fresh-install control was an unfaithful fixture — it passed a marker dir that does not exist, which is a state no real install has, and is exactly why it could not catch blocker 1. Rebuilt in the producer's own marker shape.

Verify-RED: both blocker cells fail against the old behaviour, both opposite-direction controls hold under both binaries, files restored byte-identical. Acceptance bar re-run live and read-only after the fixes — still fires on this box's real corrupt index and euthanized request.

Two limits stated rather than hidden: GitNexus keeps its own store and is not covered by this check, and test_a_worktree_index_with_a_missing_root_is_out_of_scope is a forward guard against scope creep rather than a pin on behaviour — it cannot fail for the reason its docstring states.

WingedGuardian:
Leaks / privacy review at this head. 1341 added lines scanned against the install-private and personal-context patterns: addresses (v4/v6, tailnet), hostnames, MAC, home paths carrying a username, token shapes, and household/inferential framing. **0 real hits.**

The scan carried a seeded positive control (an address plus a username-bearing home path) which was caught, so the zero is measured rather than a blind scanner reporting nothing.

This one deserves an explicit note because the module is ABOUT this install's own failure. The docstring and changelog describe the measured incident — an index request euthanized after five attempts, a 164 MB corrupt database retained for two weeks, 35 failure lines in a runner log — and those are stated as install-generic facts with no path, host or user identifying where they happened. Paths that do appear in code are either generic (~/.genesis/index-requests) or constructed at runtime from the configured target. The slug docstring's worked example is written /home/<user>/tmp/scratch deliberately.

Every test fixture is synthetic and built under tmp_path. The one place a real path could reach an observation is the marker payload, and it is escaped at ingestion (_safe_path) before it can land in a first_party observation — which is a correctness property of the module rather than a privacy artifact of this diff, but it is the reason a real repo_path in an alert is bounded rather than free text.

Also carried from the origin/main merge: two lines in tests/test_hooks/test_hook_output.py that the pre-push guard flags. They are already on public main and read /home/u/, a synthetic placeholder. Not introduced here.

WingedGuardian:
## Kicked back to draft — needs changes, and a fresh attempt rather than a fix round

Maintainer decision. This is **not** abandoned and should not be closed: the underlying problem is real and the motivating defect was measured. What is being rejected is this attempt at it.

### Why: the findings are one module, not twenty-one bugs

Counted across both reviewers at the current head — **21 open findings**, distributed:

| file | findings |
|---|---|
| `observability/snapshots/code_intel_health.py` | **15** |
| `awareness/loop.py` | 5 |
| `awareness/code_intel_health_config.py` | 1 |

So **71% of everything sit in the single new collector module this PR adds**, and they cluster by subject rather than scattering: marker validation and rejection, malformed-input reporting, matching, and the conditions under which the alarm is allowed to fire.

That distribution is the reason for the kickback. Twenty-one unrelated defects would be a fix round. Fifteen findings concentrated in one new file, about how it validates its inputs and decides to alert, is a statement about the module's design — and fixing them one at a time is the whack-a-mole shape this repo's escalation cap exists to interrupt. A fix round here would very likely produce a second round on the fix code, which is exactly what happened on three other PRs from this same session.

### What is worth keeping when it is retried

- **The motivating defect is real and was measured**: the main repository's code index had been dead for roughly two weeks and nothing surfaced it. The index request had been euthanized at max attempts and is, by design, never retried. It did not fail silently — it failed loudly into a log nothing reads, which is operationally identical.
- **The alarm/watcher shape itself was not what drew findings.** The wiring, the observation sink, and the config lever attracted 1 finding between them.
- **Two real defects were already found and fixed during this PR's own internal review**, and both are worth carrying into any retry rather than rediscovering: the `.db.corrupt` file is the indexer's retained *backup*, not a live-state flag, so keying an alarm on its existence makes the alarm permanent after a successful rebuild; and the configured index target steers only the *check*, not any of the writers, so the alarm's own suggested remedy would have created a permanent false alarm.

### Suggested shape for the retry

Start from the reviewers' 15 findings on the collector as a **specification of what its input handling must do**, and design that first — rather than porting this module and patching toward them. Several ask for the same underlying thing: be explicit and loud about a marker the collector cannot interpret, instead of silently treating it as a clean or absent signal.

The other 6 findings look separable and may not need the same rethink.

### Status

Draft, labelled `wip`. No further review rounds should be requested against this head; the next review should be of a fresh attempt.

PR #1806, opened 2026-09-06
feat(memory): capture the domain a preference is scoped to

WingedGuardian:
## What

Extraction now captures **which domain a preference is scoped to**.

A preference is not a scalar fact. "Favorite color" can be red for work materials
and blue this month because of a new car — both true, about different contexts.
Extraction already classified `speech_act="preference"`, but captured nothing about
*what the preference is about*, so two conflicting preferences could only resolve as
newest-wins, silently discarding the older context.

## The chain

`normalize_preference_domain` (open vocabulary — shapes rather than validates, since
domains are an open set; returns `None` unless the **normalized** speech act really is
`preference`) → prompt bullet + a positive JSON example → `Extraction` field →
tolerant parse → `extractions_to_store_kwargs` → `MemoryStore.store()` →
`create_metadata` → `memory_metadata.preference_domain`, present on **both** schema
build paths (timestamp migration `20260906042425` + the `_migrate_add_columns` mirror
+ the canonical CREATE).

**Write-only**, exactly like its five MW-1 siblings — nothing reads it yet. The
consumer is MW-4 recall ranking, where a conflict between two domain-scoped
preferences resolves into coexisting truths instead of an overwrite.
`# GROUNDWORK(mw-4-preference-domain)`.

## Verification

- 13 tests written RED first; final ring **93 green** (judgment, migration, store,
  base-path upgrade, extraction).
- **Mutation-verified where it counts.** The adversarial review measured that the two
  most load-bearing lines — the `create_all_tables` mirror and the `store()` threading
  line — could each be deleted with **0/89** test failures (the second would leave the
  column always-NULL in production with no signal). Both now have guards, and both
  mutations were re-run after the fix: they fail 2 and 1 tests respectively.
- Also caught by review and fixed: the only few-shot example of the new field was a
  *negative* one (`null` on a non-preference item), contradicting the instruction and
  never showing the intended shape.
- ruff, migration-id (`--base origin/main`), subsystem-map, frozen-clock and
  external-io guards all clean; privacy grep clean.

## Deliberately not done

- **`normalize_speech_act` not widened to case-insensitive.** An LLM emitting
  `"Preference"` normalizes to `"observation"` and the domain drops with it. Case-folding
  would also promote `"Rule"`/`"Decision"` into `PROTECTED_SPEECH_ACTS` — a
  protection-eligibility change owned by MW-1/MW-5, not by this satellite. The behavior
  is documented at the decision point and locked by a test so the next reader sees a
  decision rather than an oversight.
- **No CHANGELOG entry**, consistent with the five sibling columns: write-only, no
  user-visible effect. The dark-column class is now documented in the architecture map
  instead, which had six such columns and no entry despite its own dark-code convention.

## Summary by CodeRabbit

* **New Features**
  * Added support for capturing and storing an optional preference domain with memory metadata.
  * Preference domains are normalized for consistent formatting and retained only for valid preference judgments.
  * Existing databases are upgraded automatically to support the new metadata.

* **Tests**
  * Added coverage for persistence, migrations, normalization, extraction, and default behavior.

### DIFF
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -260,6 +260,19 @@ at gate time (`security/immunity.py`). Migration 0053 backfilled history
 (no owner heuristics); `scripts/backfill_origin_class_qdrant.py` mirrors the
 payloads idempotently.
 
+**Judgment axes (DARK — written, no reader):** `memory_metadata` carries six
+write-only classification columns stamped at extraction time and consumed by
+nothing yet — `speech_act` + `speech_act_confidence`
+(# GROUNDWORK(mw-5-speech-act-protection)), `assertion_provenance`
+(# GROUNDWORK(mw-4-provenance-weight)), `durability` + `expires_at`
+(# GROUNDWORK(mw-4-durability-ttl)), and `preference_domain`
+(# GROUNDWORK(mw-4-preference-domain) — the domain a preference is scoped to,
+so a later conflicting preference coexists as a different-context truth
+instead of overwriting). Contract + normalizers in `memory/judgment.py`;
+distinct from `origin_class` above, which is a pipeline-trust label. Expiry is
+opt-in (`durability='temporary'` + an elapsed `expires_at` only), so an
+unclassified row never expires.
+
 ## 2. Execution — CC sessions (DirectSession)
 
 Spawning, tracking, and recovering Claude Code sessions — Genesis's hands for
--- a/src/genesis/db/crud/memory.py
+++ b/src/genesis/db/crud/memory.py
@@ -353,6 +353,7 @@ async def create_metadata(
     assertion_provenance: str | None = None,
     durability: str | None = None,
     expires_at: str | None = None,
+    preference_domain: str | None = None,
 ) -> str:
     """Insert a row into memory_metadata. Returns memory_id.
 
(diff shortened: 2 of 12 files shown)


WingedGuardian:
@codex review

WingedGuardian:
Automated first-pass review (structural + PII/secrets/wording) on head `9059c3b`.

**Domain 1 — Structural**
Checked all 8 modified/added `.py` files (`db/crud/memory.py`, new migration `20260906042425_preference_domain.py`, `db/schema/_migrations.py`, `db/schema/_tables.py`, `memory/extraction.py`, `memory/judgment.py`, `memory/store.py`, plus 4 test files):
- No nohup/task_states-insert/`/tmp/`/force-push/reset-hard/bare-except/mutable-default/`os.killpg` patterns.
- Raw SQL only appears in `db/crud/memory.py` (the designated CRUD layer) and the two schema/migration modules — consistent with the "except in migrations" carve-out.
- Import grouping in the new migration file and new test file is correctly stdlib → third-party → local; no star imports, no circularity introduced.
- The one new `src/` file (the migration) has a matching test file (`tests/test_db/test_migration_preference_domain.py`); all four changed functions (`create_metadata`, `parse_extraction_response_full`, `extractions_to_store_kwargs`, `store()`) have corresponding test updates.
- New public function `normalize_preference_domain` is fully docstringed; changed docstrings on `create_metadata` reflect the new parameter.
- PR title carries a conventional prefix (`feat(memory): ...`).

Scanned all added (`+`) lines across the diff and the PR title/body (no `CHANGELOG.md` touched). No API keys/tokens/secrets, no private IPv4/IPv6 ranges, no SSH keys, no connection strings, no personal emails, no username-bearing absolute paths, no location-revealing tz identifiers, no cloud account IDs, no hardware identifiers or private hostnames. PR title/body describe a write-only schema/extraction change (domain-scoped preferences) with no security-incident or infra-exposure language.

VERDICT: PASS



---

WingedGuardian:
@codex review

PR #1810, opened 2026-09-06
docs(security): correct SECURITY.md to the posture that actually ships

WingedGuardian:
## What was wrong

`SECURITY.md` was last substantively touched 2026-07-10. Ninety-six commits
have landed under `dashboard/`, `autonomy/` and `security/` since, and the doc
describes the posture from before them.

The consequential one: it told operators to treat the dashboard API as
**"unauthenticated administrative access"** and stated the dashboard password
*"protects the web UI, not the programmatic API."* Neither has been true since
the API mutation gate landed. `check_api_mutation_auth` (`dashboard/auth.py`)
runs as an app-level `before_request` and requires either the internal bearer
token or an authenticated same-origin cookie with a CSRF check from
`Sec-Fetch-Site`/`Origin`/`Referer`, failing closed. `/v1` enforces its own
separate bearer.

## Why the fix is careful rather than a flip

The error ran in the **safe** direction — it told operators to rely on network
isolation, which they should. Rewriting it to "the API is authenticated" would
turn a safely-wrong doc into a dangerously-wrong one and invite someone to
relax a firewall, because the gate is conditional. Every condition is now
stated: mutations only (`GET`/`HEAD`/`OPTIONS` stay open), inert when no
dashboard password is set, exempts `/api/genesis/auth/*`, disableable via
`GENESIS_DASHBOARD_API_AUTH=off`, and does not cover the web terminal or noVNC.
Network isolation is described as the primary control, with the gate as a
second layer.

## Other corrections, each verified against the tree

| Claim | Reality |
|---|---|
| "seven levels (L0-L6)" | `AutonomyLevel` ships `L1`-`L4`; its docstring defers `L5`-`L7`; no `L0` exists |
| "six categories" | `AutonomyCategory` has four; now named |
| `os.killpg` PGID listed as a PreToolUse policy | no hook matches `killpg`/`PGID` — that is `genesis.util.proc_kill`, runtime library hardening |
| hooks "enforced by" one linter | that linter is matched on `Write`/`Edit` and lints file content; shell and web guards are separate programs |

The level/category errors predate the drift window — aspirational design-doc
language that entered the security doc as fact.

## Newly documented

These shipped with no representation at all:

- **Provenance stamping** and the two privileged-write paths gated on it
  (`memory/user_model.py`, `autonomy/dispatcher.py`, via
  `security.immunity.is_trusted_for_privileged_write`, fail-closed on `NULL`).
  Deliberately scoped — gating two auto-consume paths is *not* "untrusted
  content is isolated", and the doc says so explicitly.
- **Approval + pre-delete journal** in front of irreversible entity merges.
- **Shared session-id validation** before filesystem use.
- **External Egress**, which states plainly that mail enforces a gate today
  while the Discord and GitHub doors are observe-only. A shadow gate described
  as a control is worse than no line at all.
- The **autonomous-CLI approval gate**, the hard boundary under the whole
  autonomy ladder, previously unmentioned.

## Verification

Every statement was re-derived from code, not from the changelog. Privacy scan
of the diff is clean: no IPs, home paths, emails, hostnames or real names; the
CGNAT range stays a class reference, not an address. No exploit path or guard
defeat-condition is described.

`Supported Versions` needs no change — `v3.x` already covers v3.0b18.

Follow-up `423245b5` (MW-7 part 2). Its other half — a doc-drift guard so
security docs cannot silently lag code again — is **not** in this PR and is the
part that actually prevents recurrence.

---

E2E: none — prose only. `SECURITY.md` and `CHANGELOG.md` have no runtime path to exercise. Every corrected claim was verified against the code it describes (`dashboard/auth.py`, `autonomy/types.py`, `autonomy/shadow_gate.py`) by reading that code, which is the appropriate verification for a document whose defect was describing the system wrongly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,36 @@ Versioning follows Genesis release stages (v3.0a → v3.0b → v3.1 → v4.0a…
 
 ## [Unreleased]
 
+### Fixed
+
+- **SECURITY.md described a posture the code left behind two months ago.** The
+  security policy told operators to treat the dashboard API as
+  "unauthenticated administrative access" and said the dashboard password
+  "protects the web UI, not the programmatic API". Neither has been true since
+  the API mutation gate landed: with a password set, state-changing `/api`
+  requests require a bearer token or an authenticated same-origin cookie with a
+  CSRF check, and `/v1` enforces its own bearer. The doc now says so -- along
+  with the limits that decide whether you still need network isolation, because
+  you do: the gate covers mutations only, reads stay open, it is inert when no
+  password is set, and it has a documented kill switch.
+
+  Corrected in the same pass: the autonomy ladder was described as seven levels
+  (`L0`-`L6`) when four ship (`L1`-`L4`, the rest deferred), the per-category
+  permissions were described as six categories when there are four, and the
+  process-group-kill example was attributed to a PreToolUse hook when it is
+  runtime library hardening -- a different guarantee, since a hook cannot be
+  bypassed by the agent and a helper only protects its own call sites. The
+  guard section also named one linter as the enforcement mechanism for shell
+  and URL policy; that linter only ever sees file edits, and the shell and web
+  guards are separate programs chosen by tool matcher.
+
+  Newly documented rather than corrected: provenance stamping and the two
+  privileged-write paths gated on it, the approval requirement and journal now
+  standing in front of irreversible entity merges, session-id validation before
+  filesystem use, and an External Egress section that says plainly which
+  outbound channel actually enforces a gate today and which are only observed.
+
+
 ### Changed
 
 - **The session charter now lists every open ledger item, not just the oldest
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -43,23 +43,33 @@ layer described below exists because Genesis takes seriously the question:
 
 ### Earned Autonomy as Security Model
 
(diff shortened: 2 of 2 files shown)


WingedGuardian:
@codex review

WingedGuardian:
**Automated first-pass review — docs-only PR (CHANGELOG.md, SECURITY.md)**

Per process, structural review is skipped for a markdown-only change; PII/secrets/wording scan run instead.

Structural review: skipped (documentation-only PR — no `.py` files changed)

PII/Secrets/Wording scan: clean (no hard blocks, no soft warnings)
- No API keys, tokens, private IPs, SSH keys, or credentialed connection strings in the added lines. The `100.64.0.0/10` CGNAT range reference is pre-existing, unchanged context and is a class reference, not a specific address.
- No personal emails, username-bearing paths, revealing timezones, cloud account IDs, or private hostnames.
- PR title/body/CHANGELOG wording: describes correcting SECURITY.md to match already-shipped behavior (autonomy levels, dashboard API mutation gate and its limits, provenance stamping, egress gate status). Nothing here announces a leak/incident, names concrete private infrastructure, or discloses an exploit — including the candid note that the Discord/GitHub egress gate is observe-only, which documents an existing, already-shipped limitation rather than a new vulnerability.

No blocking findings.

VERDICT: PASS



---

PR #1811, opened 2026-09-06
feat(ci): catch a test that stops running, on both axes

WingedGuardian:
## The gap

A test that stops running is invisible. It does not turn the suite red — it turns it
**smaller**, and a smaller green suite reads exactly like a healthy one. CI had no
assertion on collection or skip counts at all.

## Two axes, because dormancy moves both ways

One number cannot see both directions, and the first version of this shipped with only
one of them:

| failure | effect on counts | caught by |
|---|---|---|
| a `skipif` breaks and becomes always-true | skip count **rises** | `--ceiling` |
| a marker lost in a refactor | skip count **falls** | `--min-collected` |
| a directory falls out of collection | both fall | `--min-collected` |

**Measured on a 4-test sandbox at `--ceiling 1`, before the floor existed:**

```
half the suite dropped from collection → "1 skipped of 2 collected. OK."   exit 0
skip marker deleted                     → "0 skipped of 4 collected. OK."   exit 0
```

So the ceiling detected one of the three modes its own docstring named, and its failure
message listed the two it categorically cannot detect — an inverted diagnosis headed for
permanent record. Each message now names only the causes its own axis can produce, and
the ceiling message explicitly redirects to `--min-collected` for the rest.

**Acceptance bar, replayed after the fix:**

```
healthy    → "1 skipped of 4 collected, ceiling 1, floor 4. OK."              exit 0
dropped    → FAIL "2 tests collected, floor is 4 … stopped being collected"   exit 1
skipif     → FAIL "3 skipped, ceiling is 1 … NOT a lost marker or a dropped
              directory — those LOWER this number; --min-collected is the
              check for them."                                               exit 1
```

## Refusal over guessing

"Could not measure" and "measured, and it is fine" must never share an exit code. Exit 2
on: a missing report, an unparseable one, a run that collected **zero** tests, a suite
omitting its counts, negative counts, a nested `<testsuite>` that would double-count, and
a nonsense bound. Every refusal names itself, because the interpreter *also* exits 2 when
it cannot open the script — a caller keying on the code alone could not tell them apart.

## Why it ships report-only

The honest thresholds are numbers measured on the CI runners. I could not obtain them
beforehand (the run log truncates at 45%, the raw API log had expired), and a threshold
invented rather than measured is one nobody trusts. **This PR's own `test` job prints
both numbers; they go in as a second commit on this branch before merge.**

The step is not inert meanwhile — a missing or unreadable report still exits 2. And
report mode refuses to offer a paste-ready baseline from a run that had errors or
failures: the step runs under `!cancelled()`, so it runs on red runs too, which is
exactly when the counts describe a corpus that did not finish.

## Also here

`~/.claude/settings.json` becomes a ninth credential-integrity target with its matching
backup entry. The **project** settings file is tracked in git; the **user-level** one is
in neither git nor the backup set, and carries its own live hooks block — so a
process-internal write that empties it disarms those hooks with nothing to notice.

It deliberately declares **no** `required_keys`: `missing_keys` is a *restorable* status,
so requiring `"hooks"` would arm a destructive restore over a healthy fresh install that
legitimately has none — the failure the `secrets.env` comment already documents.

The parity test that holds `cred_integrity.py` and `backup.sh` in sync claimed in its
docstring to check *every* target and checked **four of nine**. It now derives the
expectation from `backup.sh` for every entry, with a population floor so an empty sweep
fails instead of passing. Mutation-verified: a typo in the new target's `backup_rel` is
caught, and so are typos in two targets the old spot-check never looked at.

## Review

Two adversarial passes, both internal — no cross-model round consumed.

Round 1: one blocker (the single-axis gap) plus six findings — a blanket `# noqa`, an
ungitignored `junit.xml` carrying hostname and absolute paths, `if: always()` firing on
cancellation under `cancel-in-progress`, a vacuous test, undocumented xfail conflation,
and unchecked report counts.

Round 2: a 15-mutation battery confirmed all of those resolved (15/15 caught), and found
two more — a baseline offered from red runs, and a silent half-configured state where
setting only the ceiling reinstated the original gap while printing OK.

**Three vacuous tests were found and fixed on the way, all the same shape** — an
assertion that is also true when the mechanism is absent:

1. asserting only `returncode == 2` — the interpreter returns 2 for a missing script
2. `or "report"` — every refusal message contains the word
3. `assert "collect" in out` — argparse's own `unrecognized arguments: --min-collected`

Every refusal assertion now requires a marker only the checker emits. Separately, the
first mutation sweep reported every mutation caught **while running zero tests** — it
invoked an interpreter with no pytest and a non-zero exit was read as a kill. Sweeps now
abort unless pytest emits a result line, and carry a green baseline control.

## Verification

- `ruff` clean on all four touched Python files, including `--select S314,RUF100` (an
  earlier `# noqa justification (S314):` was a **blanket** suppression — ruff reads
  `# noqa` followed by anything but a colon as suppressing every rule on the line, and
  this repo does not select RUF100, so it was silent)
- 116 targeted tests pass · `bash -n` clean · external-io, frozen-clock, subsystem-map pass
- Workflow YAML parses; the step is last in the job and adds no new check run

## Deploy note

`src/genesis/guardian/` is a host-deploy path, so the guardian side picks this up only
after `scripts/update.sh` runs. Merged is not deployed.

Ledger: c3dedfc7ba3741b5b5d6489cea38e17f

## Summary by CodeRabbit

* **Bug Fixes**
  * CI now detects missing, malformed, or unexpectedly dormant test reports and flags excessive skipped or insufficiently collected tests.
  * Credential integrity checks now include Claude settings, rejecting empty, corrupted, or invalid configuration data.

* **Chores**
  * CI-generated test reports are excluded from version control.
  * Encrypted backups now include Claude settings.

* **Tests**
  * Added comprehensive coverage for test-report validation, enforcement thresholds, and credential backup consistency.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -584,4 +584,10 @@ jobs:
         # would kill pytest mid-suite, leave no report, and the step would
         # refuse with a cause that is not the cause.
         if: ${{ !cancelled() }}
-        run: python scripts/check_skip_ceiling.py junit.xml
+        # Baseline MEASURED on this PR's own green run (job 101515129147, head
+        # bbd11ad1): "95 skipped of 23164 collected". Exact-fit on both axes, not
+        # a padded guess — headroom would be a number nobody measured, and the
+        # whole point of shipping report-only first was to avoid inventing one.
+        # Raising either bound is a deliberate act that belongs in the same commit
+        # as the change that moves it; both failure messages say so.
+        run: python scripts/check_skip_ceiling.py junit.xml --ceiling 95 --min-collected 23164
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -559,10 +559,29 @@ jobs:
           pip install -e ".[test]"
       - name: Run tests
         run: |
+          # Never measure a stale report: if pytest dies before writing
+          # one, the checker must find nothing rather than yesterday's.
+          rm -f junit.xml
           # Ignore preserved AZ-era test files that import removed modules
           pytest -v \
             --ignore-glob='**/test_*az*.py' \
             --ignore-glob='**/test_*shim*.py' \
-            --ignore=tests/test_ui/
+            --ignore=tests/test_ui/ \
+            --junit-xml=junit.xml
         env:
           GENESIS_TESTING: "1"
+      # A test that stops running is invisible: it makes the suite SMALLER,
+      # not red. Two axes, because dormancy moves both ways — a broken
+      # skipif RAISES the skip count, while a lost marker or a dropped
+      # directory LOWERS it and is only visible in the collected count.
+      # REPORT ONLY until both bounds, measured on these runners, are
+      # recorded here — a threshold invented rather than measured is one
+      # nobody trusts. The step still has teeth meanwhile: a missing or
+      # unreadable report exits 2.
+      - name: Suite dormancy guard
+        # NOT always(): that also fires on cancellation, and this workflow
+        # sets cancel-in-progress for every non-main ref — so a re-push
+        # would kill pytest mid-suite, leave no report, and the step would
+        # refuse with a cause that is not the cause.
+        if: ${{ !cancelled() }}
+        run: python scripts/check_skip_ceiling.py junit.xml
--- a/.gitignore
+++ b/.gitignore
@@ -156,3 +156,8 @@ scripts/cops_interop.sh
 
 # Personal skills (user-level data, lives at ~/.claude/skills/)
 .claude/skills/code-voice/
+
+# pytest junit report, written by CI's test job. Carries hostname and
+# absolute local paths, and a committed copy would let the skip-ceiling
+# checker read a stale report as a live measurement.
+junit.xml
--- a/scripts/backup.sh
+++ b/scripts/backup.sh
@@ -654,6 +654,7 @@ if $_ENCRYPT_READY; then
         "$HOME/.config/gh/hosts.yml:gh_hosts.yml" \
         "$HOME/.claude/.credentials.json:claude_credentials.json" \
         "$HOME/.claude.json:claude.json" \
+        "$HOME/.claude/settings.json:claude_settings.json" \
         "$HOME/.genesis/guardian_remote.yaml:guardian_remote.yaml" \
         "$HOME/.genesis/config/genesis.yaml:genesis.yaml" \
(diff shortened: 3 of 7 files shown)


WingedGuardian:
@codex review

WingedGuardian:

**Structural review**

No pattern violations (nohup for services, direct `task_states` inserts, `/tmp/` usage, `git add -A`/`git add .`, force-push/`reset --hard`, worktree-path `pip install -e`, bare excepts, mutable default args, unchecked `os.killpg`, raw SQL against `genesis.db`) found across the changed files (`.github/workflows/ci.yml`, `.gitignore`, `scripts/backup.sh`, `scripts/check_skip_ceiling.py`, `src/genesis/guardian/cred_integrity.py`, `tests/test_guardian/test_cred_integrity.py`, `tests/test_scripts/test_check_skip_ceiling.py`).

Minor, non-blocking items:
- Import hygiene (warning): `tests/test_scripts/test_check_skip_ceiling.py`, `TestRunsTheWayCIRunsIt.test_bare_interpreter_relative_path_stdlib_only` — `import os` is done locally inside the test method instead of being grouped with the file's other stdlib imports (`subprocess`, `sys`, `pathlib.Path`) at the top.
- Documentation (note): `scripts/check_skip_ceiling.py` — `main()` itself has no docstring (the module-level docstring is extensive and covers its behavior, so this is low priority).
- Coverage: the new `scripts/check_skip_ceiling.py` has thorough matching tests in `tests/test_scripts/test_check_skip_ceiling.py`, and the `cred_integrity.py` target addition has matching updates in `tests/test_guardian/test_cred_integrity.py`. No gaps found.
- Commit/PR title carries a proper conventional prefix (`feat(ci):`).

**PII/Secrets/Wording scan**

- No API keys, tokens, passwords, private IPs (10.x/192.168.x/100.64–127.x), IPv6 ULA addresses, SSH keys, DB connection strings, or secret env values in the added lines.
- No personal emails, username-bearing absolute paths, location-revealing timezone identifiers, cloud account/project IDs, hardware identifiers, or private hostnames in the added lines. (`$HOME/.claude/settings.json` in `backup.sh` is the `$HOME` variable, not a literal path.)
- PR title/body: describes CI/test-engineering work (a skip/collection dormancy checker, backup-target parity) — no mention of a PII/secrets leak, security incident, or specific private infrastructure. No `CHANGELOG.md` changes in this diff.
- PII/Secrets/Wording: clean (no soft warnings either).

VERDICT: PASS



---

WingedGuardian:
@codex review

PR #1823, opened 2026-09-06
docs(windows): the session boundary isolates window stations, not sockets

WingedGuardian:
How Genesis reaches a Windows desktop — and why the obvious route is the wrong one.

## What changed, and why this PR shrank

This originally shipped a screen-capture pair (a shell trigger plus a PowerShell body installed as a scheduled task) alongside the reference doc. **It now ships the doc alone**, because the scripts solved a problem the machine did not have.

The scripts crossed the session boundary the expensive way: register a task with an interactive logon, trigger it over SSH, and — since the trigger cmdlet passes no arguments — invent a request/response protocol with per-invocation correlation, plus a shim so the task stops flashing a console at the operator. Real work, all of it scaffolding around one obstacle.

**The obstacle has a cheaper way round, and this doc already contained it as a footnote about CDP: a socket is not session-isolated.** MEASURED: from an SSH login confirmed to be in session 0, a loopback HTTP request returned `200` from a service in session 1 — the session that can see the desktop. So the first question is not *how do I get my code into the interactive session*, it is *is something already there I can ask*. On a daily-driver machine the answer is usually yes.

## What the doc now says

Written generically, so it applies to whatever a given install actually runs: the contract to look for in a resident agent, how to reach it without asking it to give up its loopback binding, and where install-local values belong. Port, endpoint and auth are properties of a deployment, not values for this repo.

The scheduled-task route stays as the escape hatch for when nothing is resident, and was **strengthened rather than trimmed** — a sibling branch is building on it and cites this file. Its traps are written down: the argument-passing limit and what it does to result correlation, the console window, the battery flags, the SID registration, and the fact that the trigger returns before the work does.

Measured knowledge is preserved and extended. The empty-success table gains the `MainWindowTitle` case. DPI awareness gets its own section. And one claim is corrected: the desktop is session **1 or higher**, never reliably 1 — RDP reconnects, user switching and logoff cycles all move it — so assert the session the desktop is actually found in, never a constant.

## The audit found eight things wrong with the rewrite

All fixed here. Three are worth naming, because each was a confident sentence in a document whose whole subject is confident-looking wrong answers:

- **"A triggered task cannot be passed arguments" was false as an absolute.** The cmdlet cannot; the COM interface substitutes up to 32 values into the action. Narrowed, with result correlation named as the cost that actually survives.
- **The already-set return semantics were attributed to the wrong function.** That behaviour is documented for the shcore call, not the legacy user32 one — inferred from the wrong page, in a doc whose header promises measured-not-inferred.
- **The console-hiding shim was recommended as a free win.** It is not. Under it the scheduler no longer owns the process, so the stop cmdlet stops working; six orphaned pollers accumulated in one afternoon on the sibling branch. The cost and the remedy are now stated with the recommendation.

Also fixed: the loopback example uses the spelling that was actually measured, since the doc itself warns that agents validate the `Host` header; the probe is shown in the shell a reader runs it from, with the variable-expansion trap explained for both shells; two facts that lived only in the deleted scripts are carried over; and the suggested config no longer points at the machine-identity file, where a nested key fails silently, nor collides with a namespace that is already taken.

## One boundary, stated rather than implied

Reaching a resident agent makes desktop work cheap to invoke, and the deleted script carried the reasoning that kept that in check — that the only real advantage of wiring such a thing as a tool is autonomous callability, which is the one property this capability should not have yet. That reasoning now lives in the doc, next to the gate it belongs beside, instead of disappearing with the code that carried it.

## Also

The dependabot / CI / SECURITY edits are dropped: #1960 landed the same change in a better form, carrying a measured rationale this branch's variant lacked. Those three files are byte-identical to the merge base again, so they leave this PR entirely rather than competing with what is already on main.

E2E: none — documentation. The behaviour it describes was exercised directly while writing it (session ids read on a live machine, a loopback call from session 0 to a session 1 service), but nothing in this diff runs.


WingedGuardian:
@codex review

WingedGuardian:
## Structural Review (Domain 1)

No `.py` files changed — this PR adds one doc (`docs/reference/windows-remote-execution.md`) and two scripts (`scripts/capture_desktop.sh`, `scripts/genesis-capture.ps1`). Checked the applicable cross-language pattern rules against the shell/PowerShell content:

- No `nohup`, no `/tmp/` usage (the bash script defaults to `~/tmp/genesis-captures`, correctly per convention)
- No `git add .`/`-A`, no `git push --force`/`git reset --hard`, no `pip install -e`
- No direct `task_states` inserts, no raw SQL, no `os.killpg`
- Commit/PR title uses the correct conventional prefix (`docs(windows): ...`)
- No new `.py` files, so the missing-test-coverage rule doesn't apply
- The PS1 script carries a full comment-based help block (`.SYNOPSIS`/`.DESCRIPTION`/`.PARAMETER`/`.EXAMPLE`); the bash script has a descriptive header comment explaining intent

No structural findings — no hard blocks, no warnings, no notes.

## PII / Secrets / Wording Review (Domain 2)

- No API keys, tokens, passwords, or connection strings in the diff
- No private IPv4 ranges (10.x / 192.168.x / 100.64–127.x) or IPv6 ULA (`fd00::/8`) addresses
- No SSH private keys or fingerprints
- Host/user/key are sourced from `~/.genesis/config/genesis.yaml` or env vars (`GENESIS_DESKTOP_HOST`/`GENESIS_DESKTOP_USER`) — never hardcoded, matching the PR body's own claim
- No absolute paths containing usernames — all paths use `$HOME`/`~`/`$env:USERPROFILE`
- No location-revealing timezones, AWS/GCP account IDs, MAC addresses, or private hostnames (task name `GenesisCapture` and directory `GenesisCaptures` are generic)
- PR title/body/doc describe a new documented capability and platform limitation (Windows session isolation) — no language referencing a PII/secrets leak, security incident, or exploitable infrastructure detail

PII/Secrets/Wording: clean — no soft warnings either.

VERDICT: PASS



---

WingedGuardian:
Correction to this PR's own body, and two fixes pushed.

I cited `rect=0,0,1536x864` as evidence the capture was working. It was evidence the capture was **DPI-scaled**. Measured on the target machine, both states in one run:

```
BEFORE SetProcessDPIAware : GetSystemMetrics 1536 x 864
SetProcessDPIAware=True
AFTER                     : GetSystemMetrics 1920 x 1080
monitor_dpi = 120   (96 = 100% scaling)
```

`1920 × 96/120 = 1536` exactly. Two harms: the PNG was a downscale (1536x864/41KB vs the true 1920x1080/146KB — 36% of linear resolution discarded), and `SendInput` absolute mode addresses the *physical* desktop, so any coordinate derived from a virtualised capture or UIA tree is off by the scale factor.

Then the wrapper hid the fix. Its poll comment said "a capture newer than we started" and set `START_EPOCH`, but the loop broke on the first capture it found and compared nothing — so it returned a 30-minute-old file and reported `OK` at the old resolution. That read as "the DPI fix changed nothing", and had I trusted it I'd have reverted a correct change.

Both now fixed and verified end to end through the shipping scripts: new file, `rect: 0,0,1920x1080`, confirmed on disk.

Worth naming: both defects are the exact class this PR's doc warns about — an operation that succeeds, produces a well-formed artifact, and is wrong. Shipping that warning with two live instances of it is the useful part of the story.

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:

Pushed cdc528435: merged current main and fixed HOME-unset startup in capture_desktop.sh. Existing HOME tests: 2 passed; real HOME-unset help and shell syntax passed. Fresh CI and Codex review requested. Broader Windows-capture findings remain unresolved; #1830 still depends on this PR. No merge authorized.

WingedGuardian:

Pushed c4aebe6f1681cbdd81648f5861f0d3173c7c8bf3 for the owner-requested shared audit correction. The dependency-audit step now reports failures without failing the job, matching the verified #1960 reporting policy. Setup/install failures remain blocking. All existing audit commands and ignore arguments are preserved; no dependencies or vulnerability suppressions changed. SECURITY/Dependabot descriptions now match the policy.

Independent review verified the full parsed workflow is unchanged except for audit-step continue-on-error, with negative controls for install/job-level relaxation and ignore-list drift. Fresh CI has been triggered at this commit. Existing PR findings, review freshness and required checks remain merge gates; no merge or additional review round was triggered by this propagation.

WingedGuardian:

## Deferred feature-completion handoff

This is a Windows remote desktop-capture feature, not an easy conflict repair. It is deliberately out of the current low-risk PR queue. Do not merge or request another review until the implementation is complete.

The owner accepts a selected-window capture as a **screen-rectangle crop**: visible overlapping applications may be present. Do not add a window-surface capture API solely to prevent occlusion; document that contract accurately.

A later session should complete the existing feature in this order:

1. Resolve the current `.github/workflows/ci.yml` conflict while retaining current main policy.
2. Pass requested mode and window match through the scheduled-task invocation.
3. Give every invocation a request ID and unique artifact/metadata paths. The client must poll that invocation's terminal metadata (`OK`, `ERROR`, or `REFUSE`) before retrieving its artifact; do not select an arbitrary newer PNG.
4. Require an artifact belonging to that invocation, validate effective DPI awareness before capture, and fail closed when it remains DPI-unaware.
5. Correct the bounded edge cases: literal window-title matching, positive retention counts, valid low-variance/single-colour captures, robust desktop configuration parsing, and the remote-PID diagnostic quoting.
6. Add focused regression coverage, run the Windows-target end-to-end capture verification, then request one fresh review at the completed head.

Estimated implementation: 8–14 hours plus CI and one review round. PR #1830 depends on the capture script and Windows remote-execution reference delivered here, so it remains deferred with this feature.

WingedGuardian:
@codex review

Issue #1718, opened 2026-09-05
Merge gate: require an explicit E2E declaration per PR, and turn it into a durable post-merge obligation


WingedGuardian:
## Problem

"Run the end-to-end verification after this merges" is currently a convention — every merging session has to remember it. Conventions that call sites must remember are defect generators (the repo's own standing lesson), and recent merges bear it out: of two owner-directed merges in one night, one had its post-merge E2E forgotten until asked for after the fact. Both E2Es, once run, earned their cost — one found live configuration drift on a deploy target, the other was the first production execution of a brand-new code path.

Two distinct failure modes:

- **A.** Nobody explicitly decided whether the PR needs a post-merge E2E — the need isn't labeled anywhere.
- **B.** The decision existed but evaporated at merge — no durable obligation survives the merging session.

### A — the PR body declares the decision; the merge gate refuses silence

A required PR-body line, same family as the existing `Ledger:` / `Follow-up:` conventions:

```
E2E: <one-line plan for the post-merge verification>
```
or
```
E2E: none — <reason there is no runtime surface to verify>
```

Enforced in `scripts/hooks/git_push_guard.py` (merge interception + `--check-pr`) as a new check row (`e2e-plan`), same tier as the existing `review-body` check: **block the merge when the body has no `E2E:` line at all.** `none — <reason>` passes; the point is not forcing E2Es onto docs PRs, it is making the decision explicit — the same principle as "a check that could not run must say so", generalized to obligations.

Guard classification: BLOCK at merge time only (never on push); message audience is the agent; background sessions are unaffected (they cannot merge PRs by design). Transition: apply only to PRs created after the convention lands, so the existing queue is not retro-blocked.

### B — the declaration becomes a durable row automatically

The repo-pulse worker already parses merged PR bodies for `Ledger:` / `Follow-up:` markers at session boundaries. Extend the same parse: a merged PR with a non-`none` `E2E:` line auto-creates a HOT follow-up row ("Run the declared E2E for PR #N: <plan>"), deduped one-per-PR, which surfaces until closed with evidence. Belt-and-braces: the gate's merge-time output echoes the E2E line so the merging session sees the obligation at the moment of merge.

## Non-goals (deliberate)

- No path-based auto-detection of which PRs need E2Es — that judgment belongs to the authoring session; the code only makes the judgment mandatory and durable. At most an advisory nudge when a diff touches migrations / systemd units / shell scripts but declares `none`.
[...]

PR #1824, opened 2026-09-06
fix(gate): E2E declaration is advisory, not a merge block

WingedGuardian:
## Summary

Reverses the enforcement half of #1808. The `E2E:` PR-body declaration stops blocking the merge command and becomes an advisory NOTE. The parser, the cutoff, the `--check-pr` row and every classification test are unchanged — only the verdict moves.

## Why

A block guarantees that a **sentence exists**, never that it is **true**. The judgment behind that sentence gets made the same way with or without the gate, so the gate bought no determinism — it bought compliance text. A mandatory field produces the shortest string that satisfies the parser; an optional one produces signal. The original justification was two observed cases.

The obligation belongs **after** the merge instead: a per-merge verification row, with a documentation-only diff auto-closed by **path** so no model ever decides that a prompt change "needs an end-to-end". **That half is not built** — #1718 stays open for it — and every message in this diff now says so instead of promising a safety net that does not exist. Until it ships, the declaration line is the only record there is, which the advisory states plainly.

## Changes

- **Merge arm** prints a NOTE and proceeds. It prints the **whole** message, so the remedy — both valid forms plus two copyable examples — actually reaches the operator. Printing only the first line answered a colon-promise with a restatement of the problem; an advisory minus its remedy is just noise to route around.
- **`--check-pr`** prints `advisory — …` with the guidance indented beneath, and never increments the failure count.
- Both arms branch on the **returned flag**, never on the message prose. Sharing a function while re-deriving severity from a string prefix is divergence wearing the invariant's clothes.
- `CLAUDE.md` and the PR template rewritten as encouragement — including what skipping the line on a **code** PR actually costs, since the future row's exemption is by path and cannot recover a judgment nobody made.
- Test file renamed off "blocked"/"gate" vocabulary: a test named `..._blocks` that asserts exit 0 is a trap for whoever greps next.

## Testing

- [x] `ruff check .` passes
- [x] `pytest` — 1278 passed across 15 merge-gate-adjacent hook suites plus the parser unit file
- [ ] `docs/architecture/CURRENT.md` — not applicable, no subsystem capability change

**Mutation proof, not a source grep.** Re-inserting `sys.exit(2)` after the merge-arm NOTE goes RED on exactly the two cases that encode the reversal (`e2e_declaration_absent_is_advisory_not_blocking`, `e2e_bare_none_is_advisory_not_blocking`) while 108 others stay green; restored from backup with `sha256sum -c` verified. #1808's "wiring" tests were source-string greps that stayed green while the gate was disarmed — this diff does not repeat that.

**Live:** `--check-pr 1804` against a body that declares nothing prints `e2e-plan : advisory — …` with the full guidance tail, and the report's only remaining BLOCK is `codex-at-head`.

## Residual risk, stated

Until #1718's half B lands, this strictly **removes** a record: a PR declaring nothing now merges with only stderr text nobody retains. That is the accepted tradeoff, and no message in the diff pretends otherwise.

E2E: none — the change removes a merge-time block; its behavior is fully exercised by the characterization suite driving the real command and asserting exit codes, and the live `--check-pr` run above. No runtime service surface is touched.

## Summary by CodeRabbit

* **Bug Fixes**
  * E2E-plan checks remain advisory and do not block merges.
  * Missing or invalid declarations now receive consistent remediation guidance while processing continues.
  * Compliant declarations are not rejected, and advisory messages are not treated as enforcement decisions.
  * Clarified that `none` requires a reason; omitted declarations leave the decision unrecorded.

* **Tests**
  * Expanded coverage for advisory reporting, parser fallbacks, unreadable inputs, downstream gates, and merge verdict behavior.

### DIFF
--- a/.claude/skills/genesis-development/SKILL.md
+++ b/.claude/skills/genesis-development/SKILL.md
@@ -2054,9 +2054,10 @@ merge, the Pre-Merge Gate below governs unchanged.
 
 **Canonical pre-merge check:** run
 `python3 scripts/hooks/git_push_guard.py --check-pr <N> [--repo OWNER/REPO]`
-BEFORE proposing a merge. It runs the SAME functions the enforcement gate uses
-(mergeable → CI → base-invariant → Codex-freshness → scheduled-Claude-review →
-review-body → inline findings), so the report and the gate can never disagree —
+BEFORE proposing a merge. It runs the SAME functions the enforcement gate uses, in the gate's own order
+(mergeable → CI → base-invariant → pin-receipts → e2e-plan *(advisory)* →
+Codex-freshness → head-binding → review-body → inline findings →
+scheduled-Claude-review), so the two read the same state —
 this `--check-pr` read IS the mandatory pre-merge step: **always run it and read
 the PR's automated-review comments (Codex, leak/CI, the scheduled Claude review)
 before any merge** — never hand-roll a
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -36,6 +36,10 @@ Closes #
      declaring here pre-fills that row with what YOU know instead of leaving a
      validator to reverse-engineer it from the diff days later. (This guidance
      lives inside a comment on purpose: comments are stripped before the line is
-     read, so the template can never fill itself in.) -->
+     read, so the template can never fill itself in.)
 
-E2E:
+     The marker is NOT pre-printed below. A bare `E2E:` parses as *present but
+     empty*, which is reported as undeclared — so shipping one in the template
+     would fire the advisory, with its full remedy block, on every PR whose
+     author simply left the template alone. An advisory that fires on everything
+     is one nobody reads. Type the whole line, or type nothing. -->
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -23,17 +23,19 @@ Closes #
 - [ ] `pytest -v` passes
 - [ ] `docs/architecture/CURRENT.md` updated (entry prose + `verified:` stamp) if subsystem capabilities changed
 
-<!-- REQUIRED, and the merge gate enforces it: declare the POST-MERGE end-to-end
+<!-- OPTIONAL but worth 10 seconds: declare the POST-MERGE end-to-end
      verification this change needs. Replace the line below with one of:
 
        E2E: <one-line plan for the post-merge verification>
        E2E: none — <reason there is no runtime surface to verify>
 
-     `none` is a legitimate answer for a docs/prose PR — what is not legitimate is
-     leaving the decision unmade. Note it passes the gate but does NOT release the
-     validator, which assumes every merged PR has an E2E and hunts for one anyway;
-     the line is its first lead, not a boundary. (This guidance lives inside a
-     comment on purpose: the gate strips comments before reading, so the template
-     itself can never satisfy the gate.) -->
+     `none` is a legitimate answer for a docs/prose PR. Leaving the line empty
+     does not block the merge — but nothing else records the decision yet either:
+     the per-merge verification row that will carry this automatically is not
+     built (issue #1718), so today this line is the only record. Once it exists,
+     declaring here pre-fills that row with what YOU know instead of leaving a
+     validator to reverse-engineer it from the diff days later. (This guidance
+     lives inside a comment on purpose: comments are stripped before the line is
+     read, so the template can never fill itself in.) -->
 
 E2E:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -386,17 +386,25 @@ reads as noise. You are the first line of defense; ambient
 extraction (session-manager PR-3) is only the safety net. Plan files stay
 the working documents — ledger rows are the durable index, not a duplicate.
 
-**PR-body convention — `E2E:` (required, merge-gated):** every PR body declares
+**PR-body convention — `E2E:` (encouraged, advisory):** a PR body may declare
 the POST-MERGE end-to-end verification its change needs, on its own line:
 `E2E: <one-line plan>` or `E2E: none — <reason there is no runtime surface>`.
-The merge gate blocks a PR that declares neither — and a `none` carrying no real
-reason (a bare `none`, a placeholder, or a refusal word like TBD). Merge time
-only, never on push. PRs created before the convention are exempt; if the PR's
-creation date cannot be READ, the gate blocks rather than assuming the
-exemption. `none` is a fine answer for a docs PR; what is refused is leaving the
-decision unmade. It does NOT release
-the validator, which assumes every merged PR has an E2E and hunts for one anyway
-— the line is its first lead, not a boundary.
+This does NOT gate the merge — the merge gate prints an advisory NOTE when a
+body declares neither, and proceeds.
+
+**The obligation WILL be carried after the merge, not before it — and that half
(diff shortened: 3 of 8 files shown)


WingedGuardian:
## First-pass automated review (structural + PII/secrets/wording)

**Scope:** 7 files changed (`.github/PULL_REQUEST_TEMPLATE.md`, `CLAUDE.md`, `scripts/e2e_declaration.py`, `scripts/hooks/git_push_guard.py`, `tests/test_hooks/conftest.py`, `tests/test_hooks/test_e2e_plan_gate.py`, `tests/test_hooks/test_merge_gate_characterization.py`). Not doc-only (touches `.py` logic + tests), so both passes ran.

### Structural review

- No pattern violations found (`nohup`, direct `task_states` inserts, `/tmp/` usage, `git add -A`, force-push/reset, bare `except:`, mutable defaults, unsafe `killpg`, raw SQL — none present).
- No import-hygiene issues (no new imports touched).
- No new `.py` files without tests; the one behavior-changing function (`_check_e2e_plan` in `scripts/hooks/git_push_guard.py`) has its existing test files updated in the same PR (`test_e2e_plan_gate.py`, `test_merge_gate_characterization.py`, `conftest.py`), consistent with the new `undeclared`-not-`blocked` semantics.
- Docstrings/comments were rewritten to match the new advisory behavior (not left stale).
- Verified directly (not just from the diff): grepped the full PR head of `git_push_guard.py` — only 3 occurrences of `_check_e2e_plan(` (def + merge arm + `check_pr_report`), no stray caller still branching on the old block-returning semantics, and no leftover `should_block`/enforcement code paths outside comments explaining the historical behavior.
- No conventional-commit-prefix issue (PR title uses `fix(gate): ...`).

### PII / Secrets / Wording

- No API keys, tokens, credentials, private IPv4/IPv6 ranges, SSH keys, DB connection strings, or secret env values in the added lines.
- No soft-warning items either (no personal emails, no username-bearing absolute paths, no location-revealing timezones, no cloud account/project IDs, no hardware identifiers, no private hostnames).
- PR title/body/description reviewed for the semantic wording layer: this PR is a policy reversal (making an `E2E:` PR-body declaration advisory instead of merge-blocking) and references only internal process artifacts (issue #1718, reviewer annotations like "Kimi P3", "CodeRabbit Minor", "architect SHOULD-FIX"). Nothing here describes a PII/secrets leak, a security incident, or names specific private infrastructure — it's process/tooling language, not an incident disclosure.

VERDICT: PASS




---

WingedGuardian:
## Cross-model review round — Kimi K3

Codex declined this PR with an explicit usage-limits reply three seconds after it opened (the fifth today), so the review ran through the approved fallback. It was given the scope and the intent of the change and none of my own findings, fetched its own diff, read both caller arms end to end, and ran the affected suites itself.

**Verdict: APPROVE** — no P1, one P2, three P3. All four verified against the code and fixed in `552c864f`.

### P2 — the advisory's downstream blind spot

Every `e2e_*` characterization case ran with all the other gates already green, so between them they proved "exits 0" and nothing more. The blocking version returned at that point in the merge arm. If a later edit tidies the advisory branch with an early return, Codex-at-head freshness, the TOCTOU head binding and both finding scans go unenforced for exactly the PRs that lack a declaration — and the whole suite stays green.

Verified rather than accepted: inserting `return 0` after the advisory NOTE goes RED on exactly one case, the new one, while the other 87 stay green. That is both the proof the gap was real and the proof the new case is what closes it. The case pairs an undeclared body with a stale Codex review and asserts exit 2 on the *downstream* message, which pins "advisory" to mean "the later gate decides" rather than merely "exit 0".

### P3 — three stale claims, each checked against the code first

- `_check_e2e_plan`'s docstring stated the `createdAt` fail direction unconditionally. It is false when the parser module is also missing: there a bare `E2E:` line reports *declared*. Both cases are now stated.
- The pin-receipts fixture comment still said the E2E reader fails closed and would otherwise steal that case's allow. Neither is true any more.
- Three `GUIDANCE` comments still addressed "an author blocked by this gate". The first sweep did five substitutions and missed these.

### A correction I owe the record

The runtime comment justified "the obligation row is unbuilt" with "0 of the 5 repo_pulse modules contain it". Three reviewers agreed on that denominator — it was written by one, copied by me, and confirmed by a third. All three of us were wrong: it is 4 modules by one glob and 6 if you count the crud and scripts modules. Consensus is not verification. The comment now makes the claim that does not depend on how anyone counts — zero occurrences of `parse_e2e`, `e2e_declaration` or the row's dedup key anywhere under `src/`, and `PR_FIELDS` still lacking the `createdAt` the lane needs.

### Also on this PR

A security review over the same diff returned no CRITICAL and no WARNING. Its useful finding, which I verified against `origin/main` directly: the old blocking code returned at line 6701, immediately *before* the Codex-at-head check — so an undeclared E2E was short-circuiting the Codex gate and the finding scans. Making this advisory means strictly more enforcement runs than before, not less.

610 tests green across the affected suites; ruff clean.

WingedGuardian:
## Fresh-context audit — the class, not the instance

The commit gate's mode-switch fired: two consecutive external rounds each surfacing new defects is the signature of patching named instances. Its remedy is a fresh-context audit of the whole diff, so I ran one — un-anchored, carrying none of the previous findings, tasked with enumerating classes rather than listing findings.

**Result: 7 SHOULD-FIX, 4 NOTE, no BLOCKER.** The downgrade itself was judged correctly implemented; every finding was about the guarantees the diff *claimed* around it. All 7 fixed, plus 3 of the 4 NOTEs.

### The class, and why it took four rounds

*"Advisory means the later gate decides."* Round 1 shipped exit-0-only cases. Round 2 added gate one. Round 3 added gates two through four. Each round covered the gates the previous reviewer **named**, and each stopped one short — then round 3's comment asserted "one case per remaining gate", which was false and would have stopped the next reviewer re-running the enumeration.

The arm runs **five** gates after the advisory: freshness, head binding, review-body scan, inline scan, and the scheduled-Claude review. The fifth was unpinned — skipping it left every required scheduled review unenforced with the suite green. What broke the run was counting the gates from the source rather than from the last finding, and the comment now enumerates all five with file:line and says the list is load-bearing.

The same shape already existed in two **pre-existing** `ci-override` paths, named `_continues` while nothing checked that anything continues. Covered here too — fixing a class means the population, not the member this PR happened to touch.

### Two invariants that were prose, not locks

Both measured green while violated:

- the report arm re-counting the advisory toward its verdict — its only coverage anywhere was a source grep for the row label, and no test had ever called `check_pr_report` with an undeclared body
- the merge arm re-deriving severity from the message text under a **renamed** variable, which sailed straight past the negative grep pinning the old spelling

Both now have behavioural locks that fail however the violation is spelled.

### Three false claims

An operator-facing string still read *"This gate guards every merge"* directly beneath the line saying *"Not blocking"* — my earlier sweeps kept fixing comments and missing strings. *"Both callers print the WHOLE message, whose tail is GUIDANCE"* held on one of four undeclared paths; rather than patch three call sites, all four now route through one helper that appends the remedy, so the property holds by construction. And the conftest's *"nothing here can fail a merge"* ignored that its seam is shared with the pin gate, which does block.

The PR template no longer ships a bare `E2E:` — it parses as present-but-empty, so it would have fired the advisory and its full remedy block on every PR whose author simply left the template alone.

### Verification

822 tests green, ruff clean. Four mutations that were green before this round now go RED — the scheduled-gate skip, the ci-absent override skip, the report failure-increment, and the renamed re-derivation — each re-derived rather than taken on the audit's word, each with a hash-verified restore.

Not done deliberately: issue #1718's title still asserts the half this PR reverses. Retitling is a public-surface edit, so it's raised rather than done unilaterally.

WingedGuardian:
@codex review

PR #1830, opened 2026-09-06
docs(skills): say which click path actually has the guarantee

WingedGuardian:
Seven review findings, all one class: the skill files asserted capabilities, paths and fixed defects the code does not support. Fixing them surfaced three more, and an adversarial audit **of the fix** surfaced three blockers in the fix itself — which is the part of this worth reading, since it is the same failure the PR exists to correct.

## Root cause

The branch sat **130 commits behind main**. Claims that were true when written had quietly stopped being true, in both directions — one reviewer finding was itself stale, because the defect it described had been fixed in #1828 two days before the review. The branch is merged up now, and every statement below was re-derived against the current tree.

## What changed

**Layer 4 no longer claims Genesis builds no desktop layer "by design."** The present-tense half is true and was verified by enumeration — no input injection, no capture code, anywhere in the tree. The design half was contradicted by the module the same bullet cited: the takeover gate ships first and alone *precisely so* an actuator, loop and MCP tool can follow behind it. Saying no Genesis subsystem may actuate would leave that gate governing nothing. It now describes two routes at their actual stages — ask a resident agent (available today, no new code), or the gated actuator (in flight, inert) — and keeps the prohibition that matters: **do not build a third route outside the gate.**

**The coordinate-safety warning claimed no path reads back what is under the pointer before clicking.** That is false, and it is falsified by this repo's own vendored dependency: Playwright hit-tests the point and refuses with `intercepts pointer events` on every non-stealth click. The truth is messier, so it is now a table — five click paths against two questions, *does it hit-test* and *does it read back*:

- **Stealth mode is the DEFAULT and does not hit-test.** Which is why the advice preferring a selector was wrong forty lines before the text explaining why.
- **A third and fourth coordinate path existed** that the first rewrite missed entirely — the Turnstile widget click and the shadow-DOM fallback.
- **The VNC row is the subtle one.** It verifies DELIVERY, not IDENTITY: it reads the pointer position back and warns on drift past 3px, then clicks anyway, deliberately.

A table rather than prose on purpose — a path added later is visibly missing from it, where a sentence beginning "either path" just silently stops being true.

**`SendInput` absolute mode was listed as physical pixels.** It is normalised 0–65,535, and *which rectangle* it normalises across is the part that bites: the primary monitor alone, or — with `MOUSEEVENTF_VIRTUALDESK` — the whole virtual desktop, whose origin can be **negative**. Get it wrong and the pointer lands correctly on the primary display and silently wrong on every other one.

**Two helpers were documented as implemented** under a heading instructing you to apply them in every session. `_idle_jitter()` and `_human_scroll()` exist and **nothing calls them**, so the behaviour simply does not happen. Now stated plainly, with `_human_type()` marked as genuinely wired and scoped to the path it is wired on.

**Also:** the perceive leg pointed at a script that exists on no branch; the desktop is session 1 **or higher**, not 1; the dpr fix landed 2026-09-09 in #1828, not the 09-06 the doc claimed (that was when it was *found*); material from an unlanded Windows spike now carries its provenance instead of reading as described behaviour; and the layer had three different names across the file, one of them in the `description` the skill-injection hook matches on.

## Verification

Every claim was checked against the code it describes, including reading the vendored Playwright driver rather than reasoning about Playwright's semantics. The adversarial audit's own findings were then re-derived independently before being acted on — three blockers, six should-fix, four notes, all addressed.

Two claims the audit confirmed exactly right on the first pass: the dead-helper call sites, and the #1828 attribution.

## Ordering

Depends on **#1823** for `docs/reference/windows-remote-execution.md`, which is not on main yet. That PR is clear of its gates and merges first.

E2E: none — documentation. Nothing in this diff runs.

### DIFF
--- a/src/genesis/skills/browser-automation/SKILL.md
+++ b/src/genesis/skills/browser-automation/SKILL.md
@@ -69,10 +69,24 @@ that can accomplish the task. **Token cost increases with each layer.**
 - Use when you need network inspection, Lighthouse, or performance tracing
   beyond what the built-in Genesis browser tools provide
 
-### Layer 4: Computer Use (visual fallback — V4)
-- Claude Computer Use API (screenshot-based interaction)
-- ~1000 tokens per screenshot, expensive but universal
-- Not yet implemented
+### Layer 4: Computer Use — the operator's whole desktop, not just a browser
+- Reaches what no browser layer can: native applications, OS dialogs, and any
+  window that is not a web page. Also drives a browser when the point is to
+  drive it *as the operator sees it*.
+- **Perceive leg is live.** `scripts/capture_desktop.sh` triggers a capture on
+  the operator's machine and returns the image. Window-scoped by default;
+  full-screen exists but is never the default and is never inherited from a
+  window grant (it discloses every open window at once).
+- **Act leg — mouse, keyboard, drag — is NOT live**, and is gated behind an
+  approval design. Do not attempt to construct it ad hoc.
+- Mechanism, and it is not obvious: an SSH login on Windows lands in
+  SessionId 0 while the desktop is SessionId 1, so **nothing run over SSH can
+  see or touch the desktop**. Work reaches the desktop only via a scheduled
+  task registered with an Interactive logon type.
+  Full detail: `docs/reference/windows-remote-execution.md`.
+- Cost is a screenshot per observation, so prefer a lower layer whenever the
+  target is reachable by one. A browser task is almost always cheaper at
+  Layer 2 or 3.
 
 ### Layer Selection Guide
 | Need | Layer | Why |
@@ -86,6 +100,8 @@ that can accomplish the task. **Token cost increases with each layer.**
 | Submit on user's logged-in site | Remote CDP | User's sessions |
 | Network inspection / Lighthouse | On-Demand MCP | Chrome DevTools |
 | Take action in user's banking app | Remote CDP | MUST confirm |
+| Native app / OS dialog / non-web window | Computer Use | No browser layer can reach it |
+| See what is actually on the operator's screen | Computer Use | Capture, not a page |
 
 ## When to Use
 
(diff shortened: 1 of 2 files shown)


WingedGuardian:
@codex review

WingedGuardian:
**PR type:** Documentation-only (2 `SKILL.md` files, no code changes) — per process, ran the PII/Secrets/Wording scan only and skipped structural review.

- No API keys, tokens, passwords, secrets, or credentials in the diff.
- No private IP addresses (10.x.x.x / 192.168.x.x / 100.64–127.x.x) or IPv6 ULA (`fd00::/8`) addresses.
- No SSH private keys/fingerprints, DB connection strings, or secret-bearing env var values.
- No personal email addresses, home-directory paths containing usernames, location-revealing timezone identifiers, AWS/GCP account IDs, hardware identifiers, or private hostnames.
- PR title/description reviewed for wording issues: the body describes a DPI/coordinate-space-mixing bug class and a `SendInput` zero-return bug, found and fixed in #1823/#1828. This is a functional-correctness disclosure (coordinate arithmetic, Windows session-0/session-1 isolation) — not a PII/secrets leak, not a security-incident reference, and not an exploitable-vulnerability writeup with actionable detail. No private infrastructure is named (references to "the operator's machine" and generic script paths only).

VERDICT: PASS




---

WingedGuardian:
@codex review

WingedGuardian:

Deferred from the easy-fix queue. This documentation advertises the desktop-capture capability supplied by #1823 (`scripts/capture_desktop.sh` and `docs/reference/windows-remote-execution.md`). #1823 is undergoing deferred feature completion, so do not merge this PR independently or revise it to claim the feature is live before #1823's implementation, CI, and review are complete.

WingedGuardian:
Handing this back to a build session rather than closing it from the review side.

**Why, in one line:** seven findings, and every one is the same class — the
document asserts something the code or the tree does not do. That is not a
polish pass; the PR's factual claims need re-deriving against the code, and that
is build work.

I verified the most concrete one independently rather than taking the reviewer's
word: `scripts/capture_desktop.sh` and `docs/reference/windows-remote-execution.md`
are both **absent from the tree**, while `SKILL.md:76-86` calls that leg "live"
and points sessions at them. A session following this skill fails at the first
capture.

**There is also an ordering dependency that is not modelled anywhere:** #1823 is
the PR that ships those artifacts, and it is still OPEN. Merging this first
publishes instructions for something that does not exist yet.

What I did do, and it stands on its own: `1811838be` names the boundary on the
readback rule — Genesis's VNC fallback drives raw pixel coordinates through
`vncdo`, which has no concept of an element, so there is nothing under the
pointer to read back and a wrong active window still produces a click rather
than the refusal the rule describes. Keep or drop that commit as the rewrite
requires; it is one paragraph.

The remaining six are listed in the threads above with file:line. Left open and
unresolved deliberately — they are the work, and a closed thread would hide it.

WingedGuardian:
@codex review

PR #1835, opened 2026-09-07
fix(learning): fail-closed channel allow-lists for procedure extraction and the STEERING.md write

WingedGuardian:
Two channel gates in the triage pipeline were deny-lists that fail OPEN, plus a skill rule for the class of PR a review pass cannot fix.

## Commit 1 — procedure extraction

The legacy extraction branch fired on `APPROACH_FAILURE` / `WORKAROUND_SUCCESS` for **any** channel, and on plain `SUCCESS` for anything in `_AUTONOMOUS_CHANNELS`. Its output becomes a **stored procedure that later sessions recall and follow**.

Two of those channels carry externally-written text: `mail` passes raw email **subject lines** as `user_text` (`mail/monitor.py:617-619`) and `inbox` passes raw `item.content` (`inbox/monitor.py:1841` — note the inbox's own `ContentSanitizer` is applied when building the CC prompt, not on this call). `voice` reached it too, although `_CHANNEL_ORIGIN`'s own comment excludes voice deliberately, because ambient multi-speaker STT can capture a non-owner human in the room.

Now gated on a fail-closed ALLOW-list: the owner half of `_CHANNEL_ORIGIN` plus Genesis's own cognition (`reflection`, `surplus`). `_AUTONOMOUS_CHANNELS` itself is unchanged — its two call sites were enumerated first.

## Commit 2 — skill rule

Some PRs are not a review problem. Adds a rule to the review-loop discipline: hand a structurally-conflicted, spec-contradicting, or architecture-grade PR to a foreground architecture conversation, comment naming which trigger fired, apply `needs-architecture-session`, and move on. Taking that exit does not discharge the escalation cap's own stop. Worked example: PR #1605, where main's shared settings writer replaced the two inline heredocs that PR reimplements, and the literal base-wins resolution measured 61 passed → 36 failed / 25 passed.

## Commit 3 — the STEERING.md write

Same defect class, higher stakes. The identity-file write was gated on the same fail-open deny-list, which does not list `voice` — so a voice interaction classified `APPROACH_FAILURE` could write a rule verbatim into STEERING.md, while the block's own shadow record classified that write `external_untrusted` and let it through (`immunity_shadow.record_would_block` only OBSERVES). Now allow-listed, applied inside `_extract_steering_rule`, and the refusal logs. `immunity_shadow` is untouched.

The two gates derive from **different predicates on purpose**: a procedure is Genesis's own institutional knowledge and may come from a gateway conversation or its own cognition; STEERING.md is the user's identity file. Do not collapse them.

## Scope, stated rather than implied

- **This does not close the class.** `memory/extraction_job.py` `_EXTRACTABLE_SOURCE_TAGS = {"foreground", "voice"}` still mines voice into stored procedures on the **live** path (inbox and mail are excluded there). Commit 1 hardens the deprecated fallback, whose own comment says "Remove after 2026-07-09". The remaining sites are unenumerated and tracked separately.
- `tests/test_security/test_gate_chain_e2e.py` encoded the **pre-fix** behaviour as an invariant — it asserted that voice *does* write a steering rule and *does* emit an external_untrusted shadow row. It was updated rather than the gate weakened, with the reason in its docstring, and its step 3 moved off `voice` so it still exercises the directive-shape filter instead of passing vacuously.
- A fourth commit narrowing the steering gate further — deriving from `cc/types.py::is_owner_attended_channel` (terminal + telegram) rather than the broader `_CHANNEL_ORIGIN` owner map, which also admits `web` (live OpenClaw HTTP, caller-supplied text) — is prepared and lands on this branch pending its depth-gate audit. Until it does, this PR's steering gate still admits `web` and `whatsapp`; both reached the extractor before this change, so neither is a regression.

E2E: restart genesis-server, then confirm from the live logs that a triage on a non-owner channel emits the "Steering write REFUSED" warning and writes no STEERING.md rule, and that an owner-channel (telegram) correction still writes one. The gates sit in the server's triage path, so `git pull` plus a restart is the whole deploy.

## Summary by CodeRabbit

- **Security**
  - Strengthened learning and steering controls by limiting steering updates to owner-attended channels, including terminal and Telegram.
  - Non-authorized channel requests are blocked without creating records, while permitted owner-channel requests continue to work.
  - Procedure extraction remains available only through approved owner and self-cognition channels.
  - Invalid directives continue to be rejected.

- **Bug Fixes**
  - Improved reporting of rejected steering attempts, distinguishing expected refusals from unclassified-channel attempts.

### DIFF
--- a/.claude/skills/genesis-development/SKILL.md
+++ b/.claude/skills/genesis-development/SKILL.md
@@ -1638,6 +1638,34 @@ above (full definitions in `.claude/agents/genesis-architect.md`):
   the user with a minimize-change recommendation. The ack asserts that
   step-back happened — appending it without doing the work is the same
   violation as falsifying `--clean`.
+- **Some PRs are not a review problem — hand them to an architecture session.**
+  When you ARE driving a PR toward green (per "When to DRIVE a Merge" below),
+  asking a few clarifying questions is not a substitute for the design
+  conversation. STOP and hand the PR to a foreground architecture conversation —
+  a session with the USER present, never a `genesis-architect` subagent, which is
+  review-session judgment by another name — when any of these holds: the conflict
+  is STRUCTURAL, main having superseded the mechanism the PR implements, so no
+  conflict resolution is correct on its own terms; it changes or contradicts the
+  spec at a significant level; it raises real questions about how Genesis
+  operates, its architecture, or its infrastructure; its consequences are
+  far-reaching or irreversible; or it is delicate enough that getting it wrong
+  damages a running Genesis. Treat arrival at any tier of the cap (the round-2
+  mode-switch, the round-3 hard stop, and certainly the round-7 terminal) as its
+  own trigger to ask whether this is that kind of PR rather than merely another
+  round — taking this exit does NOT discharge that tier's own stop, which is
+  still owed; the handoff is the answer you bring to it, not a way around it.
+  If you ARE the session with the user present, hold the conversation now. With
+  no user to ask, the action is: comment on the PR naming WHICH trigger fired and
+  the evidence for it, apply the `needs-architecture-session` label (create it if
+  the repo lacks it — labels are per-repo and forks do not inherit them), open a
+  `ready` follow-up naming the PR and the decision it awaits — the label is a
+  GitHub annotation nothing drains, so the row is the intake — and move on to the
+  next PR. Expect it to be uncommon: the owner's estimate, explicitly unmeasured,
+  is on the order of 1 in 10 or fewer, so a session reaching for it often is
+  mis-triaging. Worked example, PR #1605: main's shared settings writer had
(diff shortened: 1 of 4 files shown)


WingedGuardian:

Reviewed 4 changed files (`.claude/skills/genesis-development/SKILL.md`, `src/genesis/learning/pipeline.py`, `tests/test_learning/test_pipeline.py`, `tests/test_security/test_gate_chain_e2e.py`) against both domains — no ERROR/WARNING/NOTE structural findings, no HARD BLOCK/SOFT WARNING/WORDING findings. New allow-list logic in `pipeline.py` is covered by new tests in both test files; no new untested `.py` files; docstrings updated to match the new gate behavior; PR title carries the `fix(learning):` conventional-commit prefix.

VERDICT: PASS



---

WingedGuardian:
@codex review

PR #1853, opened 2026-09-07
fix(cc-tmp): reap sessions not projects, and report a severed control plane

WingedGuardian:
The temp watchgod exists so runaway temp usage cannot kill Claude Code. This
fixes three places where it did the opposite, and adds the detector for the one
failure it cannot undo.

## What was wrong

**YELLOW reaped projects, not sessions.** The layout under cc-tmp is
`claude-<uid>/<project>/<session-uuid>/`, so the sweep's depth-2 target was the
PROJECT directory — and staleness was judged by that directory's own mtime,
which tracks only its direct children. A project directory therefore goes stale
as soon as no NEW session starts under it, however busy the sessions inside are.

MEASURED on a live install: the one actively-used project directory carried an
mtime **2.1 days old while its contents had been written seconds earlier**, and
held **54 session workspaces** including every live session's scratchpad and the
output of running background tasks. Every dormant project showed a divergence of
0.0 days — the gap appears precisely on the directory that must not be deleted.
Seven quiet days and a routine 50%-usage YELLOW, a tier that pages nobody, takes
the lot.

The reap is now per SESSION, judged on contents recursively, and it **fails
closed**: a directory whose freshness cannot be determined is kept. That
direction is the real finding here — a first draft of this fix used a
relative-date predicate that a different `find` implementation rejects, and the
swallowed error made every session read as stale. A housekeeping sweep is one
broken predicate away from total data loss, so the branch that cannot evaluate
must delete nothing.

**ORANGE killed idle sessions for nothing.** It reaped unattached tmux sessions
idle over 2h. The loop-break comment guarding it already made the case against
itself — sessions are not what fills cc-tmp — so the kill freed approximately
nothing while destroying a session's entire context. Across the whole log
history (2026-08-19 → 2026-09-07, 7,563 lines, **1,029 ORANGE polls**) the kill
loop was reached **once** and killed **zero** sessions, so removing it is
behaviour-neutral on the record available. RED is unchanged and remains the
pressure valve.

**Zone B could delete a live socket tree.** Its file sweeps already spared
`*.sock` by name; the empty-directory sweep did not spare their DIRECTORIES.
Measured by running the predicate with `-print` instead of `-delete`: it matched
`/tmp/cc-socks` and a live CC daemon's empty `pty` rendezvous directory —
created up front, populated later. Latent rather than firing (that `/tmp` is not
tmpfs, so the tier reads absolute free space, which was ample), but the same
failure class as the incident below. This zone had **no test coverage at all**,
which is how an earlier audit of mine called it safe on the reasoning
"socket-holding dirs are non-empty" — true of the directory holding the socket,
false of its siblings.

## The detector

Claude Code binds one unix socket per session for cross-session messaging, at a
path that falls back into cc-tmp when `XDG_RUNTIME_DIR` is unset — the very
directory this daemon reclaims. Deleting the PATH does not stop the listener:
the process keeps the inode, so `ss` still reports LISTEN and the session looks
healthy from the inside, while every peer resolves by path and gets ENOENT.
Nothing re-binds after startup, so the session is unreachable for the rest of
its life and cannot notice. On 2026-09-05 a sweep severed 3 of 4 sessions on one
install and 6 of 6 on another.

The sweeps were taught to spare sockets, which prevents recurrence and cannot
heal what is already severed. So the state is now reported: counts in the
watchgod state file, one page on a confirmed rise, quiet across daemon restarts
until it rises again.

Deliberate edges, since each is easy to "simplify" away later:
- **Read-only.** A stale socket file is counted, never reaped. Putting socket
  deletion back into this daemon is what caused the outage.
- **Three statuses, not two.** `unknown` when the probe could not run, `empty`
  when it ran and saw nothing. Neither is health — a detector that has gone
  blind must not be indistinguishable from a clean plane.
- **Confirmation before paging.** The same elevated count must appear on two
  consecutive polls, so a session exiting between `ss`'s snapshot and its own
  unlink cannot page spuriously.
- **Scope.** It watches the per-session `cc-socks` plane only. CC's separate
  daemon tree is protected from deletion but not counted, because what severance
  means there is unestablished — protect what you do not understand, report only
  what you do.

**Expect a page on the first deploy** to any install that is already severed.
That is the feature working, not a regression.

## Verification

Not only unit tests. The real daemon loop, run against a live 4-session
severance with a sandboxed HOME: silent on the first reading, paged exactly once
on the confirming second, silent thereafter, correct counts in the state file,
and still exactly one alert after a full daemon restart.

Verify-RED is measured, not asserted. Against the pre-fix script,
`test_live_session_under_a_stale_project_survives` fails with *"a live session
was reaped because its PROJECT dir went stale"* — the hazard demonstrated
through the real tier entry point rather than described. Zone B could not use
the base as a baseline (its test seam is new), so its control is a mutant of
this branch with the two exclusions removed, and exactly the two socket-tree
tests fail against it.

Four tests are honestly labelled as NOT regression cover, and one earlier draft
of the Zone B tests was **vacuous** — pytest's basetemp can contain a `pytest-`
path component, which Zone B's own exclusion then skips wholesale, so every
survival assertion passed on a sweep that never ran. Fixed by sweeping outside
`tmp_path` and planting a canary the sweep must reclaim in every survival test.

Cost measured: `ss` at 45.6 ms on a 30 s poll (0.15% duty); the session sweep at
420 ms across 83 session directories, falling once the stale ones are reaped.

## Not in this PR

Relocating the sockets out of cc-tmp. That is the actual fix — the control plane
should not live in a directory whose purpose is to be reclaimed — but one
resolver inside the CC binary serves both bind and discovery, so a mixed-era
fleet cannot see itself during the cutover. It waits on the deploy path that
restarts resident daemons (#1703), so one deploy can move the fleet.

Region: `scripts/tmp_watchgod.sh` (all Zone A/B sweep functions, the main poll
loop) and `collect_cc_tmp_usage`. Noted here rather than messaged, because the
peer sessions that would want to know are themselves severed and unreachable —
which is the bug this PR reports.

Ledger row `1eaf5081f0fe41bebfef1df848754457` covers this work and a follow-on;
not cited as complete, because the relocation half is still outstanding.

E2E: after merge and deploy, restart genesis-tmp-watchgod and confirm the state
file's `control_plane.severed_sockets` equals the count of live sessions whose
cc-socks path is missing (cross-check against `ss -xlp`), that exactly one page
fires on the first confirmed rise, and that a YELLOW pass leaves live session
workspaces under cc-tmp intact.

## Summary by CodeRabbit

* **New Features**
  * Added control-plane monitoring for messaging sockets, including health status, stale or severed socket detection, persisted metrics, and confirmed-issue paging.
  * Added configurable temporary and socket-tool paths.
  * Improved state reporting with control-plane health metrics.

* **Bug Fixes**
  * Preserved socket-related directories during temporary-file cleanup.
  * Session cleanup now safely handles nested sessions and protects active or recently updated data.
  * Orange cleanup no longer terminates idle or attached terminal sessions.

### DIFF
--- a/changelog.d/20260907220000-fixed-watchgod-control-plane-safety.md
+++ b/changelog.d/20260907220000-fixed-watchgod-control-plane-safety.md
@@ -1,0 +1,21 @@
+- **The temp watchgod no longer damages the things it exists to protect, and it
+  now reports a broken Claude Code control plane instead of leaving it silent.**
+  Its weekly cleanup was aimed one directory level too shallow and judged
+  staleness by directory timestamp, so a project folder whose sessions were all
+  still active could be deleted whole — on one install that folder held 54 live
+  session workspaces. It now reaps individual session folders, and only when
+  nothing inside has been touched for seven days; if it cannot work out what
+  "seven days ago" means, it deletes nothing at all. The ORANGE tier no longer
+  kills idle Claude Code sessions: sessions are not what fills the temp
+  directory, so the kill freed almost nothing while destroying a session's entire
+  context. The /tmp housekeeping now leaves Claude Code's socket directories
+  alone — an empty folder inside a live socket tree is a rendezvous point the
+  daemon fills in later, not leftover junk.
+
+  And a new check reports Claude Code sessions whose messaging socket has been
+  deleted underneath them. They keep listening, so they look healthy while no
+  other session can reach them, and nothing re-binds after startup — the only
+  cure is restarting them. The count appears in the watchgod state file and pages
+  once when it rises, staying quiet across restarts until it rises again. The
+  check never deletes a socket, and says "unknown" rather than "fine" when it
+  cannot see.
--- a/docs/architecture/CURRENT.md
+++ b/docs/architecture/CURRENT.md
@@ -2149,6 +2149,56 @@ verified: f24c15e9 2026-09-05
 - **_config_overlay.py**: `.local.yaml` deep-merge (user config dir first;
   dicts merge, lists REPLACE wholesale); dependency-free by design to stay
   import-cycle-safe.
+- **`scripts/tmp_watchgod.sh`** (a standalone systemd user service, not a
+  `src/genesis` package — it must keep running when Genesis does not): dual-zone
+  temp protection, polling every 30s. Zone A is `~/.genesis/cc-tmp` against a
+  budget (GREEN <50% / YELLOW >50% / ORANGE >75% / RED >90% or free space under
+  the sacred-ground floor); Zone B is `/tmp` on time-based housekeeping. **Its
+  one job is stopping runaway temp usage from killing Claude Code — so the bar
+  for any sweep is that it must not damage correct functioning, even when it has
+  to fire.** Three edges follow from that and are easy to reintroduce:
+  * **Sockets are never deleted.** CC binds one unix socket per session for
+    cross-session messaging, at `$XDG_RUNTIME_DIR/cc-socks/<pid>.sock` — falling
+    back to the temp dir when that variable is unset, which on a Genesis install
+    is cc-tmp itself. Deleting the PATH does not stop the listener: the process
+    keeps the inode, so `ss` still shows LISTEN and the session looks healthy
+    from the inside while every peer resolves by path and gets ENOENT. Nothing
+    re-binds after startup, so the session is unreachable for life and cannot
+    notice. Measured 2026-09-05, when a RED sweep deleted them: 3 of 4 sessions
+    severed here, 6 of 6 on a sibling install. Every Zone A sweep that removes a
+    directory now goes through `reap_dir_sparing_sockets` (object-level, never
+    `-type s`); Zone B's file sweeps spare `*.sock` by name, and its empty-dir
+    sweep spares the socket TREES — an empty directory inside a live socket tree
+    is a rendezvous point CC populates later, not garbage, and the exclusion by
+    name is what covers it (MEASURED 2026-09-07: that predicate matched
+    `/tmp/cc-socks` and a live daemon's `pty` directory). `check_control_plane`
+    reports severed listeners and stale socket files into the state file, paging
+    once on a confirmed rise. It is strictly READ-ONLY — reaping a stale socket
+    would put socket deletion back in the daemon that caused the outage — and it
+    watches the per-session `cc-socks/` plane ONLY. CC's separate daemon tree
+    (`/tmp/cc-daemon-<uid>/`: control, pty and rendezvous sockets for the
+    background-spare machinery) is protected from deletion but deliberately not
+    counted: what severance means there is unestablished, and a number derived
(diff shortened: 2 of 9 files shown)


WingedGuardian:

**Domain 1 — Structural**

Files touched: `scripts/tmp_watchgod.sh` (bash), `src/genesis/observability/service_status.py`, four test files (`test_service_status.py`, `test_watchgod_control_plane.py` new, `test_watchgod_loopbreak.py`, `test_watchgod_session_reap.py` new, `test_watchgod_zone_b.py` new), `docs/architecture/CURRENT.md`, and a `changelog.d/` fragment.

- No pattern violations found: no `nohup`, no direct `task_states` insert, no `git add -A`/`--force`/`reset --hard`, no `pip install -e` worktree path, no bare `except:`, no mutable default args, no `os.killpg`, no raw SQL against `genesis.db`. (The script's own use of `/tmp`/`SYS_TMP_DIR` is not a "temp file" violation — `/tmp` housekeeping is this daemon's actual subject matter, not incidental scratch space.)
- Import hygiene: clean — no circular/star imports, no ordering issues in the one Python diff hunk.
- Test coverage: unusually thorough for a first-pass check — the PR adds/updates ~874 lines of tests against the ~372 lines of `.sh`/`.py` behavior change, including verify-RED evidence for each fixed hazard. No coverage gaps found.
- Documentation — NOTE only: `collect_cc_tmp_usage()`'s docstring (`src/genesis/observability/service_status.py:296`) still reads "Returns dict with tier, usage, budget" and wasn't extended to mention the new `control_plane` key in the return value. Cosmetic, not blocking.
- Genesis conventions: PR title carries a correct `fix(cc-tmp):` prefix; no `IntervalTrigger`/`$CLAUDE_PROJECT_DIR` misuse; no hooks touched.

- Code diff: clean — no keys/tokens/secrets, no private IPv4/IPv6 ranges, no SSH keys, no credentialed connection strings.
- Soft warnings: none. The synthetic project-dir fixture names in the new test files (e.g. `-home-dev-workrepo`, `-home-dev-retired`) mirror CC's real path-encoding convention but are clearly synthetic test fixture data, not real usernames or hosts.
- PR title/body/changelog entry: describes an internal reliability bug (a cleanup daemon deleting live session workspaces and severing local IPC sockets) in generic terms. No mention of a credentials/PII leak, no specific private infrastructure named, no exploitable vulnerability detail aimed at outsiders.

PII/Secrets/Wording: clean (no soft warnings worth noting).

VERDICT: PASS



---

WingedGuardian:
Closing in favour of #1856, which carries an identical tree as one clean commit.

Why: this PR's first commit's patch contained an install-specific identifier in test fixtures. A later commit scrubbed it and the net diff was clean, but `private_pattern_scan` deliberately scans every commit's added lines rather than the net diff — a value added then removed still lives in published history — so `leak-detector` blocked and could not be cleared by a forward commit. Force-push is not available here, so the remediation is a fresh branch.

#1856 was scanned locally with the same CI scanner before pushing: clean.

PR #1857, opened 2026-09-08
fix(hooks): refuse a push that would publish an install's own private value

WingedGuardian:
The pre-push privacy check already found install-private values in an outgoing
push and printed a warning. The warning arrived in the same tool call as the push
it described — which is too late to be useful, because a push cannot be taken
back.

## Why an advisory is the wrong shape here

MEASURED 2026-09-07. The advisory fired correctly on a real push, the push
proceeded, and the value was published. It could not then be fixed forward:
`private_pattern_scan` in CI deliberately scans **every commit's added lines**
rather than the net diff, so a later scrub commit does not clear it, and
force-push is banned. The remediation was abandoning the branch, opening a
replacement PR, and deleting the old ref.

CI is the right place to refuse a *merge*. It is the wrong place to prevent a
*publication*, because by the time it speaks the value is already public.

## What blocks, and what deliberately does not

Only a match against the install's own `release-fingerprints.txt`. Class-pattern
findings — private-range addresses, `/home/<user>` shapes, email addresses — stay
advisory, because those regexes have real false positives and this repo's own
fixtures carry a dozen legitimate ones. Blocking on those would train everyone to
route around the gate.

Two things worth stating precisely, because the first draft of this change got
both wrong:

- It is **not** an "exact literal" match. `sanitize._check_fingerprints` compiles
  each entry as a **regex**, with `re.escape` only as the fallback for a line
  that is not valid regex; on one install 4 of 7 entries carry metacharacters. So
  an entry is as precise as whoever wrote it. The reason this tier blocks is not
  "zero false positives" — it is that the patterns are curated per-install to name
  that install's own values.
- The sanitizer **already** assigns a fingerprint hit `Severity.BLOCK`
  (`sanitize.py:455`). This change makes the push hook act on a severity that was
  already decided, rather than inventing a new judgement.

Background sessions lose nothing: `git_push_guard` already hard-denies a push in
a Genesis-dispatched session, so the only sessions that reach this gate are
foreground ones where somebody can act on it.

## The substantive half: converting the input side

An adversarial review found the verdict had changed from advisory to blocking
while the helpers that decide **whether** and **what** to scan were still at
advisory tolerances — a gate failing OPEN on the common path and CLOSED on an
uncommon one. Each was fixed as one class, and each is measured:

| what | was | now |
|---|---|---|
| command parser | local `re.split` with no `\n` — **4 of 5** ordinary spellings read as "not a push" and skipped the scan | the shared `shell_parse.analyze()`, like the four sibling hooks — 5 of 5 |
| scan range | net `diff base..HEAD` — blind to add-then-scrub, the one shape this gate exists to catch | per-commit over the merge-base, matching CI's own range selector |
| sigil scope | command-wide — a comment written for a following `rm`, or on a later line, waived the block | scoped to the push segment, like 14 of the 15 sibling call sites |
| unresolvable remote | reported as PUBLIC and blocked | advises; only a confident public target blocks |
| scan budget | outside the wall-clock budget — a hook timeout is treated as a block, so an unbounded regex over a user-editable file could make the repo unpushable | inside it, failing open on overrun |

The parser one is the sharpest: `git_push_guard` runs on the same tool call and
**does** parse those spellings, so the operator would get its approval prompt —
which reads as "this push was reviewed" — with no privacy verdict behind it. That
is worse than either hook alone.

## Verification

End-to-end against the real hook script, on a real repo whose **net diff is
clean** (add-then-scrub, 0 occurrences confirmed by `git diff`):

| case | result |
|---|---|
| plain push | **exit 2**, blocked |
| multi-line commit-then-push | **exit 2** — used to bypass the gate entirely |
| sigil written for a *different* command | **exit 2** — used to waive |
| sigil on the push itself | exit 0, and the advisory announces the waiver |

Plus the acceptance replay: the value that actually leaked on 2026-09-07, against
the real fingerprint file, blocked with the value not echoed. 248 tests pass
across this hook, `shell_parse`, the value-flag consistency lock and the
hook-surface guardrail.

Quoted spoofing was probed and is **not** possible — `bash -c '… # sigil'` and
`echo '# privacy-override'` are both correctly refused. The parser's quote
tracking is sound; the defect was only the argument scope.

## What registering the sigil changes elsewhere

`shell_parse.py` asks the PR that adds a sigil to enumerate what flips, so: with
`# privacy-override` written FIRST in a comment, all **11** other sigils flip
`False → True` for a following token — `review-override`, `ci-override`,
`full-suite-ok`, `escalation-ack`, `merge-to-main-override`, `discard-override`,
`final-round-accept`, `audit-ack`, `depth-ack`, `stale-review-override`,
`scheduled-review-override`. That is the registry's intended property (an
unregistered token reads as prose and ends the leading run, silently disabling
every sigil after it), and the practical risk is low, since the second sigil must
actually be typed.

## Known limitations, stated rather than hidden

Two hooks emit verdicts on the same tool call; a deny is terminal in Claude
Code's aggregation, so exit 2 should win over `git_push_guard`'s `ask`. That
precedence is **not** independently verified against the binary. Likewise the
"a PreToolUse timeout is treated as a block" premise is this file's own
long-standing claim, inherited rather than re-measured here — it only informs the
budget fix, which is safe either way.

E2E: after merge and deploy, run a push carrying a line that matches this
install's fingerprint file and confirm it exits 2 with the location named and the
value absent from stderr; then confirm a class-only finding (a private-range
address in a fixture) still exits 0 with the advisory, and that
`# privacy-override` on the push itself passes with the waiver announced.

## Summary by CodeRabbit

- **New Features**
  - Privacy checks now distinguish advisory pattern findings from high-confidence fingerprint matches.
  - Confirmed fingerprint matches can block public pushes, while uncertain or non-public cases remain advisory.
  - Added a trailing-comment override for intentionally approved fingerprint findings.
  - Push checks now review all newly added content across the outgoing commit range.

- **Bug Fixes**
  - Privacy check failures, timeouts, missing data, and unreadable configuration continue to fail safely without blocking pushes.
  - Block messages no longer expose matched sensitive values.

### DIFF
--- a/changelog.d/20260908005218-changed-pre-push-privacy-block.md
+++ b/changelog.d/20260908005218-changed-pre-push-privacy-block.md
@@ -1,0 +1,17 @@
+- **A push that would publish one of your install's own private values is now
+  refused, not just flagged.** The pre-push privacy check already spotted these
+  and printed a warning — but the warning arrived alongside the push it was
+  describing, and a push cannot be taken back. CI scans every commit a branch
+  adds rather than its net result, so a later commit that removes the value does
+  not clear it; the only remedy left is abandoning the branch and opening a
+  replacement pull request. The check now refuses that push up front, naming the
+  file and line and never repeating the value itself.
+
+  Only a match against your own `release-fingerprints.txt` — the list naming this
+  install's private values — blocks. Pattern-shaped findings — private-range addresses, home-directory paths, email
+  addresses — stay advisory, because those patterns have real false positives and
+  legitimate test fixtures carry them. If you genuinely mean to push a matching
+  line (updating the fingerprint file itself is the usual reason), append
+  `# privacy-override` to the push command; it says so out loud rather than
+  passing quietly. An install with no fingerprint file, and any failure of the
+  check itself, leaves pushes untouched.
--- a/scripts/hooks/pre_push_privacy_review.py
+++ b/scripts/hooks/pre_push_privacy_review.py
@@ -1,14 +1,49 @@
 #!/usr/bin/env python3
-"""PreToolUse ADVISORY: surface private-data leaks in an outgoing PUBLIC push.
+"""PreToolUse gate: private-data leaks in an outgoing PUBLIC push.
 
-NON-BLOCKING. When a ``git push`` targets the public repo (origin), this hook
-scans the diff being pushed for this install's private data — install IPs,
-personal emails, and local release-fingerprints — and, if it finds any, injects
-a review prompt into the model's context (PreToolUse ``additionalContext``). It
-NEVER blocks the push: the CI ``leak-detector`` job and the branch/force gates
-in ``git_push_guard.py`` own hard enforcement. Its job is to make the author
-LOOK at suspicious lines before code goes public — the exact step that, when
-skipped manually, put real install IPs into public test fixtures.
+TWO VERDICTS, split by how confident the match is.
+
+ADVISORY (the default, and what this hook was originally): CLASS findings —
+RFC1918-shaped addresses, ``/home/<user>`` shapes, personal emails. These are
+regexes over generic patterns and they have real false positives; this repo's
+own test fixtures carry a dozen legitimate ones. They inject a review prompt
+into the model's context and never block.
+
+BLOCKING: a FINGERPRINT match — an entry from ``~/.genesis/release-fingerprints.txt``,
+the install's own curated list of its private values. The sanitizer already
+classifies such a hit as ``Severity.BLOCK``; this change makes the push hook act
+on the severity that was already assigned, rather than inventing a new judgement.
+
+Be precise about what "fingerprint" means, because an earlier draft of this
+docstring said "exact literals" and that is FALSE: ``sanitize._check_fingerprints``
+compiles each line as a REGEX, falling back to ``re.escape`` only for a line that
+is not valid regex — and on this install 4 of 7 entries carry metacharacters. So
+an entry is as precise as whoever wrote it: an escaped one is exact, an unescaped
+dot is a wildcard. The reason this tier blocks and the class tier does not is
+therefore NOT "zero false positives"; it is that these patterns are curated
+per-install to name that install's own values, while the class patterns are
+generic shapes this repo's own fixtures legitimately contain by the dozen.
+
+WHY THIS ONE BLOCKS, measured 2026-09-07. The advisory fired correctly on a real
+push and the push proceeded in the same tool call, because that is what an
+advisory does — so the warning arrived AFTER the irreversible act it described.
+A published fingerprint cannot be fixed forward: ``private_pattern_scan`` in CI
+deliberately scans every commit's ADDED lines rather than the net diff (a value
+added then removed still lives in published history), and force-push is banned
+here, so the only remediation was abandoning the branch, opening a replacement
+PR, and deleting the old ref. CI is the right place to enforce a merge; it is the
+wrong place to prevent a publication, because by the time it speaks the value is
+already public.
+
+This does not cost a background session anything: ``git_push_guard`` already
+hard-denies ``git push`` in a Genesis-dispatched session (no human to approve),
+so the only sessions that reach this gate are foreground ones where someone can
+act on it.
+
+Fail direction is OPEN, deliberately: any error, timeout, or unreadable
+fingerprint file leaves the push alone. A fresh clone has no fingerprint file at
+all, and a hook bug that blocked every push would be far worse than the leak it
+prevents — CI still refuses the merge either way.
 
 Reuses the contribution sanitizer's cheap REGEX scanners (``parse_diff`` +
 ``_check_portability`` / ``_check_emails`` / ``_check_fingerprints``) — NOT the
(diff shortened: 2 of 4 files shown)


WingedGuardian:
## First-pass review — PR #1857

**Scope reviewed:** `scripts/hooks/pre_push_privacy_review.py`, `scripts/hooks/shell_parse.py`, `tests/test_hooks/test_pre_push_privacy_review.py`, `changelog.d/20260908005218-changed-pre-push-privacy-block.md` @ `682a1a4`

### Domain 1 — Structural

No pattern violations found against the checklist:
- No `nohup`, no direct `task_states` inserts, no `/tmp/` usage, no `git add .`/`-A`, no `--force`/`--hard`, no bare `except:` (both new handlers name `SystemExit` / `Exception` explicitly), no mutable default args, no `os.killpg`, no raw SQL (n/a — hook has no DB surface).
- Import hygiene: no circular/star imports; the new `from shell_parse import analyze, has_trailing_override` follows the same local-import-after-`sys.path.insert` pattern (with `# noqa: E402`) already used for `hook_input` in this file. `import pytest` in the test file is correctly grouped after stdlib imports.
- Coverage: this is a behavior change to existing functions (`_scan`, `_outgoing_diff`, `_targets_public_repo`, `main`), not a new `src/` module, and it's backed by extensive new/updated tests (fingerprint-block path, override scoping, per-commit diff range, target-resolution uncertainty) — good coverage, no gap.
- Documentation: new functions (`_resolve_target`, `_target_resolved`, `_render`, `_push_segments`) all carry docstrings; the module docstring and `main()` were updated to reflect the new two-verdict (advisory/block) behavior.
- Genesis conventions: hook stays stdlib + `shell_parse` + the existing (pre-existing, not newly introduced) `genesis.contribution.sanitize` import; no `$CLAUDE_PROJECT_DIR` misuse; no `IntervalTrigger`; PR title carries a valid `fix(hooks):` prefix.

Structural: no errors, no warnings.

- No API keys/tokens/passwords, no private IPv4/IPv6 ULA addresses, no SSH keys, no DB connection strings with credentials, no secret env values in the added lines.
- **Soft warning:** the test-file docstring retains the pre-existing `100.64.0.1` CGNAT literal as an illustrative fixture note. This is the documented false-positive shape (synthetic address whose purpose is to exercise the sanitizer, asserted-on in the tests) — flagging per policy, not blocking. The new `_SECRET = "sw-fixture-secret-slug-42"` fixture value is likewise clearly synthetic, used specifically to assert the block message never echoes matched content.
- PR title/description/changelog: describes the mechanism generically (fingerprint-file matching, advisory vs. block tiers, override sigil). No specific private infrastructure named, no exploitable detail, no "leak/incident" wording beyond describing the fix itself in the abstract.

PII/Secrets/Wording: clean (soft warnings only).

VERDICT: PASS



---

WingedGuardian:
@codex review

WingedGuardian:
Shelved as a draft pending an architecture conversation — not abandoned, and not because the review was wrong.

Codex round 1 returned 11 findings (6 P1, 5 P2). Two were already fixed at the current head; nine are live. Read together they are not nine bugs, they are one:

`--repo` skips the repository value · `--all` / `--tags` / `--mirror` / explicit refspecs publish refs the scan never sees · `--dry-run` publishes nothing but is treated as a push · `git remote get-url --push` returns only the first of several URLs · `--no-merges` drops merge-resolution content · a sigil on one push segment globalizes to every push in the payload.

Every one of those is the same shape: **this gate's correctness depends on modelling what `git push` will actually do.** That is the argv→effect mapping the development skill names as an open-set surface, with an explicit decision test — *could a git flag you have never heard of change your guard's verdict?* Here it can, nine ways. The skill is equally explicit that the review loop IS the hardening loop for this class and does not converge, and that a single finding exposing a semantic-modelling gap means re-architect rather than patch. Shipping another patch round would be arguing with that.

The direction worth discussing, rather than deciding here:

- Narrow what the gate is willing to BLOCK to a closed set — a single push segment, no refspec other than the current branch, none of `--all` / `--mirror` / `--tags` / `--delete` / `--repo` / `--stdin` / `--dry-run`. Everything else falls back to the ADVISORY that exists today, so an unmodellable form degrades to the status quo instead of to a broken guarantee. `git_push_guard` already takes this posture next door with its `_PUSH_SAFE_LONG_FLAGS` allowlist, and the push that motivated this work stays inside the blockable set.
- One finding survives that redesign: a catastrophic-backtracking fingerprint can outlive the hook's 30s process timeout, and a killed hook reads as a block — inverting the promised fail-open. CPython will not interrupt a running regex, so the deadline check added earlier does not help; that needs the scan in a bounded child process.

The advisory shipped in `main` is unaffected and keeps working. What is parked is only the escalation to a hard block.

WingedGuardian:
Working the same surface from a later incident (2026-09-08, a push that published an
install's own LAN address and a tailnet address into public test fixtures). The design
here is right and I am not duplicating it. Three measured gaps, offered as findings:

**1. The block cannot fire for a tailnet address, because it is not a known value.**

This PR blocks only on a `release-fingerprints.txt` match, which is correct — but the
fingerprint *harvest* never collects tailnet IPv4 addresses. `contribution/fingerprints.py`
harvests live IPv6 ULA prefixes from `ip -6 addr`, which is why a Tailscale v6 prefix is
present and its v4 address is not.

An `ip -4 addr` fix would not close it either: in the incident the leaked address belonged
to a **peer**, and a local-interface scan cannot see peers at all. Reading
`tailscale status --json` for `Self` + `Peer` yields the whole tailnet — 8 addresses on the
install I measured, including the exact peer address that leaked.

Without that, this gate would have passed the push it is designed to stop.

Peer *hostnames* have the same problem and are worse: a MagicDNS name is not
address-shaped, so it trips neither the fingerprint set nor the class-pattern advisory —
no signal at all.

**2. Four ways a value still reaches origin with no output whatsoever.**

- `_outgoing_diff` always diffs `HEAD`, and no refspec reaches it. `git push origin
  other-branch:main` — pushing a branch you have not checked out — scans unrelated content.
- `_norm_url` does not canonicalise scp-style SSH against HTTPS, so the *same* repo reached
  by the other URL form compares unequal to `origin` and the scan is skipped entirely.
  Measured: `https://github.com/Owner/Repo.git` and `git@github.com:Owner/Repo.git`
  normalise to different strings.
- An unresolvable diff (shallow clone, `origin/main` never fetched, orphan branch) returns
  `""` and `main()` returns silently — indistinguishable from "scanned, clean".
- The blanket `except Exception: return` is right to fail open, but it is silent: a guard
  broken for weeks looks exactly like a guard finding nothing. I hit this for real — a
  stale stub raised `TypeError`, the hook went quiet, and only an error log identified it.

**3. Minor: the advisory path prints email literals.**

`_render` emits `finding.message`, and `_check_emails` interpolates the address into it.
The BLOCK path is safe — `_check_fingerprints` uses a constant message, as this PR's own
docstring notes — but the advisory publishes a personal email into the model's context and
the session transcript. Rendering `file:line [category]` instead closes it for every
scanner rather than the one that currently interpolates.

I have these implemented and tested (11 mutations, zero survivors; the historical leaking
commit replayed and now refused). Happy to rebase them on top once this lands rather than
open a competing PR against the same file — say which you prefer.

PR #1860, opened 2026-09-08
docs: separate "where a record lives" from "whether the work waits"

WingedGuardian:
## The problem

`CLAUDE.md`'s deferred-work rule opens with a fix-now bias, then spends a
paragraph routing by owner: repo work to a GitHub issue, user-owned work to a
follow-up, consciously-not-doing to tabled.

The routing half reads as though it answers the whole question. It does not — it
answers only *where a record lives*. Whether the work waits is the bias above it,
and that decision comes first.

Read as a single rule, "this is repo work, therefore a GitHub issue" becomes a
disposition. An in-scope defect then gets a ticket instead of a fix, and the
tracker makes that look like diligence: an untracked drop wearing a ticket
number is harder to notice than a plain one.

## The change

The two decisions are separated, in the order they should be made:

1. **Is it in scope for what this session is doing?** If yes, it gets addressed
   here. File it too if that helps someone else — the filing is *in addition*,
   never *instead*.
2. **Only then, where does the record live?** The existing owner-routing, hard
   limits, and mechanics are unchanged.

Stated directly in the rule: an issue is a public record, not a disposal;
in-scope work is filed *and* done, often in the same session. And naming the
owner of a piece of work is a classification — a classification is not a
decision.

`.claude/docs/mcp-tools-guide.md` gets the same note at the top of its mechanics
section, because that is the page you land on when you are already reaching for
the routing answer and have skipped the question before it.

## Why it is worth a change rather than a note

The failure is silent in the direction that matters. Deferring something that
should have been done leaves no artifact saying so — the issue exists, the row
exists, and everything looks handled. The only signal is someone later asking
why a known problem is still open. A rule whose first half is routinely skipped
because its second half looks complete is a wording problem, not a discipline
problem, so it gets fixed in the wording.

E2E: none — documentation only. The files changed are `CLAUDE.md`, a doc under
`.claude/docs/`, and `CHANGELOG.md`; no code path, hook, schedule, or runtime
surface is touched, so there is nothing that could behave differently after a
deploy. The change takes effect when a session next reads the instructions.


WingedGuardian:
## First-pass review (leaks/PII scheduled scan)

**Scope:** Documentation-only PR (`CLAUDE.md`, `CHANGELOG.md`, `.claude/docs/mcp-tools-guide.md`) — per process, structural review is skipped and only the PII/secrets/wording scan runs.

### PII/Secrets/Wording scan

- Code diff: no API keys, tokens, credentials, private IPs, SSH keys, connection strings, or other secrets in any added line — the diff is entirely prose clarifying the "filing is not deferring" policy distinction.
- No soft-warning items either (no personal emails, no usernamed paths, no location-revealing timezones, no account/project IDs, no hardware identifiers, no private hostnames).
- PR title/body/CHANGELOG: describes a wording/structure fix to the deferred-work rule (separating "is it in scope" from "where does the record live"). No mention of a PII/secrets leak, security incident, specific private infrastructure, or exploitable vulnerability detail.

Structural review: skipped (documentation-only PR).

VERDICT: PASS



---

WingedGuardian:
@codex review

PR #1861, opened 2026-09-08
fix(hooks): blind-spot net denies instead of prompting

WingedGuardian:
## What

The commit and push guards carry a "blind-spot net": when the shell parser cannot tokenize a command, the raw text names a gated operation, and no gated segment was resolved, the verdict is "unknown". That verdict used to be an interactive approval dialog, with a hard deny only for dispatched sessions.

It now denies in every session type, with the way out spelled in the message (Write tool for prose; pass the text as a file for real git commands). `_ask` is deleted from the commit gate.

The one prompt that survives is the deliberate egress approval on `git push` / `gh pr create`. That is structurally un-preemptable here: the net arms only when `not (push_segs or merge_pr_segs or merge_git_segs or create_segs)`, and all four `ask_reason` assignments sit inside the push or create arms, so the two conditions are mutually exclusive.

The predicate stays deliberately broad. Every narrowing conjunct was previously measured to starve the trigger, and the rationale changed rather than weakened: an over-match now costs the agent one rewrite the message names, where it used to cost a human one confirmation. An under-match still silently runs an unverified publish.

## Three defects found while doing it, each fixed as a class

**A probe crash denied everything.** The push guard wrapped `untokenizable()` so a probe defect could not crash into a silent allow, but let the crash stand in for the gated-mention check as well — so one bug in the probe would have refused *every* command in the session, `ls -la` included. That is a self-inflicted outage, not fail-closed. The commit gate has no such hole because its net sits behind `_COMMIT_PATTERN`, so a crash there can only refuse a command already naming a commit. The push guard now matches: the crash stands in for UNTOKENIZABLE only. This concedes no coverage — the normal path is bounded by the same mention predicate, so dropping the conjunct could only add coverage in a bug state.

**A probe crash was silent.** Catching the exception locally deleted `run_guard`'s own loud `GUARD ERROR (...): failing CLOSED` line (`hook_input.py`), whose docstring states the exception is "logged loudly (so a guard bug is visible, per 'never hide broken things')". Measured base-vs-new with a raising-probe shim: base `rc=2` plus the GUARD ERROR line; new `rc=0` with empty stderr. So a broken probe produced no signal at all for commands naming nothing gated — and the deny message's own advice ("if this persists, the probe itself is broken — flag it") was unreachable by exactly the shapes that would reveal it. Both guards now report the exception type and message. Advisory only; no verdict changes.

**Deferring the verdict opened a window.** The net hands its deny to the tail so that more specific hard blocks (sqlite write, `--no-verify`) still print their own sharper cause first. But the tail's malformed-payload fail-open (`except (JSONDecodeError, KeyError): return 0`) sits between the two points and would drop an already-decided deny — a window the old inline `return 2` did not have. An armed net now outranks that fail-open. Reachability is low (all 10 `json.loads` sites in the file are individually try-wrapped), but the guarantee had narrowed and is restored.

Also: the probe-failure path gets its own message, because the generic one asserted an ANSI-C quoting diagnosis the guard never made; and the rationale comment beside the mention predicate no longer argues from the retired prompt policy, since a maintainer reading a false premise would narrow the predicate — the exact failure that comment exists to prevent.

## Verification

- **620 passed / 24 skipped** across `test_guard_ansic_fail_closed`, `test_push_create_override`, `test_worktree_remote_guard`, `test_merge_review_gate`. `ruff` clean.
- **Mutation sweep 5/5 killed**, one baseline copy for the whole sweep, anchor count asserted per cycle, `py_compile` on each mutant, hash-verified restore: predicate scoping, probe-failure message, deferred ordering, and both new fixes each go RED when reverted. Two mutants first reported no output; that was the box-wide pytest lock held by a concurrent run, not a survivor — the harness refuses to score a run that produced no result line, and both went RED once queued.
- **End-to-end through the real hook stdin protocol** (not the pytest harness), 8/8: both guards x {interactive, dispatched} x {untokenizable + gated mention -> block, benign -> allow}, with no `ask` emitted on any net path in either mode. Re-run after the review fixes: still 8/8.
- Adversarial review by a fresh-context architect agent over the full branch diff. Three SHOULD-FIXes and two NOTEs, all applied in this PR. Each reviewer claim was verified against source before acting on it.

E2E: hooks land at next CC session start. In a fresh session, confirm an unparseable command naming a gated op is refused with the rewrite advice and no dialog, and that an ordinary `git push` still raises the egress approval.

Ledger: 9f3be652e23640e89ab8dea2fb331f80


WingedGuardian:
@codex review

WingedGuardian:
## Structural Review (Domain 1)

Checked `scripts/hooks/git_push_guard.py`, `scripts/review_enforcement_commit.py`, `tests/test_hooks/test_guard_ansic_fail_closed.py`, and the changelog fragment against the pattern/import/coverage/convention checklist:
- No `nohup`, no `git add -A`/`.`, no raw SQL against `genesis.db`, no bare `except:`, no mutable default arguments, no `os.killpg`, no `/tmp/` usage for real temp files — the `/tmp/x.db` string in `test_direct_sqlite_write_still_hard_blocks` is pre-existing synthetic input to a command the guard is asserted to *block*, not an actual file write.
- `except Exception as exc:` (both guards) names the exception type explicitly — not a bare `except:` — and is deliberately broad per its own `noqa: BLE001` comment, so a probe crash fails closed instead of silently allowing.
- `import json` was removed cleanly from `review_enforcement_commit.py` alongside the deleted `_ask()` helper; no import-order issues in what remains (stdlib only: `os`, `re`, `subprocess`).
- PR title carries a proper `fix(hooks):` conventional-commit prefix.
- Test coverage for the changed `main()` logic in both guards is extensive — new `TestCommitGuardProbeFailure`, `TestPushGuardProbeFailure`, and `TestBlindSpotDenyIsModeIndependent` classes, plus updated matrix/acceptance-corpus tests — no missing-coverage warning.
- No `IntervalTrigger`, no heavy imports added to hook scripts, no `$CLAUDE_PROJECT_DIR` misuse as a Python env var.

## PII / Secrets / Wording Scan (Domain 2)

PII/Secrets/Wording: clean (no findings) — scanned all added (`+`) lines across the diff plus the PR title, body, and changelog fragment.
- No API keys, tokens, credentials, private IPs (10.x/192.168.x/100.64–127.x), IPv6 ULA, SSH keys, or DB connection strings with credentials.
- No personal emails, username-bearing absolute paths, location-revealing timezones, cloud account/project IDs, hardware identifiers, or private hostnames.
- PR title/body/changelog describe the guard's ask→deny behavior change and its testing methodology in general terms; no mention of a specific leak/incident or exploitable infrastructure detail.

VERDICT: PASS



---

WingedGuardian:
@codex review

Please review the updated head 979f362ea7d81597073a3d60f37e3eafa8be69b9 after integration with current main. The conflict resolution preserves analyze_checked, both parser bounds, and override audit logging while changing syntax-blind approval to an actionable denial. The obsolete separate-probe tests now exercise the checked parser; armed denial remains intact through both payload-exception tails.

Validation: 640 targeted tests passed, 24 expected skips; Ruff and the external-I/O, subsystem-map, shared-artifact, and frozen-clock checks passed. Independent internal integration review found no blocking issues. This follows the prior clean review at 64356038be; required new-head CI and review gates remain in force.

WingedGuardian:
@codex review\n\nCurrent head `979f362ea7d81597073a3d60f37e3eafa8be69b9` is fully green in CI. GitHub requires a current review before it may merge; please review this exact head.

WingedGuardian:

Current-head review refresh for the queue-wide gate pass.

@codex review

PR #1862, opened 2026-09-08
fix(skill): "sequential" was dischargeable by killing a running reviewer

WingedGuardian:
## What happened

The reviewer-concurrency directive read **"NEVER run two review agents
simultaneously."** A session was told to switch reviewers, read that NEVER as a
hard prohibition, concluded the only way to start the new one was to terminate
the one already running, and killed it four minutes into a real analysis.

Nothing was gated on stopping it. The rule's own rationale — the second reviewer
should see FIXED code, parallel doubles spend — never implied killing anything.
**"Sequential" means WAIT.**

## The shape, which is the transferable part

This rule prohibited a **STATE** ("two running") rather than an **ACTION**. A
state prohibition can be satisfied by DESTROYING the offending state; an action
prohibition (never push to main, never hardcode the slug) can only be satisfied
by not acting. That difference is why this one was dischargeable destructively
and its neighbours are not.

Swept both files for the shape. One other instance: `CLAUDE.md`'s genesis-bridge
line ("must never run alongside the server"), where the wrong discharge is worse
— stopping the server rather than declining to start the bridge. **Left alone
deliberately:** it is a user-sovereign top-level doc, so that is the user's call,
not a drive-by edit in this PR. Noted in the new text as the second example.

## Three copies, and only the least salient was found first

The directive lived in three places. The first attempt patched only SKILL.md —
which loads *when the skill is invoked*. Review caught that the other two were
untouched:

| surface | when it loads |
|---|---|
| `SKILL.md` | when the skill is invoked |
| `scripts/review_enforcement_prompt.py` | **every user prompt** (wired in `settings.json`) |
| `.claude/commands/deep-review.md` | the command that *dispatches* the reviewers |

With the old absolute still in the per-prompt injection, the fix would have been
inert exactly where the incident happened — CLAUDE.md's own "ambiguity resolves
TOWARD the prohibition" would have resolved back to it. All three now agree;
`grep` confirms zero surviving copies of the old wording.

## Also narrowed: a second overreach nobody had noticed

The old text banned two reviewers on **different** PRs, which share no baseline
and were never a problem. Now allowed — with the constraint stated, because it
is real: external reviewers share one quota, and exhausting it strands the
cross-model gate every open PR depends on. Concurrency there is a tolerated
exception, not a throughput strategy.

Two other clauses were tightened after review flagged that they reproduced the
defect they were describing: "a peer session's reviewer is not yours to stop"
was itself an unqualified absolute (now scoped to *while that session is alive*,
with orphan cleanup explicitly allowed), and "the user may ask for both" now
says what counts as asking — the running reviewer named in the user's own words,
which a preference for a different tool is not.

## Verification

- `grep`: zero surviving copies of the old absolute across `*.py` and `*.md`.
- Injection budget MEASURED: the per-prompt reminder totals 6,604 chars against
  the 10,000-char per-hook persistence cap (66%, 3,396 headroom); this adds ~249.
  Upper bound — not every branch emits.
- `ruff` clean. Two self-referential factual claims in the first draft were
  caught by review and removed rather than corrected, since nothing in CI
  re-checks a claim a file makes about itself.

E2E: none — prompt-surface and documentation text; no runtime behaviour changes.


WingedGuardian:
@codex review

WingedGuardian:
**Automated first-pass review (structural + PII/secrets/wording)**

Scope: 3 files changed — `.claude/commands/deep-review.md`, `.claude/skills/genesis-development/SKILL.md` (both prose), `scripts/review_enforcement_prompt.py` (a string-literal prompt-text change only, no logic/control-flow touched). No new `.py` files, no changed function signatures, no CHANGELOG.md addition in this PR's own diff. Confirmed `tests/test_scripts/test_review_scope.py` (the only test referencing `review_enforcement_prompt`) doesn't assert the old wording, so the string change doesn't break existing coverage.

**Structural review:** no violations found — no `nohup`, no direct `task_states` writes, no `/tmp/` usage, no `git add .`/`-A`, no force-push/reset patterns, no bare `except:`, no mutable defaults, no raw SQL, no unsafe `os.killpg`, no import-hygiene issues (no imports touched). Commit/PR title carries a valid conventional prefix (`fix(skill): ...`).

**PII/Secrets:** scanned all added lines (`+`) across the diff — no API keys/tokens/secrets, no private IPv4/IPv6 ranges, no SSH keys, no connection strings, no personal emails, no user-bearing absolute paths, no location-revealing timezones, no cloud account IDs, no hardware identifiers, no private hostnames.

**Wording (title/body):** PR title and body describe an internal prompt/process fix (a reviewer-concurrency rule being misread as license to kill an in-flight review) — no mention of a PII/secrets leak, no security-incident framing, no named private infrastructure, nothing exploitable disclosed. Clean.

VERDICT: PASS



---

WingedGuardian:
@codex review

PR #1866, opened 2026-09-08
fix(hooks): a bound that stops the parse must not silence the worktree guard

WingedGuardian:
## What

`worktree_cwd_guard` asks `analyze_checked` once per invocation and treats a blind
spot as a fail-CLOSED trigger, alongside the `untokenizable` case it already
handled.

## Why

The guard decides by **searching** `analyze`'s segment list. A parse stopped by
the substitution-depth or command-length bound returns *no* segments — by
design, so a searching consumer cannot read "stopped looking" as "nothing
found". This guard read it as exactly that: a removal nested past the depth
bound produced no targets and was allowed, on a guard whose only job is
blocking.

The fallback is the guard's own existing answer for an unreadable command, not a
new mechanism: the legacy regex extractor reads raw text, so nesting cannot hide
a removal from it.

### Why not the allowlist

The chokepoint test offers two ways out, and the other one could not be
justified here. Every existing `_BARE_ANALYZE_ALLOWED` entry states why
blindness is *safe* for that module — `git_push_guard` filters to `depth == 0`,
so a depth bound cannot change its result; `tmux_kill_server_guard` is
advisory-only, so its worst case is advice nobody needed. Neither shape fits a
blocking guard that searches. Writing the entry would have meant recording a
known fail-open.

## This also repairs a red main

#1607 added the bare-`analyze` chokepoint test; #1675 carries this guard's bare
import. Each was green against a main that lacked the other, and neither diff
touches the other's files — so the collision merged textually clean and CI only
found it afterwards, on main. Because PR CI builds the merge commit, it was
failing every open PR, not just one branch.

## Evidence

- **Verify-RED**: reverting the new branch makes a real `git worktree remove`
  nested 8 substitutions deep exit `0` (allowed) instead of `2`. Mutant hash
  confirmed changed and compiling, restore hash-verified.
- **Negative control** stays green under the same mutation: a deep command
  carrying no removal is still allowed, so the fallback discriminates rather
  than blocking everything unreadable.
- 281 passed across both hook test directories; the chokepoint test passes.

## Not claimed

This does not make the parser see `eval`, process substitution or `env -S`.
`analyze_checked`'s docstring is explicit that it cannot report those; that gap
is pre-existing and unchanged.

E2E: confirm main's push CI goes green after merge, then confirm an open PR's
next CI run no longer fails on `test_bare_analyze_requires_being_on_the_allowlist`.

### DIFF
--- a/changelog.d/20260908212219-fixed-worktree-guard-blind-parse.md
+++ b/changelog.d/20260908212219-fixed-worktree-guard-blind-parse.md
@@ -1,0 +1,19 @@
+- **The worktree-removal guard no longer goes silent when a bound stops the
+  parse.** It decided by searching `analyze`'s segment list, and a parse halted
+  by the substitution-depth or command-length bound yields no segments at all —
+  by design, so a searching consumer cannot read "stopped looking" as "nothing
+  found". The guard read it as the latter: a removal wrapped past the depth
+  bound produced no targets and was allowed, on a guard whose only job is
+  blocking.
+
+  It now asks `analyze_checked` once per invocation and treats a blind spot the
+  same way it already treats a command shlex cannot tokenize — falling back to
+  the coarser regex extractor, which reads the raw text and so cannot be hidden
+  from by nesting. Measured by mutation: with the new branch removed, a real
+  `git worktree remove` nested eight substitutions deep exits 0 (allowed); with
+  it, 2 (blocked). A deep command carrying no removal is still allowed, so the
+  fallback discriminates rather than blocking everything unreadable.
+
+  This also restores the chokepoint invariant that every consumer of the shared
+  parser either uses the checked entry point or is listed with a stated reason
+  why blindness is safe for it. That reason could not be written here.
--- a/scripts/hooks/worktree_cwd_guard.py
+++ b/scripts/hooks/worktree_cwd_guard.py
@@ -40,7 +40,7 @@ import sys
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 from hook_input import read_payload, run_guard, tool_input  # noqa: E402
 from shell_parse import (  # noqa: E402
-    analyze,
+    analyze_checked,
     git_subcommand_index,
     untokenizable,
 )
@@ -146,7 +146,7 @@ def _legacy_targets(cmd: str) -> list[str]:
     return targets
 
 
-def _extract_worktree_targets(cmd: str) -> list[str]:
+def _extract_worktree_targets(segs: list) -> list[str]:
     """Target paths of every EXECUTED worktree-removal segment.
 
     Keyed on parsed structure, not on the phrase appearing in the text: the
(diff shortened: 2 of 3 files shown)


WingedGuardian:

**Domain 1 — Structural**

Reviewed `scripts/hooks/worktree_cwd_guard.py` and `tests/test_hooks/test_worktree_guard.py`.

- No pattern violations against the checklist (no `nohup`, no direct `task_states` writes, no `/tmp/` usage, no bare `except:`, no mutable default args, no raw SQL, no force-push/reset patterns — none applicable to this diff).
- Import hygiene: clean. `analyze_checked` swaps in for `analyze` in the existing stdlib-only import block; ordering unchanged.
- Genesis conventions: the hook stays stdlib-only (only `hook_input`/`shell_parse` locals added), consistent with the module's own "Stdlib-only" contract. Commit/PR carries a conventional `fix:` prefix. No `IntervalTrigger`/`$CLAUDE_PROJECT_DIR` concerns (not applicable).
- Coverage: no new source files; the two new tests (`test_removal_past_the_depth_bound_still_blocks`, `test_deep_nesting_without_a_removal_is_still_allowed`) exercise exactly the behavior change, including a negative control. Verified `MAX_SUBSTITUTION_DEPTH = 5` in `shell_parse.py` matches the test's comment.
- Docs: docstrings on `_extract_worktree_targets` and `_carries_a_command` were updated to match the new `segs: list` signatures; the new fail-closed branch in `_handle_bash` carries an inline rationale.

One minor NOTE (non-blocking): in `_handle_bash`, `if untokenizable(cmd) or blind is not None:` re-invokes `untokenizable(cmd)` even though `analyze_checked`'s own contract (per its docstring in `shell_parse.py`) already folds that case into `blind` whenever no length/depth bound already fired — i.e. `untokenizable(cmd) == True` implies `blind is not None` given the function's documented ordering, so the condition reduces to `blind is not None` and the extra call recomputes the same tokenization check. Purely a redundant cheap regex pass, not a correctness issue — worth a follow-up simplification but not this PR's problem to fix.

I also spot-checked the PR's premise: `_BARE_ANALYZE_ALLOWED`/the bare-`analyze` chokepoint test does exist in `tests/test_hooks/test_untokenizable_probe.py` and `shell_parse.py` on `main`, so the "restores the chokepoint invariant" claim in the changelog checks out structurally.

- Added lines: no API keys/tokens/secrets, no private IPv4/IPv6 ranges, no SSH keys, no connection strings. Test fixture paths (`/tmp/nonexistent-worktree-xyz`, `/wt/X`) are generic, non-identifying synthetic values.
- No soft-warning items either (no personal emails, no usernamed absolute paths, no location-revealing tz identifiers, no cloud account/project IDs, no hardware identifiers, no private hostnames).
- PR title/body/changelog: describes a guard-bypass fix in an internal Claude Code safety hook (worktree removal protection) in general terms (depth/length bounds, fallback to a regex extractor). No private infrastructure named, no exploitable specifics beyond what's needed to justify the fix, no incident/leak language implicating PII or credentials.

PII/Secrets/Wording: clean (no hard blocks, no soft warnings, no wording issues).

VERDICT: PASS



---

WingedGuardian:
@codex review

PR #1873, opened 2026-09-09
fix(memory): sanitize FTS5 terms before joining, not after

WingedGuardian:
fix(memory): sanitize FTS5 terms before joining, not after

Recall returned HTTP 500 on some ordinary prompts. MEASURED against a live
endpoint: 5 of 80 probe requests failed with `fts5: syntax error near ")"`,
isolated to one reproducible prompt.

Balance is not validity. `_prepare_fts5(boolean=True)` strips every character
outside `[\w\s()]` from the FINISHED expression, and its only structural check
compares open-paren and close-paren counts. A term made only of punctuation
therefore reduces to whitespace and leaves a dangling operator inside a group
that is still balanced — `(fusion OR  )` — so the check passes and FTS5 rejects
the string. The exception surfaced as a 500 from the recall endpoint.

Two producers compose such expressions, and only one of them was the case that
reproduced. The tag co-occurrence expander joins terms drawn from live indexed
data, which may hold any character. The proactive hook's file-context keywords
are appended the same way, filtered by a truthiness test that drops an empty
term but not `"---"` — and those terms are filenames, so punctuation is the norm
there rather than the exception. Fixing only the reproduced producer would have
shipped the other as the next bug.

Both now sanitize each term BEFORE joining and drop terms that sanitize to
nothing, which makes the expression valid by construction — what the expander's
docstring already claimed to rely on. Where no keyword survives it declines to
build a query rather than emit an expansion-only one, which would violate its
own precision rule.

A backstop in `search_ranked` retries with bare terms when FTS5 rejects a
composed boolean expression, so a future producer degrades instead of failing
the recall: expansion is an optimisation, and losing its precision beats losing
the query. SQLite is the validator on purpose — checking the grammar ourselves
would mean maintaining a second FTS5 parser, which is the one that disagrees
with the engine on an input nobody enumerated. It is narrowed to fts5 syntax
errors so a locked database or a missing table still raises rather than becoming
a silent empty result.

One docstring claimed `_prepare_fts5` "re-sanitizes, so raw terms can't cause an
FTS5 syntax error". That was false, and it is what made the second producer look
safe. Corrected in place with the reason.

Tests assert the invariant against FTS5 itself rather than against a hand-written
model of its grammar, parametrised over both producers, with negative controls
(a hyphenated tag, real filenames) that must keep expanding so the fix cannot
pass by discarding expansions. Mutation-verified 3/3 with hash-checked restore;
the backstop had no test at all until the sweep showed no cell reached it.

E2E: after deploy, re-run the concurrency probe against the live recall endpoint
and confirm zero HTTP 500s across the same request population (was 5/80).

### DIFF
--- a/changelog.d/20260909000735-fixed-recall-fts5-syntax-error.md
+++ b/changelog.d/20260909000735-fixed-recall-fts5-syntax-error.md
@@ -1,0 +1,1 @@
+- **Memory recall no longer fails on some ordinary prompts.** A query could be expanded into a search expression the text index refused to parse, which surfaced as an error instead of results. Expansion terms are now cleaned before they are combined, and a malformed expression falls back to a plain keyword search rather than failing the whole recall.
--- a/src/genesis/db/crud/_fts.py
+++ b/src/genesis/db/crud/_fts.py
@@ -12,7 +12,34 @@ when the AND query returned zero — it never changes a query that already hit.
 
 from __future__ import annotations
 
+import re
+
 import aiosqlite
+
+
+def fts5_term(value: object) -> str | None:
+    """One FTS5-safe bare term, or ``None`` if nothing survives sanitising.
+
+    Callers that COMPOSE a boolean expression must sanitise each term BEFORE
+    joining it, not after. ``_prepare_fts5(boolean=True)`` strips unsafe
+    characters from the finished expression, which turns a punctuation-only term
+    into whitespace and leaves a dangling operator — ``(fusion OR  )`` — that is
+    still parenthesis-BALANCED and so passes that function's only structural
+    check, then fails in FTS5 with ``syntax error near ")"``.
+
+    Returning ``None`` for "nothing left" is the point: it lets a composer DROP
+    the term instead of emitting an operator with no operand. Word characters
+    and single interior spaces survive (a hyphenated tag becomes two terms, the
+    same result ``_prepare_fts5`` already produced for it); everything else goes.
+    """
+    if value is None:
+        return None
+    cleaned = re.sub(r"[^\w\s]", " ", str(value), flags=re.UNICODE)
+    # Collapse runs so a joined expression can never contain a bare double space
+    # that reads as an empty operand to a later reviewer.
+    cleaned = " ".join(cleaned.split())
+    return cleaned or None
+
 
 # Ultra-common English stopwords dropped from the OR retry so a verbose query
 # like "where is the nonexistent service" falls back to "nonexistent OR service"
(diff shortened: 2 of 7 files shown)


WingedGuardian:
@codex review

WingedGuardian:
**Structural review** (WARNING / NOTE only — nothing blocking)

- WARNING — test coverage: `tests/test_db/test_memory.py::test_search_ranked_real_db_error_still_raises` doesn't actually exercise `search_ranked`'s new fallback path. It only asserts that `db.execute_fetchall("SELECT 1 FROM no_such_table_xyz WHERE x MATCH ?", ["a"])` raises `sqlite3.OperationalError` directly — it never calls `memory.search_ranked`, so the narrowing condition at `src/genesis/db/crud/memory.py:246` (`if not (boolean and "fts5" in str(exc) and "syntax error" in str(exc)): raise`) is untested. A future change that widened the except's scope (e.g. accidentally swallowing a locked-database or missing-table failure) would slip past this suite despite the docstring's claim. Suggest a test that forces `fetch_fts`/`db.execute_fetchall` to raise a non-fts5 `OperationalError` from inside `search_ranked` itself and asserts it propagates.
- NOTE — hygiene: `src/genesis/db/crud/memory.py` adds `import sqlite3  # noqa: F401 — used by search_ranked's FTS5 syntax-error backstop`, but `sqlite3.OperationalError` is referenced at line 232, so the import is used and the `noqa: F401` suppression is unnecessary/stale (harmless, but confusing to a future reader).

Checked and clean: no `nohup`, no direct `task_states` writes, no `/tmp/` usage, no `git add -A`/force-push/`reset --hard`, no bare `except:`, no mutable default args, no raw SQL outside crud modules, no circular/star imports, import grouping is stdlib → third-party → local throughout all four touched modules. The new public function `fts5_term` (`src/genesis/db/crud/_fts.py`) is documented; the docstring edits in `intent.py`/`retrieval.py` correct previously-false claims (the old "re-sanitizes, so raw terms can't cause an FTS5 syntax error" note) rather than leaving them stale. Conventional commit/PR-title prefix present (`fix(memory):`).

**PII/Secrets/Wording scan**

Reviewed all added lines across the 7 changed files plus the PR title, body, and changelog entry. No API keys, tokens, credentials, private IPv4 ranges (10.x/192.168.x/100.64–127.x), IPv6 ULA, SSH keys, or DB connection strings. No personal emails, username-bearing absolute paths, location-revealing timezone identifiers, cloud account/project IDs, hardware identifiers, or private hostnames. The title/body/changelog describe a query-sanitization defect and fix in ordinary engineering terms — no leak/exposure framing, no security-incident language, no naming of private infrastructure.

PII/Secrets/Wording: clean (no soft warnings, nothing to flag)

VERDICT: PASS




---

PR #1876, opened 2026-09-09
docs(skill): why a PR's CI can be missing, red, or lying

WingedGuardian:
## What

Five diagnosis findings from one reviewer session, each of which produced a
**wrong conclusion at the time**, written into the development skill where a
reader already looks.

## The non-duplication check, because it found something

These were approved on the condition that they do not already exist. A keyword
grep said they didn't — `rerun`, `workflow_dispatch`, `statusCheckRollup`,
`log-failed`, `update-branch` all returned zero across `SKILL.md` and all nine
`references/` files.

Reading found otherwise: bullet 8 of the pre-merge traps **already covers** a
CONFLICTING PR suppressing the CI suite, thoroughly. So that cause is not
re-added — the related causes hang off that bullet as corollaries instead of
becoming a rival section a reader would have to reconcile.

That is the same failure this diff documents (a check keyed on existence can
only confirm), caught on its own subject matter.

## Added

Corollaries to the existing bullet 8:

- Mergeability reads `UNKNOWN` after **any** merge, suppressing CI queue-wide.
- A PR opened against a **non-default base** never ran CI at all — `ci.yml`
  filters `on: pull_request: branches: [main]`, the filter is on the BASE, and
  retargeting emits none of the default trigger types. This one never
  self-heals.
- `workflow_dispatch` completes without appearing in the PR's check rollup.

New entries:

- **`gh run rerun` replays the ORIGINAL merge commit**, so it cannot clear a
  failure caused by a base that has since been fixed. Measured: a re-run started
  15 minutes after the fix merged still failed, and
  `git merge-base --is-ancestor <fix> refs/pull/<N>/merge` returned false.
- **`gh run view --log-failed` truncates** on a large suite and shows only
  PASSED lines, so a failing job reads as a clean log.
- **Two individually-green PRs can break `main` together** — one adds a
  repo-wide invariant test, the other adds a violation. No textual conflict, no
  signal until both merge, and then every open PR fails on an unrelated test.
  The entry states how this differs from the structural collision already
  documented, so the two are not conflated.

And in the verify-outcomes guidance: **a check keyed on existence can only
confirm.** If it would still pass when the thing is absent, stale, or the wrong
object, it proved nothing. Three worked examples, all from one session.

## Not claimed

That bullet 8's list is now complete, or that these are the only such traps. The
`--log-failed` truncation point is an observation from this repo's suite on this
date, not a documented constant — the text says "around" and the actionable half
is the remedy.

E2E: none — documentation only, no runtime surface.

### DIFF
--- a/.claude/skills/genesis-development/SKILL.md
+++ b/.claude/skills/genesis-development/SKILL.md
@@ -107,6 +107,32 @@ changes: verify the notification actually arrives. Ask: "If the system
 restarts right now, will this actually work?" If you can't answer yes
 with evidence, you're not done.
 
+**A check keyed on EXISTENCE can only confirm — it can never disconfirm.**
+Before trusting a verification, ask what it would do if the thing were absent,
+stale, or the wrong object. If the answer is "pass", it proved nothing. Three
+shapes of this, all measured in one session 2026-09-08:
+
+- **Waiting for a NEW artifact by testing that one exists.** A poll for "a
+  scheduled-review marker is present" passed instantly against a marker from six
+  days earlier. Key the wait on the artifact's IDENTITY — the head SHA it names —
+  not on its presence.
+- **Matching the first row instead of the right one.** "Is main green?" matched
+  the first `push` run for that SHA, which was the CodeQL workflow (success),
+  while the CI workflow for the same SHA had failed. Filter by the thing you
+  actually mean (`--workflow ci.yml`), and prefer reading the JOB you care about
+  over an aggregate conclusion.
+- **Counting from a listing sorted by the wrong key.** `gh pr list --state
+  merged --limit 40` sorts by CREATION, so PRs created earlier and merged today
+  fall outside the window and vanish; the count read low and looked plausible.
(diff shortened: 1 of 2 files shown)


WingedGuardian:
Automated first-pass review — this PR is documentation-only (`.claude/skills/genesis-development/SKILL.md` + a new `changelog.d/` entry, both markdown), so per process the structural review is skipped and only the PII/Secrets/Wording scan ran.

**PII/Secrets/Wording scan:** clean (no soft warnings either) — no API keys/tokens/secrets, no private IPv4/IPv6 ranges, no SSH keys, no DB connection strings, no personal emails, no usernamed paths, no location-revealing timezones, no cloud account IDs, no hardware identifiers, no private hostnames. Content is meta-documentation about PR/CI diagnosis tooling (`gh run rerun`, mergeability states, `workflow_dispatch` behavior); title/body/changelog contain no leak/security-incident/private-infra wording.

Structural review: skipped (docs-only PR).

VERDICT: PASS



---

PR #1877, opened 2026-09-09
docs(skill): three ways a test can look rigorous and prove nothing

WingedGuardian:
Three additions to the development skill's testing-discipline section, plus one correction to a stale claim already in it. All four come from defects measured while reviewing a single PR.

## What's added

**A RED for the wrong reason is not a verify-RED.** The section already lists six causes for an expected RED that came back GREEN. This is the mirror, and it passes review: the run failed, so the checklist is satisfied, but the failure came from somewhere other than the defect. Split into the two shapes because they cost differently — the primary assertion firing on a condition the *fixture* created goes green on the fix and pins nothing, while a guard-the-guard assertion firing first stays red after the fix and costs a round instead. Measured twice on one PR.

**A fixture must create the shape its docstring claims.** Deliberately framed as the same obligation the existing vacuous-shapes bullet already states, keeping that bullet's mechanical remedy (add the guard-the-guard assert) rather than introducing a second, softer one. What's new is where the shape most often goes wrong: the seam between two components. Two ways to break it, both of which pass against unfixed code — a hand-written intermediate the real producer never emits, and the fix inlined into the test so it grades its own copy.

**Two reviewers pushing opposite values for one constant is a signal about the dependency.** One review pass moved a parse to field `$4`, the next to `$5`; both were right, for different releases of the tool being parsed. The value was never the defect — depending on column position was. Scoped to constants that name a position in something external; thresholds and timeouts are explicitly excluded and routed to the existing Timeout Policy, which requires the opposite move.

## What's corrected

The section cited the cap on a subprocess's captured output as its example of a size limit that cuts silently. That is no longer true at the read — the cap now returns the true drained total and the caller declares it.

It is also not yet a worked example of a loud cut, and that turned out to be the more useful half. The notice is appended at character 50,000 of the result field, and all six onward consumers head-slice it at 200–2000 characters, including the one that feeds the next step's prompt. So the declaration is stripped in every direction it travels, and the slice that strips it declares nothing. The paragraph now carries that as the general rule: when you fix a silent cut, check the readers of the field you just made honest, or you have moved the silence one layer out.

## Review

One internal adversarial audit, which returned seven should-fix findings and no blocker. All seven applied, none accepted unfixed. The most useful was against the correction itself: the first draft promoted that cap to "the model", and the audit refuted it by reading the consumers rather than the function I had cited — which is why that lesson is now the substance of the paragraph rather than a footnote to it.

E2E: none — documentation only. The file is an instruction surface with no runtime path; the changelog fragments are validated by `assemble_changelog.py --check` (22 fragments, all valid).

Ledger: 49c493d8e6e1436cb8ee248a522c3a0b
Ledger: ac5e56f9a0064d72b0100fa57d5100ee

### DIFF
--- a/.claude/skills/genesis-development/SKILL.md
+++ b/.claude/skills/genesis-development/SKILL.md
@@ -213,12 +213,20 @@ repo bounds exactly that at a few MiB (`autonomy/executor/deterministic.py`
 2026-09-04). Losing the tail of a log beats losing the process. The obligation
 there is not to keep the bytes, it is to be LOUD about the cut — say the output
 was bounded and roughly by how much, so nobody reads a clipped log as a
-complete one. That cited cap does NOT yet meet this bar: `_read_limited`
-returns only the retained bytes with no truncation flag, and its caller reports
-the retained length as the total — a 3 MiB stream reads as "2097152 bytes
-total". The cap is right; its silence is the improvable half, cited here as a
-counterexample to "never cut", not as a model of declaring. A silent lossy cap
-is still the defect; a declared one is a resource guard doing its job. This
+complete one. That cited cap has since been fixed AT THE READ and is no longer
+the silent-cut example it was: `_read_limited` drains the whole stream while
+retaining only `limit` bytes and returns `(retained, total_size)`, so the caller
+appends `... (truncated, N bytes total)` with N the TRUE drained total (PR
+#1796; verified in `autonomy/executor/deterministic.py` 2026-09-08).
+**It is still not a worked example of a loud cut, and the reason generalises:
+a declaration has to survive the CONSUMERS, not just be emitted.** That marker
+is appended at character 50,000 of the result field, and all six onward paths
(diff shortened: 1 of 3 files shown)


WingedGuardian:
Documentation-only PR (`.claude/skills/genesis-development/SKILL.md` prose + 2 `changelog.d/` fragments, no code changes) — per process, structural review skipped; PII/secrets/wording scan run.

- No API keys, tokens, passwords, or secrets in the diff.
- No private IPs (10.x/192.168.x/100.64-127.x), IPv6 ULA, SSH keys, or DB connection strings.
- No personal emails, usernamed paths, revealing timezones, AWS/GCP IDs, hardware IDs, or private hostnames.
- PR title/body/changelog entries reference only public artifacts (PR #1796, a source file path, a session-measured lesson) — no mention of a PII/secrets leak, security incident, or named private infrastructure.

VERDICT: PASS



---

WingedGuardian:
@codex review

PR #1886, opened 2026-09-09
test(hooks): a forgotten allowlist entry can no longer pass a test vacuously

WingedGuardian:
## Summary

The merge gate scores a review finding only when its path is in the PR's changed-file set. A finding anywhere else routes to the **off-diff lane**: surfaced as a stderr NOTE, never scored, so it does not block. `_off_diff` is checked *before* the doc-path lever on both branches.

Hook tests pin that changed-file set with a **fixed allowlist**. So a test anchoring a finding on a path the allowlist forgot does not fail — it silently routes off-diff, and a bare `assert not block` then passes **vacuously**, proving nothing about the exemption it is named for. That happened on #1690.

The only defence was prose: a fixture docstring and a standing comment asking every not-block test to also assert its lane marker. A census of 106 tests found **3 that had already drifted from it**. This makes forgetting fail instead.

## This is prophylactic, not a repair — and that corrects the premise it was scheduled on

MEASURED: **zero currently-vacuous tests** (0/68 and 0/38 — every path-anchored test resolves to a listed path; #1690's instance was repaired when `changelog.d/*` was added to the allowlist). Said plainly, because "fixes a live bug" would be a false claim in the permanent record.

What is actually broken is the **lock, not the list**.

## The mechanism was chosen by measurement, and the obvious candidates both die

- Reading `capsys` after `yield` sees `""` in exactly the tests that matter — ~20 call `readouterr()` in the body, which **drains** the buffer.
- A write-through tee on `sys.stderr` recorded **0 of 4** chunks. pytest reinstalls `sys.stdout`/`sys.stderr` per test *phase*, after fixture setup, so the wrap is discarded before the body runs. This was the original design; the probe killed it.
- Wrapping `capsys.readouterr` survives both. 5/5 on a probe with a negative control and an identity check.

The lock **fails open** by design — failing closed would break every hook test on a capture hiccup — so its own self-tests are the load-bearing part, and the docstring says exactly that rather than implying a guarantee it cannot make.

## Review findings, and two are the same class the lock exists for

- My **class-level opt-in** exempted all 11 tests of `TestDiffScopedInlineFindings`, and the docstring justifying the blanket claimed "every not-block case below also asserts its lane marker". **False**: `test_off_diff_codex_p2_pair_not_scored` asserted only `not block`. My own exemption was hiding a live instance of the vacuity being fixed. The blanket is gone — declarations are per test, and the lock itself enumerated that exactly **4 of 11** need one.
- The **drift guard proved `MARKER` matches one lane**, never that all three are covered. Rewording only the CodeRabbit label — the highest-weight lane, 1.0 each against a 1.0 threshold — would have left it green while the lock went blind where it matters most. Now parametrized over all three.
- The drain test was satisfiable by **any** nonzero exit (a conftest `ImportError` exits 4). It now asserts the lock's own message.
- The docstring listed two limits and omitted the only one that fails *silently* (fail-open). Three limits now, that one first.

## Testing

- [x] `ruff check` clean
- [x] **1,352 tests green post-merge** across every hook file main touched (main moved 15 commits under this branch mid-build)
- [x] **Zero new opt-ins required** for anything main added — the lock's blast radius across 15 commits of other people's work is nil

**Mutation sweep 6/6 RED, 0 survivors**, hash-verified restores, exact anchor counts, `compile()`-checked, ABORT on a run with no result line. One is worth naming: deleting the `pytest.fail` goes RED **only** because of a child-process test that reads an exit code — without it that mutation survives. A grep proves a call exists; an exit code proves it blocks.

**Acceptance bar, both directions.** Removing `changelog.d/*.md` from `_FINDING_PATHS` reproduces #1690 exactly:

| state | result |
|---|---|
| defect reconstructed, **no lock** | `1 passed` — clean green, the vacuous pass that shipped |
| defect reconstructed, **lock on** | `1 passed, 1 ERROR` naming the finding and its path |

The test body passes either way. Only the lock separates them.

## Known and deliberate

The lock is Python-level and cannot see a guard run in a **child process**, so the subprocess integration tests are uncovered. No path-anchored finding lives there today; the gap is empty, not closed, and the docstring says so.

No changelog fragment: the CI job validates fragments rather than requiring one, and a test-only change has no user-visible effect.

Ledger: f6555d25a480465994db1e5272041944

E2E: after merge, run `python3 scripts/hooks/git_push_guard.py --check-pr <N>` to confirm the gate is unaffected (this touches no runtime code), then delete a path from `_FINDING_PATHS` in a scratch checkout and confirm the affected test ERRORs with the off-diff message instead of passing.


WingedGuardian:
@codex review

WingedGuardian:

**Scope**: test-only + one doc change — `tests/test_hooks/conftest.py` (new `offdiff_lock` autouse fixture + `OffDiffLock` class), `tests/test_hooks/test_merge_review_gate.py` (new assertions + `TestOffDiffLockItself`), `.claude/skills/genesis-development/SKILL.md` (doc addition). No `src/` changes.

### Structural review

- Pattern violations: none found (no `nohup`, no direct `task_states` writes, no `/tmp/` usage — `tmp_path` is the standard pytest fixture, no `git add .`/`--force`/`reset --hard`, no bare `except:`, no mutable default args, no raw SQL).
- Import hygiene: clean. `tests/test_hooks/test_merge_review_gate.py` adds `from tests.test_hooks.conftest import OffDiffLock` correctly grouped after the stdlib (`json, os, subprocess, sys, pathlib`) and third-party (`pytest`) imports. No star imports, no circularity (`conftest.py` doesn't import back from the test module).
- Coverage: N/A — the changed files are the tests themselves.
- Genesis conventions: commit/PR title uses a correct conventional prefix (`test(hooks): …`).
- Documentation (NOTE): `OffDiffLock.record()` in `tests/test_hooks/conftest.py` (~line 108) and the `declared` property have no docstrings, unlike the rest of the class. Minor — not blocking.

Spot-checked the load-bearing claim against the actual guard: `scripts/hooks/git_push_guard.py` prints `[off-diff CodeRabbit Critical/Major]`, `[off-diff P1]`, and `[off-diff P2]` at the three discount sites, all matching `OffDiffLock.MARKER = "[off-diff "` — the drift-guard parametrization (`TestOffDiffLockItself.test_marker_matches_the_guard_s_real_off_diff_output`) is anchored correctly, not just asserted.

### PII / Secrets / Wording

- No API keys, tokens, credentials, private IPs (10.x/192.168.x/100.64–127.x/fd00::/8), SSH keys, or connection strings in the added lines.
- No personal emails, username-bearing absolute paths, location-revealing timezones, account/project IDs, or hardware identifiers.
- PR title/body/comments: purely technical (test-lock mechanism, PR/issue numbers, measured counts) — no incident language, no naming of private infrastructure, nothing exploitable disclosed.

PII/Secrets/Wording: CLEAN

VERDICT: PASS



---

PR #1887, opened 2026-09-09
docs(skill): a reviewer session flags a dead PR, it does not retire it

WingedGuardian:
Standing user ruling, given 2026-09-08 while a reviewer session was triaging a PR whose mechanism had been shown not to work.

## The rule

A reviewer station can establish that a PR's mechanism does not work. It cannot establish that nothing in the branch is worth reviving, that the author holds no context it lacks, or that the successor design will not reuse most of it. So the disposition for a premise-level defect is a `needs-architecture-session` comment carrying the evidence, with the PR left **OPEN**. `gh pr close` belongs to the session that takes up the revival, at the moment it takes it up.

In the user's words: *"you are not the one who should be closing anything. It needs to wait until the session that actually is going to take it up is open, and that's its responsibility to close it."*

## Why it needed the terminology paragraph

The section immediately above already uses "closing" to mean **merging** — `SKILL.md:2163`, *"**Closing a green PR.** ... merging it under the standing pre-approval ... is closing, not driving."* A new rule twenty lines below saying "never close a PR" would read as forbidding the merge that rule makes mandatory. The new section therefore opens by defining RETIRING as `gh pr close` and stating that merging a green PR is unaffected.

## Why an open PR is the better handoff

An open PR carrying an evidence-bearing comment is found by listing the queue. A closed one is found only if someone remembers it existed. Retiring also costs the review record its addressability: threads on a closed PR stop being where the conversation happens, so reasoning that cost a review round stops being read.

## It cannot strand a dead PR

The rule names its own exit: if a PR should be retired and nobody is picking it up, that is a question for the user, not a judgment call for the review station. The reviving session may also retire in favour of a successor — that is explicitly its call.

## Scope

Documentation only. 34 added lines in `.claude/skills/genesis-development/SKILL.md` plus a `changelog.d/` fragment. No runtime code, no test, no config.

A first attempt at this diff copied `SKILL.md` from a worktree that was behind `origin/main` and read +49/-185 — it would have reverted 185 lines of other sessions' skill work, including the existence-check section from #1876. Caught by reading `git diff --stat` rather than assuming a copy is additive; redone against the current file. Final diff verified at 41 insertions, 0 deletions.

E2E: none — documentation-only change to a model-facing skill file, no runtime surface, nothing to exercise post-merge.

PR #1895, opened 2026-09-09
docs(skill): keep the PR the PR — adjacent findings become issues

WingedGuardian:
Standing rule from the repo owner, 2026-09-09, given after a review session asked whether an adjacent pre-existing bypass should hold up an unrelated PR. It should not, and the session should not have had to ask.

## The rule

A review session's job is to get **that** PR merged. Reviews routinely surface bugs that are real, already on `main`, and nothing to do with the diff in front of you.

One question routes every finding — **does the PR work without this fixed?**

- **No** → fix it in the PR. The builder session should have caught it; a PR that cannot function is not a merge.
- **Yes** → **file a GitHub issue and merge the PR.** Do not grow the diff, do not park it in a reply.

In the owner's words: *"the majority of issues, if they aren't blocking, if they aren't stopping the PR from being functional... those should just get filed as issues on GitHub automatically. Keep the PR the PR."*

## What this changes about approvals

CLAUDE.md requires explicit per-instance approval for every public post, because a public post is irreversible. This records a **standing** approval for one named case: an adjacent, non-blocking, already-existing defect found while reviewing. Every other public post still needs asking, and the text says so in the same paragraph so the carve-out cannot be read as a repeal.

## The exception that is not waived

**A security defect is never filed publicly before it is fixed.** An unpatched bypass, a credential exposure, anything exploitable — private record under `~/.genesis/output/` plus a `follow_up_create` row, and nothing reaches a public surface, this standing approval included. Where a finding is both adjacent and a live bypass, the security rule wins. That tiebreak is stated explicitly rather than left to inference, because this exact overlap came up twice in the session that produced the rule.

## Why it is worth a rule

A review that widens into every adjacent defect stops being a review and becomes an unbounded refactor — which is how a two-finding PR turns into a six-round loop. The PR is the unit of work; the issue queue is where the rest goes.

It also fixes the opposite failure, which was the more common one: an adjacent bug named in a PR reply and never tracked anywhere. The rule makes filing mandatory, not optional, and asks for the measurement and the file:line while the context is still in your head — an issue that only names a symptom costs the next session the whole investigation again.

## Scope

Documentation only. 37 added lines in `.claude/skills/genesis-development/SKILL.md` plus a `changelog.d/` fragment. No runtime code, no test, no config.

Placed as a `###` subsection under "When to DRIVE a Merge", immediately after the retire rule from #1887, since the three together now cover the full disposition space for a PR under review: merge it, leave it open and flagged, or file what does not belong in it.

E2E: none — documentation-only change to a model-facing skill file, no runtime surface, nothing to exercise post-merge.

### DIFF
--- a/.claude/skills/genesis-development/SKILL.md
+++ b/.claude/skills/genesis-development/SKILL.md
@@ -2224,6 +2224,43 @@ The same holds for a superseded PR: name the successor in a comment and leave it
 open. If you believe a PR should be retired and nobody is picking it up, that is
 a question for the user, not a judgment call for the review station.
 
+### Keep the PR the PR — adjacent findings become issues (standing user rule, 2026-09-09)
+
+**A review session's job is to get THAT PR merged.** Reviews routinely surface
+bugs that are real, pre-existing on `main`, and nothing to do with the diff in
+front of you. Those do not belong in the PR and they must not hold it up.
+
+Route every finding by ONE question — **does the PR work without this fixed?**
+
+- **No, the PR is broken without it** → fix it in this PR. The builder session
+  should have caught it, and shipping a PR that cannot function is not a merge.
+- **Yes, the PR works** → **file a GitHub issue and merge the PR.** Do not grow
+  the diff, do not open a discussion, do not park it in a reply and move on.
+
+**Standing approval to file, for this class only.** The user has granted
+blanket, ongoing approval to open GitHub issues for adjacent non-blocking bugs
+discovered during review. Do NOT ask per instance — asking each time was the
+friction this rule removes. That standing approval is scoped to exactly this
+case: an adjacent, non-blocking, already-existing defect found while reviewing.
+Every other public post still needs explicit per-instance approval.
+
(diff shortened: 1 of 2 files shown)


WingedGuardian:
Documentation-only PR (`.claude/skills/genesis-development/SKILL.md` + `changelog.d/*.md`, no code changes) — skipping structural review per process, PII/Secrets/Wording scan only.

- No API keys, tokens, passwords, secrets, private IPs, IPv6 ULA addresses, SSH keys, DB connection strings, or env-var secret values in the diff.
- No personal emails, username-bearing paths, location-revealing timezone IDs, cloud account/project IDs, hardware identifiers, or private hostnames in the diff.
- PR title/body describes a general review-routing policy (adjacent non-blocking findings → GitHub issue, PR merges). The "adjacent pre-existing bypass" mention is an abstract example category used to motivate the rule, not a reference to a specific vulnerability, an actual incident, named private infrastructure, or exploitable detail. No wording violation.

VERDICT: PASS



---

PR #1904, opened 2026-09-09
docs: filing a bug you found while reviewing should not need permission

WingedGuardian:
## Summary

Three review-process rules said two things at once, and the third was costing us bugs.

## The policy change

Opening a GitHub issue required **explicit approval every time** — and that is what kept adjacent bugs unfiled. Someone finds a real problem mid-review, has nowhere cheap to put it, and it evaporates.

The gate moves to **what goes IN the issue** rather than permission to open one:

- **Scrub anything personal or identifying** — names, hosts, IPs, username-bearing paths, real-world context. An issue carries technical detail only.
- **Borderline either way? Ask.**
- **Filing needs no per-instance approval.** A bug found while reviewing a PR, that does not block that PR, is the ordinary case — file it and keep the PR moving.
- **A security defect is still never filed publicly before it is fixed.** That limit does not move.

Applied to **every consumer**, not just the rule's home — a policy that lands in one file and not its readers is the drift this PR exists to remove. Swept for the old phrasing and triaged each hit: CLAUDE.md's rule and its zero-drop cross-reference, the skill's "Keep the PR the PR" grant and its security carve-out, and the `mcp-tools-guide` summary CLAUDE.md points at. Four same-words-different-subject hits were deliberately left alone (the escalation caps that *consume* standing approval, and the per-use approval for an alternate **reviewer** — a different gate).

## Two false claims, re-derived from source rather than by picking a side

**Stale-review relief is DELTA-scoped, not PR-scoped.** The docs said a hook-surface PR gets no leniency at all, while a passage further down measured the opposite. `_classify_post_review_delta` iterates the *delta's* files — so a docs-only follow-up commit on a gate PR can still be review-trivial. Not cosmetic: the false version is what a session budgets a follow-up against.

**A usage-limits reply is evidence, not proof.** It was described as unavailability, contradicting a measured case (#1484) where real inline findings arrived while the issue channel still showed only the quota message. This sentence is the *sole* stated precondition for reaching for a fallback reviewer, so the false version spends an approval on a reviewer that was never down.

## What did NOT change

Verified explicitly, because a relaxation PR is exactly where an unintended one hides:

- [x] `No unsanctioned financial transactions` — present, unchanged
- [x] `Autonomous-CLI approval gate is MANDATORY & non-negotiable` — present, unchanged
- [x] a peer's request is never approval — present, unchanged
- [x] the security-defect limit, in **both** files — present, unchanged

The only "non-negotiable" phrase removed anywhere is the issue-route one, which is the change.

## Testing

- [x] 58 tests green — the injection-budget suite (CLAUDE.md is an injected identity file under a CI-enforced ceiling) and the published-bypass-recipes guard (this relaxes a publishing rule)
- [x] Residual sweep: the only surviving matches for the old phrasing are the three new sentences stating the new rule
- [x] changelog fragment validates; runtime/test/config untouched

## Known and deliberate

`SKILL.md:1591`'s per-use approval for an alternate **reviewer** is left alone here — #1903 rewrites that same passage, and editing it in two open branches would conflict. Different gate, different PR.

All three contradictions were found while reviewing #1903, which works fine without them — which is exactly the "Keep the PR the PR" routing this repo just adopted.

E2E: none — process doctrine, no runtime surface.

### DIFF
--- a/.claude/docs/mcp-tools-guide.md
+++ b/.claude/docs/mcp-tools-guide.md
@@ -115,10 +115,14 @@ another has already scanned and approved, between the scan and the post.
 
 Two things stay with you, because they are judgment, not mechanism:
 
-**Approval, every time.** A public post is irreversible;
-`contributor_issue.py` classifies it exactly that way. Channel-driven sessions run
-with `skip_permissions=True`, so no tool confirmation appears — ASK, in words, and
-get a yes, for each issue. A prior "go ahead" does not carry.
+**What goes IN it.** A public post is irreversible — `contributor_issue.py`
+classifies it exactly that way — so the gate is the CONTENT, not permission to
+file. Filing needs no per-instance approval; that mandate was the friction that
+kept adjacent bugs unfiled, and it is gone. What replaces it is the scan below,
+plus one judgement that stays yours: **borderline, in either direction? Ask.**
+Channel-driven sessions run with `skip_permissions=True`, so no tool
+confirmation appears and nothing will stop you — which is exactly why the
+scan is a habit rather than a prompt.
 
 **Privacy-scan the title AND the body.** Both egress. Technical detail only: no
 IPs, hostnames, absolute home paths, identifiers, or raw install metrics (row
--- a/.claude/docs/mcp-tools-guide.md
+++ b/.claude/docs/mcp-tools-guide.md
@@ -28,8 +28,10 @@ consciously not pursuing → `follow_up_create` with `work_state="deferred_cold"
 
 `CLAUDE.md` carries the routing principle (Genesis-repo work → a GitHub issue;
 user-owned + purely-local operational work → a follow-up; consciously-not-doing →
-`tabled`) and its two hard limits (explicit approval every time; never publish a
-security defect before it is fixed). This section is the operational detail.
+`tabled`) and its two hard limits (scrub personal/identifying detail from anything
+public, and ask when borderline; never publish a security defect before it is
+fixed). Filing an issue does not need per-instance approval. This section is the
+operational detail.
 
 ### Filing it — use the script, not a hand-typed command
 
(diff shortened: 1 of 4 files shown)


WingedGuardian:
@codex review

WingedGuardian:
Automated first-pass review (scheduled).

**Scope**: This PR is documentation-only — all 4 changed files are `.md` (`CLAUDE.md`, `.claude/docs/mcp-tools-guide.md`, `.claude/skills/genesis-development/SKILL.md`, and a new `changelog.d/` fragment). Per protocol, structural review is skipped; only the PII/Secrets/Wording scan ran.

- Code diff (hard blocks): no API keys, tokens, passwords, bearer tokens, or base64 credentials; no private IPv4 (10.x/192.168.x/100.64–127.x) or IPv6 ULA (fd00::/8) addresses; no SSH keys or fingerprints; no DB connection strings with credentials; no secret-bearing env values. None found in the added lines.
- Code diff (soft warnings): no personal email addresses, no username-bearing absolute paths, no location-revealing timezone identifiers, no AWS/GCP account IDs, no hardware identifiers, no private hostnames. None found.
- PR metadata + changelog (critical wording check): title, description, and the new `changelog.d/20260909165024-fixed-review-doc-contradictions.md` entry describe a process/policy change (relaxing the GitHub-issue filing approval gate to a content-scrub gate, and reconciling three internal doc contradictions about stale-review leniency and reviewer-unavailability evidence). No mention of a PII/secrets leak, no security-incident references, no naming of private infrastructure, and no exploitable vulnerability detail. Clean.

VERDICT: PASS



---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

WingedGuardian:
Fixed at `4636eb066` — and you're right that this mattered more than its severity suggested.

This branch exists to remove a contradiction about issue-filing approval. It updated the guide's **summary** and left the guide's **operational section** saying "Approval, every time … ASK, in words, and get a yes, for each issue." An agent following that document reaches the operational section — that's what it's for — so the branch would have shipped having *moved* the contradiction rather than removed it, inside the one file whose whole job is telling you how to file.

The block now states what goes *in* an issue rather than who may open one. The `skip_permissions=True` sentence is kept and re-pointed: it used to justify asking, and now explains why the privacy scan has to be a habit, since nothing will prompt you.

Swept the phrase across every tracked `*.md` rather than patching the flagged line. Two files needed changing and both are changed; the two surviving matches are a different subject and stay — unsanctioned financial transactions, and the sovereignty gates where a peer's request is never approval.

WingedGuardian:
@codex review

PR #1918, opened 2026-09-10
docs(skill): fold six review-queue lessons into genesis-development

WingedGuardian:
Six rules out of one long pass through the pull request queue, folded into
`.claude/skills/genesis-development/SKILL.md`. Each landed beside the section it
extends — verification, measurement, review protocol, merge-driving — rather
than in a "lessons from a session" appendix, and two of the six extend text that
was already half there instead of adding a near-duplicate.

**1. A correctness test cannot see a resource defect** (Test-First Discipline,
new bullet). Verifying a change three ways is one verification when all three
ask the same question. An in-flight fix was checked at the parser level, at the
guard level, and against the exploit it existed for — three correctness passes,
all green — while the scan it added was quadratic in the length of a single
token: sub-millisecond on the linear form at any length, seconds and then
minutes on the pathological one. A PreToolUse hook that overruns its wall-clock
is killed, a killed hook fails open, and the parser is shared across the hook
stack — so where a blown deadline means permit, a super-linear scan over
attacker-influenced input is the defect regardless of how correct its output is.
The rule asks three questions the correctness suite never asks (complexity,
maximising input, behaviour on not-finishing) and prescribes a timing test,
because nothing else in the suite can go red on this class.

Written deliberately without the reproduction parameters — no input sizes, no
timeout tier, no affected-module count, no quoted fail-open line — and the
bullet says why, as a rule of its own: a lesson about a vulnerability does not
need the vulnerability's parameters, and "there is a fix" is not a licence for
them either, because what matters is whether the fix has landed on `main`.

**2. A null result is a lead, never a clearance** (Acceptance Bar, new block).
Three null results from one session were wrong the same way, and each was
overturned by construction rather than by more sampling — a structural claim
generalised from samples of one spelling of a construct, a "zero flips across
129,179 real commands" whose regression turned out to be constructible, and the
"verified three ways" above. A corpus measures what has been typed, never what
is typeable, so a null handed to a subagent, a peer, a PR body or the user
carries its denominator and its method's blind spot in the same breath as the
number.

**3. Generate the cells from the grammar, not from history** (extends the
existing corpus-replay/matrix bullet). Enumerating {program} × {option form} ×
{construct} × {position} and running every cell through real bash and the real
guard subprocesses on both trees produced 746 cells and a control that moved —
744/2 on the base, 339/407 at head, 534/212 after the fix — surfacing 206
over-blocks a 129,179-command corpus had shown none of. Includes the probe trap:
these guards emit ASK as JSON on stdout with exit code 0, so a probe reading only
the return code cannot tell ASK from ALLOW, which is exactly the state a new flag
usually produces.

**4. Refuting a finding is itself a claim** (review-loop discipline, new bullet,
next to "external review feedback is a set of claims to VERIFY"). A structural
refutation must cover the construct's grammar, not a sample of its instances;
where it cannot, the finding stands. The tell in the incident was inside the
refutation's own text — it said "I have NOT proven no clean construction exists"
and stated the conclusion in the next sentence.

**5. A checker's verdict depends on the tree you ran it from** (extends the
`--check-pr` bullet). `--check-pr` is (remote PR state × local gate code). Run
from a worktree carrying an unmerged gate change it reports what would be true
after that change lands — correct about the fix, false about the PR. Includes the
cheap bounding check: if `git log --since=<window> origin/main -- <checker path>`
is empty, every verdict in that window used identical logic.

**6. Which one first — prefer a fix, because it is usually smaller** (new
subsection under the merge-driving rule, which said when to drive but not which
to pick). Across the 65 open fix/feat pull requests: feat n=30, mean 2,024 added
lines, 67% over a thousand; fix n=35, mean 804, 29% over a thousand. Chained to
86% of pull requests over a thousand lines reaching 3+ external rounds against 6%
under two hundred. The findings-per-PR gap is modest (3.00 vs 2.77), so the rule
is written on the size property rather than the commit prefix — a small feature
is not deprioritised and a large fix gets no pass — and stated as one axis in a
larger calculus, not a hard ordering.

Also included: one `changelog.d/` fragment.

**Checked and deliberately not added.** The mechanism-interrogation bullet ends
"re-derive rather than adding a seventh", so lesson 1 was placed in Test-First
Discipline instead of becoming a seventh axis there. Lessons 3 and 5 extend
existing text. The verify-RED causes, the vacuous-test shapes, the
existence-check rule and the CI-diagnosis list were each read in full and cover
different ground.

E2E: none — documentation only. The file is a skill document read at agent load
time; there is no runtime surface to exercise.


WingedGuardian:
**Scheduled first-pass review** (doc-only PR — this diff touches only `.claude/skills/genesis-development/SKILL.md` and a `changelog.d/` fragment, so per protocol this runs the PII/Secrets/Wording pass only; structural review is skipped).

Note: this review would normally be posted via `gh pr review --request-changes`, but GitHub rejects a "request changes" review from the PR's own author account, so the blocking findings are posted here as a comment instead.

### PII/Secrets/Wording

**WORDING (CRITICAL) — HARD BLOCK — 2 findings**

1. `.claude/skills/genesis-development/SKILL.md:228` — states the exact working bypass form for `git_push_guard.py`'s force-push detection: a brace-range expansion with identical endpoints that expands to a single argv word, described specifically enough to reproduce against any install still on the pre-fix guard. This is a vulnerability disclosed with reproduction detail, in a document that ships in the public repo.

2. `.claude/skills/genesis-development/SKILL.md:830-856` — documents a specific algorithmic-complexity (O(n²)) DoS in `scripts/hooks/shell_parse.py`: exact input-size/timing thresholds to blow each PreToolUse timeout tier, a direct quote from the guard's own code that a SIGKILL "fails open on every gate," and the exact count (and by extension the identities, via the cited grep) of all hook modules affected. Taken together this is a ready-to-use recipe for disabling every PreToolUse security gate (push guard, destructive-command guard, financial-transaction gate, etc.) with one attacker-controlled string. This matches the CRITICAL bar for "advertises a vulnerability fix with enough detail to be exploitable" even though the underlying guard fix has apparently already landed elsewhere — the exact syntax/thresholds/quoted internals don't need to be in the permanent public record for the lesson to land.

**Soft warnings:** none — no private IPs, personal emails, or username-bearing absolute paths found in the added lines.

**Recommendation:** the underlying lessons (a correctness suite can't see a resource/complexity defect; a null result needs its blind spot stated) are valuable and worth keeping — only the two passages above need to drop the reproduction-level specifics (exact bypass syntax, exact timing thresholds, the "fails open on every gate" quote, exact affected-module count). The `changelog.d/` fragment already describes the same lesson at the right level of abstraction ("a scan that slows down super-linearly on attacker-chosen input," no syntax/numbers) — the SKILL.md passages could be brought to a similar level without losing the pedagogical value.



---

WingedGuardian:
Requesting the scheduled PII/secrets scan at head `3a1c59b197ea` — the merge gate reports `scheduled-claude: BLOCK — leaks missing at head`, and there is no earlier accepted review on this PR to carry forward.

Every other gate passes: CI green, Codex reviewed at head with no findings, mergeable, review-body ok. This is documentation-only (a skill file plus a changelog fragment), so the scan should be quick.

Noting for anyone else driving the queue: the scanner takes the single most-recently-updated open PR, so this comment is also the bump that puts this one on top. If you were about to fire it for a different PR, re-bump yours and re-run — we would otherwise race.

WingedGuardian:
Pushed `ea5fb5859` in response to the scan's two hard blocks. Both were correct
and both are addressed by removal rather than by rewording.

Removed from three passages (the third was not in the report, but is the same
class — the refuted argument was quoted in terms that named the vector):

- the working construction, now stated only as "an expansion in verb position"
  with "a degenerate case that emits exactly one word";
- the exact input sizes and per-tier timings, now "sub-millisecond on the linear
  form; seconds, then minutes, on the pathological one";
- the quoted fail-open line, now paraphrased with no file cited;
- the count of modules behind the shared parser, now "the parser is shared, so it
  does not stop at the one guard being edited".

Verified by filtering the branch's added lines for every token that carried
reproduction value, rather than by spot-check: zero matches, in the skill file
and in the changelog fragment.

Kept, as they carry the lesson and nothing else: the axis argument, the
timing-test prescription, the null-result rule, the matrix control requirement,
and the accidental-quadratic shape described as a class.

Two things the report did not cover, fixed here as well. The **PR body** carried
the identical specifics — a public surface, so redacting only the file would have
left the recipe one click away; it has been rewritten to the same register. And
the report's aside that "the underlying guard fix has apparently already landed
elsewhere" is **not correct** — the detector's identifiers are absent from
`origin/main`, so the published tree is pre-fix and the earlier text described a
live bypass, not a historical one. That raises the finding's severity rather than
lowering it.

The general rule is now in the file itself, inside the bullet whose
over-specificity caused it: a lesson about a vulnerability does not need the
vulnerability's parameters, and a fix existing as a pull request is not the fix
landing — the check is against the default branch.

Ready for the re-scan at `ea5fb5859`.

WingedGuardian:
Documentation-only PR (`.claude/skills/genesis-development/SKILL.md` + one `changelog.d/` fragment) — skipping structural review per protocol, PII/wording scan only.

**PII/Secrets/Wording scan:**
- No API keys, tokens, passwords, or secrets found.
- No private IPs (10.x/192.168.x/100.64–127.x), IPv6 ULA, SSH keys, or DB connection strings found.
- No personal emails, username-bearing paths, location-revealing timezones, AWS/GCP IDs, hardware identifiers, or private hostnames found.
- Content is abstract retrospective prose (resource-exhaustion review lessons, corpus-testing methodology, refutation discipline). One passage discusses a past "resource defect" in a shared shell-parsing hook that could fail a guard open under a timeout kill — the PR body explicitly states it was written **without** reproduction parameters (no input sizes, no timeout tier, no module count, no fail-open quote) specifically to avoid teaching an exploit. No specific guard/module is named, no repro steps, no exploit code. Read as legitimate internal engineering retrospective, not a leak disclosure or an infra-specific vulnerability advertisement.
- PR title/body/changelog: no mention of a PII/secrets leak, no security-incident references, no named private infrastructure.

PII/Secrets/Wording: clean (no hard blocks, no soft warnings)

VERDICT: PASS



---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

Round 3. Applied all three doc-path P2 findings from the previous round rather than accepting them, because this file is loaded by every genesis-development session and a wrong rule there executes:

- **Per-cell, not distribution** (SKILL.md:822-829) — the old rule said identical verdict distributions prove the harness is dead. It doesn't: cells can swap ALLOW/BLOCK while totals hold, and a resource-only fix preserves every verdict by design, which is exactly what the following bullet describes. Now compares per cell keyed by input, with a known differentiating control for liveness.
- **Compare the checker, not main's log** (SKILL.md:1672-1680) — the old bounding check inspected `origin/main` history for a file that executes from the working copy, so it returned empty in precisely the drift case it was meant to catch. Now diffs the checker across trees.
- **Timing claim** (SKILL.md:845) — said sub-millisecond where the measurement was 0.004s. Now states the measured value.

Focus this round on whether the three replacements are themselves correct, especially the per-cell rule's claim that a matching distribution proves nothing in either direction.

WingedGuardian:
**Automated review — PII/Secrets/Wording pass**

This PR is documentation-only (both changed files are `.md`: `.claude/skills/genesis-development/SKILL.md` and a `changelog.d/` fragment), so per policy this ran the PII/secrets/wording scan only; structural review was skipped.

**Code diff (added lines):** clean. No API keys/tokens/passwords, no private IPv4 ranges (10.x/192.168.x/100.64-127.x), no IPv6 ULA, no SSH keys, no DB connection strings, no personal emails, no username-bearing absolute paths, no location-revealing timezone identifiers, no AWS/GCP account IDs, no hardware identifiers, no private hostnames.

**PR title/body/changelog wording:** no security-incident references, no PII/secrets-leak language, and no private infrastructure named.

One passage got closer scrutiny: the new Test-First Discipline bullet describes a quadratic-complexity defect in a shared shell parser that a PreToolUse hook timeout can turn into a fail-open bypass. I checked it against the "advertises vulnerability fixes with enough detail to be exploitable" hard-block criterion — the PR body explicitly states this was written "deliberately without the reproduction parameters," and the merged text confirms that: no vulnerable file/module name, no triggering construct, no timeout thresholds, no affected-guard count are given, only the general shape (super-linear scan, attacker-influenced input, fail-open on kill). That's below the exploitability bar, so this is a note rather than a block — flagging for awareness, not requesting changes.

VERDICT: PASS



---

PR #1924, opened 2026-09-10
fix(security): refuse this install's own values in a public push, not just report them

WingedGuardian:
## The advisory that watched a real leak go past it

This branch was written 2026-09-08 and then sat unopened. In the two days since, a public branch on this repo carried an install's real machine name and account name — and the advisory hook saw the push and said nothing that stopped it. That is the case this change exists for.

## What changes

`scripts/hooks/pre_push_privacy_review.py` stops *reporting* this install's own values in an outgoing public push and starts **refusing** them.

The supporting half is `src/genesis/contribution/fingerprints.py`, which generates `~/.genesis/release-fingerprints.txt` from local config — the install's specific literals (hostnames, subnets, ULA prefixes, private repo name, timezone) that a public scanner must never itself contain. The tracked tree keeps only GENERIC class patterns (all RFC1918, IPv6 ULA per RFC 4193, `/home/<user>` shapes); the specific values live outside it. `sanitize._check_fingerprints` already reads that file — this generates it.

Format contract is strict because three readers consume it: one pattern per line, comments on their own line only (`_check_fingerprints` skips lines that *start* with `#`, so an inline `pat # note` would compile the note into the regex), and every emitted pattern must be valid for **both** Python `re` and GNU `grep -E` — the commit-msg hook greps it with `grep -E -f`. Only `re.escape`'d literals and `\b` are emitted, never `\d`/`\w`/`\s`, which are a literal `d`/`w`/`s` under some `grep -E` builds and would silently fail to match: a false negative on a leak.

Hand-edited patterns below the marker pair survive regeneration, so an install can add what config cannot derive — a hardware token, a cloud account prefix.

## After merging

Run `python -m genesis.contribution.fingerprints --write`, then confirm `~/.genesis/release-fingerprints.txt` is mode 0600. The generator is what turns this from a mechanism into a defence — on this install the file currently knows the account name but **not** the machine name, which is precisely the value that leaked.

## Scope, stated honestly

This closes the **value**-based half: things the install knows about itself. It does **not** yet catch a credential by SHAPE — an unknown API key, JWT or PEM block still passes. That is a real remaining gap and it is being tracked separately; this PR is not it.

## Verification

95 tests green across `test_fingerprints.py`, `test_pre_push_privacy_review.py` and `test_review_invalidate_on_commit.py` after merging 37 commits of main. ruff clean. Every harvest source is independently guarded, so a fresh install with partial config degrades rather than crashing, and generic tokens that would over-block (`user`, the container's own hostname) are excluded by an explicit deny-list.

E2E: after merge, run the generator and confirm a push carrying a known install literal is refused rather than merely annotated.


WingedGuardian:

Note: this review would normally be posted via `gh pr review --request-changes`, but GitHub refuses a self-request-changes on this account's own PR (same restriction as self-approval) — posting as a comment instead. Treat the wording finding below as blocking.

### Domain 1 — Structural Review: no pattern violations found

No ERROR-level hits against the pattern list (no `nohup`, no direct `task_states` writes, no `/tmp/` usage, no `git add -A`/`.`, no force-push/reset --hard, no bare `except:`, no mutable defaults, no unguarded `os.killpg`, no raw SQL against `genesis.db`).

Import hygiene is clean — the hook (`scripts/hooks/pre_push_privacy_review.py`) stays stdlib-only (`contextlib`, `json`, `os`, `re`, `shlex`, `subprocess`, `sys`, `time`, `pathlib`, `urllib.parse`) with the `genesis.contribution.sanitize` import correctly deferred inside `_scan()`; `fingerprints.py`'s new `ipaddress`/`json` imports are correctly grouped and alphabetized. No circular or star imports.

Coverage: no new `src/` file lacks a test — this PR's own `tests/test_contribution/test_fingerprints.py`, `tests/test_hooks/test_pre_push_privacy_review.py`, and new `tests/test_hooks/test_review_invalidate_on_commit.py` cover every changed function, including the bypasses named in the docstrings/comments (SSH-vs-HTTPS remote canonicalization, refspec-aware diffing, `_COMMIT_OK` success detection).

Documentation: every new public/module function (`_canonical_repo`, `_push_source_ref`, `_render`, `_harvest_tailnet_addrs`, `_COMMIT_OK`) carries a docstring or comment explaining intent.

Commit prefix (`fix(security):`) is conventional.

### Domain 2 — PII, Secrets, Wording

**Code diff: clean.** Every private-range literal added in this diff (`10.11.12.13`, `192.168.1.5`, `100.64.0.x`, `100.100.5.6`, `8.8.8.8`, the `fd7a:1:2::…` ULA examples) and the email fixture (`realperson@personal.example`) are synthetic values in test files whose explicit purpose is to exercise the scanner/sanitizer under test — the documented synthetic-fixture exemption applies, not a hard block. No real API keys, tokens, SSH keys, DB connection strings, or install-specific hostnames were added anywhere in the diff.

**PR description — WORDING: this is the CRITICAL-tier issue this review pass exists to catch, and it hard-blocks.**

The PR body opens with the heading "The advisory that watched a real leak go past it" and states, as fact, in permanent public record:

> "In the two days since, a public branch on this repo carried an install's real machine name and account name — and the advisory hook saw the push and said nothing that stopped it."

and later, in the "After merging" section:

> "...on this install the file currently knows the account name but not the machine name, which is precisely the value that **leaked**."

This isn't a description of the *mechanism* being fixed (advisory-only design, no blocking, same-result timing — which is fine and clearly explained elsewhere in the body) — it's an on-the-record confirmation that a real data-exposure incident occurred, on this repo, involving a real machine name and account name. A security-incident admission in a PR description is permanent and, unlike code, cannot be scrubbed out of git history later.

**Requested fix:** rewrite the motivating narrative (the opening section heading and the "In the two days since…" / "...which is precisely the value that leaked" sentences) to describe the gap in the *mechanism* generically — advisory text arrives in the same result as the completed push, so "confirm before it lands" was never actually enforceable — without confirming that a specific real leak occurred or naming what it contained. The CHANGELOG.md additions in this same diff are worded correctly this way already (generic "your own machine's addresses", no incident confirmation) and can be used as the model for the PR body's rewrite.

Everything else in this PR (the DENY/ADVISORY split, the SSH/HTTPS canonicalization fix, the refspec-aware diff, the tailnet harvest, and the commit-marker invalidation fix) looks sound on this first pass — this block is solely about the PR description wording.

---


---

WingedGuardian:
Superseded by #1942, which takes **this PR's implementation as its base** rather than replacing it.

Both PRs are the same session's work two days apart — this one landed the design on 09-08, and #1942 rebuilt the same capability on 09-10 without adopting it. Reading them side by side, this one is the better design, and one of its choices fixes a real defect in the other:

- **Two verdicts split by confidence.** DENY for a value this install actually has; ADVISORY for a generic RFC1918/CGNAT literal. The other implementation had a single verdict and measured **27.5% of merged commits blocked** before rediscovering this same distinction as a false-positive rate.
- **A renderer that never echoes the matched value.** `_check_emails` interpolates the offending address into its message; the other implementation printed it. Adopting this renderer fixes the class — a future scanner that interpolates a secret cannot leak through it either.
- `permissionDecision: deny` rather than exit 2, refspec-aware source resolution, and an explicit "NOTHING was scanned" message where silence read as clean.

#1942 folds four things onto it that this PR does not have: credential **shape** detection (this PR only recognises values the install already has, so a key belonging to anyone else passed untouched), gitleaks for the JWT/PEM/high-entropy classes, the annotation path, and whole-branch scanning — this PR kept the incremental `origin/<branch>` base, which makes already-pushed content invisible and lets a re-push report clean on a branch that is not.

Measured on the consolidated result: the motivating leak is refused (2 deny findings) while the 13 generic shapes in the same diff correctly stay advisory. Deny rate over 40 merged commits: 3/40, all test fixtures the annotation clears. All 34 tests from this PR are kept as the base, plus 22 folded on.

One gap inherited and fixed there: this PR's own `realperson@personal.example` fixture was unannotated, so the gate denied its own branch.

Closing in favour of #1942. Nothing here is discarded.

PR #1929, opened 2026-09-10
fix(memory): the dedup short-circuit supersedes, via the shared validator

WingedGuardian:
**Split 3 of 5 from #1831.** Base: `feat/memory-supersede-tool` (#1933).

> **Stacked**: `main` → #1928 → #1933 → this. CI does not run for a non-default
> base, so the targeted suites and mutation matrix are below. Review #1933 first
> — it is the redesign this PR is built on.

## The defect

`MemoryStore.store()` returns early when the exact content is already stored,
and that early return sits **~300 lines above** the supersession block. So a
`supersedes` request on that path deprecated nothing, silently. It is the
likeliest path there is: retrying a correction re-sends the same content with a
corrected id and lands exactly here.

## Two changes from the first revision — each replaces a patch with a structure

**1. This PR no longer carries its own copies of the successor guards.** The
first revision grew them inline behind a `verify_successor` flag. #1933
established `memory_supersede` and with it a shared `_validate_supersede_pair`;
two callers now need identical validation, so this one **calls it**. The property
the flag encoded — *only the dedup path validates the successor* — is now
expressed by **where** the call sits (store.py:271, dedup path only) rather than
by a boolean threaded through a method that mostly ignores it.

That validation matters here and nowhere else, because on this path the
successor is **not a memory the caller chose** — it is whatever
`find_exact_duplicate` matched, and that query selects on content length and a
200-char prefix only, reading neither the supersede target nor the `deprecated`
column. So it can hand back the target itself, or an already-deprecated row.

**2. Resolution moves above the duplicate lookup.** The first revision answered
the swallowed-exception finding by moving the supersede out of the lookup's
`except Exception`. That is handler placement. Resolving first means an
unresolvable handle **never reaches the lookup**, never reaches the handler, and
cannot duplicate anything by any route.

It also fixes something the first revision left: a request to deprecate a memory
that does not exist used to succeed quietly whenever the content happened to be
a duplicate, because the early return fired before anything checked the handle.

## Behaviour change, stated plainly

A rejected pair on the dedup path now **raises** out of `store()` rather than
being logged. Validation runs before any write, so nothing is stored when it
rejects, and the exception carries `successor_id` — the memory that already holds
this content — so the caller loses nothing by the failure. Asserted directly
rather than assumed.

An infrastructure failure **after** validation is still swallowed and logged,
mirroring the normal path: content is durable by then, and raising reads as "the
store failed" and invites the duplicating retry.

## Verification

**Verify-RED — 3 mutations, all applied, all biting:**

| mutation | red |
|---|---|
| the dedup path does nothing (`main`'s behaviour) | 3 tests |
| it supersedes but skips the shared validator | the same 3 |
| resolution moves back **below** the lookup | `test_an_unresolvable_target_never_reaches_the_dedup_lookup` — a test neither of the others touches |

Two honesty notes on the matrix itself, both also in the commit:

- The first form of the no-supersede mutation produced a **SyntaxError**, which
  proves nothing — a collection error is not a RED. It was repaired to a valid
  no-op and `ast.parse`-checked before being counted.
- Mutations 1 and 2 fail the **same** three tests. They remove different things,
  but the suite does not currently discriminate "does not supersede" from
  "supersedes without validating" — the row-state assertions catch both. A real
  limit of this evidence, recorded rather than glossed.

- 130 targeted tests green; `ruff` clean.
- The new tests drive the real dedup path through real SQLite (`memory_fts`
  populated, production `FTS5_DDL` imported rather than retyped).

`test_a_memory_cannot_supersede_itself` is **removed here, not lost**: the
unit-level case now lives with the validator it tests, in #1933; the dedup-path
case it also covered is kept.

## Not fixed here

`find_exact_duplicate` matches deprecated and subsystem rows. Every dedup caller
inherits that, and narrowing it changes what "duplicate" means repo-wide. This
PR works around it at the supersede site by validating the successor it is
handed.

E2E: `memory_store(content=<already-stored text>, supersedes=<handle>)` over MCP — confirm the target is deprecated (previously it was not); then re-send unchanged content alongside a handle for the memory holding it and confirm it raises with that memory still at `deprecated=0`.

---

<a href="https://app.devin.ai/review/wingedguardian/genesis-agi/pull/1929" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,8 @@
 # fastmcp <3 for API stability, transitive secure floors) that automatic
 # version-update PRs would fight. GitHub's Dependabot *security* alerts are
 # already enabled for the repo and cover the pip CVE side; the pip-audit CI job
-# gates the resolved tree per-PR. So dependabot only keeps the Actions current.
+# reports on the resolved tree per-PR without blocking on audit findings.
+# Dependabot version updates here only keep the Actions current.
 version: 2
 updates:
   - package-ecosystem: "github-actions"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
   workflow_dispatch:
   schedule:
     - cron: "0 6 * * 1"  # weekly Mon 06:00 UTC — surface new CVEs even without repo activity
@@ -34,10 +33,9 @@ jobs:
         run: ruff check src/ tests/ scripts/
 
   dependency-audit:
-    # Fail on any KNOWN-vulnerable dependency in the resolved runtime tree,
-    # except the triaged IDs below. A red run is a SIGNAL to bump the dep or
-    # add a triaged ignore — not a hard freeze (the ruleset has no required
-    # checks). Complements GitHub's already-enabled Dependabot security alerts.
+    # Report known vulnerabilities outside the retained ignore baseline.
+    # Audit findings remain visible in the log without blocking this job;
+    # dependency installation failures still fail it.
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
@@ -50,9 +48,10 @@ jobs:
           pip install -e .
           pip install pip-audit
       - name: Audit dependencies for known CVEs
+        continue-on-error: true
         run: |
           # Triaged --ignore-vuln IDs (rationale of record: pyproject.toml pins).
-          # A NEW / untriaged advisory is NOT in this list and so fails the job.
+          # Advisories outside this list remain visible in the log.
           #   litellm 1.78.7 — every ID is a proxy-server / JWT / OIDC / UI /
           #     MCP-gateway vuln; Genesis uses acompletion/completion_cost only
           #     (no proxy server), so none are reachable. Pinned <1.83 also dodges
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -286,7 +286,8 @@ the rule below used to fire so rarely. Three concrete triggers:
    X is not what its name suggests. Especially when you went looking for X
    expecting it to be absent.
 3. **You CORRECTED a belief** — your own, a prior session's, or a written note's.
-   Use `supersedes` to link the correction to what it replaces. This is the
+   Use `supersedes` to link the correction to what it replaces, or
+   `memory_supersede(old_id, new_id)` when both memories already exist. This is the
    highest-value trigger and the easiest to skip, because being wrong does not
    feel like a finding. It is the one that stops the next session paying for the
    same mistake.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -287,8 +287,10 @@ source — there is no `requirements.txt`). Known-CVE scanning runs automaticall
 
 - **CI** — the `dependency-audit` job in `.github/workflows/ci.yml` runs
   `pip-audit` against the resolved runtime tree on every PR/push and weekly,
-  failing on any new (untriaged) advisory. Already-triaged, not-reachable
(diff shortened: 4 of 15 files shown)


WingedGuardian:
## Automated first-pass review (structural + PII/secrets/wording)

**Scope:** `src/genesis/memory/store.py`, `tests/test_memory/test_supersede_prefix_resolution.py`, `tests/test_memory/test_supersession.py`, `changelog.d/20260910190500-fixed-dedup-short-circuit-supersedes.md`.

### Structural review
- No `nohup`, no direct `task_states` inserts, no `/tmp/` usage, no `git add -A`/`--force`/`--hard`, no worktree `pip install -e`, no `os.killpg`.
- `except Exception:` blocks are typed (not bare `except:`) — pre-existing pattern, consistent with the surrounding file (verified against the full PR-head copy of `store.py`, no bare excepts present).
- No mutable default arguments introduced.
- No raw SQL against production `genesis.db`; the new SQL in the test files builds an in-memory test fixture DB (and imports the real `FTS5_DDL` rather than retyping it), which is the correct pattern, not a violation of the "use CRUD modules" rule. `_mark_superseded`'s new successor check goes through `memory_crud.get_metadata`, consistent with existing CRUD usage.
- Import hygiene: no circular imports, no star imports, ordering intact.
- Coverage: no new `.py` file requiring a companion test; the changed function (`store()`, `_mark_superseded`) is covered by 4 new targeted tests plus the existing suite (120 targeted tests reported green per PR body).
- Documentation: `SupersedeUnresolved` docstring and `_mark_superseded` docstring both updated to describe the new `reason` values and `verify_successor` param.
- Commit/PR title carries a conventional prefix (`fix(memory): ...`).

No structural ERROR or WARNING findings.

### PII / Secrets / Wording
- No API keys, tokens, credentials, private IPs (v4/v6), SSH keys, or connection strings in the added lines.
- No personal emails, username-bearing paths, revealing timezones, cloud account IDs, or hardware identifiers.
- PR title/body/changelog describe the defect in generic technical terms (dedup short-circuit, supersede guards) with no reference to a security incident, private infrastructure, or exploitable leak wording.

PII/Secrets/Wording: clean (no soft warnings)

### Note (informational, not a finding)
This PR targets `fix/supersede-resolve-short-handles` (#1928), not `main` — CI does not run against a non-default base, so the suites above were validated locally per the PR body rather than in CI on this PR.

VERDICT: PASS



---

WingedGuardian:
CI repair published at `3c1ec5f4ada4347ebba7b4c3fe29a2e998593254` by `codex-01a091e8-d482-78a3-bc7f-4daf10bd0bdd`. The dependency-audit step now reports advisories without failing its job, matching the reviewed reporting policy in #1960. Dependency installation failures still fail; audit commands and ignore arguments are unchanged.

The pull-request trigger now also covers feature-branch bases; the push trigger remains restricted to main.

Independent internal review and the canonical commit gate passed for this narrow CI delta. Existing review findings, review-budget holds, dependencies, and merge requirements remain in force; this is not a merge recommendation or a new external review request.

WingedGuardian:

Classified the new mutating memory_supersede tool in the benchmark policy. The previously failing forcing-function test passes (10 passed) and Ruff is clean.

@codex review

WingedGuardian:
## Converting to draft

Main landed a **later revision of the same supersede machinery** this PR edits, so both sides rewrote the same regions: `SupersedeUnresolved`'s constructor and docstring in `src/genesis/memory/store.py` (3 hunks), the `memory_supersede` docstring in `src/genesis/mcp/memory/core.py`, and both supersede test files.

Which revision of each hunk wins is author judgement. A union would produce a file that compiles and describes two different designs.

PR #1930, opened 2026-09-10
fix(memory): report the supersede outcome to the caller, not to the log

WingedGuardian:
**Split 3 of 5 from #1831.** Base: `fix/supersede-dedup-short-circuit` (#1929).

> **Stacked**, two deep: `main` → #1928 → #1929 → this. It needs
> `SupersedeUnresolved` from #1928 and — for its retry advice to be **true** —
> the dedup-path supersede from #1929. **CI does not run for a non-default
> base**; the targeted suites were run locally and are recorded below.

## What this adds

The previous two PRs made the supersede resolve its target and made the dedup
path perform it. Both wrote their failures **to the server log** and returned a
memory id regardless — so the caller still read success while the stale memory
stayed live in recall. This routes the verdict to the caller.

- **`store()` stops swallowing `SupersedeUnresolved`.** Two clauses, one per call
  site, because a single test cannot reach both.
- **`memory_store` returns a report dict** when `supersedes` was passed, and the
  bare id string otherwise — so the ~all-calls path is untouched. The report says
  whether the deprecation happened, which id **did** land, why it did not, and
  which ids collided for an ambiguous handle, including the resolver's LIMIT-3
  truncation flag so three candidates are never presented as the whole collision
  set.
- **It reports rather than raises.** The memory is durable by that point, and an
  exception reads as "the store failed" and invites the duplicating retry that
  was measured twice on 2026-09-06.
- **The advice fits the reason.** "Re-send with a full 36-char id" fixes a handle
  that named nothing; for the two successor reasons the id was fine and re-sending
  reproduces the collision, which is a loop. Handing back an instruction that
  cannot work is how the original defect kept its retry going.
- **An empty or all-whitespace `supersedes` is not a request.** Both store-side
  guards are truthiness tests, so `""` and `"   "` deprecate nothing — while an
  `is None` test here would have built a success report about an operation that
  never ran. Normalized once, before the store call, so the two layers cannot
  disagree. MCP and `POST /api/t/memory_store` both take a string here, so an
  empty one arrives without any client bug.

## Blast radius — enumerated, not spot-checked

- `grep -rn 'supersedes=' src/ --include=*.py` returns exactly **two** lines: the
  parameter's own definition and `mcp/memory/core.py`. So that module is the sole
  caller that can trigger the exception, and it catches it. **No internal path
  sees it.**
- The HTTP door: `tool_api.py::_normalize_result` (read at lines 79-100) returns
  `dict` and `list` unchanged and only wraps scalars, so both shapes serialize
  correctly through `POST /api/t/memory_store`.

## Verification

**Verify-RED — 4 mutations, all applied, all biting.**

| mutation | result |
|---|---|
| delete `except SupersedeUnresolved: raise` on the **normal** path | `test_store_lets_an_unresolved_supersede_out` red |
| delete it on the **dedup** path | 3 red |
| guard reverts to `if supersedes is None` | empty-`supersedes` test red |
| collapse the reason branches to one advice string | advice test red |

The first two fail **disjoint** test sets — that is the point of having two
clauses and two tests. The third bites for a specific reason worth stating:
`""` folds to `None` and would pass either guard, but `"   "` normalizes to `""`,
truthy under `is None` and falsy under `not` — so the test loops over **both**
values, and what it pins is the guard form, not the normalization.

- 127 targeted tests green; `ruff` clean.
- `test_memory_store_without_supersedes_still_returns_a_bare_id` uses an explicit
  `isinstance` check, so a dict cannot satisfy it by comparing equal.

## Known and deferred

`superseded: true` here means "nothing raised". A supersede that half-applies
(SQLite deprecation lands, Qdrant payload write swallowed) or fails outright with
a database error still reports `true`. That is real, and it is the entire subject
of **split 4** — named here rather than left for a reviewer to find.

E2E: `memory_store(supersedes=<bogus id>)` via MCP — confirm the response is a dict with `superseded: false`, the stored `memory_id`, and a reason; and that `memory_store(supersedes="")` still returns a bare id string. Both doors: MCP and `POST /api/t/memory_store`.

### DIFF
--- a/changelog.d/20260910191000-fixed-supersede-reports-its-outcome.md
+++ b/changelog.d/20260910191000-fixed-supersede-reports-its-outcome.md
@@ -1,0 +1,19 @@
+- **A supersede that could not be performed now says so to the caller.** Until
+  now `memory_store` returned a memory id whether or not the deprecation it was
+  asked for had happened: a failure was written to the server log and nothing
+  else, so a session was told its correction had landed while the stale memory
+  stayed live in recall.
+
+  When `supersedes` is passed, the call now returns a report -- whether the
+  deprecation happened, which id did land, why it did not, and which ids
+  collided for an ambiguous handle. Without `supersedes` the return is the
+  memory id string exactly as before, so the overwhelming majority of calls are
+  unchanged.
+
+  It reports rather than raises on purpose: the memory is durable by that
+  point, and an exception reads as "the store failed" and invites the retry
+  that duplicates the memory. The advice fits the reason -- "re-send with a
+  full 36-character id" is the fix for a handle that named nothing, and the
+  wrong instruction for a correction that collided with its own target, where
+  re-sending reproduces the collision. An empty or all-whitespace `supersedes`
+  is treated as no request at all rather than as a supersede that succeeded.
--- a/src/genesis/mcp/memory/core.py
+++ b/src/genesis/mcp/memory/core.py
@@ -824,8 +824,20 @@ async def memory_store(
     room: str | None = None,
     collection: str | None = None,
     supersedes: str | None = None,
-) -> str:
+) -> str | dict:
     """Store memory with source metadata and type tag. Returns memory_id.
+
+    Return shape depends on ``supersedes``, which the caller controls: without
+    it you get the memory_id string, exactly as before. WITH it you get a dict
+    ``{"memory_id", "superseded", ...}`` reporting whether the deprecation
+    actually happened — because it can fail on its own while the store
+    succeeds, and reporting only the id is what let that failure go unnoticed.
+    ``reason`` names why a supersede did not happen, and the ``warning`` gives
+    advice that fits that reason.
+
+    This tool is also reachable over HTTP as ``POST /api/t/memory_store``
+    (``dashboard/routes/tool_api.py``), which accepts ``supersedes`` like any
+    other parameter — so the conditional shape is visible to that door too.
 
     Args:
         memory_class: Optional classification — "rule", "fact", or "reference".
@@ -836,7 +848,13 @@ async def memory_store(
             collection routing when provided (e.g. "knowledge_base").
         supersedes: Memory ID that this new memory replaces. The old memory
             will be marked as deprecated with a ``succeeded_by`` link to the
-            new one. Use when correcting stale facts.
+            new one. Use when correcting stale facts. Accepts the same short
+            ``id:<8-char>`` handles the proactive hook prints and
+            ``memory_expand`` accepts; an ambiguous handle is never guessed.
+            When the supersede cannot be performed the memory is STILL STORED
+            and the returned dict says so — do not retry the store blindly, or
+            you will duplicate the memory (measured: that retry happened twice
+            on 2026-09-06 while this path was silent).
     """
     memory_mod = _memory_mod()
     memory_mod._require_init()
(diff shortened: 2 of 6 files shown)


WingedGuardian:
## Structural review

Files: `changelog.d/…md` (new), `src/genesis/mcp/memory/core.py`, `src/genesis/memory/store.py`, 3 test files.

- No banned patterns found: no `nohup`, no direct `task_states` writes, no `/tmp/` usage, no `git add .`/`-A`, no force-push/reset, no worktree-path `pip install -e`, no bare `except:`, no mutable default args, no `os.killpg`, no raw SQL against `genesis.db` (mutations go through `memory_crud`/`memory_links_crud` as expected).
- Exception handling: `except SupersedeUnresolved` is added ahead of the existing `except Exception:` at both call sites in `store.py`, narrowing rather than widening the broad catch — correct order (specific before general).
- Import hygiene: no circular imports, no star imports. `from genesis.memory.store import SupersedeUnresolved` is a deferred (in-function) import in `core.py`, consistent with the existing deferred import of `session_origin_from_env` a few lines above.
- Coverage: this is a behavior change to `memory_store`'s return type (`str` → `str | dict`) and to `store()`'s exception behavior on two call sites — both are covered by new/updated tests in `tests/test_mcp/test_memory_mcp.py`, `tests/test_memory/test_supersede_prefix_resolution.py`, and `tests/test_memory/test_supersession.py` (including a contract-lock test for the unchanged bare-id path and a test added specifically to reach the previously-untested `except Exception` swallow site).
- Docs: `memory_store`'s docstring is updated to describe the new conditional return shape, the HTTP door (`POST /api/t/memory_store`), and short-handle acceptance. No new public symbols added without docstrings.
- Commit/PR prefix: title uses `fix(memory): …` — conventional prefix present.

No ERROR or WARNING findings.

## PII/Secrets/Wording scan

- Added lines (`core.py`, `store.py`, tests, changelog): no API keys/tokens/secrets, no private IPv4/IPv6 ranges, no SSH keys, no connection strings, no personal emails, no usernamed absolute paths, no location-revealing timezones, no account/project IDs, no hardware identifiers, no private hostnames.
- PR title/body/changelog: technical description of a return-shape and exception-handling change (measured counts, mutation-testing table, blast-radius grep). No mention of a PII/secrets leak, no security-incident framing, no named private infrastructure, no exploitable vulnerability detail.

PII/Secrets/Wording: CLEAN

VERDICT: PASS



---

WingedGuardian:
## Superseded by #1937. Nothing is dropped — most of it stopped being needed.

This PR built a report dict with four reason-specific advice branches, because
`supersedes` could fail four ways after the content was already stored, and each
needed a different thing said about it.

The redesign in #1933 removed the premise. `memory_supersede(old_id, new_id)`
makes the supersede a standalone operation, and with it #1928 and #1929 moved
resolution and pair validation **before any write**. So all four of those
failures now happen with nothing stored — they raise, and there is no partial
outcome to narrate. The only case left needing a report is: *the memory landed
and the deprecation then failed on infrastructure.* One case.

The advice collapsed for a second reason worth stating. Every branch here ended
in some form of "re-run `memory_store` with the same content", because there was
no other way to complete a supersede. That instruction was the retry loop that
**measurably duplicated memories twice on 2026-09-06**. #1937 says
`memory_supersede(old, new)` instead — an instruction the caller can actually
carry out.

Also carried into #1937 from this PR, not lost:
- the empty / all-whitespace `supersedes` guard, and its test over both `""` and
  `"   "` (the case that discriminates `not x` from `x is None`)
- the bare-id contract lock for the ~all-calls path, with the explicit
  `isinstance` check
- the review finding that the warning must not claim the old memory "is still
  live in recall" for a `not_found` target — resolution is exactly what failed to
  establish that. Fixed at the source in #1928's exception message.

+342 here → the whole of #1937 is +326, including #1931's subject as well.

Issue #1641, opened 2026-09-03
graph-db adoption: run the two engine probes and make the engine decision


WingedGuardian:
## Decision

Adopt a real graph database for the memory graph, replacing the in-process NetworkX graph rebuilt from SQLite. This supersedes the earlier comparative bake-off, which is closed by decision rather than by measurement.

SQLite remains the system of record. The engine is a **derived, rebuildable index** — one writer, many readers, with the SQLite-backed store staying as an always-correct fallback.

## Status: blocked on an engine decision, not on effort

Nothing is scheduled, and the reason is that one decision was never made. Two probes were specified to make it and never run. Both are read-only and touch no repo code:

**Probe 1 — hop-wise as-of expressibility.** Can "the neighbourhood as it stood at time D" be expressed in the engine's query language with the validity predicate applied *inside* the recursion? Filtering only at the endpoints is wrong — it admits paths through intermediates that were invalid at D. Both existing reference implementations filter per-hop. If the dialect cannot express it, the adapter must iterate hop-by-hop, which forfeits much of the reason to adopt an engine at all. That materially changes the choice.

**Probe 2 — concurrency under a single long-lived instance.** An embedded per-process engine was already measured to produce **race-dependent silent lost updates** — one run lost a write outright; a staggered re-run kept both. Embedded is therefore rejected. The open question is whether a single long-lived server instance behaves correctly under overlapping writers and readers, since consumers span multiple processes. Verify rather than assume: the failure mode here is silent.

Also unmeasured and decision-relevant: read-handle reopen latency at full scale, and peak RSS for the full projection (which sets a real memory cap rather than a guess).

## Design constraints already settled

- The adapter seam must cover **every** consumer, including the expansion path that reaches the graph directly and bypasses the main traversal module — a seam that only wraps the obvious entry point leaves that one behind.
- The data model must represent **typed parallel edges**. The link table's key is (source, target, type), so one pair legitimately carries several typed edges; a model that cannot hold them reintroduces a loss that was recently fixed in the loader.
- As-of derivation differs by layer: one layer has no edge temporality and must join endpoint validity, the other is natively bitemporal. Both need covering.
- **Fail-closed**: a graph read that cannot reach its store must raise or return a typed "unknown" — never an empty result that reads as "no neighbours".
- The existing parity oracle is reused as the equivalence gate for the migration, rather than as a contest between engines.

## Groundwork already merged

[...]

PR #1931, opened 2026-09-10
fix(memory): a partial or failed supersede must not report as complete

WingedGuardian:
**Split 4 of 5 from #1831.** Base: `fix/supersede-report-to-caller` (#1930).

> **Stacked**, three deep: `main` → #1928 → #1929 → #1930 → this. It edits the
> report #1930 introduces and both `except Exception` handlers #1929 and #1928
> established. **CI does not run for a non-default base**; targeted suites were
> run locally and are recorded below.

## What this fixes

#1930 made `memory_store` report whether the supersede happened. It derived that
answer from *"nothing raised"*, which is not the same question — two failures
were still reported as success.

**1. A supersede that HALF-applied.** The two steps after the SQLite deprecation
are best-effort and swallow their own exceptions. The vector leg of hybrid recall
filters on the Qdrant `deprecated` payload while the keyword leg filters the
SQLite column, so a swallowed payload write hides the memory from one leg and
leaves it live in the other. `_mark_superseded` now returns the steps that
**failed**, and the report names them in `partial`.

**2. A supersede that failed OUTRIGHT.** Both `except Exception` handlers caught
a database fault in `resolve_id` / `mark_superseded` / `get_metadata`, logged it,
and left the out-param empty; the unconditional `superseded: true` then claimed a
deprecation that provably had not run. A distinct marker makes that case report
`superseded: false`, because the caller's next move depends on whether the old
memory is still live.

## The point of no return

`crud.mark_superseded` commits before it returns, so past that line SQLite recall
already excludes the old memory — a failure after it can never honestly be "did
not land at all". Everything after the commit moved into `_supersede_follow_up`
and is bounded: what throws there is the `post_deprecation` **partial** and can
never escape as a total failure. Without the boundary the caller is told the old
memory is probably still live when recall already hides it, which is the same
class of lie inverted.

`_supersede_follow_up` is a pure extraction — the Qdrant block, the link create,
and the graph-cache invalidation `else:` clause (#1641 work) move together,
unchanged apart from two `degraded.append` lines.

## Found while reviewing this split, not inherited from #1831

The `qdrant_payload` and `succeeded_by_link` markers had **no test that reached
the real appends**. The MCP-level test injects the marker through the mocked
out-param, which proves the report renders it and nothing about whether
`_mark_superseded` ever produces it. **Deleting either append left the suite
green.**

Closed with two tests through real SQLite:

- `test_a_swallowed_qdrant_payload_write_is_named` — flips the target to
  `embedding_status='embedded'` first, because the fixture is `fts5_only` and that
  path **skips Qdrant entirely**, so without it the test would pass vacuously. It
  also asserts the link step still ran: one failed follow-up must not skip the
  rest.
- `test_a_failed_succeeded_by_link_is_named` — asserts **both** branches: an
  arbitrary error is a degradation, a `UNIQUE constraint` error is not (the edge
  already exists). Without the second half the marker could be appended
  unconditionally and stay green.

## Verification

**Verify-RED — 7 mutations, all applied, all biting.**

| mutation | red test |
|---|---|
| drop the `qdrant_payload` append | `test_a_swallowed_qdrant_payload_write_is_named` |
| drop the `succeeded_by_link` append | `test_a_failed_succeeded_by_link_is_named` |
| normal path stops recording an outright failure | `test_normal_path_records_an_unexpected_supersede_failure` |
| dedup path stops recording an outright failure | `test_an_unexpected_supersede_error_is_reported_not_swallowed` |
| remove the point-of-no-return boundary | `test_a_failure_after_the_deprecation_commits_is_a_partial` |
| an outright failure keeps `superseded: true` | `test_memory_store_does_not_report_success_for_a_failed_supersede` |
| drop the `partial` list | `test_memory_store_reports_a_partially_applied_supersede` |

Rows 3 and 4 fail **different** tests, verified individually — there are two
handlers and one test cannot reach both.

- 134 targeted tests green; `ruff` clean.
- A post-deprecation partial deliberately keeps `superseded: true` (the
  deprecation DID land); only the outright-failure marker flips it. A test
  asserts that `true` explicitly, so a later "simplification" flipping every
  partial to `false` goes red.
- An out-param rather than a wider return type, because `store()` returns the
  memory id at ~every call site in the tree; `supersede_degraded` is `None`
  everywhere except the one MCP call site, and every append is guarded.

## Not fixed here

Locating the Qdrant point across collections before claiming the payload update
applied. The guard it concerns (`embedding_status == 'embedded'`) is untouched by
this stack, and the remedy is new per-supersede Qdrant I/O.

E2E: with Qdrant stopped, `memory_store(supersedes=<id of an embedded memory>)` via MCP — confirm the response carries `partial: ["qdrant_payload"]` with `superseded: true` (the SQLite deprecation did land), rather than a bare success.

### DIFF
--- a/changelog.d/20260910191500-fixed-partial-supersede-not-reported-complete.md
+++ b/changelog.d/20260910191500-fixed-partial-supersede-not-reported-complete.md
@@ -1,0 +1,17 @@
+- **A supersede that only half-applied was reported as complete.** The two
+  steps after the SQLite deprecation -- the vector payload write and the
+  `succeeded_by` link -- are best-effort and swallow their own errors, so a
+  supersede whose payload write failed left the memory hidden from keyword
+  recall and still live in vector recall while the call reported success. The
+  failed steps are now named in a `partial` list on the report.
+
+  A supersede that failed OUTRIGHT with an unexpected database error was also
+  reported as successful: the error was logged and nothing else, and the
+  success report was unconditional. That case is now distinguished from a
+  partial and reported as not superseded, because the caller's next move
+  depends on whether the old memory is still live.
+
+  The two are kept apart by a boundary at the point of no return. Once the
+  deprecation has committed, recall already excludes the old memory, so a
+  failure after that point is a partial no matter what threw -- it can never be
+  reported as "did not happen at all".
--- a/src/genesis/mcp/memory/core.py
+++ b/src/genesis/mcp/memory/core.py
@@ -833,7 +833,10 @@ async def memory_store(
     actually happened — because it can fail on its own while the store
     succeeds, and reporting only the id is what let that failure go unnoticed.
     ``reason`` names why a supersede did not happen, and the ``warning`` gives
-    advice that fits that reason.
+    advice that fits that reason. A partially-applied supersede adds ``partial``
+    naming the steps that did not land; ``partial`` containing
+    ``supersede_failed`` means it did not land AT ALL (an unexpected database
+    error), and ``superseded`` is False there.
 
     This tool is also reachable over HTTP as ``POST /api/t/memory_store``
     (``dashboard/routes/tool_api.py``), which accepts ``supersedes`` like any
@@ -865,7 +868,14 @@ async def memory_store(
     # external-influenced session's writes stop landing first_party via the
     # "conversation" pipeline. None (foreground/unset) → pipeline-derived.
     from genesis.memory.provenance import session_origin_from_env
-    from genesis.memory.store import SupersedeUnresolved
+    from genesis.memory.store import SUPERSEDE_FAILED, SupersedeUnresolved
+
+    # Collector for steps that failed AFTER the SQLite deprecation landed. The
+    # Qdrant payload write and the link create are best-effort and swallow
+    # their own exceptions, and the vector leg of hybrid recall filters on that
+    # Qdrant payload — so without this a partial supersede would be reported as
+    # a complete one, which is the class of lie this whole change exists to stop.
+    degraded: list[str] = []
 
     # An empty or all-whitespace handle is not a request. Both store-side
     # guards are truthiness tests (`if supersedes:`), so "" and "   " perform no
@@ -890,6 +900,7 @@ async def memory_store(
             room=room,
             collection=collection,
             supersedes=supersedes,
+            supersede_degraded=degraded,
         )
     except SupersedeUnresolved as exc:
         # The memory IS durable; only the deprecation failed. Report both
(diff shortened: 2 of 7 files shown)


WingedGuardian:

Reviewed: `changelog.d/20260910191500-fixed-partial-supersede-not-reported-complete.md`, `src/genesis/mcp/memory/core.py`, `src/genesis/memory/store.py`, `tests/test_mcp/test_memory_mcp.py`, `tests/test_memory/test_mcp_integration.py`, `tests/test_memory/test_supersede_prefix_resolution.py`, `tests/test_memory/test_supersession.py`

**Domain 1 — Structural**
- No `nohup`/direct `task_states` insert/`/tmp/`/`git add .`/`git push --force`/`git reset --hard`/worktree-path `pip install -e`/`os.killpg`/raw-SQL-in-src violations.
- All `except Exception:` blocks are typed (no bare `except:`). The new outer-handler appends (`SUPERSEDE_FAILED`) and the bounded `_supersede_follow_up` try/except match the PR's described point-of-no-return design — a post-commit failure is converted to a named `partial` and never escapes as a raised exception or a swallowed no-op.
- No mutable default arguments (`supersede_degraded: list[str] | None = None` is correct).
- Imports: the local/deferred import in `core.py` (`from genesis.memory.store import SUPERSEDE_FAILED, SupersedeUnresolved`) matches the existing in-function import style already used there (`session_origin_from_env`); no circular or star imports introduced.
- Test coverage: every call site of `store()`/`_mark_superseded()` whose signature changed was updated (`test_mcp_integration.py`, `test_supersession.py`), plus new tests exercising the real code path (not just mocks) in `test_supersede_prefix_resolution.py` and MCP-level report-shape tests in `test_memory_mcp.py`. No new `src/` file lacking a corresponding test.
- Docstrings: `memory_store`, `_mark_superseded`, and the new `_supersede_follow_up` all carry updated/added docstrings describing the new `partial`/`supersede_degraded` behavior and the `SUPERSEDE_FAILED` marker's meaning.
- Commit prefix: PR title carries `fix(memory):` — conventional.

- No API keys, tokens, passwords, private IPs (10.x/192.168.x/100.64–127.x), IPv6 ULA, SSH keys, DB connection strings, or secret env values anywhere in the added lines.
- No personal emails, username-embedded absolute paths, location-revealing timezone identifiers, cloud account/project IDs, hardware identifiers, or private hostnames. (The commit author address is a `noreply@` GitHub address — ignored per the known-false-positive list.)
- PR title, body, and changelog entry are a purely technical description of a SQLite/Qdrant supersede-reporting bug fix; no mention of a PII/secrets leak, security incident, or exploitable infrastructure detail.

VERDICT: PASS



---

WingedGuardian:
## Superseded by #1937. The taxonomy this built is the thing that went away.

This PR added four markers to describe how far a supersede got —
`qdrant_payload`, `succeeded_by_link`, `supersede_failed`, `post_deprecation` —
plus a point-of-no-return boundary whose whole job was keeping the last two
apart, because "did not land at all" and "landed then degraded" needed different
sentences.

**Every one of those markers has the same remedy** now that
`memory_supersede(old_id, new_id)` exists (#1933): call it. It redoes the
deprecation, the Qdrant payload, and the link. A distinction that cannot change
what the caller does next is not information — and four ways to say "the
deprecation is incomplete" is four things that can be wrong.

That is not hypothetical. This PR's own review round found one already: the
`qdrant_payload` marker was silently lost when a later follow-up step raised,
so the report omitted the very degradation it was built to surface. Codex found
the same class independently. Both findings are moot rather than fixed, because
the code they concern no longer exists.

#1937 keeps the one property that mattered: a supersede that did not land is
never reported as one that did. It gets there with a single boolean written on
**both** branches, defaulted to `False` at the reader, and locked by mutation.

Carried into #1937 from this PR, not lost:
- the twin-call-site coverage — a test parametrized over the normal and dedup
  paths, because the dedup one returns ~300 lines earlier and one test cannot
  reach both
- the `test_memory_store_delegates` out-param expectation, fixed at the
  expectation rather than by loosening the assertion
- the "absent means fine" objection, which is why the outcome is written on the
  success branch too

One thing this PR did that outlived it: while reviewing it I found that the
`qdrant_payload` and `succeeded_by_link` markers had **no test reaching the real
appends** — the MCP-level test injected them through the mocked out-param, so
deleting either append left the suite green. That gap is gone with the markers,
but it is the reason the mutation matrices in this split now check that a
mutation actually *bites* rather than that a test merely exists.

+374 here → #1937 is +326 total, covering #1930's subject as well.

PR #1937, opened 2026-09-10
fix(memory): memory_store reports the one thing it could not do

WingedGuardian:
**Replaces #1930 and #1931.** Base: `fix/supersede-dedup-short-circuit` (#1929).

> **Stacked**, three deep: `main` → #1928 → #1933 → #1929 → this. CI does not run
> for a non-default base; targeted suites and the mutation matrix are below.

## Why this replaces two PRs rather than amending them

Between them, #1930 (+342) and #1931 (+374) carried **716 insertions**: a report
dict with four reason-specific advice branches, and a partial/total marker
taxonomy — `qdrant_payload`, `succeeded_by_link`, `supersede_failed`,
`post_deprecation` — plus a point-of-no-return boundary to keep the last two
apart.

Both existed to answer a question the redesign deleted. With resolution and pair
validation now running **before any write** (#1928, #1929), every rejection
leaves nothing behind and simply raises. The only outcome a caller cannot infer
is: *the memory was stored and the deprecation then failed on infrastructure.*
**That is one case.**

The marker taxonomy went with it, and not for tidiness — **every marker had the
same remedy** once `memory_supersede` existed: call it. A distinction that cannot
change what the caller does next is not information, and four ways to say "the
deprecation is incomplete" is four things that can be wrong. #1931's own review
had already found one (a marker silently lost when a later follow-up step
raised); Codex found the same class independently.

**716 insertions → 326**, and the advice went from four branches that could not
be carried out to one call that can.

## The change

- `store()` takes a `supersede_outcome` out-param; one helper (`_apply_supersede`)
  serves both call sites.
- `memory_store` returns a report dict when `supersedes` was passed, the bare id
  string otherwise, so the ~all-calls path is untouched.
- The remedy is now `memory_supersede(old, new)` — a call the caller can
  actually make, which is only true because that tool exists.
- An empty or all-whitespace `supersedes` is not a request. The store's guard is
  a truthiness test, so an `is None` test here would have disagreed with it.

### Two defaults chosen against, deliberately

1. **The outcome is written on the success branch too**, not only on failure.
   "Absent means fine" is the default that goes wrong quietly the first time a
   new path forgets to set it — and the failure it produces is precisely the
   original defect: an unperformed deprecation reported as done.
2. **The MCP layer defaults the flag to `False`**, so a store that never wrote it
   reports "not superseded". Fail toward *we cannot confirm it*.

Both are mutation-locked, so neither is decoration.

## Verification

**Verify-RED — 4 mutations, all applied, all biting, on distinct tests:**

| mutation | red |
|---|---|
| log the failure without recording it | both parametrized `..._is_recorded_not_just_logged` cases **and** the success-recording test (3 red) |
| `.get("superseded", True)` | `...does_not_claim_a_supersede_it_cannot_confirm` |
| guard reverts to `is None` | `...does_not_report_a_supersede_nobody_asked_for` |
| describe the remedy instead of naming the call | `...reports_a_deprecation_that_did_not_land` |

The first mutation's two parametrized cases are **separate code paths** — the
dedup site returns ~300 lines earlier — so one test could not have covered both.

Every mutation was `ast.parse`-checked before its run, after a SyntaxError
mutation earlier in this split produced a collection error that would have
counted as a false RED. 138 targeted tests green; `ruff` clean.

## What is deliberately given up

The caller is no longer told **which** follow-up step failed (Qdrant payload vs
`succeeded_by` link). That detail is in the server log, and it could not change
the caller's next action — `memory_supersede(old, new)` redoes all of it. If a
consumer ever needs the breakdown programmatically it belongs on the log/event
surface, not in a tool result an LLM has to parse.

E2E: `memory_store(supersedes=<valid id>)` over MCP with the deprecation forced to fail — confirm the response carries `superseded: false` plus a `memory_supersede(...)` call in the warning, and that running that call completes the deprecation without creating a second memory.

---

<a href="https://app.devin.ai/review/wingedguardian/genesis-agi/pull/1937" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,8 @@
 # fastmcp <3 for API stability, transitive secure floors) that automatic
 # version-update PRs would fight. GitHub's Dependabot *security* alerts are
 # already enabled for the repo and cover the pip CVE side; the pip-audit CI job
-# gates the resolved tree per-PR. So dependabot only keeps the Actions current.
+# reports on the resolved tree per-PR without blocking on audit findings.
+# Dependabot version updates here only keep the Actions current.
 version: 2
 updates:
   - package-ecosystem: "github-actions"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
   workflow_dispatch:
   schedule:
     - cron: "0 6 * * 1"  # weekly Mon 06:00 UTC — surface new CVEs even without repo activity
@@ -34,10 +33,9 @@ jobs:
         run: ruff check src/ tests/ scripts/
 
   dependency-audit:
-    # Fail on any KNOWN-vulnerable dependency in the resolved runtime tree,
-    # except the triaged IDs below. A red run is a SIGNAL to bump the dep or
-    # add a triaged ignore — not a hard freeze (the ruleset has no required
-    # checks). Complements GitHub's already-enabled Dependabot security alerts.
+    # Report known vulnerabilities outside the retained ignore baseline.
+    # Audit findings remain visible in the log without blocking this job;
+    # dependency installation failures still fail it.
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
@@ -50,9 +48,10 @@ jobs:
           pip install -e .
           pip install pip-audit
       - name: Audit dependencies for known CVEs
+        continue-on-error: true
         run: |
           # Triaged --ignore-vuln IDs (rationale of record: pyproject.toml pins).
-          # A NEW / untriaged advisory is NOT in this list and so fails the job.
+          # Advisories outside this list remain visible in the log.
           #   litellm 1.78.7 — every ID is a proxy-server / JWT / OIDC / UI /
           #     MCP-gateway vuln; Genesis uses acompletion/completion_cost only
           #     (no proxy server), so none are reachable. Pinned <1.83 also dodges
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -287,8 +287,10 @@ source — there is no `requirements.txt`). Known-CVE scanning runs automaticall
 
 - **CI** — the `dependency-audit` job in `.github/workflows/ci.yml` runs
   `pip-audit` against the resolved runtime tree on every PR/push and weekly,
-  failing on any new (untriaged) advisory. Already-triaged, not-reachable
-  advisories are listed with rationale in that job.
+  reporting advisories outside its retained ignore baseline in the job log.
+  Audit findings are advisory: step-level `continue-on-error` prevents them
+  from failing the job. Dependency installation failures still fail the job.
+  Existing ignored advisories are listed with rationale in that job.
 - **Dependabot** — GitHub's dependency-graph security alerts are enabled for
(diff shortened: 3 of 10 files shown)


WingedGuardian:

**Files reviewed:** `src/genesis/mcp/memory/core.py`, `src/genesis/memory/store.py`, `tests/test_mcp/test_memory_mcp.py`, `tests/test_memory/test_mcp_integration.py`, `tests/test_memory/test_supersession.py`, `changelog.d/20260910191000-fixed-supersede-reports-what-it-could-not-do.md`

- **Pattern violations:** none. No `nohup`, no direct `task_states` writes, no `/tmp/` usage, no force-push/reset patterns, no bare `except:` (the two `except Exception:` blocks in `_apply_supersede`/callers specify a type and are deliberate — the docstring explains why the out-param must swallow rather than raise), no mutable default args (`supersede_outcome: dict | None = None` defaults to `None`, populated per-call), no raw SQL.
- **Import hygiene:** clean — no star imports, no circular-import shape, the one new import (`session_origin_from_env`) follows the existing local-import pattern already used in this function.
- **Coverage:** no new `.py` files. The changed function signatures (`memory_store`, `MemoryStore.store`, new `_apply_supersede` helper) all have corresponding new/updated tests across three test files, including both call sites (normal path and dedup short-circuit path) for the infrastructure-failure case.
- **Documentation:** `memory_store`'s docstring was updated to describe the new `str | dict` return shape and the `supersedes` semantics; the new `_apply_supersede` helper has a docstring explaining the non-raise/non-silent constraint. No gaps found.
- **Genesis conventions:** N/A (not a hook script, no `$CLAUDE_PROJECT_DIR`, no `IntervalTrigger`). PR title carries a conventional `fix(memory):` prefix.

- **Code diff:** no API keys/tokens/secrets, no private IPv4/IPv6 ranges, no SSH keys, no connection strings. Test fixture IDs (`old-memory-id`, `abcd1234-0000-4000-8000-000000000001`, etc.) are synthetic UUID-shaped memory-id placeholders, not IPs or credentials.
- **Soft-warning categories** (personal emails, home-dir paths with usernames, location-revealing timezones, account/project IDs, hardware IDs, private hostnames): none present.
- **PR title/body + changelog:** no mention of a PII/secrets leak, no security-incident language, no naming of private infrastructure, no exploitable vulnerability detail. The body's "MEASURED 2026-09-06" note describes an internal data-correctness bug (stale `succeeded_by` edges from a silent failure), not a security/PII exposure.

VERDICT: PASS



---

WingedGuardian:
CI repair published at `94400f3ad60937552e426932f9c50fba4542e60a` by `codex-01a091e8-d482-78a3-bc7f-4daf10bd0bdd`. The dependency-audit step now reports advisories without failing its job, matching the reviewed reporting policy in #1960. Dependency installation failures still fail; audit commands and ignore arguments are unchanged.

The pull-request trigger now also covers feature-branch bases; the push trigger remains restricted to main.

Independent internal review and the canonical commit gate passed for this narrow CI delta. Existing review findings, review-budget holds, dependencies, and merge requirements remain in force; this is not a merge recommendation or a new external review request.

WingedGuardian:

Classified the new mutating memory_supersede tool in the benchmark policy. The shared forcing-function regression passes on the validated sibling branch; CI will validate this published head.

@codex review

PR #1942, opened 2026-09-10
feat(worktrees): the reaper archives instead of deleting, and the work is visible

WingedGuardian:
## What this changes

`scripts/worktree_lifecycle.py` **deletes nothing**. Both lanes — merged at 7
idle days, unmerged at 14 — archive into the trash as a gzip tarball plus an
append-only tombstone row, recoverable indefinitely. `_purge_old_trash` is
removed.

The reason is a measurement, not a preference: a worktree can hold the only
surviving trace of the session that produced it. The session that authored an
open PR on this repo has **no `cc_sessions` row** (4,774 rows, prefix-matched)
and **no transcript in any of 532 CC project directories** — its commits are the
entire record it ever ran. Whether a piece of context matters later is not a
judgement a daily timer is positioned to make, and the storage does not justify
guessing: 191 worktrees are ~11 GB raw, ~2.9 GB archived, against 265 GB free.

gzip rather than xz on measurement: a 58 MB worktree gave 10.6 MB in 46.6s under
xz and 14.9 MB in 6.6s under gzip, so xz buys ~0.9 GB across the whole backlog
for ~2 extra hours of CPU.

**One producer, three readers.** Every fate is decided in `_classify` /
`classify_all` and published to a cache that the session-start block and the
dashboard both read, so no surface can describe a worktree one way while the
reaper treats it another. Reading the cache is what makes it affordable —
classification costs 19–41s over 191 worktrees; the session-start block renders
in 2ms.

**Essential knowledge reports a triage, not a roster.** Newly at-risk branches by
name, the rest as a count. The old block capped at 8 and overflowed to
`(+183 more)`, which is unreadable and therefore unread. At-risk is bounded by
construction: it drains into the archive at the second threshold.

**New: `scripts/session_provenance.py`** — answers "which session produced this?"
in both directions by parsing commit BODIES. Git's trailer API is blind here:
the squash strategy appends a `Co-authored-by` block that becomes the final
paragraph, and git parses trailers only from that paragraph. MEASURED:
`%(trailers:)` matched **0 of the last 400** commits while the literal text
appears **720 times** in those same bodies; a body scan attributes **196/200
(98%)**. It is the appended block, not the concatenation, that does the damage —
even a single-commit squash loses them this way.

**Dashboard**: a worktree board modal on the overview tab, with an overview card
showing at-risk / total. The refresh endpoint invokes the reaper in report-only
mode and cannot reap.

## Review

Adversarial review found three data-destruction defects. Each was reproduced by
execution before being fixed:

- **Archive verification read one member header**, so a truncated archive passed
  and the source directory was then removed. A 61-member archive truncated to
  50% passed the shipped check while a full walk raised `EOFError`. Now walks
  every member and compares the count against the source; `EOFError` is not an
  `OSError` and is caught explicitly.
- **The name-collision guard probed the pre-compression path**, so it was blind
  to every already-archived entry — the next reap truncated an existing tarball
  through `tarfile.open(..., "w:gz")`.
- **`extractall(filter="data")` aborts partway on an absolute symlink**, and
  worktrees here symlink `secrets.env` to the repo root by standing convention.
  Three of the 48 worktrees due for archiving carry that link. Recovery now
  falls back to `filter="tar"` and relies on the per-destination containment
  invariants that were always the real defense.

Also fixed: a `UnicodeDecodeError` that aborted the whole run mid-list, a TOCTOU
widened from sub-second to minutes by classify-all-then-act-all, argv dispatch on
truthiness that silently ran a *different* command and exited 0, a fail-open that
reported a merged PR as open on a git timeout, `--dry-run` writing state, a
shared temp filename, and absolute home paths on an unauthenticated GET.

Each fix is pinned by **mutation testing** — reversing it turns the matching test
red (6/6, hash-verified restore). The pre-fix suite was green through all three
blockers, which is the point: they lived in the gaps between existing cases.

## Verification

- **76 tests** across 4 files; ruff clean; `node --check` clean; templates parse.
- **Acceptance bar, real data, twice**: archived and recovered two real
  worktrees — 4,100 and 3,847 entries restored with identical content hashes,
  including the named absolute-symlink case that would previously have produced
  an unopenable archive. Both were put back.
- Test pollution found and fixed at the class level: runs were writing into the
  real state directory (tombstone index, then board cache), both silently. An
  autouse fixture redirects every writable path by name, and a guard test
  enumerates the module's `Path` constants so a third one cannot be added
  unredirected.

## Not covered

The dashboard half (JS, templates) has 16 route tests and syntax checks but no
adversarial review — the branch grew it after the review was dispatched. Stated
rather than left to look like coverage.

E2E: archived and recovered two real worktrees including the absolute-symlink
case, verified identical content hashes, and confirmed the session-start block
and dashboard route render the same board from one cache. Post-merge: confirm
the daily timer publishes the cache and the modal renders in a browser.

Ledger: 1601788c3e5349bd83ce3f1134018617
Ledger: 00672afba8424a0cb247f2787714aec9
Ledger: 1d63177eb276469296cbf81cf2032374
Ledger: 071e90e0befa4bc59f2d637de759f65f
Ledger: ae4c84c6dfb747d0a73cf7bf56bd37e7

## Summary by CodeRabbit

- **New Features**
  - Added a dashboard Worktree Board with status, risk, merge, activity, and archive details.
  - Added board refresh, filtering, recovery guidance, and offline refresh support.
  - Added session provenance tools for tracing commits, branches, pull requests, and sessions.
  - Added durable worktree archiving with merged and unmerged handling, reporting, and recovery.
  - Added detection for credential-shaped and AWS access-key content during privacy reviews.
  - Added a dashboard Zero-Drop view with automatic updates.

- **Bug Fixes**
  - Re-pushing a published branch without an open pull request is now blocked with guidance.
  - Privacy reviews now distinguish blocking secrets from advisory findings without exposing values.


WingedGuardian:
@codex review

WingedGuardian:
## Structural Review

No ERROR-level pattern violations found (checked: nohup for service starts, direct `task_states` inserts, `/tmp/` usage, `git add -A`/`.`, force-push/`reset --hard`, `pip install -e` on a worktree path, bare `except:`, mutable default args, `os.killpg`, raw SQL against genesis.db). Note: `scripts/session_provenance.py` uses `sqlite3.connect` directly rather than `genesis.db.crud` — verified this is consistent with the established stdlib-only convention shared by 19+ other scripts under `scripts/` (the module is explicitly documented as needing to run without the genesis package), so not flagged as a violation.

Import hygiene is clean (correct stdlib/third-party/local grouping, no star imports, no circular imports) across `session_provenance.py`, `worktree_lifecycle.py`, and `dashboard/routes/worktrees.py`.

**WARNING**
- `scripts/session_provenance.py` — CLI handlers `cmd_commit` (~line 313), `cmd_session` (~line 327), and `cmd_coverage` (~line 361) have no direct test coverage in `tests/test_scripts/test_session_provenance.py`. The `scan`/`enrich`/`cmd_pr`/`cmd_branch` paths are well covered; the `--commit`/`--session`/`--coverage` CLI entry points are not.
- Dashboard state-vocabulary gap: `scripts/worktree_lifecycle.py:1240` sets `state = "archived"` on freshly-reaped rows, but `src/genesis/dashboard/routes/worktrees.py` (`_STATE_LABELS`), `partials/modals/worktrees.html:40` (chip filters), and `webui/js/dashboard.js:4834` (`rank` sort map) all enumerate only the six pre-reap states and omit `"archived"`. Non-crashing (JS falls back to defaults) but a freshly-archived row won't get a summary chip/color until the next cache regen. Consistent with the PR's own "Not covered" note on the dashboard half.

**NOTE**
- `scripts/session_provenance.py` — `installs_in` (119), `_pr_head_ref` (267), `_print_sessions` (393), and `main` (402) lack docstrings; most other functions in the file have them.
- `src/genesis/memory/open_loops.py:3-4` — module docstring still says the block covers "live git worktrees" broadly; the rewritten `_worktree_lines` now intentionally shows only the bounded at-risk triage (fresh/merged/protected excluded). Docstring wasn't updated to match the new behavior.

Test coverage for new modules is otherwise present (`tests/test_scripts/test_session_provenance.py`, `tests/test_dashboard/test_worktree_board_routes.py`, plus rewritten suites for `worktree_lifecycle.py` and `open_loops.py`). Title carries a valid conventional-commit prefix (`feat(worktrees):`).

## PII / Secrets / Wording Scan

Scanned all added lines for private IPs (10.x / 192.168.x / Tailscale CGNAT 100.64–127.x), IPv6 ULA, API keys/tokens/passwords, SSH key material, and DB connection strings with embedded credentials: none found.

One path-with-username pattern (`/home/ubuntu/genesis/secrets.env`) appears in `tests/test_scripts/test_worktree_lifecycle.py` and `tests/test_dashboard/test_worktree_board_routes.py` — this is synthetic fixture data purpose-built to exercise the absolute-symlink archive/recovery path, matching the stated test-fixture exception. Not counted as a finding.

PR body wording checked specifically against the code: the "absolute home paths on an unauthenticated GET" and `secrets.env` symlink-convention language describes fixes already shipped in this same diff, for an internal non-secret naming convention (a symlink to a repo-local file, not a live credential). `_relative_location()` in `dashboard/routes/worktrees.py` strips `Path.home()` before the JSON response reaches the wire (with a regression test asserting the home prefix is absent), and the `extractall` fallback to `filter="tar"` on `AbsoluteLinkError` only changes extraction mode — it still relies on the pre-existing per-destination containment checks in `_restore_from_dir`. No live IP, hostname, username path, or actual secret value is disclosed anywhere in the diff. This reads as ordinary engineering description of a fix that shipped in the PR, not an advertisement of a live exploitable issue.

PII/Secrets/Wording: clean (no disqualifying findings; the one soft-warning-pattern match is an excluded test-fixture false positive).

## Summary

Nothing here rises to a blocking finding. The WARNING/NOTE items above are optional follow-ups.

VERDICT: PASS



---

WingedGuardian:
@codex review

WingedGuardian:
Review notes from the PR-queue session. One request, two findings. Nothing here blocks the PR.

## 1. Request: please keep the `locked` check in `_classify`

I am building worktree ownership on top of `git worktree lock`, and this PR's `STATE_PROTECTED` branch is what makes that work. On `main` it is the first check in the reap loop; here it is preserved inside `_classify`. Flagging it because **nothing currently uses the mechanism** — measured, 0 of 198 worktrees are locked — so it would be easy to mistake for dead code in a later round.

MEASURED, with a control that moves. Scratch repo, merged branch backdated past the staleness window, the real reaper copied in so it resolves the scratch root:

```
UNLOCKED -> WOULD TRASH /.../wt-done
LOCKED   -> SKIP /.../wt-done: locked (git worktree lock)
```

Same worktree, same command, verdict flips on the lock alone. Git itself also refuses to remove a locked worktree (exit 128) without an explicit force flag, so the lock covers both the reaper and a manual removal.

## 2. Finding: the restore-contract docstring contradicts the new unmerged lane

`_restore_from_dir` justifies not reapplying uncommitted tracked modifications with:

> the reaper only trashes worktrees already merged into main, so committed content is always recoverable from main

That premise is true on `main` and **false in this PR**, which adds `STATE_REAP_UNMERGED` at `UNMERGED_STALE_DAYS`. Unmerged content is exactly what is not recoverable from main — that is why the lane exists.

Severity is low, because the data survives: `_archive` tars the whole moved directory and `_purge_old_trash` is gone, so a dirty file is inside the archive indefinitely. It just is not replayed by `--recover`, and the docstring now explains why using a reason that no longer holds. The same docstring names the fix (preserve the git admin dir at trash time rather than move-then-prune). Fix or consciously accept — I am not opening a competing PR on it.

## 3. Finding: externally-created worktrees are enumerated by this reaper

Worth knowing for the new unmerged lane. A tool that creates a worktree with `git worktree add` against this repo — even when the working directory lives far outside the repo tree — produces a `.git` file pointing back at `<repo>/.git/worktrees/<name>`. It is therefore a normal linked worktree: it appears in `git worktree list --porcelain` and is enumerated by `_list_worktrees` like any other.

Measured on an install running such a tool: its worktree appears in the repo's own worktree list alongside the rest. So a worktree belonging to a paused or long-running external workflow is a reap candidate, and the process-CWD check will not protect it — see below.

## Context for all three

The `/proc`-CWD liveness check that both this script and the worktree guard rely on is structurally inert for real sessions: measured, **zero live sessions have a CWD inside any worktree**, because sessions `cd` per command and the process CWD never leaves the main tree. That is not a defect this PR introduces, and it is not yours to fix here — it is why the lock matters more than it looks, and why finding 3 has no existing backstop.

WingedGuardian:
@codex review

WingedGuardian:
Closing in favour of four separate PRs. **No work is lost** — every code line in this branch is carried verbatim into one of them, and all 34 review findings are named as inherited and still open in the PR that now owns them.

## Why

This PR hit the 3-round Codex cap with 13 always-fix findings (8 P1 + 5 CodeRabbit Critical/Major, review score 27.5). The escalation gate's own guidance is to triage and minimise rather than request another round, so I grouped the findings by the file they land in first. The pattern was not a bad design — it was **four unrelated features being reviewed as one**, so every round found new defects in a *different* feature:

| feature | size | findings | now |
|---|---|---|---|
| reaper archives instead of deleting | +1,094 | 16 | **#2008** |
| push-time gates (privacy + push-without-PR) | +727 | 8 | **#2009** |
| `session_provenance.py` | +562 | 11 | **#2011** |
| dashboard worktree board | +445 | 12 | **#2012** (stacked on #2008) |

5,010 insertions across 22 files. This repo's own measure: 86% of PRs over 1,000 lines reach 3+ rounds, against 6% under 200.

## How the split was decided

By **commit history and imports**, not by filename guessing:

- the reaper imports none of the other three (grepped), so it is fully standalone
- the board shells out to the reaper's `--report-json` (`routes/worktrees.py:167`), a flag that **does not exist on main** — so it is stacked on #2008 and cannot merge before it
- `open_loops.py` was rewritten to read the *board cache*, so it went with the board even though it was committed under the reaper's commit message
- the two doc changes were placed by reading their **diffs**, not their titles — both are about the push-without-PR gate
- the push-gate files were grouped because all six land in the same three commits

## Before carving

Main was merged in (35 commits behind, `CONFLICTING`), which cleared the conflict and the `dependency-audit` failure. The `leak-detector` failure was a false positive — `detect-secrets` was flagging a line *inside the comment that documents this branch's own leak-detector fix* — cleared with the repo's established allowlist pragma and re-verified against CI's exact invocation: 0 findings.

## One defect found that had no finding

Running the reaper rather than reading it: `--report-json --no-network` emitted a prose note on **stdout** ahead of the JSON, making the document unparseable — and the board consumes exactly that flag through a JSON parser. Fixed in #2008 by sending the note to stderr, verified in both directions.

## What the carves do *not* do

They do not answer any of the 34 findings. Carving makes them reviewable against 500–1,100 lines instead of 5,010; none has been silently fixed or dropped, and each carve's body lists the ones it inherits. The substantive calls — whether the privacy gate should block, whether an unwired 562-line tool should ship, whether the absent-cache path in `open_loops.py` is handled — stay with whoever owns them.

PR #1944, opened 2026-09-11
fix(git): a changelog entry removed at a release cut must not come back

WingedGuardian:
`CHANGELOG.md` carried the repository's only merge driver, `merge=union`. Union takes lines from **both** sides of a conflicting hunk, so it cannot express a **removal**: when one side deletes lines the other side's hunk overlaps, the deleted lines come back — at exit 0, with no conflict reported.

That is not a corner case here. Cutting a release **moves** every unreleased entry under a version heading, so every branch open across a release cut is exposed.

## Measured

On a live branch: a release cut removed **287 substantive lines** from `CHANGELOG.md`; the branch merged the base afterwards with an ordinary `git merge`; the merge reported **no conflict and exited 0**; all 287 lines were present again.

Replayed in isolation on the same three blobs with `git merge-file`: the plain three-way merge exits **1** (the real conflict), union exits **0**, and the union output carries **78 unique substantive lines** the base had deleted. A probe line appears once in the merge base, zero times in the base branch, zero times in `changelog.d/`, and once in the union output.

**Confounder ruled out.** A branch that simply never merged the base since the release also "contains" the deleted lines — that is ordinary staleness, not resurrection. Separated by ancestry (`git merge-base --is-ancestor`): across 82 open PRs, 31 carry the lines, **30 are merely stale**, and exactly **1** has the release commit as an ancestor *and* the deleted lines. The first count is the one that looks alarming and it is the wrong question.

## The part that decided it

That branch was **compliant**. It used `changelog.d/` fragments and never edited `CHANGELOG.md` deliberately. Union hit it anyway during a routine catch-up merge, because the attribute fires on the *file*, not on whether the branch touched it. A hazard you cannot avoid by following the convention is not one a convention can contain.

## The benefit was already measured at zero

GitHub ignores a repository's `.gitattributes` in its server-side merge — tested against GitHub's own merge engine on two branch pairs built to collide: 409 Conflict **with** the union attribute present in both commits, and 409 for the no-attribute control. So the rule provably could not make a conflicting PR mergeable on github.com, which is the only place the merge gate reads. It bought local convenience and paid for it in silent data loss.

## What replaces it

Nothing new — the structural fix already shipped. `changelog.d/`, one fragment per change, assembled at release time. Two branches writing two different filenames have nothing to merge. Union was a stopgap layered on top of a fix that had already landed.

A collision now **conflicts visibly** and is resolved by taking the base's file and moving the entry into a fragment. That is the correct outcome: the failure mode worth preventing is the silent one.

## Tests invert with the rule

The old negative control — *"without the attribute the same merge conflicts"* — becomes vacuous once the attribute is gone, since it would assert the same thing as the acceptance case. It is replaced by a test that writes a **synthetic** union rule and proves it resurrects a deletion. That keeps the reason executable rather than letting it decay into a comment, and it is the one case where this fixture yields a clean exit-0 merge — which is what establishes that the conflict elsewhere is a property of the driver and not of the harness.

The population check now asserts the union set is **empty** rather than `{CHANGELOG.md}`. The paths-that-do-not-exist-yet cases are unchanged; they remain the only test that fails on an unanchored pattern.

**Verify-RED:** re-adding `/CHANGELOG.md merge=union` turns exactly the two intended tests red (scope check reports `got ['CHANGELOG.md']`; the collision test catches the silent merge). Mutation proven applied by sha256 before/after, restored from a hash-gated copy, `sha256sum -c` OK. 117 tests green afterwards.

## Not bundled, deliberately

15 open PRs still edit `CHANGELOG.md` directly and will each conflict once this lands. Converting them to fragments is separate work.

E2E: after merge, a branch and the base each adding a bullet under the same `[Unreleased]` heading must CONFLICT on `git merge` rather than resolving silently, and `git check-attr merge -- CHANGELOG.md` must report `unspecified`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Changelog merges now use Git’s default conflict handling instead of silently combining changes.
  - Removed changelog entries are no longer unintentionally restored during merges.
  - Conflicting changelog updates are surfaced for manual resolution.
  - Branches created before this update may require an additional merge to fully adopt the corrected behavior.

- **Tests**
  - Added coverage confirming that changelog collisions produce visible conflicts and deletions remain deleted.

---

<a href="https://app.devin.ai/review/wingedguardian/genesis-agi/pull/1944" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,63 +1,52 @@
-# CHANGELOG.md is an append-only list of independent bullets, so two branches
-# adding an entry under the same heading are inserting at the same position
-# rather than disagreeing — which git's default driver calls a conflict. `union`
-# takes lines from both sides instead of leaving markers.
+# CHANGELOG.md deliberately has NO merge driver. This file records why, because
+# the obvious fix is to give it `merge=union` and that was tried here and removed.
 #
-# WHAT THIS DOES NOT DO, because the measurement that motivated it is easy to
-# over-read: it does NOT make a conflicting PR mergeable on github.com. GitHub
-# ignores repository .gitattributes in its server-side merge. MEASURED against
-# GitHub's own merge engine (POST /repos/:owner/:repo/merges) on two branch
-# pairs built to collide on this file: 409 Conflict WITH the union attribute
-# present in both commits, and 409 for the no-attribute control. So a PR that
-# collides here still shows CONFLICTING, and scripts/hooks/git_push_guard.py
-# still blocks the merge on GitHub's `mergeable` field — which this rule
-# provably cannot move.
+# THE PROBLEM UNION WAS ADDED FOR, which is real: two branches that each add a
+# bullet under the same `[Unreleased]` heading are not disagreeing — they are
+# inserting at the same position, which git's default driver calls a conflict.
+# Measured 2026-09-04: of 49 open pull requests, 21 could not merge, and 18 of
+# those 21 conflicted on CHANGELOG.md and nothing else.
 #
-# AND IT DOES NOT HELP A BRANCH'S FIRST MERGE EITHER, which is the one that
-# matters for a PR that is already open. Attributes resolve from the CHECKOUT,
-# not from the commits being merged, so on a branch created before this file
-# existed, `git merge origin/main` still conflicts — the rule arrives IN that
-# merge and is not in effect during it. MEASURED on a real open PR: checked out
-# at its head and merging in the branch carrying this rule, CHANGELOG.md
-# conflicts. An earlier measurement here claimed "18/18 merge clean locally";
-# that replayed the OPPOSITE direction (merging each PR into a checkout that
-# already had the rule) and did not describe the workflow anyone performs. It
-# has been withdrawn.
+# WHY UNION IS NOT THE FIX. `union` takes lines from BOTH sides inside a
+# conflicting hunk. It therefore cannot express a REMOVAL: when one side deletes
+# lines that the other side's hunk overlaps, the deleted lines come back. That
+# is not a corner case here — cutting a release MOVES every unreleased entry
+# under a version heading, so every branch open across a release cut is exposed.
 #
-# So what this actually buys is narrow and worth stating plainly: once a branch
-# CONTAINS this file — every branch cut after it lands, and any older branch
-# after its first merge — subsequent merges of the base branch resolve the
-# changelog with no hand-editing. Measured in that direction: 18/18 clean with
-# every bullet from both sides intact. The structural fix for the collision
-# itself is one changelog fragment per change (see changelog.d/), which this
-# rule does not replace.
+# MEASURED 2026-09-10, on a real open pull request:
+#   * A release cut removed 287 substantive lines from CHANGELOG.md.
+#   * A branch merged the base afterwards with an ordinary `git merge`.
+#   * The merge reported NO conflict and exited 0.
+#   * All 287 removed lines were present in the branch afterwards.
+# Replayed in isolation on the same inputs: the plain three-way merge exits 1
+# (the real conflict), the union merge exits 0, and the union output contains 78
+# unique substantive lines that the base had deleted. A probe line appears once
+# in the merge base, zero times in the base branch, zero times in changelog.d/,
+# and once in the union output.
 #
-# The pattern is ANCHORED. A gitattributes pattern with no slash matches the
-# basename at every depth, like .gitignore (gitattributes(5)) — unanchored, this
-# would silently arm every future vendored or subproject CHANGELOG.md with a
-# driver nobody reviewed for it.
+# THE PART THAT DECIDED IT: that pull request was fully COMPLIANT. It used
+# changelog.d/ fragments and never edited CHANGELOG.md deliberately. Union hit it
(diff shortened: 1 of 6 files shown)


WingedGuardian:
## First-pass review (structural + PII/secrets/wording)

Scope: 3 changed files — `.gitattributes` (comments only), a new `changelog.d/` fragment, and `tests/test_changelog_merge_attribute.py`. No `src/` code changes. This is more than doc/config/markdown-only (it touches a `.py` test file), so I ran both review passes in full.

### Domain 1 — Structural
Read the full modified test file, not just the diff, for context.

- No pattern-violation errors: no `nohup`, no direct `task_states` writes, no `/tmp/` usage (uses pytest's `tmp_path` fixture throughout), no `git add -A`/`git push --force`/`git reset --hard`, no bare `except:`, no mutable default arguments, no raw SQL, no `os.killpg`.
- Import hygiene: clean — `from __future__ import annotations`, then stdlib (`os`, `shutil`, `subprocess`, `pathlib`), then third-party (`pytest`); no star or circular imports.
- Coverage: this change modifies an existing test file directly (it *is* the test surface for the `.gitattributes` behavior), so the "new `.py` without a test" rule doesn't apply.
- Docstrings: module and all four test functions carry substantial docstrings explaining intent; helper functions are private (`_git`, `_git_ok`, `_hermetic_env`, etc.) and pre-existing in style.
- Genesis conventions: not a hook script, no `$CLAUDE_PROJECT_DIR` misuse, no `IntervalTrigger` involved. Commit/PR title carries a conventional prefix (`fix(git): ...`).

- Code diff (added lines only): no API keys/tokens/secrets, no private IPv4/IPv6 ranges, no SSH keys, no DB connection strings with credentials. The only email-shaped string is `test@example.invalid` in the test fixture — RFC 2606 reserved test domain, not a personal address (known-false-positive class).
- No absolute paths with usernames, no location-revealing tz identifiers, no cloud account IDs, no hardware identifiers, no private hostnames.
- PR title/body/changelog fragment: describes a git merge-driver bug (measured data loss from `merge=union` resurrecting deleted CHANGELOG lines) in purely technical terms. No mention of a security incident, no naming of private infrastructure, no exploitable-vulnerability advertisement — this is public repo tooling behavior, not a leak/cleanup disclosure.

VERDICT: PASS



---

WingedGuardian:

Pushed 85bf6f5963791cf6dc809a80c7d84b9be889184e for the owner-requested shared audit correction. The dependency-audit step now reports failures without failing the job, matching the verified #1960 reporting policy. Setup/install failures remain blocking. All existing audit commands and ignore arguments are preserved; no dependencies or vulnerability suppressions changed. SECURITY/Dependabot descriptions now match the policy.

Independent review verified the full parsed workflow is unchanged except for audit-step continue-on-error, with negative controls for install/job-level relaxation and ignore-list drift. Fresh CI has been triggered at this commit. Existing PR findings, review freshness and required checks remain merge gates; no merge or additional review round was triggered by this propagation.

WingedGuardian:
## Reply sweep: no findings cleared here — both are open on purpose

A reviewer session swept the queue's unreplied findings, clearing those verified fixed at head. **Neither of this PR's two findings qualified**, so nothing was replied and both still score. Recording why, since "all findings were filed against an older head" is what put this PR on the candidate list and that turned out not to mean fixed.

### The P1 is accepted-and-documented, not eliminated

`d12fe566a` and `2e9efe4a1` pin and document the transition window; they do not close it. The hazard the finding describes is intact: a pre-fix branch's FIRST catch-up merge resolves attributes from its own checkout, where `merge=union` still exists, so the incoming removal cannot govern that merge. `.gitattributes:29` and `:47` now state this in the file.

That may well be the right disposition — from this PR's vantage the window looks structurally unreachable, since the PR cannot modify other branches, and the reviewer's own remedy ("remove the attribute in a separate pre-merge commit") is an operational step on each exposed branch rather than a change here. But that is a call for the author and the owner, not something a sweeping reviewer should clear by replying. The finding closes with **"Ready to merge: No — the targeted silent resurrection remains possible on the branches currently exposed"**, and nothing since has answered that.

If the disposition IS accept-and-document, the reply should say so explicitly and carry the structural argument — a maintainer reply clears a P1, so the reasoning needs to be on the record rather than implied by a docs commit.

### The Minor is partially addressed, but not at the gap it named

The suite now asserts repo-wide that no tracked path resolves to `merge=union` (`union_paths == set()`), which is stronger than the original per-path check for the union case. The finding's actual point survives: it asked for an exact `_merge_attr("CHANGELOG.md", cwd=REPO_ROOT) == "unspecified"`, because a future `/CHANGELOG.md merge=text` rule passes every current assertion while `git check-attr` no longer reports `unspecified`. Grepped the file at `d12fe566a`: no `== "unspecified"` assertion exists; the only occurrences of the word are in a docstring.

One observation from running it rather than reading: on `main` today, `git check-attr merge -- CHANGELOG.md` still returns `merge: union`. That is expected while this PR is open, and it is also the thing that makes the transition window above concrete rather than theoretical.

PR #1962, opened 2026-09-11
fix(cc): the slot door must hand over a temp dir it can actually use

WingedGuardian:
## What this is

One of three independent attempts at the slot-door failure class. This is the
**slot-environment usability** half: a temp directory that is *created* is not
necessarily *usable*, and a blanked `TMPDIR` is not the same as an unset one.

Nine commits, finished 2026-09-05, and never pushed until now — so it has had no CI
and no review. Opening it so the work is visible and reviewable rather than sitting on
one disk.

## The failures it addresses

- **`mkdir -p` returns success for a directory that already exists**, including a
  root-owned one left by an earlier sudo run. So the door accepted a temp directory it
  could not write to. Each candidate must now be created **and** writable **and**
  private before it is accepted, in a loop no future candidate can bypass.
- The old form ran `mkdir`/`chmod` **unguarded under `set -euo pipefail`**, so a
  failing `chmod` killed the whole door.
- The value is pinned into the session with `-e TMPDIR=`, so an unusable path was
  propagated to **every** slot rather than staying local to the script.
- **`unset`, never `TMPDIR=""`.** Blanking an already-exported variable keeps the
  export attribute, so the child receives a literal `TMPDIR=`. This script ends in
  `exec tmux`, which *starts* the server when none is running — so an empty value would
  be inherited by that server and by every slot created on it afterwards.
- `~/tmp` is treated as a **degraded** fallback (disk-hygiene prunes it at 7 days),
  never an equal one.

## How it relates to the other two attempts

`fix/slot-door-heal` attacked this failure class as one branch and took twelve rounds,
each finding a defect in a different place, because it bundled three distinct problems.
`s1` and `s2` are the decomposition of that branch:

- **`s2` (#1798)** — a slot whose Claude has died: detect, and rebuild by consent.
- **`s1` (this PR)** — the slot environment the door hands over.

They are complementary rather than competing, but they will conflict textually: both
edit `scripts/cc-slot.sh` and `tests/test_scripts/test_session_doors.py`.

## Known issues to fix on this branch

Recorded up front rather than waiting for review:

1. **It edits `CHANGELOG.md` directly.** The repo convention is a `changelog.d/`
   fragment; a direct edit is the single largest source of pairwise PR conflicts here.
   This needs converting.
2. **Base is 2026-09-05**; main has moved a long way since. Needs main merged in.
3. Its `TWIN:` comment points at an in-tmux wrapper in `scripts/bootstrap.sh` that
   **exists only on `fix/slot-door-heal`** — main's `bootstrap.sh:900` gates its wrap on
   `[ -z "${TMUX:-}" ]`, so the inside-tmux branch is not there. Either that wrapper
   lands (it is the third salvage from `heal`) or the comment needs correcting so it
   does not point at absent code.

E2E: from a slot login, confirm `TMPDIR` resolves to a writable, private directory, and
that a deliberately root-owned candidate is rejected rather than accepted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved temporary-directory handling by selecting a secure, writable location when available.
  * Prevented invalid or unavailable temporary-directory settings from being passed to launched sessions.
  * Ensured session launches can proceed cleanly when no usable temporary directory exists, with a warning.
  * Preserved conditional OAuth pane startup when optional temporary-directory cleanup is unavailable.
  * Improved manual session setup and hostname-based slot recognition.


WingedGuardian:
**Automated first-pass review** — structural + PII/secrets/wording, run on head `ebcc7fff`.

### Structural review

No ERROR-level pattern violations found in `scripts/cc-slot.sh` or the two test files:
- Temp-dir handling correctly targets `~/.genesis/cc-tmp` / `~/tmp` (never `/tmp/`), matching project convention.
- No `nohup`, no direct `task_states` writes, no `git add -A`/`--force`/`reset --hard`, no bare `except:`, no mutable default args, no raw SQL against `genesis.db`.
- Import hygiene: `import os` added to `tests/test_scripts/test_session_doors.py` is correctly grouped/ordered in the stdlib block.
- Coverage: the two new behavior branches in `cc-slot.sh` (the create/writable/chmod-guarded candidate loop, and the "leave genuinely unset" fallback) are each backed by a new test (`test_create_pins_measured_env_before_lang`, `test_no_usable_temp_dir_leaves_tmpdir_unset_not_empty`).
- Commit-prefix convention satisfied (`fix(cc): ...`).

Two items, both already called out in the PR body itself under "Known issues to fix on this branch," so flagging for visibility rather than as new findings:

- **NOTE** — `CHANGELOG.md` is edited directly in this diff; repo convention is a `changelog.d/` fragment (PR body item 1).
- **WARNING** — the `TWIN:` comment above the candidate loop in `scripts/cc-slot.sh` (~line 397) says the in-tmux wrapper in `scripts/bootstrap.sh` "carries this same loop." Verified against current `main`: `scripts/bootstrap.sh:900` gates its wrap on `[ -z "${TMUX:-}" ]`, i.e. it only fires *outside* tmux — there is no in-tmux twin on `main` today (that wrapper lives on `fix/slot-door-heal`). Until that wrapper lands or the comment is corrected, it points at code that doesn't exist on this branch's base (PR body item 3).

No other structural findings — nothing else to duplicate from ruff/CI.

### PII / Secrets / Wording

Scanned added lines across all 5 files, plus PR title/body and the `CHANGELOG.md` addition.

- No API keys, tokens, passwords, or secrets.
- No private IPv4 ranges (10.x/192.168.x/100.64–127.x) or IPv6 ULA.
- No SSH keys, DB connection strings, or secret-bearing env values.
- No personal emails, usernamed absolute paths, location-revealing timezone IDs, AWS/GCP IDs, hardware identifiers, or private hostnames (test hostnames used — `genesis-3-4`, `genesis-abc` — are generic slot-parsing fixtures, not real infra names).
- PR title/body/CHANGELOG wording is a plain technical description of temp-directory usability/ownership handling; no leak/incident/vulnerability-disclosure language and no private infrastructure named.

VERDICT: PASS



---

WingedGuardian:
@codex review

Claim: codex-01a091e8-d482-78a3-bc7f-4daf10bd0bdd. Refreshed against current main at c365be4f8; the merge is clean. Focused validation: 31 passed across tests/test_scripts/test_cc_slot_oauth.py and tests/test_scripts/test_session_doors.py.

PR #1989, opened 2026-09-13
fix(hooks): a command inside env -S is a command, not a program name

WingedGuardian:
## Problem

`env -S 'git push --force origin main'` **runs that command.** The parser read the whole operand as the *name* of a program, so every gate keying on `seg.exe` saw something matching nothing it checks. Measured against the real push guard, with a control of the same command written plainly:

| form | runs | resolver saw | push guard |
|---|---|---|---|
| `env -S '<gated command>'` | yes | the entire line, as one program name | allowed |
| the same command, written plainly | yes | the command | **refused** |

## Why this is the second attempt

The first shipped inside #1986 and drew **five P1s, all on this code**, because it modelled `-S` as a flag carrying an opaque string — read, not measured.

So the grammar was measured first this time: **49 forms across two rounds**, each executed with the carried command printing a marker, so the output is evidence of what *ran* rather than of what the string looked like. What that established, none of which the first attempt had:

- **Five spellings run**, including the short bundle `-iS 'cmd'`. That one would not have been guessed — it's the shape `bash -ce` has, which is the only reason it was probed for.
- **Option processing stops** at `--`, at a `VAR=` assignment, and at a bare word. `env FOO=1 -S 'echo hi'` exits 127: there, `-S` is a file name.
- **`-S` has its own escapes.** `\_` is a space *inside a field*, `\c` ends the string mid-field, and an **unknown escape makes env exit 125 and run nothing at all**.
- **`#` starts a comment only at the start of a field**; `a#b` is literal.
- **The split fields are re-read as env's own arguments**, so `-S '-i cmd'` runs `cmd`.
- **Operands after the string are appended** to the command.

## Fix

Two halves that ship together, because either alone is wrong:

- the **table entry** stops the operand being read as the executable;
- the **nested walk** — the same one `bash -c` already uses — shows what it carries.

A table entry by itself would *drop* the operand and hide the command rather than mis-name it, which is worse. The entry carries a comment saying so, because that is the trap a later edit would fall into.

**Where env itself refuses the string, the parser reveals nothing** rather than inventing argv from input the system rejects — the mistake #1686 removed when it stopped decoding unterminated ANSI-C spans. Unknown escape, trailing backslash, unterminated quote, leading comment: all yield no command, asserted both at the decoder and through the real guard.

Detection is restricted to **command position**, which was a round-1 finding on the first attempt: `printf '%s' env -S '<cmd>'` prints text and runs nothing, and reading it as a wrapper would have a gate refuse something that never runs.

**Declared omission, safe direction:** an `env` behind a control keyword (`then env -S …`) is not detected, so it is simply not revealed — the status quo, never a new hole.

## Verification

- **Verify-RED** at `origin/main`: 26 failed, 7 passed. Stated honestly — the 7 do *not* all pass for the right reason. The must-not-reveal cases pass on main because main reveals nothing at all, so they only bite on the green side; the genuinely two-sided controls are the plain-command rows.
- **A 23-row behavioural matrix against the real push guard**, every row also *executed* so a mis-parse of an unrunnable command could not be mistaken for a bypass: **23/23**. One row was wrong when written — I predicted a block for `env -S 'git\_push\_--force…'` and the parser was right, because `\_` makes one program name rather than four words. Corrected and kept as a test in both directions.
- **Corpus replay**, 68,703 unique real commands: **2 resolution diffs, both probe commands from this work**. Comparator validated first against an oracle of 4 must-differ and 7 must-match commands, 11/11 — a near-zero diff from a blind comparator is worth nothing.
- **676 tests green** across every consumer of this parser, plus 33 new. `ruff` clean.

## Region note

Touches `scripts/hooks/shell_parse.py` alongside open PR #1986, in different regions — that PR edits the `xargs` table entry and its lock. #1986 is ahead; this rebases behind it.

Follow-up: 100aed32bae04e37905f849180c2e1ff

E2E: after merge and deploy, re-run the five accepted spellings through the real push guard from a session on the deployed tree — each must be refused, with the plainly-written command still refused as the control and `printf '%s' env -S …` still allowed.

## Summary by CodeRabbit

- **New Features**
  - Added support for commands launched through `env -S` and `--split-string`.
  - Handles environment options, assignments, quoting, escaping, comments, and additional command arguments.
  - Supports nested command detection when `env -S` appears in executable command position.

- **Bug Fixes**
  - Invalid or incomplete split-string input is now rejected safely.
  - Improved command analysis and Git push protection for wrapped commands.
  - Preserved existing handling of redirects, command substitutions, and ordinary argument uses.

---

<a href="https://app.devin.ai/review/wingedguardian/genesis-agi/pull/1989" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


WingedGuardian:
## Converting to draft

Two overlapping hunks in `scripts/hooks/shell_parse.py`, and the first is semantic:

1. This PR inserts an `env -S` carried-command block in the nested-command extraction; main independently added `nested.extend(_embedded_commands(_argv(seg.argv_src)))` in the same region. **Whether the env-S path is still needed under main's generalised `_embedded_commands` extraction is the question** — it may now be redundant, or it may cover a case main's version misses.
2. Main changed `_nested_script(argv)` to `_nested_script(argv, interpreter)`, and this PR adds `_ENV_S_ESCAPES` immediately above that definition.

PR #2009, opened 2026-09-14
feat(hooks): the privacy gate blocks, and the push-without-PR gate tells the truth

WingedGuardian:
> **Carve 2 of 4 from #1942**, which bundled four unrelated features and hit the 3-round review cap with every round finding defects in a *different* one. Every code line here is copied **verbatim** from that branch — only the changelog fragment is new. The point is to make an already-written feature reviewable on its own. (Carve 1: #2008.)

## The grouping is the author's, not mine

Established from commit history rather than filenames — all six code files land in the same three commits:

- `feat(hooks): the privacy gate blocks, and detects credentials it has never seen`
- `fix(hooks): the push-without-PR gate's own invariant was false, and its test proved nothing`
- `fix(hooks): stop the branch failing its own CI leak-detector`

The two documentation changes were checked by **diff**, not by title: both are about the push-without-PR gate specifically.

## What this PR should be judged on

**The privacy check now REFUSES a push where it previously warned and allowed.** That is an escalation from advisory to blocking, which this repo's default says needs a specific measured reason. The reason is in the feature's own commits and is the author's to defend — I'm carrying it unchanged, not endorsing it.

The counterpart risk is a false positive stopping a legitimate push, and **the branch's own CI hit exactly that**: detect-secrets flagged a line *inside the comment that documents this gate's fix*. So the class is real, not theoretical. Fixed with the repo's established `# pragma: allowlist secret` (8 existing uses, no baseline file), justified by the line being a comment — verified by re-running CI's exact invocation: **0 findings**.

It also stops a comment vouching for the code beside it: any trailing text used to make the check treat an inline credential as a harmless reference, so a note after an assignment quietly disarmed it.

## Lane, named rather than discovered

This diff touches the **enforcement-hook surface**, so it runs the gate-fix lane and lands in the `critical` consequence lane where **two P2s block rather than four**. Stated here so reviewers read it under that lane deliberately instead of finding the threshold in a gate message.

## Inherited findings

**The 8 review findings on these files remain open.** Carving makes them reviewable in isolation; it does not answer them. None has been silently fixed or dropped.

## One thing worth an owner

`CLAUDE.md` is **48,405 characters on main** and this adds ~1,316. I looked for an enforced ceiling and there is none in `.github/workflows` — the 40k figure that appears in session tooling is a context-budget warning, not CI. So this ships, but the file is growing toward a real problem that someone should own.

## Checks

**848 passed** across `tests/test_hooks` (`-k "push_guard or shell_parse or output_scanner"`), **107 passed** on this feature's own two suites. `ruff check .` clean. Built on a branch off `main` with only these files applied, so the isolation is structural — nothing here imports the other three carves.

The pre-push privacy hook flags 14 lines in this diff; all 14 verified synthetic — the source hits are `192.168.1.5` inside prose *explaining that such an address in a test is legitimate*, and every fixture carries the repo's `genesis:verified-generic` marker on a reserved `.example` domain. The detector is flagging its own test corpus.

E2E: after merge, a push containing an inline credential must be **refused** rather than warned about, and a push of ordinary code must be unaffected. Falsifiable in the direction that matters — if the gate starts refusing pushes with no credential in them, the new shape-based rule is over-matching and should be narrowed, not disabled.

## Summary by CodeRabbit

- **New Features**
  - Pushes to public repositories are now blocked when the branch has no open pull request, with supported dry-run and explicit override exceptions.
  - Privacy checks now distinguish blocking security findings from advisory warnings and scan a broader range of pushed changes.
  - Added detection for AWS access keys and credentials embedded in URLs.
  - Security messages identify finding locations and categories without exposing sensitive values.

---

<a href="https://app.devin.ai/review/wingedguardian/genesis-agi/pull/2009" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


WingedGuardian:
**Genesis scheduled first-pass review — PR #2009** (head `25f0344`)

**Structural review**

No blocking pattern violations found: no `nohup`, no direct `task_states` inserts, no `/tmp/` usage (the new gitleaks temp dirs correctly use `~/tmp` via `tempfile.TemporaryDirectory(dir=os.path.expanduser("~/tmp"))`), no `git add -A`/`--force`/`reset --hard`, no bare `except:`, no mutable default arguments, no raw SQL against `genesis.db`.

Import hygiene: clean. All new imports (`contextlib`, `namedtuple`, `urlsplit`, etc.) are stdlib, correctly grouped ahead of the `from X import Y` lines; no star imports or circular imports spotted.

Genesis conventions: clean. `genesis.security.output_scanner` (now imported by `pre_push_privacy_review.py`) is itself stdlib-only (`re` / `dataclasses` / `logging`), so the "hooks must be stdlib-only or genesis.env" convention holds even with the new cross-import. PR title carries the `feat:` prefix.

Two coverage notes (WARNING):
- `critical_pattern_names()` (`src/genesis/security/output_scanner.py:141`) is a new public function with no caller anywhere in the repo and no direct test — looks unused in this carve.
- The two new pattern-table entries in `output_scanner.py` (`api_key_aws`, `url_credentials`) aren't directly exercised by `tests/test_security/test_output_scanner.py` — only indirectly, via the hook's own parametrized tests in `tests/test_hooks/test_pre_push_privacy_review.py`. Worth a direct unit test in the module's own suite too.

Documentation: no gaps — docstrings are thorough throughout (this diff is unusually well-documented).

**PII/Secrets/Wording scan**

Code diff: clean. Every private-range/CGNAT literal in the diff (`10.11.12.13`, `100.64.0.1`, `192.168.1.5`, `8.8.8.8`) and every credential-shaped value (`ghp_AAA…`, `sk-ant-BBB…`, `AKIADDD…`, `hunter2`, `correct-horse-battery`) is a synthetic test fixture built specifically to exercise the new scanner/gate — several explicitly `# genesis:verified-generic`-annotated. No real API keys, tokens, private IPs, SSH keys, or connection strings found anywhere in the diff. The one email-shaped fixture (`realperson@personal.example`) uses the reserved `.example` TLD and exists solely to prove the hook's renderer never echoes a matched value into output.

PR title, body, and the changelog fragment: no wording issues. The behavior change (advisory → block) is described in generic terms throughout; no specific leaked value, private hostname, IP, or exploitable detail is named anywhere in the PR metadata.

PII/Secrets/Wording: clean (soft warnings only, and none of the fixtures above really rise even to that level given the explicit annotations and reserved domains/ranges used).

No blocking findings from either pass.

VERDICT: PASS



---

WingedGuardian:
## Converting to draft

`scripts/hooks/git_push_guard.py` has two conflict regions. One is a trivial import divergence that would union cleanly. The other is a **behavioural choice on the same guard arm**: this PR inserts a hard `BLOCKED` return when a published branch has no open PR, while main independently landed a different shape on that arm — downgrading the silent allow to an *ask* keyed on `closes_pr`.

Block versus ask on a gate is a design decision with a standing owner ruling behind it, not something a merge can pick.

PR #2011, opened 2026-09-14
feat(scripts): answer which session produced a change, from git history alone

WingedGuardian:
> **Carve 3 of 4 from #1942**, which bundled four unrelated features and hit the 3-round review cap with every round finding defects in a *different* one. Every code line is copied **verbatim** from that branch; only the changelog fragment is new. (Carve 1: #2008 · Carve 2: #2009.)

A command that answers **"which session produced this?"** — and the reverse — from git history alone. Given a file, a line or a commit it reports the session that made it; given a session it reports what that session changed. It reads only what the repository already records, so it works retroactively on history nobody instrumented at the time.

## Entirely additive

**975 insertions, zero deletions, no existing file modified.** Nothing that exists today changes behaviour because of this PR. That's the strongest safety property available here, and it's checked rather than assumed — `git diff --cached --stat` reports no deletion count at all.

Independence was the cheapest thing to establish on this carve: grepping the whole source branch for the module name returns exactly **two** files — the script and its own test. It imports none of the other three carves and none imports it.

## The thing a reviewer most needs to know: it is not wired

Per this repo's own verification taxonomy, this reaches **Level 2 (substantive)** and not **Level 3 (wired)**. There is no call site anywhere — not `disk_hygiene.sh`, not a systemd unit, not a hook, not another module. Compare `scripts/worktree_lifecycle.py`, which is also a script and *is* wired.

That isn't automatically wrong: a hand-run CLI is a legitimate category, and this one reads only git history so it has no runtime to be wired into. But two things follow:

- nothing will exercise it after merge, so a regression is invisible until someone next runs it;
- **it is documented nowhere** — not in `CLAUDE.md`, not in the genesis-development skill, not in `.claude/docs`. I checked, because a hand-run tool nobody can discover is indistinguishable from dead code six months on.

I deliberately did **not** add documentation for it. Writing docs for another session's tool, inside a carve whose purpose is to make that tool reviewable, would be me deciding how it should be presented.

## Inherited findings

**Eleven of #1942's findings land on this file** — the second-largest cluster after the reaper. They're inherited, open, and untriaged here. Carving is what makes them reviewable against 562 lines instead of 5,010; none has been silently fixed or dropped.

## Why carry it at all

It could have been dropped and closed with #1942. It's here because the work exists, is tested, and silently discarding another session's 562 lines is a decision its author should make — which closing this PR now makes a one-click choice, on its own merits, instead of a side effect of retiring a bundle.

## Checks

[...]

PR #2012, opened 2026-09-14
feat(dashboard): show what the worktree cleanup is about to do

WingedGuardian:
> **Carve 4 of 4 from #1942**, and the only one **stacked** — base is `feat/reaper-archives` (#2008), not `main`. Every code line is copied **verbatim** from the source branch; only the changelog fragment is new. (Carve 1: #2008 · Carve 2: #2009 · Carve 3: #2011.)

The dashboard now shows what the worktree cleanup is about to do, before it does it — every worktree with a single plain state (`in_use` / `protected` / `fresh` / `at_risk` / `reap_merged` / `reap_unmerged`) and the reason behind it. Previously the only way to know was to read the job's log afterwards.

## The stack is necessary, and verified

Not inferred from feature names:

- `--report-json` **does not exist** on main's reaper — `git show origin/main:scripts/worktree_lifecycle.py | grep -c report-json` → **0**
- the route shells out to exactly that flag — `routes/worktrees.py:167`
- it renders states only the new classifier produces

Based off `main` this would ship a page whose data source does not exist. **#2008 must land first.**

## The property it rests on is structural

**The route computes nothing.** It runs the cleanup's own classification and renders the result, so the view and the action come from one code path and cannot disagree. A second implementation of "is this worktree stale" would be exactly the snapshot-versus-act problem that sank an earlier design in this area.

## The one behavioural risk — `open_loops.py`

This is a **rewrite**, not an addition: 111 insertions against 73 deletions. It used to measure worktree activity itself and now reads the board cache — and it feeds the **session-start summary every session sees**.

- **Good:** it stops being a second, divergent opinion about staleness, for the same reason the route computes nothing.
- **Risk:** it now depends on a cache another process writes. An absent or stale cache degrades the summary in a way its self-contained version could not.

The authors clearly considered cache *poisoning* — degraded (`--no-network`) classifications are shown but never saved as the shared view. Whether the **absent**-cache path is equally handled is the question I'd put to this PR, and **I have not answered it**: 33 passing tests are the authors' model of the states, not proof the unhandled one is absent. A reviewer has independently filed the same concern ("Expose refresh when no board cache exists"), which is corroboration rather than my discovery.

I flag rather than fix — changing another session's cache-miss behaviour inside a carve would defeat the carve.

## Inherited findings

**Twelve of #1942's findings land on this feature** (7 on `routes/worktrees.py`, 3 on `open_loops.py`, 1 JS, 1 template) and remain open and untriaged here. Several read as substantive on their titles alone: *"Do not publish `--no-network` classifications to the shared board cache"*, *"Do not return incomplete session attribution after a Git failure"*, *"Mark malformed board caches unavailable"*.

**I did not review the front end** — 98 lines of JS and 147 of templates — beyond noting that one open finding (*"Sanitize paths embedded in classification reasons"*) is exactly that class. Worktree paths are attacker-influenceable only by someone who can already create worktrees here, so severity is bounded; bounded is not checked, and I'm not claiming to have checked it.

## Checks

**33 passed** (`test_worktree_board_routes.py` + `test_open_loops.py`) from a branch based on the reaper carve with only these files applied. `ruff check .` clean.

`routes/__init__.py` is a two-line change with history: on the source branch it was the one file that conflicted when main was merged in, because main had independently added `zero_drop` to the same two sorted lists. Resolved by **unioning** both sides — dropping either would silently delete a live route registration — and verified by an AST check reconciling imports against `__all__` in both directions.

E2E: after **both** this and #2008 merge, the dashboard's worktree view must show every worktree with the same state the cleanup would act on, and the session-start summary must agree with it. Falsifiable: if the view and a `--report-json` run disagree about any worktree, the "renders, never computes" property is broken.

---

<a href="https://app.devin.ai/review/wingedguardian/genesis-agi/pull/2012" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>


WingedGuardian:

## Structural review

No pattern violations found.

- No `nohup`, no direct `task_states` inserts, no `/tmp/` usage (cache files correctly use `Path.home() / ".genesis" / "worktree-board.json"` etc.), no raw SQL against `genesis.db`.
- No bare `except:` — every handler in `worktrees.py` and `open_loops.py` names specific exception types (`OSError`, `ValueError`, `TypeError`, `json.JSONDecodeError`, `subprocess.TimeoutExpired`).
- No mutable default arguments.
- Import hygiene: stdlib → third-party → local ordering is correct in `worktrees.py`, `open_loops.py`, and `test_worktree_board_routes.py`. No star imports, nothing circular.
- Coverage: the new `src/genesis/dashboard/routes/worktrees.py` has a matching new test file (`tests/test_dashboard/test_worktree_board_routes.py`, 239 lines). The rewritten `open_loops.py` (`_worktree_activity`/`_list_worktrees` removed, `_read_board`/`_newly_at_risk` added) has its existing test file comprehensively updated to match, including a dedicated test (`test_refresh_invokes_the_reaper_in_report_only_mode`) pinning the refresh endpoint's argv to always include `--report-json` and never a reaping flag.
- Docstrings present on all new public functions/route handlers; docstrings updated where behavior changed (e.g. `_worktree_lines`, `_read_board`).
- Commit/PR prefix: `feat(dashboard):` — present.

No structural findings worth flagging at this pass.

Clean (no soft warnings either).

- No API keys, tokens, private IPs, IPv6 ULA addresses, SSH keys, or connection strings in added lines.
- No hardcoded absolute paths containing a real username — cache/trash paths use `Path.home()` dynamically, and the test explicitly resolves the real `Path.home()` at runtime rather than hardcoding a path (its comment references `/home/ubuntu` only as an illustrative counter-example of what *not* to hardcode, inside a test fixture file — known false positive per the ignore list).
- No private hostnames, AWS/GCP identifiers, or hardware identifiers.
- PR body and changelog: the security-relevant discussion (stripping the home-directory prefix from worktree paths before they reach an unauthenticated `GET` endpoint; keeping the refresh endpoint report-only) describes a safeguard the code implements, not disclosure of an actual leak or specific private infrastructure. No CRITICAL wording issue.

VERDICT: PASS



---

WingedGuardian:
## Converting to draft

This was carved stacked on #2008 (`feat/reaper-archives`), which **landed on main in a different revision** than the one this branch was cut from. `scripts/worktree_lifecycle.py` conflicts across 13 hunks where the landed reaper revision and this carve diverge.

Worse, the denominators in `.github/workflows/ci.yml`, `config/atomic_write_baseline.json` and `scripts/check_atomic_writes.py` depend on **which lifecycle revision wins** (61 vs 63 vs 64 sites) — so they need regenerating after that decision, not resolving alongside it.

PR #2034, opened 2026-09-15
docs(skill): a question about an external tool is a measurement, not a choice

WingedGuardian:
## The rule

If the reason for picking a command, a flag, a value or a procedure is a claim about how something **outside this repo** behaves — git, the shell, the harness, a provider API — then "which of these is right?" is an experiment that has not been run, not a judgement anyone is entitled to make.

The tripwire is deliberately mechanical, so it cannot be reasoned around: *you are about to write an instruction someone will execute, and your justification is a sentence about another program's semantics that you did not observe.* Reading the manual feels like evidence. It is not — scoped to picking between candidates, not to reading documentation generally.

Three clauses, each bought by a defect:

1. **Enumerate the space; do not pick cases from it.** List the axes that independently change the behaviour and sweep the cross product. The cases you would think of are the cases you already believe in, which is exactly why they pass.
2. **Pre-register the predicate, the decision rule, and what you will do if nothing passes.** That last clause is what stops the least-bad option being rationalised into the right one.
3. **Control the instrument.** An oracle arm that must score 100% and a no-op arm that must fail. If the oracle is imperfect, every number in the run is void — including the flattering ones.

## Why it earns a section

A note telling a reader how to undo a destructive git operation was wrong **four times**. Each version was reasoned out, and each was refuted by a state nobody had constructed. The fifth was swept rather than chosen — 320 states × 6 procedures, with an oracle and a no-op.

Of the 240 states where the reader's HEAD had not moved, the winner scored **240/240**; the best alternative managed **80**, and the procedure that had actually shipped managed **56**. The sweep also surfaced a boundary no amount of reasoning had found: that same winner scores **0/80** once the reader has committed, where a different command gets 72.

Twenty minutes of compute against three review rounds of guessing.

## Scope

Two files, +130, prose only. New `### Measure, Do Not Choose` in the skill's on-load mindset, the method as section 9 of `references/high-stakes-verification.md`, and one router-table row. No code, no tests, no behaviour.

## Provenance, stated plainly

This content was written on 2026-09-12 and sat **uncommitted in its worktree for two days** — finished, verified, and invisible. It is being landed now as part of a stranded-work reconciliation, which is itself the failure mode the repo's zero-drop rule exists to prevent. Worth saying out loud rather than presenting this as fresh work.

## Review

Self-review before commit found **one should-fix in my own prose**: the paragraph claimed "every single command managed at most 56." Checked against the source measurement — one alternative managed 80. That is false, and in a rule whose entire subject is *do not assert what you did not measure*, a false number discredits the instruction it is offered as evidence for. Corrected to state all three figures with their denominator, and the committed-reader boundary now names the 72/80 alternative so it reads as a trade rather than a dead end.

Also checked: the rule is not already on main (0 occurrences of its distinguishing phrases); `references/high-stakes-verification.md` on main ends at section 8, so appending section 9 cannot collide; the router row points at a heading added in the same commit, so it cannot dangle on landing. Privacy swept with a positive control — 0 hits, and the worked instance is described install-generically.

Branch was 22 commits behind; main merged in after the commit, clean, and the 44 tests that read these files pass.

**Honest limit:** nothing enforces this rule. It is prose in a prompt surface, like the sections around it, and the worked instance is n=1 presented as one example rather than a rate.

E2E: none — prose in a skill file, no runtime surface.


WingedGuardian:
**Leaks / PII / private-context scan — CLEAN.**

Scope: every added line of the branch diff against `origin/main` — 2 files, +130, prose only (`.claude/skills/genesis-development/SKILL.md` and `references/high-stakes-verification.md`).

Patterns swept: absolute home paths embedding a username, RFC1918 and tailnet addresses, email addresses, account handles, host/container names, and infrastructure tool names that would identify this deploy.

**Result: 0 hits.**

**Positive control run:** the same expression over a seeded `+/home/ubuntu/z` line returned 1 match, so the scan can detect a leak and the zero is a real zero rather than a blind instrument.

Content check beyond pattern matching: this is a prompt surface, so the worked instance matters as much as the paths. It is described install-generically — "a note telling a reader how to undo a destructive git operation" — and names no host, no path, no PR number and no session. The figures it quotes (320 states, 240/240, 80, 56, 0/80, 72/80) are procedure scores from a git-semantics sweep and carry nothing about this install.

WingedGuardian:
@codex review

WingedGuardian:
**Scheduled leak/PII review** (docs-only PR — structural review skipped per scope)

This PR changes only two `.md` files under `.claude/skills/genesis-development/` (SKILL.md, references/high-stakes-verification.md). No code files touched.

- No API keys, tokens, passwords, or secrets in the diff
- No private IPs (10.x / 192.168.x / 100.64–127.x), no IPv6 ULA addresses
- No SSH keys, DB connection strings, or env var secret values
- No personal emails, usernamed paths, location-revealing timezone identifiers, AWS/GCP account IDs, hardware identifiers, or private hostnames
- PR title and body describe an internal test-methodology writeup (sweeping git-undo command behavior across enumerated states with an oracle/no-op control) — no mention of a PII/secrets leak, security incident, or specific private infrastructure

No findings.

VERDICT: PASS



---

PR #2062, opened 2026-09-16
docs(dev-skill): fail-open-loudly, and the test is whether repair survives

WingedGuardian:
## What

Guard-failure-direction doctrine for the development skill, plus three smaller lessons and one corrected digit in the architecture map.

The occasion was an external incident at a sibling install: blocking hooks pinned a path to a JSON tool that a package manager deleted **during a successful upgrade**. The guards refused every command for 30 minutes, and the liveness canary reported OK throughout — its dependency-missing branch printed a refusal and exited **before** the line that writes the heartbeat, so a guard refusing everything left exactly the trace a guard nobody called leaves. Recovery took a human at a physical console.

Genesis is headless. The interactive session **is** the repair path, so a guard layer that fails closed at that level does not protect the machine, it bricks it.

## The test, and why it is not the one I first wrote

The first draft turned on *whose hands the refusal costs*. That is a category judgement, it cannot be applied at a keyboard, and it contradicted a shipped decision — corollary (c) records, after three narrowing rounds, that the unattended path **keeps refusing**.

Replaced with a question that has an answer: **with this failure present, can the session still repair the failure?** Name the operations repair needs; check whether your refusal predicate matches any of them.

## The test then failed its own flagship example

`degraded_exit` has **two** legs, and the draft filed both under PARTIAL:

| leg | what breaks | verdict |
|---|---|---|
| predicate-gated | a shared sibling is unimportable; the guard refuses only what its matcher names, Read/Edit survive | repair path **open** → PARTIAL, fail-closed licensed |
| unconditional | `hook_input` **itself** is unimportable or version-skewed; six guards `os._exit(2)` on `Bash` and `pretool_check` does the same on `Write\|Edit`, no payload read, no predicate consulted | neither a command nor an edit is possible → repair path **closed** → SYSTEMIC |

The shipped code fails closed on that second leg **deliberately**. So the draft's sentence *"this is what the shipped decisions already encode"* was falsified by the decision it cited, in the brick-producing direction.

The doctrine now reports what its test returns and names the divergence, rather than bending the test to fit the code. Two things keep that leg from being a total brick today and **neither is a design**: an MCP editing tool is matched by neither matcher, and the host Guardian can `REVERT_CODE` from outside the container — but nothing wakes it, because the container is up and `genesis-server` never imports `hook_input` (one grep hit under `src/`, and it is a comment). **Tracked for a separate decision; not fixed here.**

## Four rules

1. **Could-not-evaluate is its own state** — never rendered as allow/clean/healthy.
2. **Write the degraded marker before the early return**, with zero dependency on what failed. This is the mechanism of the incident above.
3. **Degrade the renderer, never the verdict.**
4. **A watcher must detect could-not-evaluate**, not merely did-not-run.

## The numbers were wrong twice

The draft said *"the two that were wired sit at 3.02% and 10.64%, and what they protect is irreversible"* — a PR's **triage** table read as the population. Six guards are wired; the code says so itself. The correction then over-reached the other way, presenting a four-guard union as the budget for six.

Final text: **1.90%–21.47%** individually, a union of **four** at **28.80%**, read as the *shape* of an operability budget and not its value, with the six-guard figure stated as unmeasured and strictly higher. "Irreversible" is true of `worktree_cwd_guard` and false of `full_suite_guard`, which its own PR calls *heavy*.

## Also in

- **An authoritative-sources rung** under *Measure, Do Not Choose*: pretraining recall is a hypothesis, not a source. Its anchor case was verified against the vendor's own documentation rather than from memory — that rule obeying itself.
- **Never replace a runtime the current session depends on**: the uninstall kills the thing performing the repair before the install runs.
- **The "recruited defect"** — one a release newly leads users *into* — gates the **release**, explicitly without overriding `Keep the PR the PR` on an individual PR.

## CURRENT.md

`check.py` runs **6** parallel probes, not 5. The sixth landed 2026-05-25, three months before that entry's `verified:` stamp, so the entry was attested while already wrong.

**The stamp is deliberately not bumped.** I re-verified one line of roughly forty, and the file's own rule is that bumping on faith writes a false claim into permanent record. The unreliable attestation is filed separately rather than papered over.

## Review

Two internal adversarial rounds, fresh context each. Round 2 was aimed specifically at defects introduced by round 1's fixes and found one — the blocker above. Round 1 found two blockers (the recruited-defect bullet reversing a standing user rule; the whose-hands test contradicting corollary (c)).

Every load-bearing finding was re-derived from source before being acted on, including the CC exit-code contract (against the official hooks reference), the six-file `degraded_exit` population, `print_json_bounded`'s `False` return, and the probe count.

Verified: **90 tests** across the three suites that read this skill; subsystem-map coverage clean; privacy scan clean across the full diff.

E2E: none — prose only, no runtime surface; neither file is imported or executed.

No changelog fragment: a development-process artifact and a doc correction, neither user-visible.


WingedGuardian:
**Review scope:** This PR touches only `.claude/skills/genesis-development/SKILL.md` and `docs/architecture/CURRENT.md` — documentation-only. Per process, structural review is skipped; PII/secrets/wording scan run on the diff and on the PR title/body.

- No API keys, tokens, passwords, secrets, SSH keys, or DB connection strings in the added lines.
- No private IPs (10.x.x.x / 192.168.x.x / 100.64–127.x.x), no IPv6 ULA (fd00::/8).
- No personal email addresses, no hardcoded absolute paths containing usernames, no location-revealing timezone identifiers, no AWS/GCP account IDs, no hardware identifiers (MAC/serial), no private hostnames.
- PR title/body describe an "external incident at a sibling install/toolkit" and reference in-repo PR #2042 — both in generic, non-identifying terms. No mention of a leak/cleanup of this install's own PII or credentials, no named private infrastructure, and no exploitable specifics about a live vulnerability. This reads as internal design-doctrine documentation (fail-open-vs-fail-closed guard semantics), not a security-incident disclosure.

No hard blocks, no soft warnings.

VERDICT: PASS



---

PR #2069, opened 2026-09-16
fix(hooks): a degraded guard must not take away the edit that repairs it

WingedGuardian:
## What

PR #2062 landed a repair-path test — *with this failure present, can the session still **repair** the failure?* Applied to our own shipped code, it found a brick on the first try. This fixes it.

## The defect

When `scripts/hooks/hook_input.py` is unimportable or version-skewed — the file itself calls skew *"a real configuration here, not a hypothetical"* — guards hit an unconditional `os._exit(2)` **before reading any payload**.

**MEASURED by execution** against a poisoned tree: **9 of the 13 PYTHON hooks** that `.claude/settings.json` wires on matcher `Bash` exit 2 (the other 4 are advisory and exit 1, correct for them), and `pretool_check` did the same on `Write|Edit`. The session could then neither run a command nor edit a file. This box is headless. That is a brick.

**Read that denominator exactly.** The matcher carries 15 hook commands across 14 entries; the 2 not counted are shell, never import `hook_input`, and so cannot be affected — which is why they are excluded, and one of them is itself a blocking guard, so the omission is not cosmetic. The repo is also not the LIVE population: this install wires a 16th on the same matcher from its own user-level settings. That restriction was unstated in four surfaces until round 3; see below.

## Why this is safe to change: the refusal was an OVER-block

`pretool_check`'s documented scope is CRITICAL-path Write/Edit in **AUTONOMOUS** sessions only. An interactive session is allowed through by design — *"the user is present and sovereign"* — and its healthy session test is `os.environ.get("GENESIS_CC_SESSION") == "1"`, **pure environment, needing nothing from `hook_input`**.

Degraded, it exited *before* reaching that test. So it refused the interactive owner — whom it would always have allowed — from editing anything at all. And that owner is who repairs the tree.

| | healthy | degraded (before) | degraded (after) |
|---|---|---|---|
| interactive + CRITICAL path | ALLOW | **BLOCK** ← the brick | ALLOW (parity restored) |
| dispatched | BLOCK on CRITICAL | BLOCK on all | BLOCK on all (unchanged) |

**The security delta is zero, and is locked by a test rather than asserted.** Provable by construction: `pretool_check.py:226-230` returns 0 on `not _is_dispatched()` *before* any pattern load, so degraded-allow ⟹ not-dispatched ⟹ healthy-allow. There is no input where the fixed degraded path permits what the healthy guard blocks.

**The 9 Bash guards are untouched and still refuse.** That is the point — fail-closed keeps everything it was protecting, and the session keeps the one route that fixes the cause.

## The notice goes to stdout, and the first draft had this wrong

Claude Code **discards stderr from an exit-0 hook**, so an allow that spoke there would reach nobody and a degraded guard would look exactly like a healthy one — the condition this mechanism exists to make visible. `hookSpecificOutput.additionalContext` is the channel that survives exit 0.

`hook_input._degraded_say` is unreachable here by construction, so the JSON is a **fixed literal**: no serializer, no interpolation, nothing to escape, nothing that can fail. Degrade the renderer, never the verdict.

## Verification

**Verify-RED, recorded before the fix existed** — the interactive test failed with `got 2` on unfixed code, while three controls passed. The healthy-interactive control is what makes the zero-delta claim a measurement: the guard *already* permits exactly that call.

**E2E on the real guard, real interpreter, real poisoned tree, driven as a subprocess:**

```
stamp unset   exit 0   431B valid JSON on stdout   0B stderr
stamp "1"     exit 2   0B stdout                   139B stderr
stamp "true"  exit 0   431B                        0B stderr
```

stderr is 0 bytes on every allow path deliberately. The `"true"` row is **predicate parity**: a non-`"1"` stamp is interactive in *both* paths, or the degraded branch would refuse where the healthy one allows — a divergence invisible to every other test row, since they only pass the two canonical values.

**238 tests pass** across `test_import_time_degraded.py`, `test_hooks/test_pretool_check.py` and `test_scripts/test_pretool_check.py`; **ruff clean.** (This supersedes an earlier body figure of 255, which counted a different file set — a stale test count on *this* PR would have been the same defect a third time.)

**A gap neither earlier round covered, closed by measurement rather than argument.** `settings.json` wires `procedure_advisor.py` on matcher `.*`, so it fires on Write and Edit too, and `behavioral_linter.py` is wired on `Write|Edit` and **does** import `hook_input`. Had either exited 2 on a poisoned tree, this PR's central claim — that fixing `pretool_check` restores the repair path — would be **false**. MEASURED: `procedure_advisor` exits 0 (stdlib imports only), `behavioral_linter` exits 1 (non-blocking; it crashes rather than refuses), and `pretool_check` is the **only** blocker on Write/Edit. The claim survived a test that could have refuted it.

## Round 3 — two counting defects on a PR about a counting defect

**A. Codex P3, and it was right.** The commit that corrected "six" to nine in the skill left two copies of the stale six in the code that passage describes. A third turned up alongside: the test header said "the four cases below" while five tests follow.

**B. Found by the audit of A — the same mistake one level down.** The corrected figure carried a silently restricted denominator, in four surfaces including this body. Fixed by **publishing the restriction, not adjusting the number**: 13→15 was considered and rejected, since it makes the numerator wrong in the other direction. The audit's own figure was corrected in passing rather than copied (it said 15 *entries*; 15 is the *command* count, 14 is the entry count).

Proven comment-only **mechanically**, not by eye: `ast.dump(ast.parse())` identical and the tokenize stream with COMMENT/NL filtered identical, for both Python files.

## Doctrine updated in the same commit

That passage in `genesis-development` describes this code by name; leaving it would make the skill contradict the shipped behaviour it cites.

It also had the **population wrong** — it said six guards, borrowed from `grep degraded_exit(`, when this leg is precisely the one where `degraded_exit` is *unreachable* and three more guards refuse from their own import handlers. Corrected to nine with the denominator, the method, the date, and the trap named inline. This is the fourth instance of that mistake class in this repo; counting the helper instead of the condition is how it happens every time.

The text also no longer implies Leg B is fully resolved: a **dispatched** session remains systemic by the same test, deliberately, because its repair route is a human interactive session — which is what this restores.

## Review

`scripts/pretool_check.py` is hook surface, so this runs the **gate-fix lane**. **No secondary reviewer is configured on this install** (`merge_gate.secondary_reviewer` is absent — checked, not assumed), so round 1 is **Codex-only** rather than dual.

Internal adversarial audit returned no BLOCKER, one SHOULD-FIX (the guard count above — re-measured by execution rather than accepting either number) and four NOTEs, all applied: the verdict-deciding env read moved inside a try with a fail-closed default; the zero-delta test row strengthened so it cannot pass on an unread payload; a fifth test added for predicate parity.

**E2E: built ≠ wired.** `genesis-hook` resolves `HOOK_ROOT` to the MAIN worktree, so this is inert in every session until it lands on main — the runs above drive the worktree copy directly, which is what makes them evidence. Post-merge the live path is the same file under the main tree; the regression marker is an interactive session being refused a Write/Edit it should have been allowed.


WingedGuardian:
Structural review: CLEAN — no pattern violations found. Bare-except usage is typed (`except BaseException:`), the fail-closed handler only touches already-imported stdlib (`os`/`sys`), import grouping and `noqa: E402` placement are correct, the changed hook logic has accompanying test coverage in the same PR, new functions carry docstrings, and the PR title carries a conventional `fix:` prefix.

PII/Secrets/Wording scan: CLEAN — no secrets, private IPs/hostnames, or personal paths in the diff. The PR body's detailed discussion of the guard's fail-open/fail-closed session logic describes pre-existing, already-public guard design rather than disclosing a new exploit or naming private infrastructure.

VERDICT: PASS



---

WingedGuardian:
@codex review

WingedGuardian:
@codex review

PR #2074, opened 2026-09-16
docs(dev-skill): an old PR might not be work any more, and the obvious test is wrong

WingedGuardian:
## What

A review session drains a queue sorted by age, so the older the PR the more
likely it has already landed under another number. Owner observation,
2026-09-15: it happened twice in one session — the PR was open, the work had
merged earlier under a different number.

This adds one section to the genesis-development skill telling a reviewer to
check that before spending effort, and giving the check as a command that
actually works.

## The numbers

Squash merges are why this is not obvious. On a two-commit branch whose work
landed as a single squashed commit:

| test | result |
| --- | --- |
| `git merge-base --is-ancestor <branch> main` | rc 1 — reads as UNMERGED |
| `git cherry -v main <branch>` | `+` on BOTH commits — reads as unique |
| `git diff origin/main..HEAD -- <files>` | **empty — detects it** |

Two dots, not three. `origin/main...HEAD` is merge-base→HEAD: it ignores
everything that landed on `main` after the fork point, including the squash
carrying the work, so it is non-empty for a superseded PR every time.

`git cherry` gets its own sentence because it compares patch-ids: it **does**
catch a squash of a single commit, and is blind exactly when the squash combined
several — the common case. The section says which case it misses rather than
claiming it never works.

## What this PR deliberately does NOT do

The first draft got three things wrong, and the corrections are most of the
value here:

- It shipped the **three-dot** form — the error the section exists to prevent,
  with a command blessing it.
- It said a superseded PR should be **closed**. The standing rule thirty lines
  below defines retiring as exactly `gh pr close` on an unmerged PR and already
  names the superseded case, routing it to the user. The draft argued its way
  around a rule that names the exact thing. The section now says to name the
  successor and leave it open.
- Its second half **restated** guidance that already exists in `closing-session`
  and in this skill's own "Date the code before classifying a red", more briefly
  and without the conditionals — which makes a shorter copy worse than none. It
  is now a pointer, adding only what neither says in this context: `rev-parse
  HEAD` is blind to uncommitted edits.

Placed immediately before `## Pre-Merge Gate` rather than among the
merge-driving rules, so the following section's "the section above" still
resolves where it did, and a top-down reader meets the retire rule first.

## Verification

Docs only; no runtime surface. `tests/test_learning/test_skills.py`,
`tests/test_hooks/test_skill_injection.py`, `tests/test_security/test_skill_scan.py`
→ 59 passed. The git behaviour in the table was measured on a constructed
squash, with the two-dot/three-dot pair as the control.

E2E: none — this adds a section to a skill file; there is nothing to exercise
after merge beyond the skill loading, which the existing tests cover.


WingedGuardian:
This is a documentation-only PR (single file changed: `.claude/skills/genesis-development/SKILL.md`, a skill markdown file — no code). Per review process, running the PII/Secrets/Wording scan only and skipping the structural review.

- No API keys, tokens, passwords, secrets, SSH keys, or DB connection strings in the added lines.
- No private IPv4 (10.x/192.168.x/100.64–127.x) or IPv6 ULA (fd00::/8) addresses.
- No personal email addresses, hardcoded absolute paths with usernames, location-revealing timezones, cloud account IDs, hardware identifiers, or private hostnames — the examples in the diff use generic placeholders (`<branch>`, `<owner>`, `<repo>`, `<sha-on-main>`, `<files>`).
- PR title/body contain no leak/incident wording and name no private infrastructure. The body's "CRITICAL finding … public issue" passage describes a stale-worktree review-process failure (false-positive findings from auditing an outdated worktree), not a PII/secrets leak or exploitable vulnerability — nothing here is exploitable or references a real security exposure.

VERDICT: PASS



---

PR #2093, opened 2026-09-16
fix(code-intel): pin GitNexus without reintroducing OOM loops

WingedGuardian:
## Summary

- centralize the reviewed GitNexus version at 1.6.12 and prevent bootstrap from reinstalling the incompatible 1.6.8 reader
- fail closed for stale, newer-unreviewed, malformed-version, unsupported-Node, and PATH-blind GitNexus binaries
- remove floating `npx gitnexus` resolution and route MCP sessions through a version-gated launcher
- split index memory limits: GitNexus gets the measured 8 GiB rebuild scope while Codebase indexing retains 2 GiB
- make the machine-local Codebase MCP kill switch cover installation, registration, MCP launch, and indexing without creating an endless marker retry loop
- correct skills and code-intelligence guidance, including the fact that linked worktrees intentionally skip snapshot indexing and should use Serena for live branch truth

## Incident evidence

The overnight full rebuild completed on GitNexus 1.6.12 and wrote LadybugDB storage v42. Bootstrap later enforced the repository pin of 1.6.8, whose reader supports storage v41, making the valid index unreadable. A real forced rebuild peaked at 4,874,166,272 bytes, or 4.65 GiB, and completed under an 8 GiB swapless scope. The previous shared 2 GiB cap was therefore guaranteed to kill GitNexus before completion.

The installer now upgrades absent or older versions to the reviewed pin but refuses to automatically downgrade a newer installed version, because that version may already have written a newer storage format. Unreviewed versions remain on disk but cannot index or start through the Genesis MCP launcher.

## Safety and rollout

- Codebase MCP remains disabled. This change does not remove the live sentinel or re-enable it.
- An intentional Codebase disable is treated as a successful skip, so a `both` marker can complete its GitNexus leg and be consumed instead of retrying forever.
- GitNexus remains per-session stdio in this PR. The shared HTTP service and multi-client canary are separate follow-up work.
- The 8 GiB limit is for the short-lived rebuild job, not resident MCP readers.

## Verification

- 148 focused installer, launcher, pin, indexer, watchdog, and Node-contract tests passed after rebasing onto current main
- Ruff passed
- `bash -n` passed for all changed shell entrypoints
- ShellCheck error-level pass succeeded
- atomic-write, shared-artifact, frozen-clock, and diff checks passed
- live version-gated launcher verified GitNexus 1.6.12 status, query, and Cypher against the rebuilt index
- live Cypher returned 65,075 nodes
- Codebase launcher refused via the active sentinel; no real Codebase/index-worker process remained
- cgroup OOM counters stayed unchanged at `oom=73`, `oom_kill=33`

## Follow-up

Canary the optional shared GitNexus HTTP service with explicit repository/worktree arguments, reconnect testing, and reader-memory measurements. Separately complete the durable Codebase memory/concurrency fix before removing its kill switch.

## Summary by CodeRabbit

* **New Features**
  * Added GitNexus version pinning to 1.6.12 with validation for installed versions and supported Node.js releases.
  * Added a machine-level disable switch for Codebase Memory.
  * Added separate memory limits for Codebase Memory and GitNexus indexing.
  * Added launch safeguards for disabled, unsupported, missing, or mismatched installations.

* **Bug Fixes**
  * Prevented unsafe GitNexus downgrades and unverified executable launches.
  * Updated setup requirements and guidance for Node.js 22 or newer.

* **Tests**
  * Expanded coverage for disable switches, version validation, executable discovery, launch rejection, and resource limits.

---

<a href="https://app.devin.ai/review/wingedguardian/genesis-agi/pull/2093" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>


WingedGuardian:

**Scope:** 25 files changed — all shell scripts (`scripts/*.sh`, `.claude/mcp/*`), Markdown docs/skills, one changelog fragment, and Python tests. No new files under `src/`.

### Structural review

No pattern violations at blocking severity (no `nohup`, no direct `task_states` writes, no `git add -A`/`--force`/`--hard`, no bare `except:`, no mutable defaults, no raw SQL, no `npx gitnexus` left behind anywhere). Import hygiene in the touched test files is clean (stdlib-only, correctly ordered). Test coverage for the new shell logic is thorough: `.claude/mcp/run-gitnexus` and `scripts/lib/gitnexus_version.sh` are both new and both have dedicated new/expanded test files (`tests/test_scripts/test_run_gitnexus_launcher.py`, large additions to `test_dependency_pins.py`), and `test_every_gitnexus_package_resolver_uses_the_shared_pin` guards against any doc/script regressing to a floating `npx`/`npm exec` resolver.

One non-blocking maintainability note (WARNING level, not correctness):

- The ~10-line "source `gitnexus_version.sh`, then verify `GENESIS_GITNEXUS_VERSION` + the four `declare -F` helper checks" readiness block is duplicated verbatim in `scripts/bootstrap.sh`, `scripts/install.sh`, and `scripts/lib/code_intel_index.sh`. It's logically correct in all three call sites, but a future change to the helper contract (e.g. adding/renaming a required function) has to be caught and updated in three places by hand. Worth extracting into a single `genesis_gitnexus_pin_ready` predicate inside `gitnexus_version.sh` itself as a follow-up — not a reason to hold this PR, since `test_gitnexus_pin_is_single_sourced_and_current` and `test_gitnexus_registration_always_drift_heals_to_fail_closed_launcher` already lock down the current shape.

Traced through the version-gate logic in `scripts/lib/gitnexus_version.sh` (`genesis_gitnexus_ensure_pin`, `genesis_gitnexus_version_is_newer_than_pin`, `genesis_gitnexus_node_version_supported`) against the new test tables (`test_gitnexus_node_engine_contract_matches_upstream_boundaries`, `test_gitnexus_unreviewed_version_is_never_automatically_replaced`, `test_gitnexus_ensure_pin_install_upgrade_and_postconditions`) — the rc 0/1/2/3 contract (pinned / failed / newer-installed / unrecognized-version) is internally consistent and matches the fail-closed intent described in the PR body.

### PII / Secrets / Wording

Scanned all added lines (scripts, docs, tests) against the Domain 2 list — no API keys/tokens, no private IPv4/IPv6 ranges, no SSH keys, no connection strings, no personal emails, no usernamed paths, no location-revealing timezones, no cloud account IDs, no hardware identifiers, no private hostnames.

PR title, body, and the new changelog fragment (`changelog.d/20260916144547-fixed-gitnexus-pin-downgrade.md`) describe the incident (storage-format mismatch, OOM cap sizing, node/query counts) in plain technical terms with no reference to a PII/credential leak, a security incident, or named private infrastructure.

VERDICT: PASS




---

WingedGuardian:
Measured on a live install today, in case it bears on the pin choice.

The GitNexus MCP server is currently unusable against the main checkout. Every
call returns:

```
LadybugDB unavailable ... Trying to read a database file with a different
version. Database file version: 42, Current build storage version: 41
```

Details:

- installed CLI reports `1.6.12`
- the on-disk index was written earlier the same day by a build using storage
  version 42
- reproducible across separate calls (`impact`, `check`), so it is not the
  "another process may be rebuilding" race the error text also mentions

So the installed binary is older than whatever wrote the index — a downgrade —
and the storage versions do not match. Relevant here because pinning to a version
whose storage format is older than an existing index leaves that index
unreadable: the pin may need to carry an index rebuild (or an explicit
version-skew check) rather than only fixing the install.

Not blocking anything on my side — Serena covered the blast-radius question I
needed. Filing it because it is a silent capability loss: the tool is wired and
answers every query with an error, which reads as "no results" to anything that
does not check.

WingedGuardian:
@codex review

WingedGuardian:
Superseded by #2120, which carries identical content on a clean history.

This branch's history included a commit whose patch added an absolute home path in a test. The CI private-pattern scan reads added lines across **every commit's patch** on purpose — its own comment notes that a value added then removed "leaves the net diff clean yet stays in published history" — so the follow-up commit that removed it could not clear the gate. Force-push is blocked on this install by design, so the branch was re-cut rather than rewritten.

Content equivalence verified rather than assumed: same 25 files, same 1,673 diff lines, content lines identical. The only differences between the two branches' diffs against main are hunk line numbers, because main moved underneath.

Everything established here still holds and is restated in #2120 — the memory-headroom bound with its measurements across install sizes, the registration drift-heal with the launcher's fail-closed behaviour pinned by a test, and the ReDoS fix. Closing in favour of that one.

PR #2097, opened 2026-09-16
fix(install): three guards that validated less than they claimed

WingedGuardian:
#2079 and #2085 each shipped with an accepted finding I deferred rather than looping their review rounds. This closes all three. They are one shape: **a check standing in for something it did not actually verify.**

| | was | now |
|---|---|---|
| unit-path allowlist | read only `argv[0]` | every path-looking token |
| installer enablement | `is-enabled` **exit code** | requires the literal `enabled` state |
| subcommand bypass | `[ -d "$HOME/tmp" ]` | the slot path's own validated resolution |

## The part worth reading

**Widening the first one moved the narrowness rather than removing it.** Going from a positional check to a classifier put the same flaw inside the predicate, where it failed silently in *both* directions. Measured against a table built from systemd's grammar — not from the diff — **12 of 12 wrong**:

- Dropping a token that matched no start-character is **fail-open**, and systemd allows `@ + ! !! :` beyond the `-` I stripped. `ExecStart=+%h/genesis/x.sh` produced *no tokens at all*.
- `.split()` leaves quotes attached, so `/bin/bash -c '%h/genesis/x.sh'` was invisible — and that `-c '...'` shape **already ships** in `config/genesis-guardian-watchman.service`.
- The `$`/`%` catch-alls flagged `ExecReload=/bin/kill -HUP $MAINPID` — *the* documented reload idiom — and advised replacing it with `__REPO_DIR__`. That advice ends with the token allowlisted or the test weakened.

Fixed by unquoting with `shlex`, stripping the full prefix set from `argv[0]` only, and requiring a separator before judging a root. **25/25 correct** after; 0 offenders across all 23 shipped sources, unchanged.

The cells are now parametrized from the **format's grammar**. That distinction is the lesson: *"all 23 shipped sources still pass"* only measures the corpus the author already handled, which is exactly why the first version looked right.

## The other two

**`is-enabled` exit 0 ≠ durable.** It also succeeds for `enabled-runtime`, which lives under `/run` and is gone after a reboot. Verified live: real units read `enabled`; two real `static` units exit 0 and are now correctly rejected.

**The bypass accepted any directory that existed.** Existing ≠ ours, writable, or private — a root-owned `sudo` leftover would have been handed to `claude install`/`update`. The slot path already did this properly, so that loop moved into one `_cc_resolve_tmpdir` both call rather than becoming a third copy; the defect being fixed *is* drift between copies. It also no longer fails silently, which made the scenario the block exists for the one that failed undiagnosably.

## Also removed: a comment that was false

`cc-slot.sh` claimed the bashrc wrapper carries a twin of the temp loop needing hand-sync. It does not and never did — that block has no temp handling, and the loop exists nowhere else in the repo. It sat four lines under new text explaining that a second copy *was* the defect.

Two smaller things the audit caught in my own tests: a drift guard whose `count() >= 3` could not trip (one of the four mentions is prose), and a comment describing a rejection branch the test never reaches.

## Verification

- **Verify-RED on each fix**, restores hash-verified: reverting the state check, the `-d` check, and the old `argv[0]` helper each fails the right test for the right reason.
- **Live e2e** against the real script and real `claude`: ephemeral subcommand → 7 servers, zero tmux calls; `gateway` → keeps its slot; bare session → still pins `TMPDIR=~/.genesis/cc-tmp`, so the extraction preserved the door's behaviour.
- **200 tests pass**; ruff clean; `bash -n` clean on both scripts.

**Not claimed:** that `_path_tokens` is complete. It is an allowlist over a grammar I enumerated, not a systemd parser — a construct outside that grammar has no cell and would pass in silence. Stated so the next reader knows what the table covers.

E2E: the live run above is the end-to-end proof; `mcp list` through the door and a bare session both exercise the restructured resolution against the real binary. Post-merge, a bypassed `claude update` on a box with an unusable `~/tmp` should now print the temp-dir warning rather than proceeding silently.


WingedGuardian:

### Structural review
No pattern violations (ERROR). No nohup, raw SQL, bare `except:`, mutable defaults, `git add -A`/`--force`/`reset --hard`, or `/tmp/` usage — the change strengthens the existing `~/tmp` / `~/.genesis/cc-tmp` validation, consistent with project convention. Import hygiene in the modified test files is correctly grouped (stdlib, then third-party). New logic (`_path_tokens`) is docstringed; test coverage is exactly what's being added.

NOTE (non-blocking): in `scripts/cc-slot.sh`, the comment above the bypass's temp-dir check —
> `` `|| true` is load-bearing: without it `set -e` aborts before `exec claude` when no candidate is usable. ``

— doesn't match the code below it: there's no `|| true` there, and none is needed, since `if [ -z "${TMPDIR:-}" ] && ! _cc_resolve_tmpdir; then` is already exempt from `set -e` as the condition of an `if`. Functionally harmless, but the comment is confusing/inaccurate and worth a follow-up cleanup.

### PII/Secrets/Wording scan
PII/Secrets: clean (no hard blocks, no soft warnings). No secrets, credentials, or private IPs in the diff. The one placeholder path (`/home/someuser/genesis/x.sh`) is synthetic test-fixture data exercising the allowlist test (per the known-false-positive rule for scanner/sanitizer test fixtures), not a real path. PR title/body describe strengthening validation logic — no leak/incident disclosure, no naming of private infrastructure.

VERDICT: PASS



---

Issue #2145, opened 2026-09-17
update.sh records a successful deploy from three sites, but only one can flag that the server was never restarted


WingedGuardian:
`scripts/update.sh` writes a `status='success'` row into `update_history` from three
places. Only one of them can record that the deploy did **not** restart the server, so
the other two can produce a success row naming the new HEAD while nothing is serving it.

## The mechanism

```
$ grep -n '_record_update_history "success"' scripts/update.sh
1528:        _record_update_history "success" "" "$HOST_CC_DEGRADED"
1532:        _record_update_history "success" "" "$HOST_CC_DEGRADED"
2093:_record_update_history "success" "" "$_p6_degraded"

$ grep -n "_OPERATOR_STOP=" scripts/update.sh
1744:_OPERATOR_STOP=false
1773:        _OPERATOR_STOP=true
```

`genesis-server-not-restarted` is appended in exactly one place (`:2085`), gated on
`_OPERATOR_STOP`. That variable is first assigned at `:1744`. The block containing
`:1528` and `:1532` ends with `exit 0` at `:1536` — 208 lines earlier. **The marker is
structurally unreachable from the no-delta path.**

The second half is the gating. At `:1525` the `if` requires both a leftover
failure artifact and `genesis-server` in `WERE_RUNNING`. The `elif` at `:1530` requires
neither:

```bash
    if [ -f "$HOME/.genesis/last_update_failure.json" ] \
        && [[ " ${WERE_RUNNING[*]} " == *" genesis-server "* ]]; then
        ...
        _record_update_history "success" "" "$HOST_CC_DEGRADED"
    elif [ -n "$HOST_CC_DEGRADED" ]; then
        _record_update_history "success" "" "$HOST_CC_DEGRADED"
    fi
```

So on a no-delta run with the server stopped: `WERE_RUNNING` is empty, the restart loop
at `:1474` iterates zero times, `_sync_deploy_targets` at `:1512` sets
`HOST_CC_DEGRADED` during drift healing, and a success row lands carrying only a
host-alignment marker such as `container_cc_sync`.

The script already documents this hazard for the *other* path, at `:1742`:

> Without the recovery push, the restart+health block below is skipped and we'd record
> "success" with the server down and no health check.

That reasoning applies to the no-delta path too, and the guard does not.

## Why it matters

`update_history` is the only durable record of a deploy. Anything asking "is the merged
code actually running here?" has to read it, and a success row that overstates the
[...]

PR #2147, opened 2026-09-17
fix(observability): resolve a stored short commit against the object store, not the ref namespace

WingedGuardian:
`update_history.new_commit` and the guardian state file both hold **abbreviated** commit names, and two collectors in `deploy_health.py` each expanded one by hand with `cat-file -e '<name>^{commit}'`. That form consults the **ref namespace** first, and the consequence is not a refusal but a wrong answer.

MEASURED on git 2.43.0, with a branch named after an 8-hex prefix of a different commit:

```
$ git rev-parse --verify --quiet 'fc20dbf4^{commit}'
91d380f102b758280ffc2f8c7b07256247e88a23    # the BRANCH's commit
rc=0   stderr=(empty)

$ git rev-parse --disambiguate=fc20dbf4
fc20dbf4b1a8aafd5a1af0928443085559a92a8b    # the object
```

`--quiet` suppresses the ambiguity warning entirely, so the caller gets a confident, silently wrong SHA. A tag shadows identically. Both collectors then interpolated that name into a diff range, so a shadowing ref quietly changed which range was compared — `collect_tier2_pending` would report no pending tier-2 work, and `collect_host_gateway` no guardian drift, with nothing anywhere saying why.

## The fix

`resolve_commit()` replaces both copies:

- resolves with `--disambiguate`, which reads only the object store, so no ref can shadow it, and which **lists** every match so an ambiguous prefix is an error rather than a silent pick;
- confirms the result with `cat-file -t`, because `--disambiguate` returns blobs and trees too — "resolved" must not be allowed to mean merely "exists";
- reports a git timeout or exec failure as "could not check" rather than "not here", because those have different remedies.

The abbreviation floor is **4**, git's own minimum (`core.abbrev=4` prints a 4-hex name, `=3` errors), not the 8 this install happens to store. A higher floor would refuse every stored commit on an install configured lower.

## Behaviour preservation

The new form refuses six input classes the old one accepted: `HEAD`, `HEAD~1`, branch names, tag names, uppercase hex, and the empty string. Neither live input can produce them — the guardian value is regex-gated `^[0-9a-f]{7,40}$` at its source or the literal `unknown` (handled earlier), and `new_commit` is `rev-parse --short` output, 8-9 lowercase hex on all 91 rows here.

Verified against the live repo and the live state file: `resolve_commit` expands 8- and 9-hex names correctly, passes a full SHA through after verifying it, refuses an absent commit and refuses `HEAD` by shape; `collect_tier2_pending` returns real pending paths; and `collect_host_gateway` still reports `status: ok` rather than `unknown_commit`, which is the state a broken resolver would produce.

## What is deliberately not here

A deploy predicate built on this resolver — "is this merged commit actually running?" — was in two earlier revisions of this branch and has been removed. It has no consumer until `pr_verifications` stores merge commits, and it rested on the assumption that a `status='success'` row implies the server was restarted. **That assumption is false**: `update.sh` writes a success row from three sites and only one of them can record that no restart happened, because the variable gating that marker is first assigned past the other two sites' `exit 0`. Filed as #2145 against the deploy script rather than worked around in an observability module.

## Verification

Tests 0 → 19, covering both shadowing namespaces, an ambiguous prefix built by forcing a real 4-hex collision, a blob whose name is hex, the timeout and exec branches, and both refactored call sites. All seven mutations of the new mechanisms are killed (restore hash-verified).

An eighth survived at first and is worth naming: the shadowing-branch fixture for `collect_tier2_pending` pointed the ref at a commit whose `..HEAD` range still contained the change under test, so the correct and the shadowed path returned the same answer and the test bound nothing. The fixture now points at HEAD, where the two ranges genuinely disagree, and carries guard-the-guard asserts.

E2E: after merge, confirm the deploy-staleness snapshot still reports `host_gateway` status `ok` rather than `unknown_commit` — that is the state a broken resolver would produce, and it is visible on the dashboard and in the awareness tick's deploy-staleness check.

---

<a href="https://app.devin.ai/review/wingedguardian/genesis-agi/pull/2147" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


WingedGuardian:
## Structural review

No pattern violations found. Specifics:

- No `nohup`, no direct `task_states` inserts, no `/tmp/` usage, no `git add .`/`-A`, no force-push/reset, no worktree `pip install -e`, no bare `except:` (both `except Exception:` sites are typed and pre-existing), no mutable default args, no `os.killpg`, no raw SQL (uses `genesis.db.crud.update_history`).
- Import hygiene: clean. `import re` and `import importlib` are both correctly stdlib-grouped and alphabetized; no circular/star imports.
- Coverage: `resolve_commit` (new, `src/genesis/observability/snapshots/deploy_health.py:151`) is exhaustively covered — 11 new unit tests plus 2 call-site integration tests in `tests/test_observability/test_deploy_health.py`. One minor gap (WARNING, non-blocking): `collect_host_gateway`'s call site gets only the pre-existing tests plus indirect coverage via `resolve_commit`'s own shadowing tests — unlike `collect_tier2_pending`, there's no dedicated "shadowing branch" integration test through `collect_host_gateway` itself. Low risk given how thoroughly the shared resolver is tested, but worth a follow-up test if this path sees more churn.
- Documentation: `resolve_commit` has a thorough docstring (rationale, measured git behavior, sentinel semantics). No changed-behavior/doc mismatches.
- Genesis conventions: not a hook script (no heavy-import concern), no `$CLAUDE_PROJECT_DIR` misuse, no `IntervalTrigger`, commit/PR title carries a conventional prefix (`fix(observability):`).

Nice defensive touch: `resolve_commit` shape-validates the input against `_ABBREV_SHA` before it ever reaches `subprocess` argv, closing an argument-injection angle (`--upload-pack=…` etc.) even though both current callers only pass trusted-format DB/state-file values — exercised directly in `test_resolve_refuses_an_option_shaped_name`.

## PII/Secrets/Wording scan

Clean. No API keys/tokens/secrets, no private IPv4/IPv6 ranges, no SSH keys, no connection strings, no personal emails, no username-bearing paths, no location-revealing timezones, no cloud account/project IDs, no hardware identifiers, no private hostnames — in the diff, the PR title, or the PR body. Body discusses git internals (`--disambiguate`, `cat-file -t`) and references issue #2145 for cross-context; nothing here describes a PII/secrets leak, a security incident, or names specific private infrastructure beyond terms already documented in the repo's own CLAUDE.md (guardian, host_gateway).

VERDICT: PASS



---

Issue #2080, opened 2026-09-16
Guard replay: give guards a pure classify() entry point, so purity needs no proving


WingedGuardian:
`scripts/replay_guard_corpus.py` replays recorded `(command, cwd)` pairs through a
Bash guard to measure `blocked k/N`. It has to invoke each guard's **production
entry point** — `main()` — because that is the only entry point there is, and
`_run_python_guard` then emulates three ambient channels (`read_payload`, the
process cwd, `sys.argv`) to make `main()` behave like a classifier.

That is the whole reason replay-safety was ever a question. PR #2025 deleted the
apparatus that tried to answer it by reading the guard, and #2036 proposes
answering it by confinement. This issue is the third option, which the same
reviewers raised and which is cheaper for most of the table: **give the guards a
pure classification entry point and call that instead.**

These guards *are* classifiers — `(command, cwd) -> blocked`. The writes, spawns
and `gh` calls are production hook behaviour fused into the same `main()`. Split
them and nothing needs proving: you call the pure half.

## Per-guard separability (investigated, not assumed)

| guard | separable | evidence |
|---|---|---|
| `git_discard_guard` | **already is** | `_clean_violation(cmd)` and `_submodule_recurse_violation(cmd)` are the classifier, both documented "Pure argv", and a test already drives one directly. `git stash create` runs only on the allow path, after the verdict. |
| `worktree_cwd_guard` | yes, small | the verdict is `bool(targets)`, computed before the `/proc` read; every branch after that read returns 2, so the scan feeds only the message |
| inline `settings.json` blob | already pure | stdin→stderr, no writes, no state |
| `protected_paths_guard` | partial | trivial to extract, but the result still depends on `HOME` and `$VARS`, so calling it "pure" would repeat the overclaim #2025 removed |
| `bash_safety_hook.sh` | no — restructure | zero shell functions, 18 bare `exit 2`, and `gh pr view` output feeds the verdict |
| `git_push_guard` | **no, permanently** | its decision *is* a question about the world (`git ls-remote`, `gh`), and `push_allowlist.record()` writes on the decision path into the store `is_recorded()` reads earlier in the same function — so replaying row *N* changes how row *N+k* classifies, and corrupts the real allowlist |

So: three of six are small or already done, one is partial-with-a-caveat, and
two need confinement rather than a classify seam. The two approaches are
complementary, not competing.

## Contract

`hook_input.run_guard` already fixes the shape: **return the exit code (0 allow /
[...]

Issue #2036, opened 2026-09-15
Guard replay: prove replay-safety by confinement, not by reading the guard


WingedGuardian:
`scripts/replay_guard_corpus.py` will not replay a guard until someone declares what running it once per recorded command does to the machine — writes, spawns, network calls, and anything it delegates to. PR #2025 makes those declarations machine-checked instead of prose.

That work established, across two external review rounds and 11 findings, that **the instrument is wrong for half the question**, and this issue is the record of the better one.

## What static checking can and cannot do here

A declaration says a guard does not write, does not spawn, does not call out. Verifying that by reading the code is an absence claim, and absence claims over a Turing-complete language are open-set. The review history is the evidence rather than the argument:

- Round 1 found that the walk missed `Path(...).open("w")`, `os.fdopen`, `io.open`, a non-constant mode, `from subprocess import run`, `import subprocess as sp`, `os.posix_spawn`, and three more — ten realistic spellings, all missed.
- Round 2 found that the fix for those introduced its own gaps: an alias reused across lexical scopes collapses module-wide, so `spawns` reads False on a module that calls `x.run(...)`; a decorated method nested in a class falls back to raw source; a missing root module yields an empty closure and an all-`False` declaration about a deleted artifact passes silently.
- Next round would be `getattr`, `exec`, a rebound name, a metaclass.

The shell half is worse, because there is not even an AST. The scan cross-references tracked script basenames and a published program list, which is sound as far as it goes and **cannot see a redirection**. MEASURED: appending a `printf` redirect into a file to the currently replayable inline guard changes neither scanned set and passes verification, so the write would run once per recorded command.

PR #2025 responds by narrowing the CLAIM — the checker may refute a declaration, never confirm one — which is sound but strictly weaker. It catches a declaration that denies something the walk can see. It cannot certify that a guard is safe to replay.

## The instrument that answers the actual question

Do not prove purity by reading. Make the effects impossible while replaying, and the question stops being a proof obligation:

- run each guard invocation in a child with filesystem writes and network access denied — a namespace/seccomp confinement, or at minimum a preload that fails the relevant syscalls;
- a guard that is genuinely a classifier is unaffected, so the corpus rate is unchanged and directly comparable;
- a guard that is not fails LOUDLY on the first row that trips it, naming the syscall — which is a far better artifact than a green declaration, because it is evidence about the actual guard rather than about our model of the language it is written in;
[...]

PR #2151, opened 2026-09-17
fix(scripts): the corpus tool no longer overstates what it measured

WingedGuardian:
`scripts/replay_guard_corpus.py` reports a `blocked k/N` rate over a cached corpus of recorded commands. Three ways that output claimed more than it had measured, all silent — one claim type, which is why they ride together.

## What changed

**A cached corpus was returned with nothing said about its age.** A rate over a stale corpus read exactly like a rate over the whole tree. The run now prints the cache's build date, its age, and how many transcript files have changed since — on **stdout**, beside the corpus size, because that is the number people quote and a qualifier that does not travel with it is a qualifier nobody reads. The file's own caveat print already carries that reasoning.

**`--limit` truncated without a word.** `corpus: 500 unique real commands` against a far larger corpus is a subset wearing the grammar of a full sweep, printed one line above a rate. It now names the denominator, the flag, and the fact that the subset is the first N in *sort order* rather than a sample — the corpus is sorted, so a limited run is biased toward whatever sorts lowest.

**The cache is now stamped with the moment its build STARTED.** `os.replace` stamps the write, which is the walk's *end*, so a transcript appended during a multi-minute build carried an older mtime than the cache and read as unchanged — the corpus was short and the line said it was current. Reproduced before fixing: a file appended 2s into a walk produced `0 of 1 transcript files have changed since`. The stamp is applied to the temp fd *before* the replace, so it travels atomically with the content and two concurrent rebuilds cannot interleave into a wrong one.

## Deliberately not done

**No staleness threshold and no auto-rebuild.** A strict freshness check fires on *every* run, because the newest transcript is always the caller's own session and is seconds old; a full rebuild walks the entire tree. The other unusable-cache branches rebuild because the cache cannot be READ — a stale cache reads fine and is merely shorter. The defect is the silence, not the staleness, and naming it is the fix.

## Every failure path degrades to a stated unknown

An unreadable subdirectory used to shrink the denominator invisibly: a directory walk swallows that error and raises nothing, so a partly-seen tree produced a confident count. The walk now counts what it could not read and marks both figures floors. A future cache timestamp is reported rather than clamped to `0 minutes ago, 0 changed`. A cache that exists but cannot be read says so instead of vanishing from the output.

**Two cases remain unreported and are named in the changelog rather than implied:** a transcript whose own mtime moves backwards still reads as unchanged, and a far-past cache stamp renders an absurd age rather than an error. Both degrade toward over-reporting staleness, which is the safe direction.

## Why the timestamp is UTC

This line exists to be pasted into a pull request, which makes it a public surface. A local timestamp with a timezone abbreviation would publish the operator's region and the hour they were at the machine. For the same reason two error branches render `strerror` alone — an `OSError`'s `str()` embeds the absolute filename — and the `--rebuild` notice, which interpolates the cache path, moved to stderr. That last one was pre-existing; it is fixed here because it contradicted the invariant this change's own tests assert, in the same file.

## Also

The `bash_safety` replay declaration listed the hook beside five that this repo wires in `.claude/settings.json`, and it wires none for that one — it never has, in any ref, and no install script provisions it. A test binds the claim to the settings file rather than to its own prose, with a vacuity guard so the walk cannot silently start finding nothing.

## Honest scope

**The `corpus:` line is unreachable via the CLI today.** `main()` refuses before `load_corpus()` while every guard is refused pending #2036/#2080, so the provenance and `--limit` output are correct but LATENT until a guard becomes replayable again. They are verified by direct invocation and by tests, not by a full CLI run.

One pre-existing test changed: its assertion followed the `--rebuild` notice from stdout to stderr. Its subject is that the rebuild path is *reachable*; the stream is incidental to that.

## Verification

- **verify-RED: 26/26 mutations killed**, control arm green, both sources restored and confirmed by hash. The sweep aborts on a box-wide pytest-lock refusal and on a selector that collects nothing — the first run of an earlier sweep would otherwise have reported every mutation killed without executing one assertion.
- **97 tests pass**, all pre-existing ones unedited but for the assertion above.
- `ruff check` / `ruff format` clean; `--list` rc=0 with citations still machine-resolving; `assemble_changelog.py --check` valid; the CI guard scripts all rc=0, with `atomic-writes` reporting 0 new sites.
- Private-data scan over the diff: 0 hits across 903 added lines.
- Reviewed by `genesis-architect`, then `genesis-security-reviewer` on the fixed code. Both internal (same-model), so the cross-model streak is untouched and a Codex round is still owed.

E2E: verified by direct invocation against a live cache in both states — a six-day-old cache reporting its age and drift, and a freshly rebuilt one reporting minutes — plus the full test suite. A CLI-level end-to-end run is not possible until replay is re-granted (see Honest scope); that is #2080's job, not this PR's.

Ledger: 68e955f5d47f4f73a1b706bd82c4295d

---

<a href="https://app.devin.ai/review/wingedguardian/genesis-agi/pull/2151" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added cache provenance reporting, including cache age, transcript drift, missing or unreadable files, and timestamp issues.
  * Corpus reports now distinguish complete runs from limited subsets, show available totals, and identify sorted-prefix bias.
  * Rebuild completion notices are now reported separately from corpus output.
  * Privacy-safe reporting avoids exposing home-directory paths.


WingedGuardian:
## Automated first-pass review — PR #2151

**Scope:** 3 files changed (`scripts/replay_guard_corpus.py` +214/-3, `tests/test_scripts/test_replay_guard_corpus.py` +791/-1, one new changelog entry). Not doc-only, so both passes ran.

- **Pattern violations:** none found. No `nohup`, no direct `task_states` writes, no `/tmp/` usage, no `git add .`/`-A`, no force-push/reset patterns, no bare `except:`, no mutable default args, no `os.killpg`, no raw SQL against `genesis.db` (n/a to this file).
- **Import hygiene:** `import time` (script) and `import errno` / `import time` (test) were inserted in correct stdlib alphabetical position, correct group. No circular or star imports.
- **Coverage:** the new public function `_cache_provenance` is exercised by a large, dedicated test block in the same PR (age bands, future-mtime, unreadable cache/dir/file, empty tree, stamp-source, path-leak allowlist across every branch, etc.) — coverage is thorough, not just present.
- **Documentation:** `_cache_provenance` carries a full docstring explaining behavior and rationale (no staleness threshold, no rebuild-on-stale, cost). The two modified call sites (`load_corpus`, `main`) have explanatory comments for the non-obvious parts (mtime stamped on the fd before `os.replace`, provenance read *after* `load_corpus`).
- **Genesis conventions:** commit/PR title carries `fix(scripts):` prefix. Not a hook script, so the stdlib-only hook constraint doesn't apply (and everything added is stdlib regardless). No `$CLAUDE_PROJECT_DIR`/`IntervalTrigger` involved.

Structural review: no ERRORs, no WARNINGs, no NOTEs.

- **Code diff (added lines):** scanned for keys/tokens/passwords, private IPv4 ranges (10.x/192.168.x/100.64–127.x), IPv6 ULA, SSH keys, DB connection strings, secret env values — none present.
- **Soft-warning categories:** no personal emails, no hardcoded username-bearing absolute paths (docstrings reference `/home/<user>` only as a generic pattern, not an actual path), no location-revealing tz identifiers (the change explicitly avoids `%Z`/localtime for exactly this reason and is tested for it), no cloud account IDs, no hardware identifiers, no private hostnames.
- **PR title/body/changelog wording:** the body discusses privacy-by-design choices in detail (why UTC not localtime, why `strerror` not `{exc}`, why a rebuild notice moved to stderr) — this is describing defensive engineering in a script's own output, not disclosing or advertising an actual PII/secrets leak, security incident, or naming any real private infrastructure. No exploitable vulnerability detail. Reads as clean.

### Note
CodeRabbit's automated review was rate-limited on this PR (no findings to reconcile against). Devin Review was still analyzing at review time.

VERDICT: PASS



---

PR #2153, opened 2026-09-17
fix(file_tracker_issue): eliminate home directory and exception leak in lock refusal (#2152)

### LINKED ISSUE #2152: file_tracker_issue.py writes a home-directory path to stderr when the lock directory cannot be created
`_lock_path()` in `scripts/file_tracker_issue.py` renders an absolute path under the
operator's home directory into its refusal message, and that message goes to stderr —
terminal scrollback, session transcripts, CI logs.

```python
    base = Path.home() / ".genesis" / "locks"
    try:
        base.mkdir(parents=True, exist_ok=True)
    except OSError as exc:
        raise Refused(
            f"cannot establish the per-install lock directory ({base}): {exc}. "
            ...
```

Two values leak, not one. `{base}` is `Path.home()/…` by construction, and `{exc}` is an
`OSError`, whose text carries the filename it failed on.

## Why it matters

The account name in a home path is exactly the class of identifier the contribution
sanitizer exists to keep out of public artefacts. Leaking it into a local log is milder
than publishing it, but the same value ends up somewhere durable that nobody audits — and
for an agent session, stderr is routinely captured into a transcript that may later be
summarised, pasted, or attached.

It is not exploitable and nothing bypasses a gate, so this is a privacy defect rather
than a security one.

## Related, already fixed elsewhere

The same class was found by review in two other refusals in this file and corrected: they
now render `type(exc).__name__` and a static remediation hint instead of the exception's
text. This site was left alone deliberately to keep that change scoped, and is filed here
rather than folded in.

An AST test in `tests/test_scripts/test_file_tracker_issue.py` now walks every
`Refused(f"…{exc}…")` in the module and fails on new instances. This site is on its
exemption list, by message rather than by line number, with the reason recorded — so
fixing it means removing one entry from that list and watching the test go green.

## Suggested fix

Render the failure without either value:

```python
        raise Refused(
            f"cannot establish the per-install lock directory ({type(exc).__name__}). "
            "Refusing rather than using a TMPDIR-dependent fallback, which would "
            "let two sessions lock different files and both post."
        ) from exc
```

The location is already documented in the function's own docstring, so an operator who
needs it is not left guessing; the message's job is to say what failed and why the script
refuses rather than falling back.

Please check the other two exempted sites while you are there, and say whether you agree
with the exemption: `_gh_json`'s and the duplicate check's `JSONDecodeError` carry no
path, `create_issue`'s `SubprocessError` text is the script's own argv, and the
privacy-scanner `ImportError` is a module name. Those looked benign; a second opinion on
that judgement is worth more than the fix itself.

## Acceptance

- `_lock_path`'s refusal contains neither an absolute path nor the exception's text.
- The site is removed from the AST test's exemption list and that test passes.
- A test drives the failure — e.g. by pointing the lock base at an unwritable location —
  and asserts the captured stderr contains no `/home/` substring.

## Provenance

Read in `scripts/file_tracker_issue.py` at commit
`a1e0a6d90def`; found by an AST enumeration of the class
rather than by inspection, after two instances of it were reported by review.

## E2E

None — this changes the text of an error message on a path that requires an unwritable
lock directory. The acceptance test above is the verification; there is no runtime
surface beyond stderr.

## Remote-buildable

Yes. One message, one test, no live database, no deployed service, no state from any
particular install.

chenzeyan54-commits:
## Summary
Fixes #2152

In `scripts/file_tracker_issue.py`, `_lock_path()` was rendering `{base}` (which resolves to `Path.home() / ...`) and `{exc}` (the raw `OSError` string, which may contain file paths) into the `Refused` exception message on `mkdir` failure. This leaks the local operator's home directory username into logs, stderr, and session transcripts.

## Changes
- Updated the `Refused` error string in `_lock_path()` to only render `({type(exc).__name__})`, consistent with the other error refusals in this module.
- Retained the explanatory rationale ("Refusing rather than using a TMPDIR-dependent fallback...").
- Enhanced regression test `test_lock_directory_failure_refuses_rather_than_using_a_tmpdir_fallback` in `tests/test_scripts/test_file_tracker_issue.py` to assert that neither `read-only file system` nor `Path.home()` appears in the refusal message, and that `(OSError)` is rendered instead.

## Verification
- Ran `pytest tests/test_scripts/test_file_tracker_issue.py -k "lock_directory"` -> **PASSED** (1 passed, 0 failures).

E2E: none — the only runtime surface is one stderr string on a path that needs an unwritable lock directory; verified pre-merge by driving main() to that path and scanning real stderr (see review comment).

---

<a href="https://app.devin.ai/review/wingedguardian/genesis-agi/pull/2153" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lock directory failure messages by showing the exception type without exposing filesystem paths or raw system error details.
  * Lock setup refusals now provide clearer, safer error information.


WingedGuardian:
Verified before merge. Ran the fix end-to-end rather than only reading it.

**CLI-level E2E** — drove `main()` to the lock-directory failure with a stubbed runner and captured real stderr at fd 2:

```
REFUSED: cannot establish the per-install lock directory (OSError). Refusing rather than
using a TMPDIR-dependent fallback, which would let two sessions lock different files and
both post.
```

Exit 2. No absolute path, no raw OSError text, no traceback. The `from exc` chain cannot resurface either: `_lock_path` is called from inside the `try` whose `except Refused` prints `str(exc)` via `_warn`, so `__cause__` is never rendered.

**Verify-RED** — reverting the one changed line makes the same probe print the home path and the raw strerror, and turns the new unit assertions red. So the test binds the fix rather than passing alongside it.

**Non-vacuity** — worth stating because two of the three assertions could easily have been hollow. `_PathWithFailingMkdir` subclasses the real `Path` and overrides only `mkdir`, so `Path.home()` resolves to the genuine home and the OSError carries genuine text. Both `not in` assertions therefore have something real to fail against.

Full suite on this head: 60 passed.

This addresses the CodeRabbit note in substance — that comment asks for a CLI-level stderr regression test, and the behaviour it would assert is confirmed above. Locking it in as a test is worth doing but is not a condition of this merge; the class-level work is moving separately in #2146 and a follow-on.

Merging with `review-override` (Codex reports its code-review budget exhausted, which is a separate budget from CLI usage) and `scheduled-review-override`. `leak-detector` is green at this head, and the diff is six lines that remove a path from an error string.

E2E: none — the runtime surface is one stderr string on a path that requires an unwritable lock directory; verified above rather than deferred.

Thanks for the clean fix, and for not pretending to satisfy the issue's second acceptance bullet — the AST guard it refers to is not on `main` yet, it is in flight in #2146.

PR #2157, opened 2026-09-17
feat(awareness): a revisit condition that waits to be remembered is not a trigger

WingedGuardian:
A follow-up whose revisit condition reads *"when we reach 200 stars"* is a sentence, and nothing in this system evaluates sentences. The row sits in the blocked lane until somebody happens to read it and happens to know the number has moved — the shape of thing this install has lost work to before. This turns the sentence into an event: one unauthenticated read a day, and an observation when the count crosses.

## It deliberately leaves the follow-up alone

An earlier design flipped the row's `work_state` from `blocked_on_trigger` to `ready` so it would enter the actionable queue by itself. That needed a new crud function — `work_state` is set only at creation and `kind` is *derived* from it, so the two must move together — and it hands the row to readers, ego dispatch among them, whose behaviour I had not traced. The observation names the row instead. Announcing costs nothing if it is wrong; dispatching might not.

## The permanent type is load-bearing, not a detail

`repo_milestone_reached` is registered in `_PERMANENT_TYPES`. An unregistered observation type gets the 14-day default TTL, and the observation *is* the wake-up — so under the default it would expire before anyone acted on it and the watcher would be a no-op nobody noticed. It is deliberately **not** in `INTERNAL_OBS_TYPES`: a milestone the user never sees cannot wake anything up. Both properties are pinned by a test, because the type and the TTL policy live in different files and nothing else would catch them drifting apart.

## Two defects found by running it once, not by the fifteen tests that passed first

**It was inert on the very install it was built for.** `env.github_user()` returns `''` here — this box's `genesis.yaml` has no `github:` section at all, because the interactive generator that writes those keys has effectively never run. So the first live run reported "nothing to watch" about a repo we had been pushing to all day. A watcher that silently does nothing is precisely the failure it exists to prevent, reproduced inside itself. Fixed by *removing the dependency* rather than repairing one machine's config: the slug now falls back to the checkout's own `origin`, parsed for both git URL shapes, refusing a non-GitHub remote rather than guessing.

**The write path did not import at all.** `genesis.db` exports no `get_db`; every unit test passed because they all stub the announcement whole, so the suite proved the decision logic and never executed the code that does the work. Now uses the short-lived `aiosqlite.connect` shape from `repo_pulse_worker`.

Raised separately, out of scope here: the same missing config key leaves `mcp/health/contributor_issue.py:97` returning a bare ownerless repo slug.

## Fail directions, not symmetric, each pinned

| situation | behaviour | why |
|---|---|---|
| count unreadable | exit 1, no state, no announcement | a run that verified nothing must not resemble one that found no news |
| install unconfigured | exit 0, silent | bootstrap enables every shipped timer on every clone; a fresh one has nothing to watch, and a daily red unit there is a defect we'd be shipping |
| state file lost or corrupt | announce | a duplicate is an annoyance, a silent miss is the whole failure; the real guard is the stable content hash |
| announcement failed | exit 1, do **not** record it | recording it means the crossing is never announced at all |
| `stargazers_count` absent | raise, never 0 | zero compares below every milestone and reads as "no news" |

## Verification

97 tests green; ruff clean. **Mutation sweep 7/7 RED** against an asserted-green baseline with `py_compile` on each mutant and hash-verified restores: boundary `>=`→`>`, announce lowest instead of highest, record state after a failed announce, unreadable count returns 0, corrupt state read as already-announced, malformed override silently falls back, and the observation type losing permanence.

**Live acceptance, both directions** — real API, real writes, against a scratch DB built from the real `observations` schema rather than production (a forced threshold would otherwise have written a false *permanent* row into the real store): silent at 97 stars against the default 200, and with the threshold forced below the current count, announced once, silent on the second run, with the stored row read back showing `priority=high` and `expires_at IS NULL`.

E2E: none — verified live before merge rather than after, as described above.

---

<a href="https://app.devin.ai/review/wingedguardian/genesis-agi/pull/2157" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added automated monitoring for repository star milestones.
  - Announces newly reached milestones and records them as permanent observations.
  - Supports configurable milestone thresholds, with sensible defaults.
  - Added scheduled daily checks with missed-run recovery and randomized execution timing.
  - Handles unavailable repository information, invalid counts, and repeated checks without creating duplicate announcements.


WingedGuardian:

### Domain 1 — Structural

No ERROR-level violations found against the checklist:
- No `nohup`, no direct `task_states` writes, no `/tmp/` usage (state file correctly lives at `~/.genesis/star_milestones.json`), no `git add .`/`-A`, no force-push/reset, no worktree-path `pip install -e`, no bare `except:` (every catch names concrete exception types, including the deliberately-broad `except Exception as exc:  # noqa: BLE001` in `main()`, which is annotated and justified), no mutable default args, no `os.killpg`, and no raw SQL — the one DB write goes through `observations.create(...)` in `genesis.db.crud.observations`; the only bare statement on the connection is a `PRAGMA busy_timeout` for connection tuning, not a data query.

Import hygiene: clean. `scripts/star_milestone_check.py` groups stdlib imports correctly at module level and defers the package-local (`genesis.env`, `genesis.db.*`) and third-party (`aiosqlite`) imports to function scope with explicit `# noqa: PLC0415`, which is a deliberate and already-annotated choice, not disorder. No star imports, no circular-import shape.

Coverage: good. The new script has a dedicated, thorough test file (`tests/test_scripts/test_star_milestone_check.py`), and the one-line addition to `src/genesis/db/crud/observations.py` (registering `repo_milestone_reached` in `_PERMANENT_TYPES`) is directly pinned by `test_the_observation_type_is_permanent`, which also asserts it's absent from `INTERNAL_OBS_TYPES`. No changed function with existing tests left unaddressed.

Genesis conventions: the commit carries a conventional prefix (`feat(awareness): ...`). No `IntervalTrigger` (uses systemd `OnCalendar` timer, correctly with `Persistent=true` and no `OnBootSec`). No `$CLAUDE_PROJECT_DIR` misuse. Not a hook script, so the heavy-dependency-in-hooks rule doesn't apply.

NOTE (docs, non-blocking): several private helpers in `scripts/star_milestone_check.py` (`_milestones`, `_star_count`, `_already_announced`, `_remember`, `_announce`) don't have their own docstrings — though the module-level docstring and inline comments already cover the "why" for each, so this is low-value nit-picking rather than a real gap.

Scanned all added lines plus PR title, body, and the changelog entry.

- No hard-block material: no API keys/tokens/passwords, no private IPv4/IPv6 ranges, no SSH keys, no DB connection strings with credentials, no secret-bearing env values. The script is deliberately unauthenticated (public GitHub API, no token), and `GENESIS_STAR_MILESTONES` only ever carries integers.
- No soft-warning material either: no personal emails, no username-bearing absolute paths (state path uses `os.path.expanduser("~/.genesis/...")`), no location-revealing tz identifiers (uses UTC throughout), no cloud account/project IDs, no hardware identifiers, no private hostnames.
- PR title/body/changelog wording: no mention of a PII/secrets leak, no reference to a security incident, no naming of specific private infrastructure, and no exploitable vulnerability detail. The body's discussion of a missing `github.user` config key on "this box" is an ordinary root-cause writeup, not an infrastructure exposure.

VERDICT: PASS




---

PR #2168, opened 2026-09-19
fix: resolve cc-slot root from checkout path

### LINKED ISSUE #2138: cc-slot.sh hardcodes the repo path, so the slot door is dead on any clone not at ~/genesis
## Summary

`scripts/cc-slot.sh:36` hardcodes the repository location:

```sh
GENESIS_ROOT="${HOME}/genesis"
```

Every later use depends on it — the working directory the slot session starts
in, and the interpreter path used for the probes (`${GENESIS_ROOT}/.venv/bin/python`
at several sites). On a clone that lives anywhere else, all of them point at a
directory that may not exist.

## Why this matters now

This is the same defect class that has been fixed twice recently in adjacent
code: a unit whose `ExecStart` assumed `~/genesis` was installed dead on any
other clone, and the installer wrote a login hook and a wrapper with the same
assumption. In each case the artifact was created successfully and every check
passed, because nothing verified the path it named.

The slot door is the remaining instance. It fails more quietly than the others:
the probes simply return nothing when the interpreter is absent, so a session
starts with its diagnostics silently missing rather than erroring.

## Fix

Derive the root from the script's own location rather than assuming it. The
script lives at `<root>/scripts/cc-slot.sh`, so resolving its real path and
taking the parent of its directory gives the root for any clone. Resolve
symlinks, since the script is reached through a wrapper on some installs.

Keep an environment override if one is wanted, but the default should be
self-location rather than a fixed path.

## Acceptance

Run the script from a checkout that is **not** at `~/genesis` and assert it
resolves the correct root — and that the probes actually run rather than
returning empty. A test asserting only that `GENESIS_ROOT` is set would pass
today; it must assert the resolved value against a clone at an arbitrary path.

Worth checking the same assumption in `scripts/bootstrap.sh`, which contains two
further references to the same fixed location.

mikemikimike:
Fixes https://github.com/WingedGuardian/GENesis-AGI/issues/2138

## Summary
- Resolve the checkout root from cc-slot.sh's real entrypoint, including symlinked invocations.
- Render the bootstrap tmux wrapper with the actual checkout path and shell-safe quoting.
- Extend the script harnesses to cover arbitrary checkout paths and symlinked doors.

## Tests
- bash -n scripts/cc-slot.sh scripts/bootstrap.sh
- ruff check .
- pytest tests/test_scripts/test_cc_slot_cap.py tests/test_scripts/test_session_doors.py -q

---

<a href="https://app.devin.ai/review/wingedguardian/genesis-agi/pull/2168" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when the project is installed outside the default home-directory location.
  * Commands now locate the project from its actual installation path, including when launched through symbolic links or from paths containing spaces and special characters.
  * Improved shell-wrapper reliability when environment settings could redirect path resolution.
  * The shell wrapper safely handles special characters, ampersands, and newlines in paths.

* **Tests**
  * Added coverage for alternate checkout locations, symbolic links, path handling, and generated shell-wrapper behavior.


WingedGuardian:
Confirming the review-bot finding on the pane launch line (`scripts/cc-slot.sh:1142`). I read it at this head and it's real.

That argument is re-parsed by a fresh shell in the pane — which is why `\$?` and `\$__ec` are escaped on the same line — and the two `${GENESIS_ROOT}` fragments are the only unprotected interpolations left on it:

```
"${_OAUTH_SRC}cd ${GENESIS_ROOT} && … ; ${GENESIS_ROOT}/scripts/cc_exit_capture.sh ${SLOT} \$__ec …"
```

With a checkout at `/srv/Genesis Work`, the pane runs `cd /srv/Genesis` and claude starts in the wrong directory. The ephemeral-subcommand path already quotes its `cd "$GENESIS_ROOT"` (line 157), so this is the one spot that didn't get the same treatment — and it's reachable now precisely because this PR makes the root arbitrary.

The fix is the pattern you already used in bootstrap.sh: `printf '%q'` on the two root-derived fragments.

For the test, `_run_cd_guard` in `tests/test_scripts/test_cc_slot_oauth.py` already evaluates the real launch string with a configurable root, so a case with a space in the path drops straight into it.

Your call whether to fold it in here or leave it to us.

WingedGuardian:
Still open pending CI, but a heads-up first: **maintainer edits were pushed onto your branch.** `maintainerCanModify` was enabled, so I fixed a few things directly rather than sending them back for another round.

Three changes, all now on `fix/cc-slot-path-2138`:

1. **`_OAUTH_SRC` embedded `${GENESIS_ROOT}/.venv/bin/python` inside escaped double quotes.** The pane shell re-parses that command string, and double quotes do not stop `$(...)`, backticks, or a quote that closes the quoting — so a checkout path containing those executed in the pane. It is `printf '%q'`-quoted now, like the other two checkout-derived values on that line. Verified with the pre-fix line as a control: pre-fix executes the payload and never reaches the interpreter; the fixed line runs the interpreter at the literal path.

2. **`$(cd … && pwd)` in both `bootstrap.sh` and `cc-slot.sh`.** With `CDPATH` exported and a relative entrypoint, `cd` *searches* CDPATH — so the capture can both collect an extra echoed line and resolve into a **different checkout entirely**. Silencing the echoed line alone is not enough (it turns a loud failure into a silently wrong root), so both now clear CDPATH inside the substitution. The regression test covers two arms for that reason: the benign one (`CDPATH="."`) only exercises the echo, and a decoy checkout holding the same relative path is what pins the real fix.

3. **The oauth test harnesses re-implemented the script's `%q` assignments** instead of extracting them, so a regression in the assignment itself would have gone unnoticed. They extract the shipped lines now.

Your `c83cf990` is what made the first one visible: the two values already on that command line were handled correctly — it was the third call site that had been missed. Thanks for the careful work on this.

PR #2193, opened 2026-09-20
docs(claude): a question must be asked, and must carry its own context

WingedGuardian:
## What this changes

One new rule in `CLAUDE.md`, directly after the existing `AskUserQuestion`
constraint because it governs the same tool and the same surface.

## Two failures, and they are one loop

**A question posted as prose gets missed.** A question in the body of a message
competes with everything else in that message and loses. The reader is not
obliged to catch it, and treating it as asked is a fiction the asker maintains
alone.

**A question that gets no answer then gets CARRIED, not re-asked.** Restated
each turn as an open item, worked around, never forced.

MEASURED on this install's session ledger: one row carried an unanswered
architecture question for **12 days** while being read out every time the
ledger was read; another carried a design question flagged in a PR body for
**11 days**. Neither was ever re-asked. Both were narrated.

**The half that is easy to miss is why they went unanswered.** They were not
answerable as asked. Each named a decision and supplied no current state, no
tradeoff per option, and no cost of deciding nothing — which asks the reader to
reconstruct the analysis the session was supposed to have done already. So the
two failures are not independent: an unanswerable question is exactly the one
that gets no answer, and a question with no answer is the one that gets carried.

## The rule

- If an answer is needed to proceed, **ask through the tool**; if it goes
  unanswered and still blocks, **re-ask**. A question you keep mentioning and
  never force is a status line, and it will sit for as long as you are willing
  to narrate it.
- Every question **carries the facts the decision turns on**: state VERIFIED
  NOW rather than recalled from a plan, a summary or a ledger row — those go
  stale, and a decision made on stale state is one the asker caused — the
  concrete cost of each option, and the cost of deciding nothing. Say which
  parts were measured and which were not.
- **An option whose description does not say what it costs is a label, not an
  option.**
- Scoped to any channel requesting a decision, not just the tool: a dispatched
  session's report and a PR comment owe the same context.

## Why a rule and not a gate

There is no mechanical signal for *this question still blocks*. That is a
judgement about the state of the work, and a detector keying on a heuristic
would fire on rhetorical questions and on questions already answered. The
measurement above prices the **problem**, not a detector for it — which is the
standing bar here for proposing an enforcement surface, and it is not met.

## Notes for review

- **No identifiers, no PR or issue numbers.** `CLAUDE.md` is durable
  instruction that ships to every clone; numbers go stale and are
  install-specific. The failure shape is stated; the two duration figures
  describe repo work only.
- **Placement** verified by reading the surrounding bullets, not by line number.
- **No CI size ceiling applies.** Checked `.github/workflows/`: the only
  `CLAUDE.md` reference in `ci.yml` is a comment pointing at the targeted-test
  policy. The hard byte ceiling in this repo governs the injected identity
  context files, not this one.
- **No test.** A prose rule has no assertable surface, and inventing one would
  be the vacuous-assertion defect this repo keeps catching.

E2E: none — prose instruction, no runtime surface. The live exercise is the
next decision this repo's sessions put to a human; there is no code path to
exercise.

---

<a href="https://app.devin.ai/review/wingedguardian/genesis-agi/pull/2193" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


WingedGuardian:
@codex review

PR #2206, opened 2026-09-21
fix(scripts): honour --dry-run in --recover instead of silently ignoring it (#2188)

### LINKED ISSUE #2188: worktree_lifecycle: --recover silently ignores --dry-run and performs a real recovery
## Summary

`--dry-run` is documented as "Show what would happen without doing it". It is
honoured by the reap path. It is **not** honoured by `--recover`: combining the
two performs a full, real recovery with no warning that the flag was ignored.

## Reproduction

```
worktree_lifecycle.py --recover <entry-name> --dry-run
```

Observed: the command restores the worktree, re-registers it with git, and
prints success lines in the past tense —

```
Recovered to .../worktrees/<name> (branch: <branch>)
Restored 302 untracked file(s) from trash
Archive kept at .../<name>.tar.gz (recovery copies; it does not consume)
```

Measured on one install: 4,110 files written to disk and one live git worktree
registered, from a command whose flag says nothing will be done.

## Why this matters more than a normal flag bug

`--dry-run` on this tool is not a convenience. It is the documented way to ask a
destructive module what it is about to do, and the reap path teaches you it works
("WOULD TRASH ..."). That is precisely what makes the silent inconsistency
dangerous: the flag is trustworthy on one subcommand and inert on another, so the
habit the tool builds is the habit that burns you.

The recovery direction is the less harmful one — nothing is deleted, and the
archive is retained. But the same class of bug on a path that removes something
would be unrecoverable, and there is nothing structural preventing that: the flag
is simply not consulted.

There is a secondary trap in undoing it. `git worktree remove` is blocked by the
repo's own `worktree_cwd_guard` hook (correctly — worktree lifecycle is owned by
this module), so an accidental recovery cannot be reverted by the obvious command
and has to be left for the reaper to collect at the normal 7-14 day interval.

## Suggested fix

Either honour `--dry-run` in `_recover` (print the target path, branch, and file
count that WOULD be restored, and return before writing anything), or reject the
combination outright with a nonzero exit and a message saying `--recover` does not
support `--dry-run`. Silently ignoring it is the one option that should be off the
table.

If the flag is honoured, the printed lines should also move to the conditional
tense on that path — the present wording reports completed actions, so a reader
who missed the flag being dropped has no second signal that anything happened.

sup3dev:
## Summary

`--recover` ignored `--dry-run` and performed a full real recovery — #2188 measured 4,110 files written and a live worktree registered from a command whose flag promises nothing will be done. The reap path honours the flag ("WOULD TRASH ..."), which is exactly what makes the silent inconsistency dangerous: the habit the tool teaches on one subcommand is the one that burns on the other.

## Related issue

Closes #2188

## Changes

`_recover` now takes `dry_run`, and `main()` passes `args.dry_run` through. The resolution and ambiguity checks are read-only and run unchanged; under dry-run the new `_describe_recovery` reports before anything is written — no extraction, no `git worktree add`, no moves. It reads the entry's `.trash_meta.json` from the directory, or from inside the archive in memory via `tarfile` without extracting anything to disk, and prints in the same conditional voice as the reap path: `WOULD RECOVER <path> (branch: ... / detached at ...)`, `WOULD RESTORE <n> untracked file(s)`, and whether the entry would be consumed (legacy directory form) or the archive kept. Pre-flight refusals — missing or incomplete metadata, an occupied destination — use the real run's exact messages, so the preview cannot promise a recovery the real run would refuse.

## Testing

- [ ] `ruff check .` passes
- [ ] `pytest -v` passes
- [ ] `docs/architecture/CURRENT.md` updated — not needed: the flag now matches its documented contract; no subsystem capability change

Added `test_recover_dry_run_previews_and_writes_nothing`, following the existing legacy-entry pattern (real git repo fixture, manual trash entry + meta): it asserts the WOULD lines appear on stdout, that the destination is NOT recreated, and that the trash entry is NOT consumed. Existing recover tests are unchanged — `dry_run` defaults to False and the signature is backward compatible. Neither ruff nor the suite was run locally; CI runs both on this PR.

E2E: `worktree_lifecycle.py --recover <entry> --dry-run` prints the WOULD lines; `git worktree list` and the trash directory are unchanged afterwards.

---

<a href="https://app.devin.ai/review/wingedguardian/genesis-agi/pull/2206" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added dry-run support for worktree recovery.
  * Recovery previews show what would be restored without changing the destination or trash contents.
  * Previews report the number of recoverable files and identify validation issues that would prevent recovery.

* **Bug Fixes**
  * Dry-run recovery no longer performs restoration actions, including recreating the destination or modifying stored recovery data.


WingedGuardian:
Thanks for this — you found a real defect and put the fix in the right place. #2188 was a command that wrote 4,110 files while its flag promised it would write nothing, and the reap/recover inconsistency you identified is the part that made it dangerous rather than merely surprising.

I want to be straight with you about the state of this PR, because the bot comments here and the existence of #2208 both need explaining.

**What happened.** A Genesis session reviewed this PR and had fixes for the three Devin findings. That session can't push to your fork, so it staged them in a separate branch — #2208 — built on your commit. You then committed `f10bf89b` here, so the two have diverged: #2208 carries four further corrections (`8d98946c`, `b9a8cf55`, `7c0dd5b8`, `5e663d7d`) on top of the same preview logic.

**Where I land: yours is the PR that should land.** Closing an outside contributor's work in favour of a maintainer's near-identical copy of it is the wrong trade, and #2208 exists only because a session could not push here. Your authorship of `cc9ab44a` stays in the history on either path.

**Two ways to finish it — your call:**

1. **You take the rest.** Cherry-pick those four commits from #2208's branch onto `fix/recover-dry-run`. #2208 then gets closed as superseded.
2. **I push them.** `maintainer_can_modify` is enabled on this PR, so a maintainer can add those commits to your branch directly. Less work for you, same outcome.

**What is still outstanding before merge, so you aren't surprised by it.** CI is green and the diff is mergeable. Two of *our* gates are open on this head — a Codex review at `f10bf89b`, and the scheduled leaks scan. Both are maintainer-side; nothing to do on your end. I'm holding off on requesting them until we've settled which branch carries the final state, since a review at an about-to-change head just goes stale.

One review note worth understanding rather than just inheriting, whichever path you pick: CodeRabbit flagged that the preview counted all stored files rather than restorable ones (`.git` paths, `.trash_meta.json` basenames, and files the checkout itself would recreate), and that the archive metadata member wasn't pinned to the canonical entry. Both are addressed in the #2208 commits.

WingedGuardian:
@codex review

WingedGuardian:
Correction on my `@codex review` above — that was my mistake, not a change of plan.

A queue-sweep session (me) saw this PR reading `ci: green`, `mergeable`, `inline-findings: ok`, with the only two open gates being a Codex review and the leaks scan, and fired the review request as routine. I had not read the comment directly above it, which had deliberately deferred exactly that:

> I'm holding off on requesting them until we've settled which branch carries the final state, since a review at an about-to-change head just goes stale.

That reasoning is correct and still stands. If the answer to the question above is option 2 (a maintainer pushes the four commits to this branch), the head moves and the review I just requested is spent on code that is about to be replaced.

Nothing here needs action from @sup3dev because of this — the question in the previous comment is unchanged and is still the only thing outstanding on your side: do you want to cherry-pick those four commits yourself, or would you rather a maintainer push them to your branch?

Recording it so a later reader does not take the review request as evidence the deferral was resolved. It was not; I overrode it by not reading first.

WingedGuardian:
@coderabbitai review

PR #2208, opened 2026-09-21
fix(scripts): accurate --dry-run recovery preview (PR #2206 review fixes)

### LINKED ISSUE #2188: worktree_lifecycle: --recover silently ignores --dry-run and performs a real recovery

WingedGuardian:
## Summary

Supersedes #2206 (which lives on a fork this session cannot push to) — carries its commit plus fixes for the three Devin Review findings on it. The dry-run preview now reports what a real recovery would actually do instead of reading stored metadata verbatim.

## Related issue

Closes #2188

## Changes

- `_describe_recovery` takes `repo_root` and checks whether the recorded branch still exists (`rev-parse --verify refs/heads/<branch>`); when it is gone the preview reports `detached at <commit> — branch 'x' no longer exists`, matching `_restore_from_dir`'s real retry path.
- The "WOULD RESTORE n untracked file(s)" count now mirrors the restore walk — `.git` paths, `.trash_meta.json` basenames, and files the checkout would recreate are excluded via `git ls-tree -r --name-only <ref>`; when the ref is unreadable it reports "up to N" rather than a confident wrong number.
- Archive metadata lookup prefers the root member (`<name>/.trash_meta.json`, depth 2) so a worktree's own nested `.trash_meta.json` cannot hijack the preview.

## Testing

- [x] `ruff check` passes (`ruff==0.15.10`, same pin as CI)
- [x] `pytest` — `test_recover_dry_run_previews_and_writes_nothing` passes with a tightened assertion (`WOULD RESTORE 1 untracked file(s)`); the 28 other failures locally are environmental — this box's git 2.34 lacks `worktree list -z` — CI's `test` job is the arbiter
- [ ] `docs/architecture/CURRENT.md` updated — not needed; preview-only change inside the existing recovery surface

Manual verification: a fabricated archive with the nested `.trash_meta.json` ordered FIRST still resolves the root metadata, prints the deleted-branch detached note, and counts only the one restorable file.

E2E: `worktree_lifecycle.py --recover <entry> --dry-run` on an archive whose branch was deleted prints the detached fallback and the accurate untracked count; trash and `git worktree list` unchanged.

Link to Devin session: https://app.devin.ai/sessions/c03b914aecb1432abad44d8bd86554df
Open in Devin Desktop: https://app.devin.ai/desktop/session/c03b914aecb1432abad44d8bd86554df?variant=devin
Requested by: @WingedGuardian

---

<a href="https://app.devin.ai/review/wingedguardian/genesis-agi/pull/2208" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added a read-only recovery preview that validates recovery requirements without creating files or consuming recovery data.
  - Recovery previews now show planned file restoration, fallback behavior, partial recovery, and potential conflicts.

- **Bug Fixes**
  - Improved validation for metadata, destinations, references, archive contents, tracked files, symlink paths, and file collisions.

- **Tests**
  - Added coverage confirming dry-run recovery reports planned changes without modifying the destination or recovery data.


WingedGuardian:
**HELD — do not merge. Converted to draft deliberately.**

This PR carries an outside contributor's commit (`cc9ab44a`, @sup3dev) plus four maintainer corrections on top of it. It exists only because the session that wrote those corrections could not push to the contributor's fork — not because this is the branch that should land.

The decision is already recorded on #2206: **#2206 is the PR that should merge.** Closing a first-time contributor's work in favour of a maintainer's near-identical copy is the wrong trade, and #2206 is waiting on nothing but @sup3dev's answer to a question about how the four commits get onto their branch.

This PR reads `mergeable`, and a queue-sweep session has already acted routinely on this pair once today without reading the thread first (see #2206). Draft status is the guard against that happening again with a merge instead of a review request.

Disposition: closes as superseded once #2206 carries the final state. It reopens as a live candidate only if @sup3dev declines both options or goes quiet long enough that the fix should not wait — and that is a call to make out loud on #2206, not silently here.

